PR #10, opened 2021-06-25
terms: typo

arno01:


arno01:
oh, figured there is already a fix there https://github.com/osmosis-labs/osmosis-frontend/pull/3

PR #47, opened 2021-09-19
Clean tx error messages, to simplify over what the chain outputs

### LINKED ISSUE #46: Introduce a "CleanTxErrorMessage" function, replace common hard to parse error messages
We should clean error messages from the state machine to be far more human readable.

In these lines, https://github.com/osmosis-labs/osmosis-frontend/blob/master/src/tx/tx-tracer.ts#L160-L161 we should clean the error message before creating it into an Error.

We should in the beginning do:
1) Find/Replace ibc denoms strings with human readable denoms from the asset list
2) Run the following regex: Find Regex `signature verification failed; please verify account number \([0-9]*\), sequence \(([0-9]*)\) and chain-id \(osmosis-1\): unauthorized` Replace Regex `You have too many concurrent txs in flight! Try resending after your prior tx lands on chain. (We couldn't send the tx with sequence number $1)`. You can play with the regex here: https://regex101.com/r/ZkRbXL/1/
The way you can do this regex substitution before displaying the error in JS per my understanding is string_to_replace.replace(findRegex, replace)

ValarDragon:
Closes #46 

However I have not tested this on a front-end deployment, working on getting the test setup working properly

### DIFF
--- a/src/tx/tx-tracer.ts
+++ b/src/tx/tx-tracer.ts
@@ -158,7 +158,9 @@ export class TxTracer {
 						if (this.txSubscribes.has(obj.id)) {
 							if (obj.error) {
 								// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-								this.txSubscribes.get(obj.id)!.rejector(new Error(obj.error.data || obj.error.message));
+								const errMsg = cleanErrorMessage(obj.error.data || obj.error.message);
+								// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+								this.txSubscribes.get(obj.id)!.rejector(new Error(errMsg));
 							} else {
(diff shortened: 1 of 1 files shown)


ValarDragon:
hrmm, looks like this is the wrong spot for the sequence number issue, were still getting the old message. Trying to figure out where the response for attempting to broadcast a tx is rn

ValarDragon:
I think for improving the sequence number error message, that would have to be done here: https://github.com/osmosis-labs/osmosis-frontend/blob/master/src/stores/osmosis/account/index.ts . Perhaps what should be done is defining this `cleanErrMsg` function in utils, and then using that in every msg send's response & query responses?

ValarDragon:
Closing the PR for now!

PR #50, opened 2021-10-03
Update JUNO explorerUrl

giansalex:


src/config.ts
```diff
@@ -619,6 +619,6 @@ export const EmbedChainInfos: ChainInfoWithExplorer[] = [
 			},
 		],
 		features: ['stargate', 'ibc-transfer'],
-		explorerUrlToTx: 'http://junoscan.com/tx/${txHash}',
+		explorerUrlToTx: 'https://junoscan.com/transactions/{txHash}',
```
ValarDragon, line 622: This should need the `$`, no?

giansalex: nope 
https://github.com/osmosis-labs/osmosis-frontend/blob/87494115a41ca3391b2d6b59559275c575f42a94/src/config.ts#L246

ValarDragon (approved):

src/config.ts
giansalex: @ValarDragon Juno is available on mintscan, will it be used?

ValarDragon: We can use that instead, I don't really have an opinion on junoscan vs mintscan, other than mintscan is convenient since its the same as most other chains

ValarDragon (approved):

PR #65, opened 2021-10-18
rowan

sunnya97:


sunnya97:
Blocked on Osmosis and Sifchain upgrade to ics20-v2

PR #77, opened 2021-11-02
krt

sunnya97:


sunnya97:
@Thunnini is this blocked on allowing Keplr to "re-suggest chain"?

sunnya97:
I tested it, seems to be working fine

PR #86, opened 2021-11-15
Issue #45: Add HALF button to swap screen

### LINKED ISSUE #45: Add HALF button to swap screen
Currently, there is a MAX button to select your total balance for swap, other Dex'es like Solana Raydium have a half option as well which is very useful when converting funds for 50/50 pools.

image

Osmosis would benefit from this to help users swap tokens for liquidity pools, we could go a step further and offer buttons like:
10% 25% 50% 75% MAX which should cover most pool configurations.

### LINKED ISSUE #72: FakeFeeConfig logging to console
I noticed this recently while browsing the app; any reason the [fee primitive should be logged](https://github.com/osmosis-labs/osmosis-frontend/blob/399e3dca6feb40759faf0f6b5ce7399cd6907e16/src/hooks/tx/fake-fee-config.ts#L63)? I'm happy to open a PR removing this unless there's a specific reason to keep this

jonator:
### Problem
Osmosis is an AMM _for_ LPs: a common task is reinvesting earned OSMO into 50/50 pools, which necessitates dividing earnings in half.

### Solution
We should lower the friction for LPs halving their balances to participate in 50/50 pools. Code changes support dividing by arbitrary values should we feel the need to support other quick-divides later.

Screen Shot 2021-11-14 at 5 58 15 PM

Fixes #45 
Also fixes #72 (separate revert-able commit)


daniel-farina:
Loving this! This is very useful!

ValarDragon (approved):
Works great! Just tested it! The toggling works really well when flipping between max, half, and editing the number

sunnya97:
+1

Thunnini:
I think it's more intuitive to work based on multiplying rather than dividing. @jonator Please take a look at the changes.

Issue #95, opened 2021-11-19
Add swap pairs as URL params


Callum-A:
Hi there,

Would love it on the swap page if the currently selected pair was appended to the URL as params. Something along the line of:
https://app.osmosis.zone?from=ATOM&to=OSMO

Would allow users to bookmark a link to directly take them to the swaps they perform daily, i.e. when rewards are paid out.

PR #99, opened 2021-11-23
Issue #95: Add swap pairs as URL params

jonator:
### Problem

Swap functionality, the core of Osmosis, is not very shareable or save-able. Right now, every time you navigate to Osmosis, you're required to specify the tokens you wish to swap. In addition, if you want to share a swap pair with a friend they have to select the relevant token pair as well, which is especially discouraging for first-time users.

### Solution 

`http://osmosis/?from=ION&to=OSMO`

Create a custom hook to sync the current Trade state's swap pairs with browser window's URL query params. The bulk of this functionality was achieved with the [`use-query-params`](https://github.com/pbeshai/use-query-params#readme) package.

Fixes issue #95


sunnya97:
Just tested it, works swimmingly!

PR #103, opened 2021-11-23
Cheaper IBC deposit fees

assafmo:
Reduced fees on IBC Deposit from Secret, Cosmos & Terra by setting more appropriate gas limits.

### DIFF
--- a/src/stores/root.ts
+++ b/src/stores/root.ts
@@ -102,7 +102,7 @@ export class RootStore {
 					if (chainInfo.chainId.startsWith('osmosis-')) {
 						gas = 1_350_000;
 					} else if (chainInfo.chainId.startsWith('secret-')) {
-						gas = 20_000;
+						gas = 25_000;
 					} else if (chainInfo.chainId.startsWith('columbus-') || chainInfo.chainId.startsWith('cosmoshub-')) {
 						gas = 100_000;
 					}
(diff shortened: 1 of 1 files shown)


ValarDragon:
err, how were these gas constants obtained? They are oddly, vastly different

assafmo:
Os Secret we have a lower gas setup, on other chains from observations. Why was 500k decided for other chains? Why is it 1.35M on Osmosis?

ValarDragon:
Oh, odd. I didn't notice it was 1.35M vs 500k earlier, my bad!

src/stores/root.ts
```diff
@@ -98,13 +98,18 @@ export class RootStore {
...
 				},
 				chainOpts: this.chainStore.chainInfos.map(chainInfo => {
+					let gas = 500_000;
+					if (chainInfo.chainId.startsWith('osmosis-')) {
+						gas = 1_350_000;
+					} else if (chainInfo.chainId.startsWith('secret-')) {
+						gas = 25_000;
+					} else if (chainInfo.chainId.startsWith('columbus-') || chainInfo.chainId.startsWith('cosmoshub-')) {
+						gas = 100_000;
+					}
```
ValarDragon, line 108: Can we revert this clause? Then looks good to merge to me

assafmo: Yes, done. :+1:

ValarDragon (approved):
LGTM!

PR #112, opened 2021-12-02
Anticipate the lbp of comdex

Thunnini:
I expect that the pool id of comdex may be 600.
Please check that the chain info and channel info is correct and the pool id is correct.


src/stores/root.ts
```diff
@@ -229,6 +229,13 @@ export class RootStore {
 					spotPriceDestDenom: 'uosmo',
 					destCoinId: 'pool:uosmo',
 				},
+				{
```
Thunnini, line 232: The price of base coin (comdex) will be shown on the LBP section. And, it should be important to user who is considering buying comdex. So, please check that this is correct.

src/config.ts
```diff
@@ -48,16 +48,12 @@ export const PromotedLBPPoolIds: {
...
-	/*
-		{
-			poolId: '21',
-			name: 'Regen Network',
-			baseDenom: DenomHelper.ibcDenom([{ portId: 'transfer', channelId: 'channel-8' }], 'uregen'),
-			destDenom: DenomHelper.ibcDenom([{ portId: 'transfer', channelId: 'channel-0' }], 'uatom'),
-		},
-		 */
+	{
+		poolId: '600',
```
Thunnini, line 53: Please check that the pool is correct.

src/config.ts
```diff
@@ -458,6 +454,12 @@ export const IBCAssetInfos: {
 		destChannelId: 'channel-2',
 		coinMinimalDenom: 'boot',
 	},
+	{
```
Thunnini, line 457: Please check that the channel info is correct.

src/stores/root.ts
sunnya97: Is correct.

src/config.ts
sunnya97: Is correct.

src/config.ts
sunnya97: Is correct.

PR #125, opened 2021-12-08
osmosis-testnet-4

sunnya97:


ValarDragon:
(this should not be merged, but please leave open for a few days)

PR #136, opened 2021-12-11
Only allow token select drop down with > 1 token

jonator:
## Problem
It seems strange to be able to select from only one available token that is also one that is already selected.

Screen Shot 2021-12-11 at 5 12 57 PM

## Solution
In LBP scenarios and from pool detail view, it is cleaner to not allow the empty token select dropdown to be used in the swap clipboard.

Screen Shot 2021-12-11 at 4 55 27 PM

Additionally, I've removed the already selected token from the dropdown list.


sunnya97:
This looks great

sunnya97 (approved):

ValarDragon (approved):
LGTM

PR #147, opened 2021-12-20
Add Lum Network resources

lebascou:


src/config.ts
```diff
@@ -1517,4 +1524,42 @@ export const EmbedChainInfos: ChainInfoWithExplorer[] = [
...
+			{
+				coinDenom: 'LUM',
+				coinMinimalDenom: 'ulum',
+				coinDecimals: 6,
+				coinGeckoId: 'lum-network',
+				coinImageUrl: window.location.origin + '/public/assets/tokens/lum.svg',
+			},
+		],
+		coinType: 118,
+		features: ['stargate', 'ibc-transfer', 'no-legacy-stdTx', 'ibc-go'],
```
joeabbey, line 1562: Should there be gas price steps here?

Noticed they are present for $HUAHUA

https://github.com/osmosis-labs/osmosis-frontend/pull/146/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R1560-R1564

joeabbey (commented):
Just a comment about gas price steps...

Also, I assume pool creation and configuration also comes soon (given there's a 100 $OSMO cost that comes with that)

src/config.ts
sunnya97: Yes

lebascou: Do you need me to do something here? We have a min gas pice set by all validators to `minimum-gas-prices = "0.001ulum"`
Not sure on how to interpret this setting in your file though

joeabbey: Closing the loop, I think it just informs Keplr "reasonable" values for low/medium/high gas

PR #151, opened 2021-12-21
Add Juno/Ust (Pool 580) to trade router

UnityChaos:


ValarDragon:
I don't think we should do this, the pool has no liquidity atm $671

UnityChaos:
It got added to incentives. Without being part of the trade router it will essentially never get trade volume even when there is Juno/Ust activity. 

I guess if the trade router has a functional minimal pool size we should either factor that into governance, or fix it. How large would it have to be for you to expect it to work well?

ValarDragon:
Governance already voted to remove any pool with under $1k of liquidity: https://www.mintscan.io/osmosis/proposals/85

I think that should be a hard minimum. Governance doesn't really match my view on this (by incentivizing Juno/UST !)

Playing around with the deploy, seems like the router doesn't route to this by default due to slippage which is good! Probs best to leave it to Tony to opine. (I actually don't know how the front-end router works atm)

UnityChaos:
Hmm yeah forgot about prop 85... Okay, well I can make sure that it has more than $1k of liquidity pretty easily (tomorrow). Fine with not merging it for a while, just thought we should by Friday-ish so that it can actually collect a share of Juno/Ust trade volume and eventually get to the appropriate levels of incentives.

RowanUtrecht:
@ValarDragon This JUNØ/UST pool now has enough liquidity (100k +)
But the deploy still doesn't use it. Weird indeed, even with small amounts.

But what be would be against merging this one?

ValarDragon:
i think merging this is fine now then!

UnityChaos:
Lol guess it's good we didn't merge that yet... (uusd not uust)

UnityChaos:
@ValarDragon is there a way to make it redeploy so we can see if it actually works now?

ValarDragon:
the link gets edited every commit

RowanUtrecht:
Tested it,  the pool is now used

PR #172, opened 2022-01-05
injective

sunnya97:


JeremyParish69 (commented):
I believe this can be replaced with [#294](https://github.com/osmosis-labs/osmosis-frontend/pull/294)

PR #177, opened 2022-01-09
reorder

sunnya97:


RowanUtrecht:
@sunnya97 Is it possible to sort the asset page and the swap selection page the same way?

This PR was only for the Assets page, but the swap page (drop down selection) is still the old order. And for me it's confusing that both are different

sunnya97:
That's a good point.  I'm not sure.  @jonator @Thunnini do you know how to sync these?

PR #182, opened 2022-01-13
Added Desmos

sunnya97:


RiccardoM:
@sunnya97 Will this PR make Osmosis work with Desmos even if the assetlist PR (osmosis-labs/assetlists#78) is not yet merged, or do we need to wait for that first?

PR #195, opened 2022-01-18
band again

sunnya97:


RowanUtrecht:
@sunnya97
https://app.osmosis.zone/pool/626
I think this PR needs to be merged. Since 626 is live and we have Band available on Osmosis

sunnya97:
Oh, thanks for the bump!

PR #204, opened 2022-01-22
Rework coingecko credit

jonator:
More subtle design.


sunnya97:
Ooh, I like this a lot better

PR #215, opened 2022-01-26
OSMO price showing 3 digits

### LINKED ISSUE #194: OSMO price showing 3 digits
Hi,
I'm a QA engineer in my day job so I wanted to point something very small out to see if this is a bug on osmosis. It looks like we are only displaying 3 digits for the OSMO price on the pools page at the top of the page. For example right now the price of OSMO is under $10 so it displays nicely to the hundredths place. It's showing as $9.98

Yesterday when OSMO was over $10 we were only displaying the price to the tenths place. For example when the price was $10.26 it would only show as $10.2. I'd be happy to submit a PR to the repo if someone can send me a link.
Aaron

Link to OSMO reddit page where I posted this:
https://www.reddit.com/r/OsmosisLab/comments/s6eu7p/osmo_price_showing_3_digits/

MendezAndrewM:
Fixes #194 

Added a new method to the `PoolIntermediatePriceStore` class with witch the previously used, `calculatePrice` method, can easily be swapped. Optional params include the currency type and the number of decimal points to include in the result (defaults to 2 if not provided).

### DIFF
--- a/src/pages/pools/components/LabsOverview/OsmoPrice.tsx
+++ b/src/pages/pools/components/LabsOverview/OsmoPrice.tsx
@@ -7,7 +7,7 @@ import { useStore } from 'src/stores';
 export const OsmoPrice = observer(function OsmoPrice() {
 	const { chainStore, priceStore } = useStore();
 
-	const price = priceStore.calculatePrice(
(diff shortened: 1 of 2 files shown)


jonator (approved):
LGTM

Issue #203, opened 2022-01-21
Primitive UI components


jonator:
[Figma](https://www.figma.com/file/wQjMyxY0EnEk29gBzGDMe5/Osmosis-Component?node-id=3762%3A14625)

PR #216, opened 2022-01-26
Add primitive UI components

jonator:
Implements issue #203


jonator:
- [ ] Apply these components to any existing refactor code (w/out adding functionality)
- [ ] Remove demo page at pages/assets
- [ ] New slider
- [ ] Implement info elements ([Figma](https://www.figma.com/file/wQjMyxY0EnEk29gBzGDMe5/Osmosis-Component?node-id=3770%3A13992))
- [ ] Implement table ([Figma](https://www.figma.com/file/wQjMyxY0EnEk29gBzGDMe5/Osmosis-Component?node-id=4258%3A2050))

jonator:
This PR also fixes issue #223

packages/web/components/input/input-box.tsx
```diff
@@ -0,0 +1,101 @@
...
+  rightEntry?: boolean;
+  /** Will only render the first two. If `clearButton` is enabled, will show that as long as `currentValue !== ""`. */
+  labelButtons?: Button[];
+  /** Show a clear button when `currentValue !== ""`. */
+  clearButton?: boolean;
+}
+
+export const InputBox: FunctionComponent<Props> = ({
+  currentValue,
+  onChange,
```
delivan, line 27: here is the same opinion as mentioned above.

packages/web/components/input/search-box.tsx
```diff
@@ -0,0 +1,49 @@
...
+import classNames from "classnames";
+import { InputProps, Disableable, CustomClasses } from "../types";
+
+interface Props extends InputProps<string>, Disableable, CustomClasses {
+  state?: "error";
+}
+
+export const SearchBox: FunctionComponent<Props> = ({
+  currentValue,
+  onChange,
```
delivan, line 12: here is the same opinion as mentioned above.

delivan (changes requested):

package.json
```diff
@@ -18,6 +18,7 @@
     "build:clean": "yarn clean && yarn install --frozen-lockfile && lerna run build",
     "lint": "lerna run lint",
     "lint:fix": "lerna run lint:fix",
+    "prettier:all": "prettier 'packages/**/*.{ts,tsx,js,scss,css,json,md}' --write",
```
Thunnini, line 21: Isn't it the same as `lint:fix` with this?

packages/web/components/control/menu-dropdown.tsx
```diff
@@ -0,0 +1,41 @@
...
+    )}
+  >
+    {options.map(({ id, display }, index) => (
+      <span
+        className={classNames(
+          "px-2 cursor-pointer w-full text-ellipsis overflow-hidden hover:text-clip hover:bg-white-faint",
+          {
+            "bg-white-faint text-white-full": id === selectedOptionId,
+            "text-iconDefault": id !== selectedOptionId,
+            "rounded-t-[8px]": index === 0,
```
delivan, line 30: It would better use `rem` rather than `px` to manage size proportionally. Also In here, it would be better to use a predefined class in the tailwind like `rounded-t-lg`.

delivan (changes requested):

packages/web/components/buttons/button.tsx
```diff
@@ -1,29 +1,47 @@
...
+  color = "primary",
+  size = "md",
+  type = "block",
+  className,
+  disabled = false,
+  children,
+}) => (
+  <button
+    className={classNames(
+      "flex justify-center items-center rounded-lg text-base",
```
delivan, line 25: It needs to be added styles for hover and loading state.

delivan (changes requested):

packages/web/components/control/slider.module.css
```diff
@@ -0,0 +1,109 @@
+.slider {
```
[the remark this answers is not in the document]
jonator: Now I know to do this. Thanks!

packages/web/components/control/menu-dropdown.tsx
jonator: Actually this one is weird, the container needs to be 9px border radius, and the inner elements need to be 1 px less since they are 1px smaller with the container 1px border. Otherwise, we're left with a small gap on the corners of the inner elements.

jonator: See:
image

jonator: @victortokaito pointed this out. That screenshot is with tailwind `rounded-lg` I believe

delivan: @jonator I can't find the inner elements' border radius need to be 8px on figma. It looks to be auto layout.

delivan: @victortokaito Speaking of '9px', I prefer width, padding, border-radius etc's value(px) to be even number. In tailwind, There are predefined classes like `rounded-lg` and their value is almost even. If the value is odd number, developer should make new class and calculating `px` to `rem` is little bit tricky and weird(9px -> 0.5625rem).

@jonator How think about this?

jonator: I added the `lginset` class. LMK if this is fitting for using tailwind.

PR #217, opened 2022-01-26
Update CHEQ token gauges for pool 617

ankurdotb:
Added new gauges for CHEQ token pool 617

* Gauges for pool 617 begin Thursday (27 January)
* Updated link to official block explorer for cheqd network


sunnya97:
Thanks!

PR #218, opened 2022-01-26
grayscale coingecko logo

sunnya97:


jonator:
@sunnya97 we just merged #204 which removes the logo entirely

sunnya97:
Yeah, I know I saw.  I was just messing around to see if there's a way to have a gray tiny version of the logo, so it still flows.  The colored one stood out too much

PR #219, opened 2022-01-27
Added pool 617 for CHEQ and fixed coinDecimals

ankurdotb:
* Added pool configuration for CHEQ:ATOM pool 617
* Fixed coinDecimals in pool config 602

I'm creating this as a draft PR, as I'm not sure what the implications of changing coinDecimals is when changed in root.ts. Is this a visual-only change? The correct coinDecimals are in the chain config descriptions, but I don't know whether this could end up rebalancing pool weights, for instance.


sunnya97:
Still necessary?

PR #225, opened 2022-02-02
update helper message - Single Asset LP

zramsay:
When first coming across this feature, It is initially confusing + requires some digging to find out that it's a convenience feature for not having to bother with a swap when adding to a pool (rather than something else...)

with osmosis-labs/feature-requests#54 i think this helper could also be expanded / clarified


JeremyParish69:
Not bad, but still missing a period at the end of the second sentence.

How about something like this?:

"
Single Asset LP will incur price impact upon entry and is subject to Impermanent Loss.

A portion of the offered asset is swapped via this pool to obtain the exact appropriate ratio of Pool Catalysts, which will then be provided as liquidity into the pool.
"

Although I do make assumptions. Does this work for pools with more than two assets? Does it LP the _exact_ value? or could it leave token dust behind?

Issue #209, opened 2022-01-24
Modal component


delivan:
- Base component
  - Title
  - Close button
- Implement connect wallet modal 
- Implement Terms agreement modal

jonator:
Have you opened a PR? Please open a PR in order to move it from in progress to done. The automation should do it automatically and if it doesn't LMK. Thanks

PR #226, opened 2022-02-03
Add showing my lp share & Update modal component

delivan:
This PR fixes issue #209


packages/stores/src/currency-registrar/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ChainStore } from "@keplr-wallet/stores";
+
+export class LPCurrencyRegistrar<C extends ChainInfo = ChainInfo> {
+  constructor(protected readonly chainStore: ChainStore<C>) {
+    chainStore.addSetChainInfoHandler((chainInfoInner) => {
+      chainInfoInner.registerCurrencyRegistrar(this.registerLPCurrency);
+    });
+  }
+
+  protected readonly registerLPCurrency = (coinMinimalDenom: string) => {
```
jonator, line 11: Add a `Currency?` Type here since that seems to be what you're returning

jonator (approved):

PR #238, opened 2022-02-07
Add BOOT to priceStore

JeremyParish69:
Add BOOT to priceStore via pool 597. On assets page, there is no usd price under boot token.


sunnya97:
I think this was not done because it was causing issues due to the USD price of 1 boot being so small (fractions of a penny).  @jonator @Thunnini has this been fixed?

JeremyParish69:
Replaced with #333

PR #253, opened 2022-02-11
github action

faddat:
This adds a basic github action to the front end repository so that teams making pull requests will get immediate feedback.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,7 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    continue-on-error: true
     strategy:
       matrix:
         node: [ '14', '16', '17' ]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,22 @@
+name: build
+
+
+on: [push, pull_request]
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [ '14', '16', '17' ]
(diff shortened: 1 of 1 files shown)


faddat:
If we want to add anything else to this, please let me know. One thing that I realized from watching it run is that it’s looking for an osmosis instance on local host. We dealt with a lot of similar stuff while building blurt. I’m definitely interested in helping as much as I can to create an end and testing environment that can be reused with this and other front ends.

jonator:
@faddat I'd reach out to @czarcas7ic who is currently working on setting up a test net.

czarcas7ic:
This is a very interesting concept @faddat , I'm going to cc @daniel-farina as well since I know he would find this idea intriguing.

Only things that comes to my mind is the github action would also have to do the statesync workaround, in addition it feels like state sync is getting longer and longer (recently had one take up to an hour long). I think this might need its own investigating as well. Was going to reach out to you @faddat if you had any idea what could be going on since you said you had experienced something similar.

faddat:
Hi guys, 

so, OMM has gotten things started with someone from Russia, who will host in Russia.

We are hoping to host from Vietnam, as well, both of these localized.  So right now I am making a docker-compose so that we can all work on the web front end more easily, and yep, it will state sync

faddat:
Daniel,

You're right about state sync taking longer and longer, and I don't know why that is. I just found a little time now to work on the compose setup, I think this repo is the best place for it.

czarcas7ic:
Since v7 happens tomorrow which breaks state sync, I kind of thought we would table it until the wasm fixes are implemented (unless you feel otherwise)

PR #261, opened 2022-02-15
fix: Trade page's extra 0 in From amt

landiehan:


sunnya97:
@landiehan thanks for the PR!  Do you mind adding a screenshot to show the changed effect?

landiehan:
@sunnya97 Yes!
before:
before
after clicking the switch button:
after

PR #267, opened 2022-02-20
Change Crypto.org to Cronos

orfeas0:
As per the official re-branding.


JeremyParish69:
Not a big deal, but I believe this is technically incorrect. Although we may want to shift from describing the originating chain (e.g., Cronos chain vs Crpyto.org Chain) to describing the name of the token (e.g., Crypto.com Coin vs Cronos), that should be done at a system-wide level, and not by changing the 'chain name'. In this case, this CRO (Cronos) is not coming from the Cronos chain--these are coming from Crypto.org chain, and thus may result in confusion.

sunnya97:
Agreed with @JeremyParish69 

The name currently is the *chain* from which the token is originating.  Currently we only support deposits from Crypto.org chain, not Cronos

Once we add a way to display the asset name as well (not just chain name), we can update this.

Issue #281, opened 2022-02-28
Port & Implement Pool detail page


jonator:
* Configure next js to use /pool/[id] endpoint

PR #293, opened 2022-03-10
Refactor pool detail page (- pool actions)

jonator:
Closes: #281 

Does not include dialog pool actions.

Notable differences from pre-refactor
* Use loaders for each of the individual pool metrics
* Move some metrics/logic to the gamm pool share store
  * Add `getAvailableGammShareRatio`, `getShareAssets`, & `getShareLockedAssets`
  * Utilize `RatePretty` type for ratios
* `ExtraGaugeInPool` -> `PoolDetailExtraGaugeAllowList`
  * Only supports lists (simplifies code)
  * Not sure what the logic to filter and sum a set of gauge Ids in the old `ExtraGauge.tsx` component was for @Thunnini ??


packages/stores/src/queries/pools/pools.ts
```diff
@@ -69,6 +69,18 @@ export class ObservableQueryPools extends ObservableChainQuery<Pools> {
     }
   );
 
+  readonly poolExists: (id: string) => boolean | undefined = computedFn(
+    (id: string) => {
+      if (!this.response) {
+        return undefined;
+      }
+
+      return this.response.data.pools.find((raw) => raw.id === id)
```
delivan, line 78: It can be short like this: `this.response.data.pools.some((raw) => raw.id === id)`

packages/web/components/assets/generate-gradient.ts
```diff
@@ -0,0 +1,13 @@
+export default function generateGradient(num?: number) {
```
delivan, line 1: It would be better to pull this file from `components` and put other folder like `utils`.

jonator: Moved it to `components/utils.ts`

delivan (approved):

Issue #298, opened 2022-03-12
Server side rendering (fix Next.js build error)


jonator:
### Problem
Next.js runs as much react code on the server as possible before serving to clients. This achieves performance and SEO benefits. However, currently the root store (which depends on data in the Keplr extension) is initialized on the server, resulting in an error.

### Solution
We should init the store on the browser (probably by wrapping constructor in `useEffect` in `web/stores/index.tsx`). This may require us to update the pages to handle a `null` store being returned. We should try to render as much html as possible without the store in place. A `null` return from `useStore` should be treated the same as when data is loading (using placeholder UI).

jonator:
Will be addressed on Keplr-wallet side

PR #297, opened 2022-03-12
Port & Implement Assets page (- ibc transfer)

jonator:
Closes: #242 

* Added the osmosis account at `@osmosis-labs/stores/assets` but did not add to root store at `root.ts`
  * Includes message composition code
* Refactored IBC transfer logic, but did not implement (due to problem above)
* Tweak some components, namely pool-card and search-box
* Wrapped some `@osmosis-labs/stores` objects with an observable mobx class: `ObservableAssets` (at `packages/web/stores/assets` 
  * Pros: globally accessible memoized calculations 
  * Cons: Organization/architecture doesn't seem fit. Was considering writing a hook with `useMemo` called `useUserAssets` or the like, but wouldn't get the pro above.

PR #299, opened 2022-03-12
Server side rendering & fix nextjs build error

jonator:
Closes: #298 (see issue for details)

Blocked by PR #287 & PR #297.

### DIFF
--- a/packages/web/components/cards/incentivized-pool.tsx
+++ b/packages/web/components/cards/incentivized-pool.tsx
@@ -11,25 +11,25 @@ import { useRouter } from "next/router";
 export const IncentivizedPoolCard: FunctionComponent<{
   pool: ObservablePool;
 }> = observer(({ pool }) => {
-  const { chainStore, queriesOsmosisStore, priceStore } = useStore();
-
-  const chainInfo = chainStore.osmosis;
-  const queryOsmosis = queriesOsmosisStore.get(chainInfo.chainId);
-
+  const store = useStore();
   const router = useRouter();
-
   const deterministicInteger = useDeterministicIntegerFromString(pool.id);
 
-  const poolTVL = pool.computeTotalValueLocked(
-    priceStore,
-    priceStore.getFiatCurrency("usd")!
-  );
+  let poolTVL: string | undefined;
+  let apr: string | undefined;
 
-  const apr = queryOsmosis.queryIncentivizedPools.computeMostAPY(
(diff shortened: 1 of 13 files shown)


jonator:
@dogemos @victortokaito @delivan @Thunnini 

The refactor/nextjs build issue is ultimately fixed. However:

After investigating, it appears we won't have CloudFlare preview builds until we merge `refactor` to `master`.
* Our current nextjs/lerna setup on `refactor/*` branch(es) has the build output folder set to `packages/web/.next`
* From the [cloudflare settings page](https://dash.cloudflare.com/4a3d442ed4cfa3990a66d2065a8e81f5/pages/view/osmosis-frontend/settings/builds-deployments), it appears we can only set the output directory once per CloudFlare page (and GitHub repo)
  * This means, we're going to need to configure CloudFlare to the path above just before merging from `refactor` to `master`
  * A hacky workaround is we could potentially use [environment variables](https://dash.cloudflare.com/4a3d442ed4cfa3990a66d2065a8e81f5/pages/view/osmosis-frontend/settings/environment-variables) for testing the refactor branches as preview builds. Of course, this would break non-refactor preview builds/PRs.

delivan:
@jonator How about move from Cloudflare pages to Vercel? Vercel is compatible with nextjs project and provide good experience.

delivan:
Is the error you said `window is not defined`? Then, It's up to `@keplr-wallet/store`. not on this repo. Tony saids It's gonna fixed on next version. 
How about ignoring this issue until it is fixed? the build error also will be fixed on pools page PR.

jonator:
@delivan
* On moving to Vercel, I'm a fan. Seems we could get a lot of [advanced features out of the box](https://nextjs.org/docs/deployment#managed-nextjs-with-vercel) just by deploying to Vercel. @dogemos what are your thoughts on this ([here's an overview](https://vercel.com/workflow))? Seems we'd have the same workflow.
* It's true that the error is in `@keplr-wallet/store` and could be fixed there. My thought process was that since we know all of that code works well pre-refactor in the browser, then we should just have next.js only run that code in the browser. We would still get the benefits of statically rendered html for each of the pages, just there would not be data and there never will be as we need the Keplr wallet extension to get most data. I have considered the possibility of letting next.js load non-user data on the server, such as all pools & aprs, prices, etc. Is this what @Thunnini wants to do? Not sure what that would look like right now.

PR #300, opened 2022-03-13
Change custom withdraw warning from tooltip -> textbox

jonator:
### Problem

There have been complaints from the community support DAO that users have missed the warning and have deposited and lost funds to a CEX address.

image

### Solution

Change warning from tooltip to large and prominent text box below the address boxes.

Screen Shot 2022-03-12 at 8 57 53 PM


delivan (approved):

Alien901:
will you add a memo option too, that would solve any CEX problem!

image

PR #318, opened 2022-03-28
Correct coingeckoId for Cerberus & Switch pool endpoint Imperator

IbrarMakaveli:


JeremyParish69 (commented):
Not sure how to verify the pool financial data, but the cerberus CoinGecko Id LGTM.

PR #319, opened 2022-03-28
Add new BitSong gauges

angelorc:


JeremyParish69 (commented):
Can we remove the old gaugeIds?

angelorc:
Yes we can delete the old gauges

Il Sab 2 Apr 2022, 22:40 JeremyParish69 ***@***.***> ha
scritto:

jonator (approved):

PR #323, opened 2022-03-31
Merge fix change BTSG price reference

Giorgionocera:
Merge fixes to change the BTSG price reference for Osmosis, from pool 574 (ATOM/BTSG) to pool 573 (BTSG/OSMO).


JeremyParish69 (approved):
Looks good

PR #328, opened 2022-04-01
RAC CW20 Token to Osmosis Front End

racoon-supply-rac:
RAC CW20 Token to Osmosis Front End


JeremyParish69 (commented):
Changes seem correct, but this should be going to frontier first

PR #333, opened 2022-04-05
Add Bostrom CionGecko id

JeremyParish69:
Add Bostrom CoinGecko id
{"id":"bostrom","symbol":"boot","name":"Bostrom"},


JeremyParish69:
I can see an accurate price on the Assets page.

PR #335, opened 2022-04-05
Add BLOCK CW20 Juno Network from Marble DAO

the0developer:


sunnya97:
Let's repoint this to frontier

JeremyParish69:
replaced by #379

PR #339, opened 2022-04-05
Pools Page: Create Pool dialog

### LINKED ISSUE #330: Pools Page: Create Pool Multi-step Dialog
[Figma](https://www.figma.com/file/UXgL2cPiw5I4GALDOodfCy/Osmosis-Web-(Master-File)?node-id=449%3A30413)

jonator:
Closes #330


packages/stores/src/ui-config/index.ts
```diff
@@ -1,3 +1,6 @@
 export * from "./manage-liquidity";
 export * from "./amount-config";
+export * from "./create-pool";
 export * from "./fake-fee-config";
+
+export const CREATE_POOL_MAX_ASSETS = 3;
```
delivan, line 6: Why is the maximum number of assets 3? the current is 8.

packages/web/components/complex/pool/create/step1-set-ratios.tsx
```diff
@@ -0,0 +1,70 @@
...
+                  const currency = config.remainingSelectableCurrencies.find(
+                    (currency) => currency.coinDenom === token.coinDenom
+                  );
+                  if (currency) {
+                    amountConfig.setCurrency(currency);
+                  }
+                }}
+              />
+              <div className="flex items-center gap-2.5">
+                <InputBox
```
delivan, line 35: how about adding type props for input box? For example, it would be better for ux to be able to handle numbers only for `<input type="number" ... />`.

packages/web/components/complex/pool/create/step2-add-liquidity.tsx
```diff
@@ -0,0 +1,80 @@
...
+                      type="outline"
+                      size="xs"
+                      onClick={() => amountConfig.setFraction(1)}
+                    >
+                      MAX
+                    </Button>
+                  </div>
+                  <div className="flex place-content-end items-center gap-2.5">
+                    <InputBox
+                      className="w-44"
```
delivan, line 64: this `InputBox` also seems to need `type="number"`

packages/web/components/complex/pool/create/step3-confirm.tsx
```diff
@@ -0,0 +1,107 @@
...
+                );
+              }
+            )}
+          </div>
+        </div>
+        <div className="flex p-3.5 items-center place-content-between border border-white-faint rounded-2xl">
+          <span>Set Swap Fee</span>
+          <div className="flex items-center gap-4">
+            <InputBox
+              className="w-44"
```
delivan, line 84: this `InputBox` also seems to need `type="number"`

delivan:
If the swap fee is set to an invalid value and return to step 1, we cannot click next button.
스크린샷 2022-04-15 오후 5 20 56

delivan:
1. We can delegate to one validator only for superfluid staking because of chain logic. So, we should make like that on frontend side. In current version, If user already did superfluid staking to some validator, It doesn't show validator selection view. we also should do that on refactor version.
2. In current version, Intentionally we shows the validator that user already staked(not superfluid) at the top. It'd better to do that on refactor version.
3. If we make a difference between the background color when user do mouse hover on validator and the background of the validator that user already staked, it'll be more clear.

delivan (approved):

delivan:
LGTM 🚀

PR #343, opened 2022-04-06
Use `MemoryKVStore` on nodejs environment (on build)

Thunnini:
We can't access `window` on nodejs environment because there is no `window` global object on nodejs.
It makes "window is not defined" error on execution.
To avoid this problem, we should avoid the access to `window` by checking the typeof `window`.
Ex)
```typescript
if (typeof window === "undefined") {
  // In server-side (nodejs), use memory kv store (volatile kv store).
  return new MemoryKVStore(prefix);
}
return new LocalKVStore(prefix);
```
After keplr-wallet@v0.9.18, `getKeplrFromWindow()` function checks the window object before access like above way, and it returns `undefined` if `typeof window === "undefined"`

On osmosis frontend, we should avoid the more usage of `window` access, and the main points are usages of kv stores. `IndexedDBKVStore` and `LocalKVStore` requires the browser environment and it accesses `window`. To avoid this, use the `MemoryKVStore` on nodejs environment on build time. `MemoryKVStore` is volatile kv store and the main aim of that store is testing. However, we doesn't support the server-side rendering and it should be enough to generate pre-rendered build output.

### DIFF
--- a/packages/web/stores/root.ts
+++ b/packages/web/stores/root.ts
@@ -10,7 +10,12 @@ import {
   QueriesStore,
 } from "@keplr-wallet/stores";
 import { EmbedChainInfos, IBCAssetInfos } from "../config";
-import { IndexedDBKVStore, LocalKVStore } from "@keplr-wallet/common";
+import {
+  IndexedDBKVStore,
+  KVStore,
+  LocalKVStore,
+  MemoryKVStore,
+} from "@keplr-wallet/common";
 import EventEmitter from "eventemitter3";
 import { ChainStore, ChainInfoWithExplorer } from "./chain";
 import {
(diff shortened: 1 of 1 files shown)


jonator (approved):
LGTM

Thunnini:
@jonator keplr-wallet libraries are built for the extensions, thus ssr has not been taken into account. Also, I do not have plans as of right now to implement ssr.
However, if it were to be implemented, I would pass on the server side response as props to the client in the create root store stage and create a logic where they are initialized in the observable query class. I haven’t tried this myself so there’s no guarantee that this approach would work

jonator:
@Thunnini  
So perhaps we could create methods (from a SSR interface) on the non-user-data stores to trigger an axios load and return the result? Then we pass those as props to the page components which can init the stores with the SSR-fetched data, thus saving them an axios request on browser load.. I'll create an issue.

PR #356, opened 2022-04-08
reactivate juno

sunnya97:


sunnya97:
Wait, was this tested to work?

sunnya97:
Yeah, I think the deposits/withdrawals still weren't working properly

PR #361, opened 2022-04-08
Unactivate juno frontier

sunnya97:


jonator:
@sunnya97 Why was this done?

sunnya97:
Because the channels were clogged and relayers weren't back up and running yet

PR #363, opened 2022-04-09
Add Fetch to Osmosis frontend

Galadrin:
Signed-off-by: David Pierret <dapie@cros-nest.com>


JeremyParish69:
Pending initial liquidity + adding pool to trade router

JeremyParish69:
Replaced with #376

PR #390, opened 2022-04-15
Frontier - add Axelar pools

milapsheth:


JeremyParish69:
Pool incentives look good. Trade router works as expected. LGTM!

PR #400, opened 2022-04-18
Frontier - Add HASH

Taztingo:


src/config.ts
```diff
@@ -3035,4 +3041,39 @@ export const EmbedChainInfos: ChainInfoWithExplorer[] = [
...
+		stakeCurrency: {
+			coinDenom: 'hash',
+			coinMinimalDenom: 'nhash',
+			coinDecimals: 9,
+			coinImageUrl: window.location.origin + '/public/assets/tokens/hash.svg',
+		},
+		bip44: {
+			coinType: 118,
+		},
+		bech32Config: Bech32Address.defaultBech32Config('pb'), // Please check
```
JeremyParish69, line 3058: 'pb' agrees with the registration here: https://github.com/satoshilabs/slips/blob/master/slip-0173.md

src/config.ts
```diff
@@ -3035,4 +3041,39 @@ export const EmbedChainInfos: ChainInfoWithExplorer[] = [
...
+			{
+				coinDenom: 'hash',
+				coinMinimalDenom: 'nhash',
+				coinDecimals: 9,
+				coinImageUrl: window.location.origin + '/public/assets/tokens/hash.svg',
+			},
+		],
+		coinType: 118,
+		features: ['stargate', 'ibc-transfer', 'no-legacy-stdTx', 'ibc-go'],
+		explorerUrlToTx: 'https://explorer.provenance.io/tx/${txHash}',
```
JeremyParish69, line 3077: Please remove the $ dollar sign

src/config.ts
```diff
@@ -3035,4 +3041,39 @@ export const EmbedChainInfos: ChainInfoWithExplorer[] = [
...
+		],
+		feeCurrencies: [
+			{
+				coinDenom: 'hash',
+				coinMinimalDenom: 'nhash',
+				coinDecimals: 9,
+				coinImageUrl: window.location.origin + '/public/assets/tokens/hash.svg',
+			},
+		],
+		coinType: 118,
```
JeremyParish69, line 3075: don't need CoinType this here since it's already specified above.
Can we please add the gas price step?

JeremyParish69:
Pending changes, initial liquidity, pool for price data + router.

JeremyParish69 (changes requested):
Please address earlier comments, add initial liquidity, then define price data + add pool to router.

src/config.ts
```diff
@@ -3035,4 +3041,43 @@ export const EmbedChainInfos: ChainInfoWithExplorer[] = [
...
+		feeCurrencies: [
+			{
+				coinDenom: 'HASH',
+				coinMinimalDenom: 'nhash',
+				coinDecimals: 9,
+				coinImageUrl: window.location.origin + '/public/assets/tokens/hash.svg',
+			},
+		],
+		gasPriceStep: {
+			low: 1905,
```
JeremyParish69, line 3076: note, this is in coinDenom (so, HASH), not minimalCoinDenom (nhash). So gas will cost thousands of HASH tokens? I suspect you may mean to put 0.000001905, etc.

JeremyParish69:
I see that alternative coin Id 'pool:nhash' was defined, but this id isn't being used. Please define this id as the 'coinGeckoId' for each time it's listed as a currency (stake, currency, fee) under EmbedChainInfos in config.ts.

Taztingo:
We are not listed on coin gecko. Should I remove alternativeCoinId?

src/stores/root.ts
```diff
@@ -1629,6 +1636,21 @@ export class RootStore {
 					},
 				],
 			},
+			{
+				poolId: '692',
```
iramiller, line 1640:
```suggestion
				poolId: '693',
```
Recommend switching to pool 693 as it has a much lower swap fee.

Taztingo: Sure, that's a good idea.

PR #407, opened 2022-04-20
Frontier - Add MOBIX token

Galadrin:
- Token image is already existing
- coingecko id need the pool to be created


Adamjeean:
Scam project. Why not listing?

jonator:
We are no longer deploying frontier from `frontier` branch

Issue #349, opened 2022-04-07
Pool Detail Page: Swap Dialog


jonator:
Blocked by Issue #144

[Figma](https://www.figma.com/file/UXgL2cPiw5I4GALDOodfCy/Osmosis-Web-(Master-File)?node-id=516%3A22321)

Issue #240, opened 2022-02-07
Too many numbers cause an error and site crash


machinsoft:
Throws an error when there are too many digits

![img](https://i.ibb.co/p0C4jZQ/i01-1.jpg)

jonator:
Hey! Thanks for finding this. If you have the capacity, we'd love to see a pull request to address this!

machinsoft:
I did not make a pool request, but as an option, you can simply limit the number of digits and the problem will disappear.

jonator:
On `optimized-routes` branch? I'll check now

PR #410, opened 2022-04-25
Improve main page on refactor

Thunnini:
What has been implemented
- Clipboard component on main page
- Clipboard component has more ability on token search
- Token list on clipboard shows the current balances and prioritize non zero balance token
- Support simple multihop like main

TODO
- Button should reflect the warning/error/loading state
- Swap tx should notify pending/success or failure


packages/pools/src/interface.ts
```diff
@@ -16,11 +16,13 @@ export interface Pool {
     amount: Int;
     weight: Int;
   }[];
+  get poolAssetDenoms(): string[];
```
jonator, line 19: IMO, we don't need this getter. It's easy enough to just `map(asset => asset.coinDenom)`

packages/pools/src/common.ts
```diff
@@ -0,0 +1,30 @@
+import { Dec, Int } from "@keplr-wallet/unit";
```
jonator, line 1: Regarding this file, have you looked at `weighted.ts` in the `/math` package? Perhaps this would fit well there.

packages/pools/src/routes.ts
```diff
@@ -0,0 +1,387 @@
...
+        break;
+      }
+    }
+
+    // No enough liquidity
+    if (totalLimitAmount.lt(tokenIn.amount)) {
+      throw new NotEnoughLiquidityError();
+    }
+
+    // TODO: ...
```
jonator, line 260: Please remove or clarify this comment

packages/pools/src/routes.ts
```diff
@@ -0,0 +1,387 @@
...
+    effectivePriceInOverOut: Dec;
+    effectivePriceOutOverIn: Dec;
+    swapFee: Dec;
+    slippage: Dec;
+  } {
+    if (paths.length === 0) {
+      throw new Error("Paths are empty");
+    }
+
+    let totalOutAmount: Int = new Int(0);
```
jonator, line 285: Could this logic fit in the maths package? Perhaps a static `Routing` class that does math related to routes? And accepting primitive unit types as parameters. Just an idea.

packages/stores/src/account/index.ts
```diff
@@ -485,9 +485,12 @@ export class OsmosisAccountImpl {
               tokenOutCurrency.coinMinimalDenom
             );
 
+            console.log(pool.swapFee);
```
jonator, line 488: Please remove log

packages/stores/src/queries/pools/pools.ts
```diff
@@ -128,6 +145,9 @@ export class ObservableQueryPools extends ObservableChainQuery<Pools> {
     }
   );
 
+  /**
+   * @deprecated
+   */
```
jonator, line 150: Since we haven't published, let's just remove instead of deprecate.

packages/web/hooks/use-slippage-config/index.ts
```diff
@@ -0,0 +1,148 @@
+import { useState } from "react";
+import { Dec, DecUtils, RatePretty } from "@keplr-wallet/unit";
+import { action, computed, makeObservable, observable } from "mobx";
+import { computedFn } from "mobx-utils";
+
+export class SlippageConfig {
```
jonator, line 6: In the stores package I created a `ui-config` folder that contains mobx objects that are for storing UI state. This class would be a good fit for that (but not the hook)

packages/web/hooks/use-slippage-config/index.ts
```diff
@@ -0,0 +1,148 @@
...
+        return new Error("Invalid slippage");
+      }
+    }
+
+    return;
+  });
+}
+
+// CONTRACT: Use with `observer`
+export const useSlippageConfig = () => {
```
jonator, line 144: Using a hook like this to keep the state instances is a better idea then what I do on the pool detail page with `useMemo`. What if we create a more generic `useConfig` hook that accept any UI config obect and will `useMemo` to keep it there?

packages/web/hooks/use-trade-token-in-config/index.ts
```diff
@@ -0,0 +1,323 @@
...
+import {
+  CoinPretty,
+  Dec,
+  DecUtils,
+  Int,
+  IntPretty,
+  RatePretty,
+} from "@keplr-wallet/unit";
+
+export class TradeTokenInConfig extends AmountConfig {
```
jonator, line 20: Same comments as in `use-slippage-config`

packages/web/components/token-select.tsx
```diff
@@ -0,0 +1,288 @@
...
+import { AppCurrency } from "@keplr-wallet/types";
+import { observer } from "mobx-react-lite";
+import { useStore } from "../stores";
+import classNames from "classnames";
+import { useBooleanWithWindowEvent } from "../hooks";
+
+/**
+ * TokenSelect's dropdown is attached to the nearest "relative" element.
+ */
+export const TokenSelect: FunctionComponent<{
```
jonator, line 18: Have you seen my version of [Token Select](https://github.com/osmosis-labs/osmosis-frontend/blob/jonator/refactor/mobile-foundation/packages/web/components/control/token-select.tsx)  on the `/mobile-foundation` branch? It provides the option for showing balances via `CoinPretty` objects. Also, it uses `useFilteredData` to cleanly operate the search bar.

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -0,0 +1,481 @@
...
+import { Pool } from "@osmosis-labs/pools";
+import { observer } from "mobx-react-lite";
+import { useStore } from "../../stores";
+import AutosizeInput from "react-input-autosize";
+import { Currency } from "@keplr-wallet/types";
+
+export const TradeClipboard = observer<
+  {
+    containerClassName?: string;
+    containerStyle?: React.CSSProperties;
```
jonator, line 19: Should we use `ClassNames` for passing tailwind classes instead?

packages/web/pages/index.tsx
```diff
@@ -1,18 +1,92 @@
...
+        width: number;
+        height: number;
+      }
+    | undefined
+  >(undefined);
+  const [clipboardWidth, setClipboardWidth] = useState<number | undefined>(
+    undefined
+  );
+
+  useEffect(() => {
```
jonator, line 24: In the `mobile-foundation` branch, I created a `useWindowSize()` hook that may serve the needs we have here. Perhaps add a TODO to utilize that instead.

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -0,0 +1,481 @@
...
+
+          <div className="relative flex justify-end w-full h-11 mb-[1.125rem]">
+            <button
+              className="relative"
+              onClick={(e) => {
+                e.stopPropagation();
+                setIsSettingOpen(!isSettingOpen);
+              }}
+            >
+              <img
```
jonator, line 125: Should we use `next/image` here?

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -0,0 +1,481 @@
...
+                onClick={(e) => e.stopPropagation()}
+              >
+                <div className="subtitle1 text-white-emphasis">
+                  Transaction Settings
+                </div>
+                <div className="flex items-center mt-2.5">
+                  <div className="body2 text-white-disabled mr-2">
+                    Slippage tolerance
+                  </div>
+                  <Tippy
```
jonator, line 150: We do have a `<Tooltip />` component that may serve your needs here.

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -0,0 +1,481 @@
...
+                    className={classNames(
+                      "text-white-full text-xs py-1 px-1.5 rounded-md ml-2",
+                      tradeTokenInConfig.fraction === 1
+                        ? "bg-primary-200"
+                        : "bg-white-faint"
+                    )}
+                    onClick={(e) => {
+                      e.preventDefault();
+
+                      if (tradeTokenInConfig.fraction !== 1) {
```
jonator, line 250: Seems like common functionality, we could have a `toggleFraction()` method on the config object?

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -0,0 +1,481 @@
...
+              <div className="subtitle2 text-white-high">
+                Estimated Slippage
+              </div>
+              <div className="subtitle2 text-white-high">
+                {tradeTokenInConfig.expectedSwapResult.slippage.toString()}
+              </div>
+            </div>
+          </div>
+
+          {/* TODO: Styling warning/error case. Render loading state. Handle rejection. */}
```
jonator, line 402: Note: we do have an `<Error />` component that contains the style you need.

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -0,0 +1,481 @@
...
+          <button
+            className="mt-[1.125rem] flex justify-center items-center w-full h-[3.75rem] rounded-lg bg-primary-200 text-white-full text-base font-medium shadow-md"
+            disabled={
+              tradeTokenInConfig.error != null &&
+              tradeTokenInConfig.optimizedRoutePaths.length > 0
+            }
+            onClick={async (e) => {
+              e.preventDefault();
+
+              if (tradeTokenInConfig.optimizedRoutePaths.length > 0) {
```
jonator, line 412: Seems like common logic, should this live on a method on `tradeTokenInConfig`'s class?

packages/web/pages/index.tsx
```diff
@@ -34,7 +108,24 @@ const Home: NextPage = observer(function () {
...
           </g>
         </svg>
-        <TradeClipboard className="absolute left-[min(calc(920*(100vh/1080)),calc((100vw_-_206px)*0.8_-_520px))]" />
+      </div>
+      <div className="absolute w-full h-full flex items-center overflow-x-hidden overflow-y-auto">
+        <TradeClipboard
+          ref={clipboardRef}
+          containerClassName="w-full max-w-[32.5rem]"
+          containerStyle={{
+            ["--clipboardMinLeft" as any]: `calc(${clipboardMinLeftPx} * (100vh / 1080))`,
```
jonator, line 117: Is tailwind not able to achieve this justification?

jonator (changes requested):

jonator:
Also- would be nice to add an auto resize option (from `react-input-autosize` to the `<InputBox />` component so others may benefit in the future.

packages/pools/src/common.ts
Thunnini: Yeah Agreed.

packages/pools/src/routes.ts
Thunnini: This was to be added / modified after this PR. Keeping this for now.

packages/pools/src/routes.ts
Thunnini: I don't have a clear design on the final architecture. This is a temporary design I'm trying out, there's high chance that this is going to change in the future.

packages/stores/src/account/index.ts
Thunnini: My bad!

packages/stores/src/queries/pools/pools.ts
Thunnini: This was intended to change before publishing. Put deprecated as a note for me to modify in the future.

packages/web/hooks/use-slippage-config/index.ts
Thunnini: Agreed. Didn't notice that there was such folder. Will take a look.

packages/web/hooks/use-slippage-config/index.ts
Thunnini: The reason `useState` is used here instead of `useMemo` was because `useMemo` requires deps. We also had the option to pass in a empty array of deps but then it would show a warning. I wanted to make it explicit that we do not need to pass deps. 

For the generic hook approach, I think it's over engineering tbh

packages/web/hooks/use-trade-token-in-config/index.ts
Thunnini: Ditto

packages/web/components/token-select.tsx
Thunnini: Would take a look.

packages/web/components/trade-clipboard/index.tsx
Thunnini: I made this bc there were certain styles that cannot be passed on using tailwind.

packages/web/components/trade-clipboard/index.tsx
Thunnini: I'll take a look to see if I could possibly change this.

packages/web/components/trade-clipboard/index.tsx
Thunnini: Agreed.

packages/web/components/trade-clipboard/index.tsx
Thunnini: `fraction` accepts any value between 0 and 1. This code itself isn't much repeated imo, I think we could leave it like this.

packages/web/pages/index.tsx
Thunnini: Let's leave this problem for latter discussions.

jonator:
@Thunnini Could you also implement #240 for the refactor?

jonator:
- [ ] Also make sure to add URL params. Double check that they are still there when navigating from another page via the menu.

jonator:
@delivan think you could also add mobile since we have this channel open? ^ Please request my review when done! Thanks

jonator:
#349 Please also add the dialog to the pool detail page in this PR

delivan (commented):
@jonator feel free to review!(I can't see the request review button)

jonator (changes requested):
- [x] Seems there's still an integer overflow error
Screen Shot 2022-05-15 at 12 23 58 PM

- [x] (Minor) Could we also remove the ability to select tokens if in a pool detail swap tokens dialog? This is optional (or post refactor) because you'd need to handle pools with >2 assets.
Screen Shot 2022-05-15 at 12 26 30 PM

- [x] Consider catching the exception from Keplr should a user choose to reject a transaction.
Screen Shot 2022-05-15 at 5 56 03 PM

- [x] Let's disable the swap button if trading amount is 0

- [x] As with the old app, let's make button & text red when slippage is > 10(?)%
Screen Shot 2022-05-15 at 6 11 57 PM

- [x] Let's include the network name under the denoms. This is critical for bridge assets.
Screen Shot 2022-05-15 at 6 21 23 PM


### [Mobile](https://www.figma.com/file/wQjMyxY0EnEk29gBzGDMe5/Osmosis-Component?node-id=4449%3A2029)
- [x] "Estimated slippage" & % should have `wireframes-lightGrey` color (+ all text in that box)

- [x] "Available" should still be present on mobile
Screen Shot 2022-05-15 at 6 22 07 PM

Overall, great work! Getting super close.

jonator:
Also @delivan make sure that this is accounted for => https://github.com/osmosis-labs/osmosis-frontend/issues/501

jonator:
@delivan noticed another issue 

- [ ] The URL query params don't work. They seem to only navigate to the default ATOM->OSMO pair.

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -0,0 +1,635 @@
...
+      // Whenever the setting opened, give a focus to the input if the manual slippage setting mode is on.
+      manualSlippageInputRef.current?.focus();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isSettingOpen]);
+
+  const router = useRouter();
+  const firstQueryEffectChecker = useRef(false);
+
+  useEffect(() => {
```
jonator, line 85: Let's create a `hooks/query-params` folder and put this logic in `useTokenQueryParams`

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -0,0 +1,635 @@
...
+  const tradeTokenInConfig = useTradeTokenInConfig(
+    chainStore,
+    queriesStore,
+    chainStore.osmosis.chainId,
+    account.bech32Address,
+    undefined,
+    pools
+  );
+
+  const availableBalance = tradeTokenInConfig
```
jonator, line 57: Let's wrap this and below in `useMemo`. I've found w/ mobx, as queries load data from API the render fn runs quite often.

packages/stores/src/ui-config/trade-token-in-config.ts
```diff
@@ -0,0 +1,327 @@
...
+  ) {
+    super(chainGetter, queriesStore, initialChainId, sender, feeConfig);
+
+    this._pools = pools;
+
+    makeObservable(this);
+  }
+
+  @action
+  setPools(pools: Pool[]) {
```
jonator, line 50: Related to `use-token-in-config`, we should remove these simple `set*` methods and have everything be passed into the constructor instead.

If there's a reason to change these core objects that are in the constructor later in the objects lifecycle, LMK and we can keep them.

jonator (changes requested):
- [ ] When scaling the page down, the clipboard starts to get cut off. Seems like it should be fixed to right edge of window.
Screen Shot 2022-05-18 at 10 59 38 AM

- [ ] Remove `packages/web/components/token-select.tsx`. It's unused.
- [ ] For some reason, the "Available" label is cutting off the full balance. I'd set `maxDecimals` to the number configured on the token.
- [ ] The overflow issue is fixed, but for some reason it seems the estimates give up and go to zero after a certain large amount? May be problematic for large amounts of cheap tokens
- [ ] It seems on mobile the "Estimated Slippage" is faint white but not the "x%" value
- [ ] If there's no multihop (1 route) could we send the `sendSwapExactAmountInMsg` message on that one pool instead? It's just more semantically correct

jonator (approved):

PR #417, opened 2022-04-26
Frontier - Update BLOCK to be after MARBLE

Marble-Block:


Marble-Block:
Also, added external incentives.

src/config.ts
```diff
@@ -76,6 +76,7 @@ export const LockupAbledPoolIds: {
 	'670': true,
 	'678': true,
 	'679': true,
+	'691': true,
```
JeremyParish69, line 79: Can we please put the External incentive gauge in a different PR?

Marble-Block: Sure thing, removed externals from this PR.

JeremyParish69:
Block will not be moved up. External Incentives can be added in a new PR

PR #430, opened 2022-04-28
Remove dupe grav b embed

JeremyParish69:
Remove duplicate gravity bridge object under EmbedChainInfos. Removed the one where the bridged assets don't have the g- prefix for the symbol.


JeremyParish69:
@sunnya97 in this PR ( #354 ) you added a new object for Gravity Bridge under EmbedChainInfos but didn't modify the existing one. Are we OK to remove the old one without the prefixes?

PR #439, opened 2022-05-02
Main - Add usdc and usdc/osmo pool

JeremyParish69:
Add usdc and usdc/osmo pool + external incentives


jonator:
@JeremyParish69 seems to be some merge conflicts

PR #440, opened 2022-05-02
add crbrus to main

JeremyParish69:
add Cerberus CRBRUS token and osmo/crbrus pool to main site


cerberus-zone:
Our pool is no longer showing external incentives https://app.osmosis.zone/pool/662 but there is 147 epochs left. Anyway for this to be fixed. Thanks

PR #444, opened 2022-05-03
Update pool 625 guages

jkilpatr:
This pool's last guage has expired and two new gauges have been created

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -571,7 +571,11 @@ export const ExtraGaugeInPool: {
 	],
 	'625': [
 		{
-			gaugeId: '2511',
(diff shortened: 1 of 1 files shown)


JeremyParish69:
LGTM. Should we also publish the same change to Frontier branch?

jkilpatr:
> LGTM. Should we also publish the same change to Frontier branch?

As far as I know this grav/osmo pool is not a frontier pool and shouldn't be one.

If we where updating incentives on any gAsset pool then we would need to update frontier but there are no other guages to update at this time.

So I think we're good to go.

PR #445, opened 2022-05-04
Update LIKE config with new prefix and no-legacy-stdTx

williamchong:
We are seeing `Request failed with status code 501` when depositing LIKE to Osmosis after upgrading our chain with sdk 0.44 and dual prefix support. Would this PR fix the issue?

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1593,7 +1593,7 @@ export const EmbedChainInfos: ChainInfoWithExplorer[] = [
 		bip44: {
 			coinType: 118,
 		},
-		bech32Config: Bech32Address.defaultBech32Config('cosmos'),
+		bech32Config: Bech32Address.defaultBech32Config('like'),
 		currencies: [
 			{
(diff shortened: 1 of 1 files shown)


jonator:
Looks like you may need to open a duplicate PR for updating frontier as well

williamchong:
Sure, made https://github.com/osmosis-labs/osmosis-frontend/pull/446 and thanks for the reminder

JeremyParish69:
Tested this change. Works without removing LikeCoin from Keplr, but the cosmos... prefix remains. Removing LikeCoin from Keplr and re-adding updates your address to like... prefix. BigDipper Explorer doesn't seem to work yet.

JeremyParish69 (approved):

JeremyParish69:
Please update the bech32 prefix in the chain registry
https://github.com/cosmos/chain-registry/blob/master/likecoin/chain.json

PR #450, opened 2022-05-05
Add CRBRUS 662 External Incentives

JeremyParish69:
Add CRBRUS 662 External Incentives
Cerberus was re-added to main because it became incentivized, but only the asset and pool was added, and the incentive gauge was forgotten.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -68,6 +68,7 @@ export const LockupAbledPoolIds: {
 	'643': true,
 	'648': true,
 	'651': true,
+	'662': true,
 };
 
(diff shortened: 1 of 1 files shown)


cerberus-zone:
> Add CRBRUS 662 External Incentives Cerberus was re-added to main because it became incentivized, but only the asset and pool was added, and the incentive gauge was forgotten.

Also, one other thing that was noticed is Cerberus is not showing up in the "App" info/stats site here https://info.osmosis.zone. Looks like this may have been forgotten when it was moved back over from Frontier. Would you be able to help with that? Thanks, we appreciate the help.

PR #461, opened 2022-05-07
Add functionality to support toggling of zero balances showing up in the asset list

Njerschow:
Add a button to toggle showing and hiding coins with 0 balance:

Toggled:
image

Untoggled:
image

### DIFF
--- a/src/pages/assets/AssetBalancesList.tsx
+++ b/src/pages/assets/AssetBalancesList.tsx
@@ -1,7 +1,7 @@
 import styled from '@emotion/styled';
 import { AppCurrency, IBCCurrency } from '@keplr-wallet/types';
 import { observer } from 'mobx-react-lite';
-import React from 'react';
+import React, { useState } from 'react';
 import { Img } from 'src/components/common/Img';
 import { ButtonFaint, ButtonSecondary } from 'src/components/layouts/Buttons';
 import { Text, TitleText } from 'src/components/Texts';
@@ -21,6 +21,8 @@ export const AssetBalancesList = observer(function AssetBalancesList() {
 	const { chainStore, queriesStore, accountStore, priceStore } = useStore();
 
 	const { isMobileView } = useWindowSize();
+
+	const [showZeroBalances, setShowZeroBalances] = useState<boolean>(false); 
 
 	const account = accountStore.getAccount(chainStore.current.chainId);
 	const queries = queriesStore.get(chainStore.current.chainId);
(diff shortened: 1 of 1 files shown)


jonator:
@victortokaito thoughts on design?

Miniminitrue:
> @victortokaito thoughts on design?

Make it a symbol to save space (eye open/closed for example).

jonator:
@Miniminitrue for the refactor we include this functionality. We use a simple check box to implement this.

This may not be merged since we are very close to completing the refactor.

dylanschultzie:
This is great, looking forward to this.

Njerschow:
@jonator shall I close this then? would love to see it merged asap - I can add eye open/closed instead of button if it means it will be merged.

victortokaito:
Because the suggested change uses a button design, it becomes the dominant call-to-action here (draws more attention than the deposit/withdraw actions). We'd also have to consider how it is positioned relative to the "search" and "sort by" features. I'd like to revisit this as I think this may be a nice place to consider adding a general deposit button as we've discussed briefly.

> Make it a symbol to save space (eye open/closed for example).

This would be nice to have just an icon, but it may not be immediately apparent what is being toggled to appear/disappear. I would suggest an accompanying text description to join the icon, but this will take up as much room as what we have implemented now. 

Screen Shot 2022-05-10 at 1 45 37 PM

victortokaito:
Also, keep in mind how this might work on the mobile version.

ankurdotb:
> Because the suggested change uses a button design, it becomes the dominant call-to-action here (draws more attention than the deposit/withdraw actions). We'd also have to consider how it is positioned relative to the "search" and "sort by" features.

I agree with the sentiment here. Can I suggest maybe using a "[slider toggle](https://v5.material.angular.io/components/slide-toggle/overview)" (the linked example is Angular, but I wanted to just illustrate a point) as that gives a clean and compact way of toggling zero balance displays on/off vs having a button.

victortokaito:
Hi ankurdotb, thanks for bringing this up.

We have a slider toggle implemented on the mobile version of the UI update. For mobile, we wanted to make use of the perceived swiping affordance, but went this route on the desktop version to introduce fewer UI components (the checkbox is used in several other areas). 

@jonator, I can see how changing "hide zero balances" to a slider toggle could help bring the mobile more closely in line with the desktop. Side note: Switches and checkboxes are both used to enable settings. But the immediacy of when users expect those settings to activate are different. And the slider toggle has more of an instantaneous feeling.

ankurdotb:
> And the slider toggle has more of an instantaneous feeling.

Yes, I agree with you on this. It has a more "instantaneous" feel like you said. I associate checkboxes with needing to click a button to "submit" and then carry out an action, since checkboxes are more common in forms where a box might be ticked / unticked without any action happening until an explicit accept/submit is done.

But these are minor differences really. 

The other switch I'd do is bump the checkbox or slider to the top, so that it's above the fold/doesn't require me to scroll too much to get to it.

jonator:
https://refactor.osmosis.zone

PR #476, opened 2022-05-10
[Frontier] Add Another.Software Validator Token

angrymouse:


ankurdotb (changes requested):
@angrymouse I can't see the ASVT token on the [CW20 asset listi on Mintscan](https://www.mintscan.io/juno/assets) yet, which is likely because I don't see the token on the [Cosmos chain registry assetlist for Juno](https://github.com/cosmos/chain-registry/blob/master/juno/assetlist.json). Please add your asset/token to the Cosmos chain registry first, as a lot of Cosmos apps like the Osmosis frontend and others rely on this as the canonical asset list in their apps.

You could set this PR to draft while the PR upstream on chain registry goes through the merge/approval process, and then come back to this one to add to Frontier.

angrymouse:
Thank you! WIll do, don't close PR yet

JeremyParish69:
Can we try to get some liquidity added before listing?

angrymouse:
> Can we try to get some liquidity added before listing?

Like create first pool and add first liquidity here?

angrymouse:
How can we do it before listing? Is there another UIs or should we do through cosmjs?

angrymouse:
Ah, yes, clone this repo and add to config locally, then create pool

JeremyParish69:
> How can we do it before listing? Is there another UIs or should we do through cosmjs?

You can also use the preview from the vercel deployment link above.

angrymouse:
Conflict of changes for merging, will re-open in a day with created pool

PR #482, opened 2022-05-11
add weth

sunnya97:


src/stores/root.ts
```diff
@@ -1404,6 +1404,21 @@ export class RootStore {
 					},
 				],
 			},
+			{
+				poolId: '703',
```
JeremyParish69, line 1408: FYI, 704 is otherwise identical but has more TVL. Used 704 in Frontier

sunnya97: Oh ok

Huh, someone else made one too

sunnya97, line 1408:
```suggestion
				poolId: '704',
```

Issue #481, opened 2022-05-11
Add stable coin info/disclaimer


jonator:
In light of the current UST situation, we should help our users make the right decision around the use of stable coins. We should add links next to stable coins in the assets page that jump to a docs page explaining the risks associated with each type of stable coin.

This issue tracks various tasks
- [ ] Design - warning in add liquidity dialog @syed8rahman @victortokaito
- [ ] ~~Docs - a page on docs.osmosis.zone providing info about the types of stables (w/ a breakdown of the types of collateralized assets) @daniel-farina (necessary?)~~
- [ ] Development - adding the warning to the refactor development stream @jonator 
- [x] Config - Add a flag in ibc config to indicate that this transferred asset is a stable coin @JeremyParish69

JeremyParish69:
Regarding the Config flag,
-considering possible locations:
--could have under IBCAssetInfos; this seems to be limited to data directly pertaining to depositing or withdrawing IBC assets to/from Osmosis. Information about whether it's a stablecoin doesn't really fit here. If we put it here, it only needs to be specified once.
--could have under EmbedChainInfos; the seems to be mostly limited to data pertaining to Keplr's Sggest Chain feautre. Information about whether it's a stablecoin isn't required for that. However, I think specifying the token logo is also not used for Keplr's Suggest Chain feature, and yet we put the URL in here. It would also beg the question of whether the flag must be repeated wherever the stablecoin is used as a staking currency or fee currency (e.g., UST is used as a Fee currency).

---I'm leaning to put it under EmbedChainInfos, and only under the Currencies section.

-considering what//how to indicate the flag:
--simplest option is something like a boolean-- "isStable": true,
--there is opportunity to include more information about the stablecoin, such as index or mechanism
---e.g., "stableMechanism": "aglorithmic" (or "debt", or "reserve"?)--this opens a lot of questions about how to categorize, and could quickly become outdated due to special blends of mechanisms or now mechanisms
---e.g., "index": "United States Dollar" -- this could apply beyond to non-fiat or synthetics

--currently leaning to keep it simple for now with just "isStable" boolean, although could be worth discussing these other options later.

jonator:
Two points
* `isStable` may be misleading, and may have readers think it refers to the stability of the associated chain. IMO a better alternative would be `pegMechanism`
[...]

PR #483, opened 2022-05-12
Frontier - Add Stablecoin Flag

JeremyParish69:
add isStable: true, to all stablecoins (terra, gbridge, axl)
one task of #481


jonator:
Added this to refactor branch. LMK if you need to change/add more flags

jonator:
Added this to refactor branch. LMK if you need to change/add more flags

PR #494, opened 2022-05-14
use pool price for ust and luna

sunnya97:


src/config.ts
```diff
@@ -1246,14 +1246,14 @@ export const EmbedChainInfos: ChainInfoWithExplorer[] = [
...
-				coinGeckoId: 'terra-luna',
+				coinGeckoId: 'pool:uluna',
 				coinImageUrl: window.location.origin + '/public/assets/tokens/luna.png',
 			},
 			{
 				coinDenom: 'UST',
 				coinMinimalDenom: 'uusd',
 				coinDecimals: 6,
-				coinGeckoId: 'terrausd',
+				coinGeckoId: 'pool:uluna',
```
JeremyParish69, line 1256: should be pool:uusd, not luna. Will fix

sunnya97: Thanks

PR #500, opened 2022-05-16
Main - Add WBTC asset n MEME price

JeremyParish69:
-Add axlwbtc to chain info so it will look okay on it's pools, but not to Assets page
-Add Meme price to main so its pool reflects accurate liquidity


JeremyParish69:
Confirmed with generated build that Pool 701 OSMO/MEME and 712 OSMO/WBTC look correct now.

PR #502, opened 2022-05-16
Main - Add mntl

JeremyParish69:
Add AssetMantle MNTL
add to Assets page (chain infos was already there)
add to trade router (690 OSMO/MNTL)
add altCoinId


JeremyParish69:
Deployment LGTM.

PR #506, opened 2022-05-17
Frontier - Add Rizon

jayjay-crypto:


JeremyParish69:
@jayjay-crypto fantastic job with these PRs. The only issue is that we can't add atolo to the Trade Router with such low liquidity. Are we able to get this up to at least $1000? alternatively we can add it without adding to the Trade page...

Rizonian:
Hi Jeremy, we will check the swap fee. 
Does $1000 means pool size?

JeremyParish69:
> Hi Jeremy, we will check the swap fee. Does $1000 means pool size?

Thanks for checking, as 2% is unusual and much higher than all other pools in the trade router.
Yes, the $1000 is referring to total pool size.

JeremyParish69:
pausing Frontier additions. Will create a new PR to add as a Frontier token to refactored branch.

JeremyParish69:
Can we please use secure (https) for the endpoints? The current ones do not respond to https...

jayjay-crypto:
Hi @JeremyParish69 sorry for the late feed back. Rizon's team have open a new pool with good fees, 0.2%, and more liquidity, it's pool 729. Can we re-add Rizon to the router and finish the PR

PR #508, opened 2022-05-18
Legacy Main - Change old logo with recent logo

hannydevelop:


JeremyParish69:
I'm not sure the significance of the transparent padding. Will the Cosmos Chain Registry also need to be updated likewise?

hannydevelop:
> I'm not sure the significance of the transparent padding. Will the Cosmos Chain Registry also need to be updated likewise?

Yes, the Cosmos chain Registry needs to be updated too. The reason for this is the disparity in size of the SOMM logo and other logo in the chain registry.

Also, since we changed our logo, this PR updates the logo to the recent one.

JeremyParish69:
> > I'm not sure the significance of the transparent padding. Will the Cosmos Chain Registry also need to be updated likewise?
> 
> Yes, the Cosmos chain Registry needs to be updated too (I sent a PR for that also). The reason for this is the disparity in size of the SOMM logo and other logo in the chain registry.
> 
> Also, since we changed our logo, this PR updates the logo to the recent one.

Apologies, I forgot Osmosis was still using the old logo. It looks like this image is a little different from the Chain Registry; should we keep that in sync?
Also, square images tend to not look as nice on the Osmosis frontend; can we turn this into a circular shape?

hannydevelop:
> > > I'm not sure the significance of the transparent padding. Will the Cosmos Chain Registry also need to be updated likewise?
> > 
> > 
> > Yes, the Cosmos chain Registry needs to be updated too (I sent a PR for that also). The reason for this is the disparity in size of the SOMM logo and other logo in the chain registry.
> > Also, since we changed our logo, this PR updates the logo to the recent one.
> 
> Apologies, I forgot Osmosis was still using the old logo. It looks like this image is a little different from the Chain Registry; should we keep that in sync?

I just replaced it with the circular logo.

Also, I have a PR here, with the same logo, that should cover the chain registry right?
https://github.com/osmosis-labs/assetlists/pull/180

JeremyParish69:
> right?

The chain registry would be here:
https://github.com/cosmos/chain-registry/pulls

hannydevelop:
> > right?
> 
> The chain registry would be here: https://github.com/cosmos/chain-registry/pulls Then will have to go to new location

Yes. However, there's no provision to put in the logo here. I believe it's in the assetlist

JeremyParish69:
> > > right?
> > 
> > 
> > The chain registry would be here: https://github.com/cosmos/chain-registry/pulls Then will have to go to new location
> 
> Yes. However, there's no provision to put in the logo here. I believe it's in the assetlist

Please confirm the following PR for Chain Registry: https://github.com/cosmos/chain-registry/pull/381
I've created another PR to add the new images to the new frontend: https://github.com/osmosis-labs/osmosis-frontend/pull/533

JeremyParish69:
This has now been updated on the Main app

PR #514, opened 2022-05-19
Add ASVT token to Frontier

angrymouse:
Re-opens #476 
Now pool ROWAN/ASVT created, and ics20 fixed


angrymouse:
So what about merging 🤔

JeremyParish69:
> So what about merging 🤔

was adding to the Trade router and debugging a pricing issue, but should be soon.

PR #515, opened 2022-05-19
Legacy Main - Add asvt pricing data

JeremyParish69:
Frontier asset. will add pricing when visiting the pools page, but will not show on Assets list


JeremyParish69:
was included in #529

PR #525, opened 2022-05-23
Legacy Main - add xprt/usdc pool external incentives.

puneet2019:
Created gauge: incentives kick in from 24th May 2022.
https://www.mintscan.io/osmosis/txs/AAF999B38F8AD79CF44D8EFA8038C3C56452AADA32F40EA92B941E8C34351EF4


JeremyParish69:
will be included in #530

PR #544, opened 2022-05-26
Fix PSTAKE (multihop IBC) & ibc hash query denoms

jonator:
Two main improvements:
* Fix PSTAKE (multihop IBC assets)
  * Show balances in assets page
  * Show balance available to deposit
* Use IBC hash's for to/from trade query params


JeremyParish69 (changes requested):
Although many functions with PSTAKE have been fixed, the Withdraw prompt still shows 0 balance and hitting Max fills in the number of OSMO I have, and Withdrawing tries to send OSMO to Persistence chain, which is not desired behaviour.

JeremyParish69 (approved):

PR #553, opened 2022-05-28
Frontier - Remove Inject URL Overrides

JeremyParish69:
because they added Keplr integration


JeremyParish69:
Replaced with #622

PR #558, opened 2022-05-28
Frontier - kava, hard, swp to frontier

sunnya97:
Needs IBC info


JeremyParish69:
Added IBC Asset info. Liquidity of each on Osmosis would be nice so we can test IBC transfers and define a pools for the Trade Router

PR #563, opened 2022-05-31
Main - add KAVA, HARD, SWP

JeremyParish69:
Kava verified,
HARD, SWP to Frontier,


JeremyParish69:
IBC transfers work, but 730 is not yet included in the trade router

PR #567, opened 2022-05-31
Jp69/add LINK

JeremyParish69:
-add LINK images
-add LINK chain info under Axelar, and IBC asset info


jonator:
Is there a reason to include both png and svg? My preference is SVG

PR #578, opened 2022-06-02
custom chain names

sunnya97:


sunnya97:
Allows the ability add a custom source chain name (so it can say Ethereum instead of Axelar)

jonator (commented):
Could instead of introducing this change now, we add the idea to the tracking issue for the work Tony's going to be doing with native bridge integration? Or is it high prio?

My rationale is that Tony is probably going to introduce some standard type-enforced keys for what bridge and what chain an IBC asset originates. They could be used to determine what integration to use.

sunnya97:
> What about something like "Ethereum via Gravity Bridge" for non-canonical bridged tokens? or is that just too long?

I tried "Ethereum via Axelar" but it was too long and looked awkward.  But maybe its okay for the non-canonical ones on Frontier

> My rationale is that Tony is probably going to introduce some standard type-enforced keys for what bridge and what chain an IBC asset originates. 

What does this mean?  We should be preparing to support more deposit/withdrawal methods for different assets instead of locking them in?  (Like the UIs @victortokaito has been working on for choosing which chain to withdraw to).  Also, this seems somewhat orthogonal to just having a specific display name for a chain?

sunnya97:
I'm going to merge this sooner.  And then we can revert it once the larger native bridge integration is ready, and there's an alternative way we want to do this.  But for now, showing everything as Axelar, is not great UX, when really the UX flow is coming from Ethereum and the like

PR #581, opened 2022-06-02
Add 'cosmwasm' feature to juno

JeremyParish69:


jonator:
The old frontier frontend is on `v0.9.12-rc.3` of `@keplr-wallet`, and it only requires `wasmd_0.24+` as seen [here](https://github.com/chainapsis/keplr-wallet/blob/v0.9.12-rc.3/packages/stores/src/query/cosmwasm/contract-query.ts#L31-L32) to query contract balances with correct URL.

However we're running `^0.10.4` of `@keplr-wallet`, and it requires `cosmwasm` feature as well to query with the correct URL, seen [here](https://github.com/chainapsis/keplr-wallet/blob/v0.10.4/packages/stores/src/query/cosmwasm/contract-query.ts#L32). Is it a problem elsewhere to enable `cosmwasm` feature? It seems Keplr should be using `||` logic there.

jonator:
Merging for now, will undo if @Thunnini raises concerns with enabling `cosmwasm` feature for Juno

PR #583, opened 2022-06-03
PREVIEW ONLY: Unbond native assets button+tooltip in assets table

### LINKED ISSUE #580: Create UI to allow unbond of native assets
### Problem
A few users reported accidentally bonding native OSMO from an old bug in the UI. 

### Solution
We should create a simple UI that allows users to unlock native OSMO or ION. If there is a native asset currently unlocking, show that instead. Show nothing if there's no native assets locked or unlocked.

@victortokaito @syed8rahman feel free to attach Figma UI link. I think it makes sense to show this as a simple button for unlocking, and a simple faint label to show unlocking asset balances, and nothing for neither. It should be located in the withdraw column of the asset table.

jonator:
### Problem
Some users, due to a fixed refactor UI bug, managed to lock native OSMO assets.

### Solution
Added a simple tooltip in assets table showing current unbonding amounts and end times for native OSMO and ION, as well as an "Unbond" button to let the user initiate an unlock for any outstanding native locks.

We will NOT be merging this into prod, we will just use this PR to generate Vercel Preview links that users can get from the support team on discord.

Will close #580 on open.

### DIFF
--- a/packages/web/components/table/assets-table.tsx
+++ b/packages/web/components/table/assets-table.tsx
@@ -14,6 +14,7 @@ import {
   AssetNameCell,
   BalanceCell,
   TransferButtonCell,
+  UnbondNativeAssetCell,
   AssetCell as TableCell,
 } from "./cells";
 import { useStore } from "../../stores";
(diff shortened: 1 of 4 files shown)


jonator:
6/21 update - all users have used it to unbond native assets. Leaving up to let those users see how much time is left.

jonator:
Closing since incidents have been resolved

PR #584, opened 2022-06-03
Update tailwind.config.js

victortokaito:
Subtitle 2: 600 -> 400 font weight

### DIFF
--- a/packages/web/tailwind.config.js
+++ b/packages/web/tailwind.config.js
@@ -149,10 +149,6 @@ module.exports = {
       caption: 400,
       overline: 400,
     },
-// body1: 500 -> 400
-// body2: 500 -> 400
-// button: 700 -> 600
-// caption: 500 -> 400
     backgroundImage: {
       none: "none",
       "gradients-socialLive":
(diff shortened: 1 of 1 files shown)


packages/web/tailwind.config.js
```diff
@@ -142,13 +142,17 @@ module.exports = {
...
-      button: 700,
-      caption: 500,
+      subtitle2: 400,
+      body1: 400,
+      body2: 400,
+      button: 600,
+      caption: 400,
       overline: 400,
     },
+// body1: 500 -> 400
```
jonator, line 152: Could we remove these comments? We can see preview values in git history.

jonator (changes requested):

packages/web/tailwind.config.js
victortokaito: yeah lol, I'm kinda OCD about keeping notes. I'll update

victortokaito:
@jonator am getting fails for the build--I know that's normal, but which of these are okay?

jonator:
@victortokaito `osmosis-frontend-old` and `osmosis-refactor` are expected to fail since they used to rely on a `vercel.sh` folder to determine what branch they are on when we used to also deploy with cloudflare (`vercel.sh` is now removed).

PR #590, opened 2022-06-04
Add DOT and WGMLR from Moonbeam via Axelar

JeremyParish69:


sunnya97:
Why is it called xcDOT?

JeremyParish69:
> Why is it called xcDOT?

Similar to how the Ether is technically wrapped into WETH, so is the DOT into xcDOT. At least I'm pretty sure that's how it works--I'm no expert at Polkadot.

JohnnyWyles:
> > Why is it called xcDOT?
> 
> Similar to how the Ether is technically wrapped into WETH, so is the DOT into xcDOT. At least I'm pretty sure that's how it works--I'm no expert at Polkadot.

Wouldn't Proposal #249 mean that this was added to Osmosis as DOT rather than xcDOT though?  Considering that entire route, including the wrapping, as what Osmosis calls the canonical DOT.

JeremyParish69:
> > > Why is it called xcDOT?
> > 
> > 
> > Similar to how the Ether is technically wrapped into WETH, so is the DOT into xcDOT. At least I'm pretty sure that's how it works--I'm no expert at Polkadot.
> 
> Wouldn't Proposal #249 mean that this was added to Osmosis as DOT rather than xcDOT though? Considering that entire route, including the wrapping, as what Osmosis calls the canonical DOT.

I think the decision is still up in the air; however, ubiquitously throughout Moonbean on various dApps, 'xcDOT' seems to be used as the symbol every time (I wasn't able to find any counterexamples).
To simply call an asset by its index's symbol is basically to ignore all the transfers and risks taken to get the representation to where it is. Wouldn't it be disingenuous to imply that 'mGME' is the actual GameStop stock? That's why the 'm-' prefix is important.
Osmosis has canonical Ether, but it's still shown as WETH, because it is technically wrapped.

sunnya97:
We show WETH on Osmosis, because currently you can only deposit the WETH erc20 token.  Once Axelar updates to allow depositing the unwrapped normal ETH, we'll rename WETH to ETH on Osmosis as well

JeremyParish69:
> We show WETH on Osmosis, because currently you can only deposit the WETH erc20 token. Once Axelar updates to allow depositing the unwrapped normal ETH, we'll rename WETH to ETH on Osmosis as well

Right, but currently a user needs to already have Ether in their Ethereum wallet (for gas) to be able to unwrap the WETH from Osmosis to send it around or use it--this I'd argue is a good reason to not call it ETH on Osmosis (yet). However, once the wrapping is built into Axelar, a user will then be able to send the WETH from Osmosis to Ethereum, and this time won't need to already have Ether in their Ethereum wallet to be able to use it, because it'll be unwrapped automatically, and therefore the recipient will have gas--this I'd argue is good reason to then be able to label it just ETH on Osmosis, even if it's technically wrapped behind-the-scenes.

With DOT/xcDOT, the user is required to have GLMR on Moonbeam (again, for gas) to be able to move the xcDOT around--this requirement to obtain gas through some other means I'd argue is good reason to imply (via the xc- prefix) that it's not quite native DOT.

An alternative way to think about it is: after sending xcDOT from Moonbeam to Polkadot, is some other source of DOT required for gas for some kind of wrapping? And then answer is: No. Which actually supports the idea of referring to it as just DOT.

So, maybe it can be labelled as DOT so long as the Chain is clearly labelled as Moonbean, and not Polkadot. There might be a bit of a surprise if user try to use dApps on Moonbeam and discover that they all refer to it as xcDOT, but this can probably be quickly assumed and shouldn't cause any significant issue.

sunnya97:
My recommendation is that we call it xcDOT to start, and then revisit to decide to rename it to DOT in the future

JeremyParish69:
Changed xcDOT to DOT

jonator:
@JeremyParish69 there appears to be merge conflicts

PR #592, opened 2022-06-04
use ".grv" suffix instead of "g" prefix for gravity assets

sunnya97:


sunnya97:
@JeremyParish69 can you update assetlist to match as well

jonator:
@sunnya97 should we also add source chain name overrides?

JeremyParish69 (approved):

JeremyParish69:
Frontier Info site now set to uses .grv suffix. I think we're probably OK to merge this now, eh @jonator ?

PR #599, opened 2022-06-07
Support osmosis v9 for frontier

Thunnini:
For osmosis v9, there aren’t many major changes to the front-end.
Without this PR, the create pool will not work in osmosis v9.
However, all other functions are expected to work.
Merge this PR when osmosis v9 goes up.


jonator:
@Thunnini we are not deploying frontier from the master branch as well, but with an env var set to true to put it in frontier mode.

jonator:
Closing since we are no longer deploying frontier-old

PR #608, opened 2022-06-10
Temp testnet

jonator:


jonator:
DO NOT MERGE

jonator:
@czarcas7ic do we still need this? Should we update the old testnet frontend to the new codebase?

czarcas7ic:
We no longer need this, and if the testnet beach could get synced up to main that would be awesome

Issue #412, opened 2022-04-26
New user: link to chrome ext page if Keplr not installed


jonator:
If a new user explores our app without knowing about or having Keplr installed, they may not know that it is an extension that they use to interact with the chain.

jonator:
4/27 meeting- Med. priority due to recent scammy results when googling "keplr" wallet to install

CMarshall92:
I have raised a PR that makes the non install flow a bit clearer for a user that doesnt have keplr installed in the browser. You can find it at the following link. [PR here](https://github.com/osmosis-labs/osmosis-frontend/pull/613)

CMarshall92:
Let me know if its something that works cant see any of the designs if there are any so i've kind of went with what the scan QR flow looks like for consistency

PR #613, opened 2022-06-12
Added user redirect modal to official keplr site to install extension

CMarshall92:
This resolves the issue raised [https://github.com/osmosis-labs/osmosis-frontend/issues/412](here) It creates a new user modal that only appears when the keplr extension is not installed in the users browser. If this is the case it will appear with a new modal that contains a redirect to the official keplr website. You can see the result in the staging env and also in the gif below.

Jun-12-2022 23-36-53

### DIFF
--- a/packages/web/hooks/use-keplr/context.tsx
+++ b/packages/web/hooks/use-keplr/context.tsx
@@ -10,6 +10,7 @@ import { getKeplrFromWindow } from "@keplr-wallet/stores";
 import {
   KeplrConnectionSelectModal,
   KeplrWalletConnectQRModal,
+  KeplrInstallModal,
 } from "../../modals";
 import EventEmitter from "eventemitter3";
 import { BroadcastMode, StdTx } from "@cosmjs/launchpad";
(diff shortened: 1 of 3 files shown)


jonator (changes requested):
@CMarshall92 Appreciate the contribution! I left a few small code-level comments. One more thing: it would be nice if when the user returned to the web app, the wallet select modal was open, ready for them to select Keplr assuming they installed it.

Also need approval on design:
@victortokaito @syed8rahman what do you think of the UX here (use Vercel preview links above to test)? I think it works and is a good idea, my only concern is perhaps we should be explaining why the user needs to install Keplr on the first modal, instead of on the second modal after they've selected Keplr option. I have a feeling some users who don't know what Keplr is won't select the Keplr wallet connection option to see the "install keplr" modal w/ the explanation of Keplr purpose.

jonator:
@victortokaito yes. This PR should only show Keplr homepage link if it detects no Keplr extension is installed on browser. Otherwise if Keplr extension is installed/detected, it will prompt to connect Keplr.

CMarshall92:
It might be worth removing that new modal and just detecting if the extension is present at the time the first one opens. Only issue with that it could get wordy in the first modal if you want I can make that change though

jonator:
Perhaps exclude the wordy explanation and just show "Install Keplr 🔗" ??

CMarshall92:
> Perhaps exclude the wordy explanation and just show "Install Keplr 🔗" ??

Yup can do will have a look when I get a few minutes

CMarshall92:
> I've made some updates from the review- instead of showing another modal after clicking "Keplr Wallet", it will show this instead with a link to the official site.
> 
> Screen Shot 2022-06-22 at 1 42 33 PM
> 

Sorry I've been really ill the past week, will see how things go this week but I agree that looks better

jonator:
@CMarshall92 Thanks for the contribution!

PR #629, opened 2022-06-16
Update gauges

angrymouse:
Add gauge to pool 1, update gauges for pool 716, add gauge to pool 717


JeremyParish69:
The gauge for 716 is 4303 and the gauge for 717 is 4302 so those still need to be switched

JeremyParish69 (approved):

angrymouse:
When merge 😢

PR #633, opened 2022-06-17
Add LUNC, USTC moved info banner to main site

jonator:
Some users have posted on forums freaking out that they can't see LUNC, USTC balances on their assets page. We should add a banner to guide those users to the frontier site, where they can view those assets.

Should remove it in about a week or so.

I also made the banner code generic and reusable for future temp banners (instead of just for the frontier banner). Banners that are closed remain closed unless the browser clears localStorage.


victortokaito:
@jonator I'm good with this. Please just update to match [the testnet and support site banners](https://www.figma.com/file/7iuO8MZeiYezTnF7suwTNN/draft-UI?node-id=930%3A36920)

PR #654, opened 2022-06-28
update bitsong gauges

angelorc:


JeremyParish69 (approved):
yep, gauge ids are vaild. LGTM

PR #376, opened 2022-04-11
Frontier - Add fet

Galadrin:
FET is the governance token on fetchhub chain.
see https://fetch.ai/

MOBX is a secondary native token from Fetchhub chain
see https://mobix.ai/

replace #363

PR #259, opened 2022-02-14
Adding Gravity bridge token

JeremyParish69:
-assuming the token symbol (GRAV) to be used for 'coinDenom', even though coin denom is 'graviton'.

PR #671, opened 2022-07-08
Add Tgrade

abefernan:
This PR adds the Tgrade mainnet network and its native TGD token.

Maybe some of the features on the array can be removed:
- We are not adding any IBC token for now, so maybe "ibc-transfer" and "ibc-go" are not needed.
- We do use wasmd 0.27, so I guess "wasmd_0.24+" should be there.
- We are not adding any CW20 token for now, so maybe "cosmwasm" is not needed even if our network is cosmwasm capable.


Homelander91:
Hi I am just a community member, I see the preview of this PR still doesn't show tgrade, is something wrong ?

I see in previous PRs for example: https://github.com/osmosis-labs/osmosis-frontend/pull/259 the preview shows it all

abefernan:
> Hi I am just a community member, I see the preview of this PR still doesn't show tgrade, is something wrong ?
> 
> I see in previous PRs for example: #259 the preview shows it all

Thanks, maybe the previous PRs I followed as example were not complete. I'll check out that one

Homelander91:
Thanks !

Usually I see it require changes in `src/config.ts` where we add source/destination ibc channels and in `src/stores/root.ts` where poolId and stuff is mentioned. Here is a more recent example: https://github.com/osmosis-labs/osmosis-frontend/pull/376

JeremyParish69:
Hi @abefernan , thanks for the submission. I'm excited for Tgrade joining Osmosis.

FYI, the other PRs you're looking at for reference was quite a while ago when we were using a different code base. By now the structure of files has changed. You could look at #665 if you want to follow an example.

There are few reasons that TGD won't appear on the generated link. Mainly, the `config/ibc-assets.ts` is missing the tgrade token listing.

Still, we will not list a token if it cannot be transferred to and from Osmosis, so an IBC connection between Osmosis and Tgrade with active relayers is required. We have guides on setting up a relayer, as well as a list of entities willing the support or offer service of running a relayer, if needed. Finally, we'd like to register to the Cosmos Chain Registry that IBC connection once it's made (under the /_IBC/ folder), with which I don't mind helping.

Once that is sorted out, we should be OK to list Tgrade.

abefernan:
Thank you @JeremyParish69 for your help! I have added the IBC data to `config/ibc-assets.ts` and you can see the connection registered on the chain registry here: https://github.com/cosmos/chain-registry/blob/master/_IBC/osmosis-tgrade.json

homelander-91:
I think we are still missing a PR to https://github.com/osmosis-labs/assetlists (to make it appear in trade page) and creation of a liquidity pool as per docs here: https://docs.osmosis.zone/integrate/token-listings.html#token-listings

JeremyParish69 (approved):

webmaster128:
The relayer is up and running. A pool was successfully created. Is this ready to get merged?

By the way, what does `isVerified: true,` mean? Seems like it is not set for Tgrade.

PR #675, opened 2022-07-12
remove steth

JeremyParish69:
because it doesn't rebase on Osmosis, so users don't get staking rewards. Hence, virtually pointless, and potentially confusing.


sunnya97:
Also, there is currently no balance of steth on Osmosis, so is okay to remove for now without impacting users.

@JeremyParish69 please also remove from assetlist

PR #683, opened 2022-07-14
Fixed IBC channels + Added Odin logo with white stroke

AddySiz:


AddySiz (commented):
@JeremyParish69 file rename as discussed

PR #715, opened 2022-08-08
Add Crescent

nodebreaker0-0:


JeremyParish69 (changes requested):
The data is good and the intent is appreciated. But we no longer use the frontier branch. All (Main and Frontier) assets are now listed on the master branch. Care to take on adding to this to the master branch (which is formatted a bit differently)? Also would appreciate an IBC file for Crescent and Osmosis in the Chain Registry if there exists an IBC channel between them.

nodebreaker0-0:
Thank you. I submitted a new request again, so I will close this PR. I also submitted the IBC file to the Chain Registry Please review it!

PR #716, opened 2022-08-09
Trade UX Refresh

### LINKED ISSUE #658: Non-standard display of slippage vs price impact
### Problem
It was brought to [Sunny's attention (slack convo)](https://osmosis-network.slack.com/archives/C027ELA2ZM1/p1656453110061789) that the way we present slippage on the Trade UI is unconventional, yet still accurate. Our warning about slippage will result in a high probability (but not guarantee) of failure for slippage exceeding price impact (common in low liquidity pools).

### Proposed Solution
We shouldn't let users submit txs that will likely fail. Instead, we should display the price impact and warn the user if that value is high, thus communicating the low liquidity of the pool they're swapping within. If they have a high slippage tolerance, the tx will pass, otherwise it will fail. We should still also warn about and display the price impact.

jonator:
Quickly made updates to trade clipboard to reflect new trade design and to properly display slippage as price impact.

Fixes #658 .

* Updated clipboard to reflect [this Figma design](https://www.figma.com/file/om93lu8iXOhxwGHSUpAXEP/Trade-UI?node-id=93%3A1765)
* Changed pool math and estimates API to use `priceImpact` instead of `slippage`. Slippage now only (properly) refers to on chain price changes from submit->processing time. Price impact refers to price curve implications of currently selected trade.

TODO on merge:
- [ ] Publish new major FE SDK packages since `slippage` was renamed to `priceImpact` throughout public API


sunnya97:
Minor nit

- The dropdown asset selection window has a bigger width than the trade box

CleanShot 2022-08-09 at 16 55 29@2x

sunnya97:
When there's really high price impact, the button turns Red and says swap anyway

But it doesn't give me any indication as to what the issue is

CleanShot 2022-08-10 at 13 22 48@2x

sunnya97:
Sometimes the "minimum received after slippage" isn't right.  I'm noticing it seems to primarily be on WETH and MKR (so might have something to do with the number of decimals)

CleanShot 2022-08-10 at 13 24 23@2x

CleanShot 2022-08-10 at 13 27 06@2x

jonator:
Promoting this PR to complete Trade UX project

jonator:
@sunnya97 I've pushed changes according to your review

sunnya97:
The switch button animation is amazing 😍

sunnya97:
Oh what I meant was just to move swap fee above the line, not swap the two sides of the line

Like this
CleanShot 2022-08-11 at 16 50 02@2x


Fixed in #729

sunnya97:
Or maybe even this makes more sense

CleanShot 2022-08-11 at 19 10 12@2x

dogemos:
Two small edit requests:
1. Price impact/swap fee information detail gets collapsed when swap input number is cleared out and new number is put in
trade-collapse
2. It'd be good to show the estimated $ value of the token even when the field is empty
estimated-dollar-trade-ui

jonator:
@dogemos Regarding second point made above; we think the $ value fields in each To/From box should only be used for the $ value of the amount immediately above. It may mislead users if it suddenly shows the $ spot price of 1 token, even though it may be convenient to see.

sunnya97 (approved):

jonator:
Pending merge while we figure out the animation. @sunnya97 feel free to check out the animation when switching assets and give us your thoughts

sunnya97:
animation looks great to me

PR #731, opened 2022-08-11
Bridge Integration

jonator:
Integrates Axelar transfers directly into frontend, with the option to connect MetaMask and/or EVM WalletConnect.

Key components
* Observable MetaMask and observable WalletConnect objects for interfacing with respective wallet.
* `/integrations/` folder containing an `/axelar/` folder with a modal view, queries, config, and types needed by Axelar.
* `BridgeTransfer` modal that dynamically imports all Axelar transfer dependencies. Saves ~317kb from initial loaded bundle.
* Integration is enabled for an IBC asset by adding a `AxelarBridgeConfig` to the `originBridgeInfo` member of an entry in ibc-assets.ts's `IBCAssetInfos` list
* IBC transfers now use the new and improved UI + animation. Maintains support for copying addresses on mobile, and editing the withdraw address.
* For mobile, I'm simply going to scale the view down

TODO: 
- [x] Add lottie/bridge animation from Nikita and Syed
- [x] Mobile design
- [x] Add hardcoded Axelar bridge fee to `AxelarBridgeConfig`
- [x] Ensure user has correct network selected in wallet 
- [x] Add preferred network to eth clients then autoswitch/add network before tx
  ~~- [ ] Add Axelar supported network configs~~ > query configs from chainids.network endpoint (as chainlist.org does)
  * Currently, users have to make this correction & add networks on their own. We should automate the switching and adding of networks in metamask.
  * TEST TODO: 
    * - [x] Test adding new RPC configs
    * - [x] Test walletconnect client, with above
- [x] Fix josh's [finds](https://osmosis-network.slack.com/archives/C03JT90769K/p1662551173655559)
  - [x] Point 3) the cached address may be different from an address the user changes to in metamask elsewhere
- [x] Investigate complete eth tx toast
- [x] Help user connect cosmos wallet + display in UI
- [x] Don't let users transact with insufficient fee
- [x] #740 
- [x] KVStore clone error when performing IBC transfer
- [x] Disable selection of wallets that are not installed
- [x] Warn user that withdraw address is different from last used deposit address
- [x] When editing a withdraw address, a long bech32 prefix causes the address to disappear. It should truncate the prefix instead.
- [x] Disable selecting a wallet if it's not installed/accessible by web app. Update select wallet styles to match figma.
- [ ] Shrink select non-Keplr wallet modal to fit grid (center)


packages/web/config/ibc-assets.ts
```diff
@@ -27,12 +28,22 @@ export const IBCAssetInfos: (IBCAsset & {
     sourceChannelId: "channel-208",
     destChannelId: "channel-3",
     coinMinimalDenom: "uusdc",
-    depositUrlOverride:
-      "https://satellite.money/?source=ethereum&destination=osmosis&asset_denom=uusdc",
-    withdrawUrlOverride:
-      "https://satellite.money/?source=osmosis&destination=ethereum&asset_denom=uusdc",
     sourceChainNameOverride: "Ethereum",
     isVerified: true,
+    originBridgeInfo: {
```
thomasWos, line 33: Only for USDC for now?

jonator: Initially. We will gradually roll out others as we test and fix any issues that may arise.

dogemos:
I think we need to implement logic to check what the user’s active network on Metamask is set to. If it isn't Ethereum, we should request the user to change it to Ethereum (https://docs.metamask.io/guide/ethereum-provider.html#basic-usage)

jonator:
Yeah, I also realized that needed to be added to the outstanding to do list.

jonator:
Note: the large number of line changes are largely from the lottie animation files; 38k and 18k lines.

packages/web/config/ibc-assets.ts
```diff
@@ -39,10 +39,6 @@ export const IBCAssetInfos: (IBCAsset & {
       method: "deposit-address" as const,
       sourceChains: [
         AxelarSourceChainConfigs.usdc.ethereum,
-        AxelarSourceChainConfigs.usdc.bnbChain,
-        AxelarSourceChainConfigs.usdc.avalanche,
-        AxelarSourceChainConfigs.usdc.polygon,
-        AxelarSourceChainConfigs.usdc.fantom,
         AxelarSourceChainConfigs.usdc.moonbeam,
```
sunnya97: Lets remove moonbeam for now as well.  There's almost no Axelar USDC on moonbeam

packages/web/components/animation/bridge/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+    if (!animData) {
+      (async () => {
+        const data = bridge
+          ? await import("./lottie-bridge.json")
+          : await import("./lottie-ibc.json");
+        setAnimData(data);
+      })();
+    }
+  }, [loadingAnimData, animData, bridge]);
```
delivan, line 57: 👍

packages/web/components/animation/bridge/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+            : longFromName
+            ? "left-[90px] md:-left-[4px]"
+            : "left-[122px] md:left-[10px]",
+          { "opacity-30": bridge?.isLoading }
+        )}
+      >
+        From{" "}
+        {truncateString(from.networkName, bridge ? (isMobile ? 10 : 12) : 18)}
+      </span>
+      <style jsx>
```
delivan, line 87: How about writing css code in a css file inside `styles/` folder?

delivan (approved):
LGTM!!

PR #733, opened 2022-08-14
Added GEO price fetch from POOL 787

AddySiz:


packages/web/config/price.ts
```diff
@@ -404,4 +404,14 @@ export const PoolPriceRoutes: IntermediateRoute[] = [
     spotPriceDestDenom: "uosmo",
     destCoinId: "osmosis",
   },
+  {
+    alternativeCoinId: "pool:geo",
+    poolId: "787",
+    spotPriceSourceDenom: DenomHelper.ibcDenom(
+      [{ portId: "transfer", channelId: "channel-258" }],
+      "mgeo"
```
JeremyParish69, line 412: the 'g' in 'mgeo' should be capitalized, since the minimal denom is 'mGeo'

packages/web/config/chain-infos.ts
```diff
@@ -1704,7 +1704,7 @@ const chainInfos = (
           coinDenom: "GEO",
           coinMinimalDenom: "mGeo",
           coinDecimals: 6,
-          coinGeckoId: "geodb",
+          coinGeckoId: "pool:geo",
```
JeremyParish69, line 1707: although the geodb cgid does have a price, it looks like it can go stale from time-to-time, so I agree that a locally determined price feed may be more useful

packages/web/config/price.ts
AddySiz: done

JeremyParish69 (approved):

PR #736, opened 2022-08-16
Trade UX tweaks

jonator:
* Fix estimates content clip in WebKit browsers
* Speed up animations


dogemos:
Looks much better to me!

PR #742, opened 2022-08-20
Sunny/assetlist

sunnya97:


jonator:
Halting work on this for now, can reopen should it resume

PR #759, opened 2022-08-30
Remove CW20 tags from assets page (frontier) #758

Loa212:


jonator (changes requested):
I think it may be useful to still allow users to query "CW20" and get results.

Let's stick to just removing the badge from the view.

jonator (approved):

jonator:
🔥

PR #761, opened 2022-08-30
add cudos ext gauge

JeremyParish69:


sunnya97:
@JeremyParish69 is this Ready for review?

PR #769, opened 2022-09-06
Add stride assets

JeremyParish69:


packages/web/config/ibc-assets.ts
```diff
@@ -841,6 +841,36 @@ export const IBCAssetInfos: (IBCAsset & {
...
+    counterpartyChainId: "stride-1",
+    sourceChannelId: "channel-391",
+    destChannelId: "channel-0",
+    coinMinimalDenom: "stujuno",
+  },
+  {
+    counterpartyChainId: "stride-1",
+    sourceChannelId: "channel-391",
+    destChannelId: "channel-0",
+    coinMinimalDenom: "stuscrt",
```
sunnya97, line 873:
```suggestion
```
Let's remove these until these are actually available

PR #770, opened 2022-09-06
get LUNC price from local pool

### LINKED ISSUE #767: [Testcase - Swap page] - Wrong price data for LUNC on frontier
The current exchange rate for LUNC is too high, if you swap $1 in ATOM or anything else for LUNC, you will get like $4 in LUNC
However, if you the same amount of LUNC swap back to another token, you will get the accurate price.
### Steps
1. https://frontier.osmosis.zone/
2. Try to swap $1 of ATOM or OSMO to LUNC

image
image

Swap back the same LUNC amount
image

JeremyParish69:
pool 561 OSMO/LUNC
resolves #767


packages/web/config/chain-infos.ts
```diff
@@ -76,7 +76,8 @@ const chainInfos = (
           coinDenom: "LUNC",
           coinMinimalDenom: "uluna",
           coinDecimals: 6,
-          coinGeckoId: "terra-luna",
+          //coinGeckoId: "terra-luna",
+          coinGeckoId: "pool:uluna",
```
sunnya97, line 80: This is not going to get confused b/w LUNA vs LUNC right?

sunnya97: Is there a way to rename this pool to `pool:ulunc` by any chance?

JeremyParish69: yes, makes sense. Done

PR #778, opened 2022-09-08
Add user analytics events to use on Amplitude

delivan:
We already are using Matomo for user analytics, but Josh and I recommend using Amplitude. It is scalable and has a lot of useful data analytics tools. Also, [It is highly secure and has a strict privacy policy.](https://amplitude.com/amplitude-security-and-privacy) Keplr Wallet is already using it. So Let's start with using both Matomo and Amplitude and eventually transitioning to Amplitude would be good.

## TODO
- [x] Add events
- [x] Add user properties

### DIFF
--- a/packages/web/config/user-analytics-v2.ts
+++ b/packages/web/config/user-analytics-v2.ts
@@ -2,6 +2,7 @@ export type EventProperties = {
   fromToken: string;
   toToken: string;
   isOnHome: boolean;
+  percentage: string;
   poolId: string;
   poolName: string;
   poolWeight: string;
@@ -50,7 +51,7 @@ export const EventName = {
     pageViewed: "Swap: Page viewed",
     maxClicked: "Swap: Max clicked",
     halfClicked: "Swap: Half clicked",
-    settingClicked: "Swap: Setting clicked",
+    slippageToleranceSet: "Swap: Slippage tolerance set",
     switchClicked: "Swap: Switch clicked",
     swapClicked: "Swap: Swap clicked",
   },
--- a/packages/web/config/user-analytics-v2.ts
+++ b/packages/web/config/user-analytics-v2.ts
@@ -10,6 +10,7 @@ export type EventProperties = {
   toToken: string;
   isOnHome: boolean;
   percentage: string;
+  isMultiHop: boolean;
   poolId: string;
   poolName: string;
   poolWeight: string;
@@ -60,7 +61,8 @@ export const EventName = {
     halfClicked: "Swap: Half clicked",
     slippageToleranceSet: "Swap: Slippage tolerance set",
     switchClicked: "Swap: Switch clicked",
-    swapClicked: "Swap: Swap clicked",
+    swapStarted: "Swap: Swap started",
+    swapCompleted: "Swap: Swap completed",
   },
   // Events in Sidebar UI
   Sidebar: {
--- a/packages/web/config/user-analytics-v2.ts
+++ b/packages/web/config/user-analytics-v2.ts
@@ -22,6 +22,7 @@ export type EventProperties = {
   isOn: boolean;
   tokenName: string;
   tokenAmount: number;
+  bridge: string;
   hasExternalUrl: boolean;
 };
 
@@ -114,9 +115,9 @@ export const EventName = {
     assetsListMoreClicked: "Assets: Assets list more clicked",
     assetsItemDepositClicked: "Assets: Assets item deposit clicked",
     assetsItemWithdrawClicked: "Assets: Assets item withdraw clicked",
-    depositIbcAssetStarted: "Deposit IBC asset: Deposit started",
-    depositIbcAssetCompleted: "Deposit IBC asset: Deposit completed",
-    withdrawIbcAssetStarted: "Withdraw IBC asset: Withdraw started",
-    withdrawIbcAssetCompleted: "Withdraw IBC asset: Withdraw completed",
+    depositAssetStarted: "Deposit asset: Deposit started",
+    depositAssetCompleted: "Deposit asset: Deposit completed",
+    withdrawAssetStarted: "Withdraw asset: Withdraw started",
+    withdrawAssetCompleted: "Withdraw asset: Withdraw completed",
   },
 };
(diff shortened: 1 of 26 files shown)


jonator:
Converting to draft for outstanding todo items

jonator:
Is there a place where the events we track are documented? We don't necessarily have to have a notion page, but I will share [this](https://www.notion.so/osmosiszone/8bbbabce67fa4c4289989632633b9052?v=7123ab7319054d9aa63ba9e40b6d6c51) one for you to see if there's anything else you missed that we should track. @syed8rahman, @daniel-farina and I jammed in a meeting to come up with these.

delivan:
Yes, [this document](https://docs.google.com/spreadsheets/d/18w8VwJmmRdb_E-XkE1UjkqhLxCyhqVVhWlzDgTtbRWo/edit#gid=771066228) lists the events to be tracked, user properties, and example values. Google Workspace Osmosis members can add/modify/delete items. I would love that if you can check out what I missed.

jonator:
Great. I'll add comments to the document.

packages/web/integrations/axelar/transfer.tsx
```diff
@@ -233,6 +236,20 @@ const AxelarTransfer: FunctionComponent<
...
       if (depositAddress) {
+        logEvent([
+          isWithdraw
+            ? EventName.Assets.withdrawAssetStarted
+            : EventName.Assets.depositAssetStarted,
+          {
+            tokenName: isWithdraw
+              ? withdrawAmountConfig.sendCurrency.coinDenom
+              : originCurrency.coinDenom,
+            tokenAmount: isWithdraw
```
jonator, line 247: You can simply use the `amount` variable here.

packages/web/integrations/axelar/transfer.tsx
```diff
@@ -233,6 +236,20 @@ const AxelarTransfer: FunctionComponent<
     const [transferInitiated, setTransferInitiated] = useState(false);
     const doAxelarTransfer = useCallback(async () => {
       if (depositAddress) {
+        logEvent([
+          isWithdraw
+            ? EventName.Assets.withdrawAssetStarted
+            : EventName.Assets.depositAssetStarted,
+          {
+            tokenName: isWithdraw
+              ? withdrawAmountConfig.sendCurrency.coinDenom
```
jonator, line 245: Similarly, I'd just user `originCurrency` here.

packages/web/pages/assets/index.tsx
```diff
@@ -310,19 +357,41 @@ const PoolCardsDisplayer: FunctionComponent<{ poolIds: string[] }> = observer(
         if (aFiatValue.toDec().lt(bFiatValue.toDec())) return 1;
         return 0;
       });
+    const { logEvent } = useAmplitudeAnalytics();
 
     return (
       <>
         {pools.map(([pool, _, metrics]) => (
-          <PoolCard
+          <div
```
jonator, line 365: Can we figure out a way to log this event without wrapping in a div?

packages/web/pages/pools/index.tsx
```diff
@@ -328,19 +336,45 @@ const Pools: NextPage = observer(function () {
                   }
 
                   return (
-                    <PoolCard
+                    <div
```
jonator, line 339: Again, could we try to avoid this wrap? Perhaps add an onClick handler to `PoolCard`

packages/web/pages/_app.tsx
```diff
@@ -14,13 +14,24 @@ import relativeTime from "dayjs/plugin/relativeTime";
...
 import { IbcNotifier } from "../stores/ibc-notifier";
-import { IS_FRONTIER, NavBarEvents } from "../config";
+import {
+  AmplitudeEvent,
+  EventName,
+  IS_FRONTIER,
+  NavBarEvents,
+} from "../config";
+
+import * as amplitude from "@amplitude/analytics-browser";
```
jonator, line 24: We could add this func to the hook, as we did with the matomo hook.

Then we could isolate this dependency.

jonator (changes requested):

packages/web/config/user-analytics-v2.ts
```diff
@@ -0,0 +1,122 @@
...
+    twitterClicked: "Sidebar: Twitter clicked",
+    mediumClicked: "Sidebar: Medium clicked",
+    commonwealthClicked: "Sidebar: Commonwealth clicked",
+    discordClicked: "Sidebar: Discord clicked",
+    telegramClicked: "Sidebar: Telegram clicked",
+    supportClicked: "Sidebar: Support clicked",
+  },
+  // Events in Pools page
+  Pools: {
+    pageViewed: "Pools: Page viewed",
```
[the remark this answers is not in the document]
delivan: We should log page view events manually on Amplitude. :(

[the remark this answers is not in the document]
jonator: Sounds good. Let's make sure we log all page loads! In that case, let's also dynamically load the Amplitude react package using `next\dynamic`.

packages/web/config/user-analytics-v2.ts
```diff
@@ -0,0 +1,135 @@
...
+  },
+  // Events in assets page
+  Assets: {
+    pageViewed: "Assets: Page viewed",
+    depositClicked: "Assets: Deposit clicked",
+    withdrawClicked: "Assets: Deposit clicked",
+    myPoolsCardClicked: "Assets: My pools card clicked",
+    myPoolsMoreClicked: "Assets: My pools more clicked",
+    assetsListFiltered: "Assets: Assets list filtered",
+    assetsListSorted: "Assets: Assets list sorted",
```
jonator, line 125: Why do we have separate events for sorting assets? I think in general with events we should decouple them from the UI a bit, so I think it's more fitting to have something like "assets sorted" as a more general event.

Not important.

jonator (approved):

packages/web/config/user-analytics-v2.ts
delivan: I think you are right. I just wanted to figure out what sorting UI(in dropdown or table head) user using but It can be figured out by adding event property not separated events.

PR #786, opened 2022-09-10
Feat/codegen for v12 protos

pyramation:
@jonator we'd probably not merge this til later, because of bundle size (we're generating all the types that cosmjs-types generate, and every osmosis module). So to make this on par with current bundle size, we probably want to collab on ways to leverage `await import()`

however, I still think it could be good to try and see how we can incorporate telescope. I also added some tests for the new amino encoders for v12 protobufs.


pyramation:
I also included RPC gen, as well as amino encoders so we can experiment how they may be able to be integrated.

jonator:
Closing until bundle optimization work is complete

PR #791, opened 2022-09-13
Optimize coingecko API queries

Thunnini:
Due to a problem with the prior keplr libraries, there are too many cancellations in the coingecko API query. Unfortunately cancellation also seems to count towards the rate limit. This causes the user to be blocked by the coingecko for a while.

To reduce this, @keplr-wallet/* libraries have been updated.
The major updates are as follows. https://github.com/chainapsis/keplr-wallet/pull/504

Through the update, the number of queries is reduced because cancellation is reduced and the last coingecko id used previously is reused. And by default, it has a debounce interval of 250ms for coingecko requests.

A more detailed check is required because bugs may occur due to changes in the core library.


jonator (changes requested):
Screen Shot 2022-09-13 at 9 14 24 AM
Seems I'm getting reference errors to the `AbortController` on the server.

Should we be checking for the window?

Thunnini:
@jonator Hmm... no problem with my computer. https://nodejs.org/api/globals.html#class-abortcontroller If you look at the history of AbortController in this link, it seems that AbortController exists in a fairly recent version in nodejs v14. Could you update nodejs on your computer? And could you check the nodejs version used by CI/CD?

Thunnini:
@jonator Hmm... no problem with my computer. https://nodejs.org/api/globals.html#class-abortcontroller If you look at the history of AbortController in this link, it seems that AbortController exists in a fairly recent version in nodejs v14. What is nodejs version on your computer? Could you update nodejs on your computer? And could you check the nodejs version used by CI/CD?

jonator:
I'm on `v14.20.0`

jonator:
Yep that seemed to be it. Using v16 removes the issue.

jonator (approved):

PR #796, opened 2022-09-15
Fixes

### LINKED ISSUE #771: Pool detail page: handle prev lock gas workaround when querying locks
### Problem
@mattverse pointed out that, with the refactor, the frontend lost support for displaying locked positions that were created as a workaround to an old chain bug. The bug caused gas requirements to ramp up linearly with each consecutive LP position lock a user creates with the same duration. The old workaround was to create locks with durations slightly offset from the 1/7/14 day amounts. Since locks are ID'd by duration, it avoided the issue.

More detail:
https://www.notion.so/osmosiszone/8-30-Matt-Park-Lock-Jitter-d75b8b13d37d41ce8ddb4014b4f2b1e6

### Solution
We should add a tolerance to the query of up to + 1 hour after the standard 1/7/14 day durations.

### LINKED ISSUE #787: Liquidity removal issue pool 802. Calculated amount is lesser than min amount (at msg 0)
Last liquidity provider in pool 802 ATOM-stATOM cant remove his funds. He is the only one left in the pool and gets an error withdrawing. Withdrawing only 50% did not work.
Tx: https://www.mintscan.io/osmosis/txs/3EBA57060A4059923E0901C4A13B01DCFF65AC500452AE5B76DEE2BAB39B5F54

IMAGE 2022-09-11 19:54:01
IMAGE 2022-09-11 19:54:16

### LINKED ISSUE #788: Pool cards overflow: reduce sig figs
### Problem 

Cards will begin to overflow when fees and liquidity reach 9-10 significant figures.

### Solution 
Begin abbreviating in millions to avoid overflow on cards (e.g. $170,680,543 -> $170.6M)

_ex)  large fee edgecase_
IMG_1688

### LINKED ISSUE #795: Open Link in New Tab : open pools in new tab
Adding html tags to pools so users could right click and "Open link in New Tab", or command-click to open new tabs

image

jonator:
* Closes #788 
* Closes #795 
* Closes #771 
* Fixes #787 
  * Turns out we forgot to handle the exit fee in the FE since it's sparsely used
* Changed IBC transfer estimated time quote down from 3 mins
* Fixed centering text above bubbles in bridge transfer animation on desktop and mobile
* Refactored the pool detail page by pulling out some calculations into a pool-detail and superfluid pool store. This should help us with our redesign of the pool detail page, coming soon.
* Slightly reduced bundle size by dynamically importing packages, notably Lottie animation.
* Small code cleanliness improvements, such as using console.error for logging exceptions vs console.log.


jonator:
@delivan actually wait on review a bit more. I'm going to investigate the exit pool bug a bit more.

jonator:
Actually @delivan I think the root cause to the exit pool bug is they added an exit fee to pool 802.

delivan:
@jonator Yes, you're right. we missed exit fee calculation on estimateExitPool. Thank you for figuring out!

delivan (approved):
LGTM 🚀 Just need to resolve conflicts.

PR #808, opened 2022-09-20
Enable GKEY coinGeckoId

Rizbe:


packages/web/config/chain-infos.ts
```diff
@@ -468,7 +468,7 @@ const chainInfos = (
           coinMinimalDenom:
             "cw20:juno1gz8cf86zr4vw9cjcyyv432vgdaecvr9n254d3uwwkx9rermekddsxzageh:GKEY",
           coinDecimals: 6,
-          // coinGeckoId: "pool:gkey",
+          coinGeckoId: "pool:gkey",
```
JeremyParish69, line 471: this id is not defined (but you can do so in /config/price.ts). No point in uncommenting this without it defined.

JeremyParish69 (changes requested):
please first define this id

Rizbe:
Updated the PR

I assume once this is merged we will get gkey price data using: https://api-osmosis.imperator.co/tokens/v2/price/gkey 

Currently it says: `{"message":"Oops! Something went wrong with gkey ..."}`

JeremyParish69 (approved):
LGTM

PR #819, opened 2022-09-21
Fix Amplitude event name typo & Add an event to track

### LINKED ISSUE #814: Add user analytics events to use on Amplitude
https://github.com/osmosis-labs/osmosis-frontend/commit/76cb6161605329b998ebf45116ec1b397688d8e6

* seems clicking both the "Global Withdraw" and "Global Deposit" at the top actually triggers `Assets: Deposit clicked` . Is it possible to split this up between deposit/withdraw?
Screen Shot 2022-09-20 at 12 53 43 PM

Screen Shot 2022-09-20 at 12 52 33 PM

* requested tag: Deposit completed (both IBC and non-IBC assets)
– we currently have `Deposit IBC asset: deposit completed` , but we don't yet have this tag for non-IBC assets
– want to distinguish between `Deposit started` and `Deposit completed` and track all events where user actually clicks confirm.
– no need to query if asset actually went through, just want to know if user completed the user flow.

image
image

delivan:
- Close #814 
- Add `Swap: Input box clicked` event
- Add missing event properties

### DIFF
--- a/packages/web/components/trade-clipboard/index.tsx
+++ b/packages/web/components/trade-clipboard/index.tsx
@@ -382,16 +382,6 @@ export const TradeClipboard: FunctionComponent<{
                     );
                   }}
                   onFocus={() => slippageConfig.setIsManualSlippage(true)}
-                  onClick={() =>
-                    logEvent([
-                      EventName.Swap.inputBoxClicked,
-                      {
-                        fromToken: tradeTokenInConfig.sendCurrency.coinDenom,
-                        toToken: tradeTokenInConfig.outCurrency.coinDenom,
-                        isOnHome: !isInModal,
-                      },
-                    ])
-                  }
                   inputRef={manualSlippageInputRef}
                   isAutosize
                 />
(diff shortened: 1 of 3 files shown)


packages/web/components/trade-clipboard/index.tsx
```diff
@@ -379,6 +382,16 @@ export const TradeClipboard: FunctionComponent<{
                     );
                   }}
                   onFocus={() => slippageConfig.setIsManualSlippage(true)}
+                  onClick={() =>
+                    logEvent([
+                      EventName.Swap.inputBoxClicked,
```
jonator, line 387: IMO instead of a click event, we should be using a focus event as we did with matomo.

As though the user intends to add input. Also, this wouldn't trigger if a user tabs into the input box.

Please remove changes to input-box as well.

packages/web/config/user-analytics-v2.ts
```diff
@@ -60,6 +60,7 @@ export const EventName = {
     pageViewed: "Swap: Page viewed",
     maxClicked: "Swap: Max clicked",
     halfClicked: "Swap: Half clicked",
+    inputBoxClicked: "Swap: Input box clicked",
```
jonator, line 63: Again, this should be a focus event.

jonator (changes requested):

packages/web/components/trade-clipboard/index.tsx
delivan: Now auto focus is enabled on input box so I think we don't know what user really did focus. So I think It would be better to track input entered.

jonator (approved):

PR #844, opened 2022-09-26
Add seJUNO, SEASY gauges; prices of SEASY, seJUNO from OSMO pools

udit-gulati:


packages/web/config/price.ts
```diff
@@ -480,4 +480,24 @@ export const PoolPriceRoutes: IntermediateRoute[] = [
     ),
     destCoinId: "stargaze",
   },
+  {
+    alternativeCoinId: "pool:sejuno",
```
JeremyParish69, line 484: can we please add these alt coin ids to the chain-infos.tx file? (as: "coinGeckoId": "pool:sejuno", and "coinGeckoId": "pool:sejuno", in their respective token definitions)

udit-gulati: Fixed, thanks for pointing out.

JeremyParish69 (approved):
LGTM

PR #847, opened 2022-09-27
Re-add old LBP UI

jonator:
Add back the LBP UI from the pre-refactor frontend so that users can view the moving weights of any active and configured LBP pool.

Note/disclaimer: this is not a specific effort or addition for the Rebus chain. We are not affiliated and our frontend and chain remains permission-less.


packages/web/pages/_app.tsx
```diff
@@ -28,7 +29,7 @@ dayjs.extend(utc);
 enableStaticRendering(typeof window === "undefined");
 
 function MyApp({ Component, pageProps }: AppProps) {
-  const menus = [
+  let menus: MainLayoutMenu[] = [
```
delivan, line 32: It seems that It don't need to change declare type.
```suggestion
  const menus: MainLayoutMenu[] = [
```

delivan:
@jonator It seems that it is also necessary to disable to access pool detail page and swap tokens related LBP.

dogemos:
also can we make sure this only gets enabled via frontier and not through app.osmosis.zone? or would this require a ton of work?

jonator:
> also can we make sure this only gets enabled via frontier and not through app.osmosis.zone? or would this require a ton of work?

Trivial work. And it's added.

delivan:
@jonator I have something misunderstood. 😅

jonator:
@delivan what do you mean?

RowanUtrecht:
@jonator 
It looks like the price isn't correct. Since the swap ratio between Osmo and Rebus is going down. But the weight of the pool stays at about 96/4. That's not correct

Screenshot_20220930-080448_Brave
Screenshot_20220930-080439_Brave

RowanUtrecht:
Sorry, checked again. Price looks good. The LBP is just very slowly moving to the 50/50 ratio. Probably some people keep buying small amounts

PR #860, opened 2022-09-29
Osmosis v12 upgrade - frontend changes

daniel-farina:
- These are the changes made by @Thunnini applied to the latest version of the frontend as of 09-29-2022
- ⚠️  This should not be merged until the frontend nodes are upgraded ⚠️


jonator:
@daniel-farina I don't think the other builds would fail, but the txs would not succeed, right?

daniel-farina:
> 

Correct, removed the comment above.

jonator:
Queries would cause type errors, however.

Thunnini:
In the main environment, users have caches in indexed db. However, the previous caches may cause problems because the query results are changed to snake case in v12. To solve this problem, I changed the prefix of kv store of queries store.

PR #862, opened 2022-09-30
Add other .Axl asset bridging to Ethereum

JeremyParish69:
Adds Axelar Bridging for other Axelar assets

Main issues:
-Breaks WGLMR img (easy fix, just accidentally renamed the file glmr.svg -> wglmr.svg)

Other changes:
-Also adds chain-infos.ts and axelar/integrations/types.ts entries for all other Axelar-bridged assets, but it will not display in Assets because they are unchanged in ibc-assets.ts

Todo:
-Still need to figure out Eth->Axelar chainId mapping.
-Still need to add Moonbase logo to /networks/
-Fix Issues written above


packages/web/integrations/axelar/types.ts
```diff
@@ -29,10 +29,15 @@ export type SourceChain =
...
   | "Avalanche"
+  | "Avalanche FUJI C-Chain"
   | "Fantom"
+  | "Fantom Testnet"
   | "Polygon"
+  | "Mumbai"
   | "Moonbeam"
-  | "Binance";
+  | "Moonbase Alpha TestNet"
+  | "Binance Smart Chain"
```
jonator, line 39: Looks like we need to update `EthClientChainIds_AxelarChainIdsMap` below to be "Binance Smart Chain"

packages/web/integrations/axelar/types.ts
```diff
@@ -71,40 +76,680 @@ export type SourceChainConfig = {
...
+        : "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+      logoUrl: "/networks/ethereum.svg",
+    },
+    avalanche: {
+      id: IS_TESTNET
+        ? ("Avalanche FUJI C-Chain" as const)
+        : ("Avalanche" as const),
+      erc20ContractAddress: IS_TESTNET
+        ? "0x3613C187b3eF813619A25322595bA5E297E4C08a"
+        : "",
```
jonator, line 153: Can we add production contracts too?

jonator (changes requested):
Actually, given how big it's gotten, could we move `SourceChainConfigs` to a config file in integrations/axelar?

jonator:
@JeremyParish69 could we prioritize this PR? Including your changes to code to support new config? By EOW (10/7). 

Bankless wants to publish a tutorial in their newsletter on bridging ETH->Swap to Stars->Stargaze and buying a Bad Kids NFG.

packages/web/integrations/axelar/types.ts
JeremyParish69: It was left out because the data is not available. I take it that this can only be done in Testnet so far.

jonator:
@JeremyParish69 please close this when you can

PR #870, opened 2022-10-03
Revert "Add halt UI"

daniel-farina:
Reverts osmosis-labs/osmosis-frontend#869


jonator:
Instead of reverting, we should set the IS_HALTED var to false. If/when we have another halt, we can simply enable the flag.

PR #873, opened 2022-10-04
Add teritori asset on Osmosis frontend

Galadrin:
- declare chain info
- declare ibc asset
- declare pool:utori

Signed-off-by: dpierret <dapie@cros-nest.com>


JeremyParish69 (approved):
LGTM

marsiking:
looks good to me

100hp4ever:
LGTM, ready to be merged.

packages/web/config/chain-infos.ts
```diff
@@ -2016,6 +2016,29 @@ const chainInfos = (
...
+          coinDecimals: 6,
+          coinGeckoId: "pool:utori",
+          coinImageUrl: "/tokens/tori.png",
+          isStakeCurrency: true,
+          isFeeCurrency: true,
+        },
+      ],
+      features: ["stargate", "ibc-transfer", "no-legacy-stdTx", "ibc-go"],
+      explorerUrlToTx: "https://explorer.teritori.com/teritori/tx/{txHash}",
+    },
```
ankurdotb, line 2041: This is not mandatory, but I highly recommend adding your gas price steps (see the chain schema file). If someone already has the chain config previously with the gas prices from elsewhere, this isn't a problem but if you *don't* explicitly specify this, if someone acquires the chain config via Osmosis then you'll end up picking up the default gas price steps rather than your custom ones.

ankurdotb (changes requested):

PR #878, opened 2022-10-05
Use testnet historical data (Imperator) API on testnet

jonator:
Testing: you should be able to see in the frontier/main preview, it has the correct historical data for mainnet. For testnet preview, it has testnet data (I think it's returning all 0 right now).

### DIFF
--- a/packages/stores/src/queries-external/store.ts
+++ b/packages/stores/src/queries-external/store.ts
@@ -3,11 +3,10 @@ import { HasMapStore, ObservableQuery } from "@keplr-wallet/stores";
 import { DeepReadonly } from "utility-types";
 import { ObservableQueryPoolFeesMetrics } from "./pool-fees";
 import Axios from "axios";
-import { IMPERATOR_API_DOMAIN } from "./index";
 
 export class QueriesExternalStore extends HasMapStore<QueriesExternal> {
-  constructor(protected readonly kvStore: KVStore) {
-    super(() => new QueriesExternal(this.kvStore));
+  constructor(protected readonly kvStore: KVStore, feeMetricsBaseURL?: string) {
+    super(() => new QueriesExternal(this.kvStore, feeMetricsBaseURL));
   }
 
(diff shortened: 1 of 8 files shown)


packages/stores/src/queries-external/store.ts
```diff
@@ -19,17 +18,23 @@ export class QueriesExternalStore extends HasMapStore<QueriesExternal> {
 export class QueriesExternal {
   public readonly queryGammPoolFeeMetrics: DeepReadonly<ObservableQueryPoolFeesMetrics>;
 
-  constructor(kvStore: KVStore) {
-    this.queryGammPoolFeeMetrics = new ObservableQueryPoolFeesMetrics(kvStore);
+  constructor(
+    kvStore: KVStore,
+    feeMetricsBaseURL = "https://api-osmosis.imperator.co"
+  ) {
```
ankurdotb, line 24: Probably better to declare this as a constant so that it can easily be changed or injected in Vercel as an environment variable, rather than hardcoding here?

ankurdotb (changes requested):
I suggest sticking to declaring the base URL as a constant and even specifically not hardcoding it in the top-level file, so this can be injected as an environment variable. Perhaps set the default value as a value in constants file, and override it with value from environment if present?

packages/stores/src/queries-external/store.ts
jonator: Yeah it's a default param, I actually prefer hardcoding vs creating 100s of env vars to keep track of since we have such a tight CI/CD pipeline

daniel-farina (approved):
lgtm, I guess we can test it on the testnet once it's merged.

packages/stores/src/queries-external/store.ts
daniel-farina: I do like env variables for everything but mainnet vs testnet can be hardcoded for now. It makes this entire process a lot easier.

PR #884, opened 2022-10-06
Fix stJUNO

riley-stride:
- show stJUNO on the asset pane
- show the stJUNO price on pool 817
- add gauge to pool 817

### DIFF
--- a/packages/web/config/chain-infos.ts
+++ b/packages/web/config/chain-infos.ts
@@ -1974,7 +1974,7 @@ const chainInfos = (
           coinDenom: "stJUNO",
           coinMinimalDenom: "stujuno",
           coinDecimals: 6,
-          // coinGeckoId: "juno-network",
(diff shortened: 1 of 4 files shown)


packages/web/config/chain-infos.ts
```diff
@@ -1974,7 +1974,7 @@ const chainInfos = (
           coinDenom: "stJUNO",
           coinMinimalDenom: "stujuno",
           coinDecimals: 6,
-          // coinGeckoId: "juno-network",
+          coinGeckoId: "juno-network",
```
JeremyParish69, line 1977: please replace "juno-network" with "pool:stujuno"

riley-stride: fixed, thanks!

JeremyParish69 (approved):
LGTM

PR #886, opened 2022-10-06
Display external incentive APR

srph:
This is for the pool gauges in the pool page (`/pool/[id]`):

image

For Stride, we'd like to display the sum of external and internal incentives.

This is how it looks like with the mentioned changes:

image

Relevant changes I've introduced are in `packages/web/pages/pool/[id].tsx`. In addition, I thought not every pool might like this change, so I've thought of putting this configuration behind a flag.

I'd like to kindly ask for some pointers on how to implement this properly. Apparently, I've hard-coded some values (like dividing the values by 1m); and I'm not very familiar with `Keplr`'s `Dec` and `RatePretty` as well.

If there are variables or files I should take note of - that would be helpful.

I'll mark this as a _draft_ PR once I have more idea on how to implement this properly


jonator:
I'd def appreciate it if you could implement this, as my plate has been quite full as of recently.

I've started work on calculating external incentive here:

https://github.com/osmosis-labs/osmosis-frontend/blob/23ecb4e93392bd1b5adee28dd53e8fc597b8b029/packages/stores/src/queries/pool-incentives/incentivized-pools.ts#L258

You could continue my work there. I'd like for it to be a callable function that we could then use anywhere.

srph:
Hey @jonator thanks for advice. I have 2 questions:

- Are we expecting this to replace the current APRs in the PoolGaugeCards?
- Is `computeExternalIncentiveGaugeApr` complete? If so, are there any specific variables we had in mind that should be passed to it?

srph:
I have another question:

So, we have a variable called allowedGauges which is an array that displays the "external incentives" (in our case STRD) which is this:

image

Which we'll use to add to the APRs:

image

What I did was pretty simple - add `allowedGauges[0].rewardAmount` to each APR gauge (1/7/14).

What about when there are more than allowedGauges?

srph:
@jonator Hey, thanks for the discussion earlier!

I've updated the PR based on how the direction I understood we were taking. It's likely I missed something - let me know if there's anything to change or improve.

Here's how the UI looks now for pool 803:

image

packages/stores/src/queries/pool-incentives/incentivized-pools.ts
```diff
@@ -118,6 +118,52 @@ export class ObservableQueryIncentivizedPools extends ObservableChainQuery<Incen
...
+      const chainInfo = this.chainGetter.getChain(this.chainId);
+
+      const mintCurrency = chainInfo.findCurrency(mintDenom);
+
+      if (!mintCurrency) {
+        return apy;
+      }
+
+      const amount = externalGauge.rewardAmount.moveDecimalPointLeft(
+        mintCurrency.coinDecimals
```
srph, line 160: Diving the value by `1_000_000` gives me the value I expect, but I figured hard-coding doesn't make sense. So I took this from `computeAPYForDuration`. Let me know I should be getting this _value_ elsewhere.

jonator:
for pool 803, I believe the 1 day and 7 day gauges should be 0% since there's no incentives for those gauges.

The only gauge I expect to see an APR for external incentives is the 14 day gauge.

packages/web/config/feature-flag.ts
```diff
@@ -55,6 +61,14 @@ export const PromotedLBPPoolIds: {
...
   : [];
 
+// The pool ids white listed that would merge their internal and external gauges
+export const MergesInternalAndExternalIncentiveGaugeList: Record<
+  string,
+  boolean
+> = {
+  "803": true,
+};
+
```
jonator, line 71: IMO you could remove these configs

We should just include APR calculations for gauges in the allow list

This would be more config that @JeremyParish69 would maintain.

packages/stores/src/queries/pool-incentives/incentivized-pools.ts
```diff
@@ -118,6 +118,52 @@ export class ObservableQueryIncentivizedPools extends ObservableChainQuery<Incen
     }
   );
 
+  readonly computeAPYWithExternalIncentives = computedFn(
```
jonator, line 121: As Aidan and I agreed in the meeting, IMO we should calculate APR for a duration for external gauges separately from internal gauges.

jonator: Then, we add them in the view layer.

This way, we would be able to separately display the APR for just external gauges of a certain duration.

jonator (changes requested):

packages/stores/src/queries/pool-incentives/incentivized-pools.ts
jonator: Yeah we can use that code snippet for our current needs, but we'd also have those bullet points available as an option.

jonator: Reminder: we're currently redesigning the pool detail page in a major way, so flexibility in our APIs is extra critical.

I know for a fact we will be displaying external incentive APRs in a different way.

srph:
@jonator Thanks for the notes. I've updated the PR again based on what I understood. Let me know if we're headed in the right direction or if there's anything else to improve.

JeremyParish69:
This is exciting!
Due top the temporal nature of external incentives, I'd hope that the numbers aren't simply just added together, but rather that each gauge gets it's own APR figures. Might be confusing to advertise just a 200% APR when the 190% APR external incentive gauge will end after just one more epoch. A way to show internal and external separately would be better for users.

srph:
@JeremyParish69 Yea, I think that should be the effect when this PR is merged. This is the effect at least for pool 803:

| Before | After |
| ----- | ---- |
| image | image |

Given that it only has one external incentive:

image

jonator (changes requested):
@srph It appears there's an import build error in the pool detail page `[id].tsx`

srph:
@jonator Should be fixed now; I verified it by running `yarn build:frontier`. Let me know if there's anything I may have missed.

jonator:
What was the rationale for how you calculated the APR? Seems like a clever approach.

srph:
@jonator I don't think I have a solid rationale about it. Last time I tried, (existing APR from computeApy) + `(rewardAmount / 1_000_000)` gave me the value we expected. I figured `1_000_000` came from `coinDecimals`, and closest I found where I could get it was `mintCurrency` (I took references from `computeAPYForSpecificDuration`)

jonator:
https://github.com/chainapsis/keplr-wallet/blob/master/packages/stores/src/query/cosmos/supply/inflation.ts#L176

This is the code Keplr uses to compute stake APR. May be a helpful reference. I know it helped my fix an issue with calculating swap fee APR.

srph:
@jonator I consulted with the folks at Stride and have updated the formula. How does this look @jonator?

image

packages/stores/src/queries/pool-incentives/incentivized-pools.ts
```diff
@@ -237,6 +335,7 @@ export class ObservableQueryIncentivizedPools extends ObservableChainQuery<Incen
                   mintPrice.toString()
                 ).mul(yearProvisionToPot.toDec());
 
+                // yearProvisionToPotPrice / tVL
```
jonator, line 338: Can we please remove all comments in this existing function? And otherwise any comments that don't contribute to code comprehension

packages/web/pages/pool/[id].tsx
```diff
@@ -191,6 +191,36 @@ const Pool: FunctionComponent = observer(() => {
...
+          pool?.id ?? "",
+          gauge.duration,
+          priceStore,
+          fiat,
+          allowedGauges
+        );
+
+      // if (externalApy.toString() !== "0%") {
+      //   debugger;
+      // }
```
Comment (jonator) [inline packages/web/pages/pool/[id].tsx:214]: Please remove

jonator (changes requested):

srph:
Oops, good catfch. Should be fixed now.

jonator (approved):
LGTM

PR #890, opened 2022-10-08
Correction of Gravity bridge deposit/withdraw link

luisqagt:
Current link points to https://spacestation.zone/ it should point to https://bridge.blockscape.network/


luisqagt:
Wanted to add that the https://www.gravitybridge.net/ site has updated the link as well.

luisqagt:
@JeremyParish69 do we need anything else for this?

JeremyParish69:
Thanks for this. It's hard to verify this as the 'official' portal site, since it looks like one validator's creation--except the fact that it is linked from the gravity bridge site. I'd like to ask, does this site allow for url parameters?

JeremyParish69 (approved):
LGTM

PR #897, opened 2022-10-12
Add Juno pairs to router

JohnnyWyles:
Allows use of stJUNO/JUNO and seJUNO/JUNO in the router by including pools with at least 10,000 JUNO present to be used.


jonator:
@JohnnyWyles I think you need to update your branch from our master, there's a build error.

PR #899, opened 2022-10-12
Localization

DLatouche:
Added translations in the app.
Added user setting to change language.
Added the component to change the language in the navigation bar.
Added two files with English and French translations.
Added two files to translate dayJs.

The majority of the application text has been changed.
The translation is done from a function retrieved from a hook (useTranslation) which takes the translation key as a parameter
and an object for integrating data into the translation.


jonator (changes requested):
Before I review- could you remove package-lock.json? We use yarn (I recommend using yarn).

packages/web/components/control/language-select.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { MenuDropdownIcon } from "./menu-dropdown-icon/menu-dropdown-icon";
+import { MenuDropdownIconItem } from "./types";
+import { MenuOptionsIconModal } from "../../modals";
+import { useTranslation } from "react-multi-lang";
+
+export type LanguageSelectProps = {
+  options: MenuDropdownIconItem[];
+};
+
+export const LanguageSelect: FunctionComponent<LanguageSelectProps> = observer(
```
jonator, line 16: May pain you for me to say this, but I think we should at least change the name and make it seem generic. I posted a comment in the Figma, I think asset select is not the right term given our work with crypto assets. That way we'd be more able to reuse this component. Seems like it won't require too many logic/type changes.

packages/web/components/control/language-select.tsx
```diff
@@ -0,0 +1,96 @@
...
+          .getUserSettingById("language")
+          ?.setState({ language: value });
+      },
+      [setDropdownOpen]
+    );
+
+    return (
+      <div className="relative">
+        <button
+          className="flex items-center border border-osmoverse-200 rounded-xl bg-osmoverse-900"
```
jonator, line 42: I think the default/rest state has a different border. This osmoverse-200 border seems to be for when the dropdown is open.

See rest state design:
https://www.figma.com/file/kYSFzlWoiMFrqkF46dtZQV/Osmosis-Design-Guidelines-2022?node-id=528%3A3785

packages/web/components/control/menu-dropdown-icon/menu-dropdown-icon-item.tsx
```diff
@@ -0,0 +1,51 @@
...
+  currentValue,
+  optionLength,
+}: Props) => {
+  const t = useTranslation();
+  return (
+    <button
+      className={classNames(
+        "px-4 py-2 cursor-pointer hover:bg-osmoverse-700 flex items-center ",
+        {
+          "text-rust-200": value === currentValue,
```
jonator, line 37: Don't believe the text rust applies to this type of dropdown.

See:
https://www.figma.com/file/kYSFzlWoiMFrqkF46dtZQV/Osmosis-Design-Guidelines-2022?node-id=579%3A4778

packages/web/components/control/language-select.tsx
jonator: See figma comment:

https://www.figma.com/file/kYSFzlWoiMFrqkF46dtZQV?node-id=312:3892#285175175

packages/web/components/control/index.ts
```diff
@@ -9,4 +9,6 @@ export * from "./switch";
 export * from "./tab-box";
 export * from "./toggle";
 export * from "./token-select";
+export * from "./language-select";
+export * from "./menu-dropdown-icon/menu-dropdown-icon";
```
jonator, line 13: I think we can just import from `./menu-dropdown-icon/`

packages/web/components/control/types.ts
```diff
@@ -20,3 +20,9 @@ export interface NumberSelectProps extends InputProps<number> {
   min: number;
   max: number;
 }
+
+export interface MenuDropdownIconItem {
+  value: string;
+  display: string;
+  image: string;
```
jonator, line 27: Could we change this to `iconUrl`? I think it clarifies it's a url.

packages/web/pages/_app.tsx
```diff
@@ -66,28 +83,42 @@ function MyApp({ Component, pageProps }: AppProps) {
...
         icon: "/icons/chart-white.svg",
         amplitudeEvent: [EventName.Sidebar.infoClicked] as AmplitudeEvent,
       }
     );
 
     return m;
-  }, []);
+  }, [t]);
 
+  // // Localization
```
jonator, line 108: If this is not needed, please remove.

packages/web/stores/user-settings/show-dust.tsx
```diff
@@ -23,10 +23,14 @@ export class ShowDustUserSetting implements IUserSetting<ShowDustState> {
   };
 
   constructor(fiatSymbol: string) {
-    this.displayLabel = `Show pools/assets < ${fiatSymbol} 0.01`;
+    this.fiatSymbol = fiatSymbol;
```
jonator, line 26: Low prio: we can save some lines of code by declaring a readonly property directly in the constructor as `protected readonly fiatSymbol: string`

It removes the need to declare in class body and initialize in constructor.

packages/web/stores/user-settings/language.tsx
```diff
@@ -0,0 +1,55 @@
...
+    image: "/images/flag-english.png",
+  },
+  {
+    value: "fr",
+    display: "settings.languages.fr",
+    image: "/images/flag-french.png",
+  },
+];
+
+export class LanguageUserSetting implements IUserSetting<LanguageState> {
```
jonator, line 22: Nice! Let me know if you can think of any other improvements that can be made to this user setting abstraction.

jonator (changes requested):

packages/web/pages/_app.tsx
```diff
@@ -23,31 +23,48 @@ import {
...
 import { useAmplitudeAnalytics } from "../hooks/use-amplitude-analytics";
+import {
+  setDefaultLanguage,
+  setTranslations,
+  useTranslation,
+} from "react-multi-lang";
+
+import en from "../localizations/en.json";
+import "../localizations/dayjs-locale-en.js";
+import fr from "../localizations/fr.json";
```
jonator, line 34: Can we try to use next/dynamic or something to import these as needed, instead of including it in the main bundle? It doesn't appear to be very large right now, but I'm more concerned about when we add more text and languages. IMO we should not have any other translations in the browser except the ones the user needs. May need to wrap it in a hook or something.

jonator (changes requested):

PR #913, opened 2022-10-20
Remaining erc20 Axl integration

JeremyParish69:
Added assets:

Ethereum (mainnet only):
WBTC
DAI
USDT
FRAX
LINK
AAVE
APE
AXS
MKR
RAI
SHIB
UNI
XCN

Moonbeam:
DOT
GLMR (Moonbeam + Moonbase)


jonator:
@arispech have we confirmed the integration works as expected for each asset?

arispech:
> @arispech have we confirmed the integration works as expected for each asset?
> 
> This is super exciting. Also, @JeremyParish69 I'd appreciate it if you listed out each asset you're adding for reference.

Yeah, the list of assets would be really helpful. I will retest every assets added to make sure

JeremyParish69:
@jonator @arispech expanded the description to list out all the added assets.

packages/web/integrations/axelar/types.ts
```diff
@@ -37,6 +38,7 @@ export const EthClientChainIds_AxelarChainIdsMap: {
   "Avalanche C-Chain": "Avalanche",
   "Binance Smart Chain": "Binance",
   "Fantom Opera": "Fantom",
+  "Moonbase": "Moonbeam",
```
jonator, line 41: It's called moonbase on Metamask?

arispech:
The integration works smoothly!

packages/web/integrations/axelar/types.ts
JeremyParish69: Moonbeam's docs call it 'Moonbase Alpha':
https://docs.moonbeam.network/builders/get-started/networks/moonbase/
Axelar docs call it 'Moonbase':
https://docs.axelar.dev/resources/testnet

Which is the preferred source of truth?

jonator: https://github.com/osmosis-labs/osmosis-frontend/blob/JP69/stage-erc-axl-xfer/packages/web/integrations/ethereum/types.ts#L28

Actually, it doesn't matter to me, we just need to make sure we add `"Moonbase"` as an entry in this mapping of Ethereum chain IDs (given from MetaMask/EVM wallets) to display names. These mappings are actually related in the code. This isn't clear though; could you please add to the comment above `EthClientChainIds_AxelarChainIdsMap` to clarify this? That the entries are related?

https://github.com/osmosis-labs/osmosis-frontend/blob/0aadd61fc774fc244dd382b11406d45fd2572e68/packages/web/integrations/ethereum/types.ts#L28-L44

JeremyParish69: Added comments. and updated Testnet name for Moonbase -> Moonbase Alpha. and removed ropsten -> Ethereum mapping

packages/web/integrations/axelar/types.ts
```diff
@@ -24,21 +24,23 @@ export type SourceChain =
...
+ *  ethClientChainIDs must be specified in ../ethereuem/types.ts::ChainNames{} to map the name to a chainID, which is in turn used to add the network to EVM-compatible wallets, like Metamask.
+ *
+ * AxelarChainIds must be specified in SourceChain{} and are used in ./source-chain-configs.ts::SourceChainConfigs{} as <asset>::<network>::id values.
  *
  *  Values not included as keys are assumed to be the same across chainlist and Axelar.
  */
 export const EthClientChainIds_AxelarChainIdsMap: {
   [ethClientChainIds: string]: SourceChain;
 } = {
-  "Ropsten Test Network": "Ethereum",
```
jonator, line 36: As mentioned in slack, we do need this to support MetaMask and Axelar on testnet

jonator (changes requested):

packages/web/integrations/axelar/types.ts
JeremyParish69: Looks like Axelar accepts 'ethereum-2' for Goerli as that is now the network being used, not Ropsten.

jonator: True. Ropsten is getting deprecated.

packages/web/integrations/ethereum/types.ts
```diff
@@ -24,23 +24,29 @@ export interface EthWallet
 
 export type SendFn = Pick<EthWallet, "send">["send"];
 
-/** Ethereum chains: https://docs.metamask.io/guide/ethereum-provider.html#chain-ids */
 export const ChainNames: { [chainId: string]: string } = {
-  "0x1": "Ethereum",
+  /** Ethereum chains: https://docs.metamask.io/guide/ethereum-provider.html#chain-ids */
+  "0x1": "Ethereum Main Network",
```
jonator, line 29: Why "Ethereum" -> "Ethereum Main Network"?

If these are different from `SourceChain` type, things will break. We reverse-map (right -> left, or name -> hex ID) to cache EVM chain IDs in the browser ([see](https://github.com/osmosis-labs/osmosis-frontend/blob/JP69/stage-erc-axl-xfer/packages/web/integrations/axelar/transfer.tsx#L78))

jonator (changes requested):

packages/web/integrations/axelar/types.ts
```diff
@@ -15,29 +17,46 @@ export interface AxelarBridgeConfig {
...
  *  Testnet: https://axelartest-lcd.quickapi.com/axelar/nexus/v1beta1/chains?status=1
  */
 export type SourceChain =
-  | "Ethereum"
-  | "Ropsten Test Network"
+  | "aurora"
   | "Avalanche"
+  | "binance"
+  | "Ethereum"
+  | "ethereum-2"
```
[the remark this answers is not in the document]
JeremyParish69: These values are exactly equal to what Axelar accepts. (As seen here: https://github.com/axelarnetwork/axelarjs-sdk/blob/main/src/chains/supported-chains-list.ts)
From what I could tell, the SourceChain is both: the Axelar-accepted value, as well as the displayed network name in this list. But I agree, this should instead have a form optimized for human readability.

packages/web/integrations/ethereum/types.ts
JeremyParish69: The reason for the change is because of the name for it shown here: https://docs.metamask.io/guide/ethereum-provider.html#chain-ids
The corresponding change was also made in the map: EthClientChainIds_AxelarChainIdsMap, corresponding the new EthClientChainId of Ethereum Main Network to the Axelar-accepted value of "Ethereum", so I don't think it will break. Upon testing, it still prompts my Metamask to switch to Ethereum Mainnet.
I can easily change it back to just 'Ethereum' if you like.

jonator (approved):

PR #916, opened 2022-10-21
Add up all external incentive APYs for a specific duration

srph:
In continuation of #886, this adds up all external incentive APYs for the given duration rather than just calculating it from the first incentive.

image

### DIFF
--- a/packages/stores/src/queries/pool-incentives/incentivized-pools.ts
+++ b/packages/stores/src/queries/pool-incentives/incentivized-pools.ts
@@ -119,7 +119,7 @@ export class ObservableQueryIncentivizedPools extends ObservableChainQuery<Incen
   );
 
   /**
-   * Computes external incentive APY for the given duration
+   * Computes all external incentive APY for the given duration
    */
   readonly computeExternalIncentiveAPYForSpecificDuration = computedFn(
     (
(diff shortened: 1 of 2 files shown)


jonator:
Actually- can we change it to find the incentives for a specific external gauge (given an ID), rather than adding them together for the duration?

Our new design breaks down the APRs by gauge:
Screen Shot 2022-10-21 at 3 38 43 PM

srph:
@jonator Some updates:

- Replaced allowedGauges with gaugeId and denom. I wasn't sure how to get the denom without having to explicitly pass it as a parameter.
- Moved sum computation in the pool page component.

Let me know if there's anything I may have missed. Thanks!

packages/stores/src/queries/pool-incentives/incentivized-pools.ts
```diff
@@ -121,49 +123,52 @@ export class ObservableQueryIncentivizedPools extends ObservableChainQuery<Incen
   /**
    * Computes external incentive APY for the given duration
    */
-  readonly computeExternalIncentiveAPYForSpecificDuration = computedFn(
+  readonly computeExternalIncentiveGaugeAPR = computedFn(
```
jonator, line 126: Nitpick: can we please update the comment for this function as well to use APR?

jonator (changes requested):
Looks great!

packages/stores/src/queries/pool-incentives/incentivized-pools.ts
```diff
@@ -119,51 +121,54 @@ export class ObservableQueryIncentivizedPools extends ObservableChainQuery<Incen
...
       priceStore: IPriceStore,
-      fiatCurrency: FiatCurrency,
-      allowedGauges: ExternalGauge[]
+      fiatCurrency: FiatCurrency
     ): RatePretty => {
-      if (!allowedGauges.length) {
+      const observableGauge = this.queryGauge.get(gaugeId);
+
+      if (
+        duration.asMilliseconds() !==
```
jonator, line 138: I'm realizing that since we take in the gaugeId, there's no real reason to take in the duration as well and perform this check. Can we remove the duration from the params?

jonator (changes requested):
Sorry noticed one more thing. Should be good after that.

packages/stores/src/queries/pool-incentives/incentivized-pools.ts
srph: @jonator Good catch, working on it =D

jonator: Please make sure the comment/docs is updated as well.

srph: I think the main reason why we need duration is - on the pool page - we're iterating all through the `ExternalIncentiveGaugeAllowList[pool.id]`. Would you like me move duration checking there?

srph:
Let me know if there's anything else I might have missed @jonator

jonator (approved):
Great work! Appreciate it

PR #917, opened 2022-10-21
Axelar config and handle mobile bridge integration wallets

jonator:
Publishing stage

Let me know if there are any major concerns @arispech .


arispech:
The UI looks good to me!

PR #918, opened 2022-10-23
Add Gravity Externals

JohnnyWyles:
As per Proposal 347 https://www.mintscan.io/osmosis/proposals/347 Will go live Epoch of 24th October.

### DIFF
--- a/packages/web/config/feature-flag.ts
+++ b/packages/web/config/feature-flag.ts
@@ -618,12 +618,7 @@ export const ExternalIncentiveGaugeAllowList: {
   ],
   "625": [
     {
-      gaugeId: "3327",
-      denom:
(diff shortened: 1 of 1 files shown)


JeremyParish69 (commented):
There are already objects for 633 and 625, so those should be modified rather than creating new ones. In addition, would be nice to keep pool numbers in order

JeremyParish69 (changes requested):
There are already objects for 633 and 625, so those should be modified rather than creating new ones. In addition, would be nice to keep pool numbers in order

JohnnyWyles:
> There are already objects for 633 and 625, so those should be modified rather than creating new ones. In addition, would be nice to keep pool numbers in order

Didn't realise they had previous incentives sorry, replaced these objects.
Was also going by Gauge order, will do pool order in future

JeremyParish69 (approved):

Issue #655, opened 2022-06-28
Deposit assets from trade page


jonator:
[Figma discussion](https://www.figma.com/file/wQjMyxY0EnEk29gBzGDMe5?node-id=4449:2029#216756478)

We should remove the roadblock for users loading the app for the first time having to go to the assets page to deposit before the go to the trade page to swap.

jonator:
6/28 sprint planning -

Waiting on design work @victortokaito

milesz8:
Is there a design you want implemented? I cant access the figma discussion. Otherwise I have some ideas

jonator:
@milesz8 don't think we have a design for this yet. Or decided we want to do this.

We're currently undergoing a frontend redesign. May want to look into another issue for now.

jonator:
One idea- what if when a user has a "from" asset selected with a positive balance, we can have the button change to deposit? They can click that, then the UI is locked until the tx completes, then the button says swap. Seems like a simple workflow that doesn't require another control @miles

jonator:
9/29 meeting update -

We decided the best design would be to link to the assets page from the swap page if they have a 0 balance in Osmosis for the "from" asset. This would help inform users of the assets page, and would serve as a tutorial for how assets are deposited, withdrawn, and displayed.

It would work as follows from the swap page:
1. If a user has a 0 Osmosis balance for the asset in the "from" page, the swap button (or a link near it, design TBD) would show: "Deposit x ASSET from _network_" if they've approved the counterparty in Keplr, or "Deposit from _network_" otherwise
2. If they click the deposit button from the swap widget, we'd link to the assets page with query parameters in the URL to determine which transfer modal to have automatically opened.
3. User can carry out standard transfer from the transfer modal.
4. (Optional/stretch) when the transfer completes we'd keep the modal open and the "Transfer" button would be replaced with a "Back to Swap" button to let the user go back to the swap page with the query parameters populated to the relevant assets. (We've also considered having a link to Swap page in the success toast message)

victortokaito:
['Begin a deposit from the swap page' Prototype](https://www.figma.com/file/om93lu8iXOhxwGHSUpAXEP/Trade-UI?node-id=667%3A5573)
* link to figma [frame](https://www.figma.com/file/om93lu8iXOhxwGHSUpAXEP/Trade-UI?node-id=667%3A6751)

PR #925, opened 2022-10-26
Deposit from trade page

milesz8:
https://github.com/osmosis-labs/osmosis-frontend/issues/655

In this pull request
- Adds the ability to share link to specific swap modals on assets page
- Adds ability to deposit from the swap page by directing to assets page
   - Uses the share link to navigate to assets page for the selected swap asset


packages/web/pages/assets/index.tsx
```diff
@@ -157,18 +169,26 @@ const Assets: NextPage = observer(() => {
...
-        onDeposit={(chainId, coinDenom, externalDepositUrl) => {
+        onDepositIntent={() => {
+          transferConfig.startTransfer("deposit");
+          setTransferQueryParams("deposit");
+        }}
+        onWithdrawIntent={() => {
+          transferConfig.startTransfer("withdraw");
+          setTransferQueryParams("withdraw");
+        }}
+        onDeposit={(coinDenom, externalDepositUrl) => {
```
jonator, line 180: I think the cause of the bug was removing the first param here. I'd add it back with an `_` rather than go into the assets table and edit the code. Chain ID could be useful in the far future if a chain is insistent on having the same token denom as a token on another chain.

Applies to `onWithdraw` below as well.

packages/web/pages/assets/index.tsx
```diff
@@ -33,6 +33,7 @@ import {
 } from "../../hooks";
 import { WalletConnectQRModal } from "../../modals";
 import { EventName } from "../../config";
+import { useTokenTransferQueryParams } from "../../hooks/window/use-assets-transfer-query-params";
```
jonator, line 36: Add `use-assets-transfer-query-params` to the index file in the hooks/window folder. Then just import it in the existing `../../hooks` import above.

packages/web/stores/assets/transfer-ui-config.ts
```diff
@@ -129,15 +129,9 @@ export class ObservableTransferUIConfig {
   /** ### DEPOSIT/WITHDRAW FROM TABLE ROW
    *  User wants to transfer asset with counterparty IBC chain and coinDenom known. */
   @action
-  async transferAsset(
-    direction: TransferDir,
-    chainId: string,
-    coinDenom: string
-  ) {
+  async transferAsset(direction: TransferDir, coinDenom: string) {
```
jonator, line 132: Let's update the comment above to say `with the direction and coinDenom known`.

Could be beneficial to add an additional note that the coinDenom is expected to be unique across chains.

package.json
```diff
@@ -55,6 +55,5 @@
     "prettier": "^2.5.1",
     "ts-jest": "^27.1.2",
     "typescript": "4.5.4"
-  },
-  "dependencies": {}
```
jonator, line 59: Please also remove this change. (All changes to package.json files, for that matter.)

jonator (changes requested):
* When opening global deposit/withdraw buttons, only set `transfer=` once an asset is selected (currently I see `transfer=` with nothing to the right of the `=`)

packages/web/pages/assets/index.tsx
milesz8: clean

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -232,6 +236,28 @@ export const TradeClipboard: FunctionComponent<{
     [tradeTokenInConfig.expectedSwapResult.amount]
   );
 
+  var isDepositToken =
```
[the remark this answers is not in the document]
milesz8: done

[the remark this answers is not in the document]
jonator: Actually, my linter is complaining, please use `const` wherever a variable is not changed.

packages/web/stores/assets/transfer-ui-config.ts
milesz8: does that look good?

packages/web/components/trade-clipboard/index.tsx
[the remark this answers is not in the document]
milesz8: makes sense. I switched to const where possible

packages/web/localizations/en.json
```diff
@@ -15,6 +15,7 @@
     "available": "Available",
     "MAX": "MAX",
     "HALF": "HALF",
+    "DEPOSIT": "DEPOSIT",
```
jonator, line 18: Please add `fr` translations as well. Deposit should be found elsewhere in the `fr.json` file.

jonator (changes requested):

packages/web/localizations/en.json
milesz8: done

arispech:
@milesz8 As discussed on slack: 

Aris
  I did the test of deposit assets on the swap page. The deposit went through and it redirected me to the assets page to check the amount you just deposit. However, the transfer history table show "withdraw" in the type column.
image

Aris  
The ability to share links to specific swap models on the assets page - works:white_check_mark:
The share link to navigate to the assets page for the selected swap asset - works:white_check_mark:

Jon Ator
 There may be some design changes. Currently, I think it's a bit jarring to immediately link the user to the assets page, we may need some sort of pop-up to let the user know we are linking. Do you agree 
@arispech @victortokaito 

Aris
 Yeah! It feels a bit weird to me when it bumps me to the assets page, to be honest. Kind of like oh I'm still swapping and have to click back to the swap page to continue swapping. Maybe that's just me, as a regular user I trust the DEX so I do multiple swaps and do not check the asset every time after a swap

victor
I really like the deposit button blending in, but I think it makes sense to some sort of message appear before linking. Should deposit appear as an option any time a user has assets detected on source chain? Or should it only appear if they have 0 of selected asset on Osmosis?
If we decide the latter, we can remove half and max if user has 0 assets. That might help address this edge case as well (photo)
image

Aris
 I do like the deposit button as well when the selected asset has 0. Just the redirect to the assets page that I'm not sure about

jonator (changes requested):
Please fix merge conflicts

jonator:
Dev work seems fine.. putting on hold since none of us are satisfied with jumping user to assets page. We're thinking (@milesz8 and I) it may be better to open transfer modal directly in swap page.

jonator:
Expired for now

PR #926, opened 2022-10-26
Update nodes for Secret Network

assafmo:


JeremyParish69:
Thank you @assafmo . Please, what is the purpose of this change? Should this instead be added to the Chain Registry?

delivan:
It should've been updated on Keplr proxy server, not osmosis frontend. Keplr endpoints changed rpc/lcd at that time.

PR #938, opened 2022-10-28
Add dockerfile

honsontran:


daniel-farina:
@jonator @niccoloraspa 
This PR is for docker files needed to deploy Osmosis on decentralized cloud provider Flux.  I think we should probably make a folder to categorize this sort of files. 

What do guys think would be a good place for it? 

/services/flux 
or
.github/cloudproviders/flux ? 


Thanks

niccoloraspa:
There are no Flux specific files in this commit so there is no need for additional folders.

The `Dockerfile` and the `docker-compose.yaml` can stay both at the root level.

We should add environment variables to both files though.

jonator:
I agree. Also having the Dockerfile at the root may be needed for proper function by Docker.

honsontran:
> Thanks @honsontran for the contribution.
> 
> There are no Flux specific files in this commit so there is no need for additional folders.
> 
> The `Dockerfile` and the `docker-compose.yaml` can stay both at the root level.
> 
> Both files need to expose the environment variables used in the deployment.

Hi @niccoloraspa, just wanted to make sure, what environment variables are you referring to? I can make the adjustments if needed.

PR #939, opened 2022-10-29
Update feature-flag.ts

manolius148:
Add gauge #29872 for pool gamm/pool/788


manolius148:
Vercel check fails, so I can't merged it when approved. How can I override this?

manolius148:
Awesome, thanks!

PR #941, opened 2022-10-30
Add IST

zmanian:


JeremyParish69 (changes requested):
Not sure that we can put 'isVerified'.

Otherwise, it LGTM. The chain-infos data and img were already uploaded, so this should suffice. Price is still undefined, but that's OK for now.

zmanian:
Why would BLD be verified and IST not be verified?

jonator:
That's our way of marking incentivized/strategic assets to be included in the main app vs frontier.

zmanian:
Changed

Are we good to merge?

the Decentralized Cooperation Foundation is planning incentivize the IST pools asap.

Issue #789, opened 2022-09-12
Transfer modal: status indicator


victortokaito:
### Problem
IBC status indicators on info.osmosis.zone are useful, but not easily discoverable because they're accessed offsite.

### Solution
Implement "just-in-time" ibc status indicators on the transfer modal:
* This information comes to the user right at the step it becomes relevant to them (at least on the osmosis app). Indicators' utility improves along with its relevance by helping users understand why transactions may be slowed.
* The animation transforms to match the error code (using new design guidelines’ negative gradient to indicate network is congested). 

https://www.figma.com/file/qze5yJO9TEw7vysInR07Ui?node-id=1832:113709#262708753

PR #943, opened 2022-10-31
Add status indicator to transfer modal

etian511:
This PR addresses this issue: [link](https://github.com/osmosis-labs/osmosis-frontend/issues/789)

This PR will query for the IBC status of a chain and display if the chain is congested or not, informing the user at the step it becomes relevant to them if they will experience delays in the transfer or not.

Currently this PR is a draft.

- [x] Add query
- [x] update view based on status

### DIFF
--- a/packages/stores/src/queries-external/ibc-status/index.ts
+++ b/packages/stores/src/queries-external/ibc-status/index.ts
@@ -6,9 +6,17 @@ import { ObservableQueryExternalBase } from "../base";
 import { IbcStatus } from "./types";
 
 export class ObservableQueryIbcStatus extends ObservableQueryExternalBase<IbcStatus> {
-  constructor(kvStore: KVStore, baseURL: string, _counterPartyChainID: string) {
+  constructor(
+    kvStore: KVStore,
+    baseURL: string,
+    protected readonly counterPartyChainID: string
+  ) {
     super(kvStore, baseURL, `/ibc/v1/all`);
     makeObservable(this);
+  }
+
+  protected canFetch(): boolean {
+    return this.counterPartyChainID !== "";
   }
 
   readonly getIbcStatus = computedFn((counterPartyChainID: string): string => {
--- a/packages/stores/src/queries-external/ibc-status/index.ts
+++ b/packages/stores/src/queries-external/ibc-status/index.ts
@@ -29,14 +29,21 @@ export class ObservableQueryIbcStatus extends ObservableQueryExternalBase<IbcSta
 }
 
 export class ObservableQueryIbcStatuses extends HasMapStore<ObservableQueryIbcStatus> {
-  constructor(kvStore: KVStore, baseUrl: string) {
+  constructor(
+    kvStore: KVStore,
+    ibcStatusBaseUrl = "https://api-osmosis-chain.imperator.co"
+  ) {
     super(
       (counterPartyChainID) =>
-        new ObservableQueryIbcStatus(kvStore, baseUrl, counterPartyChainID)
+        new ObservableQueryIbcStatus(
+          kvStore,
+          ibcStatusBaseUrl,
+          counterPartyChainID
+        )
     );
   }
 
-  get(counterPartyChainID: string): ObservableQueryIbcStatus {
-    return super.get(counterPartyChainID);
+  get(counterPartyChainID: string) {
+    return super.get(counterPartyChainID) as ObservableQueryIbcStatus;
   }
 }
--- a/packages/stores/src/queries-external/ibc-status/index.ts
+++ b/packages/stores/src/queries-external/ibc-status/index.ts
@@ -3,47 +3,46 @@ import { HasMapStore } from "@keplr-wallet/stores";
 import { makeObservable } from "mobx";
 import { computedFn } from "mobx-utils";
 import { ObservableQueryExternalBase } from "../base";
-import { IbcStatus } from "./types";
+import { IbcMetrics, IbcStatus } from "./types";
 
-export class ObservableQueryIbcStatus extends ObservableQueryExternalBase<IbcStatus> {
-  constructor(
-    kvStore: KVStore,
-    baseURL: string,
-    protected readonly counterPartyChainID: string
-  ) {
-    super(kvStore, baseURL, `/ibc/v1/all`);
+export class ObservableQueryIbcDepositStatus extends ObservableQueryExternalBase<IbcMetrics> {
+  constructor(kvStore: KVStore, baseURL: string, counterPartyChainID: string) {
+    // If we are depositing, destination is osmosis
+    super(
+      kvStore,
+      baseURL,
+      `/ibc/v1/source/${counterPartyChainID}/destinationosmosis?minutes_trigger=-1`
+    );
     makeObservable(this);
   }
 
-  protected canFetch(): boolean {
-    return this.counterPartyChainID !== "";
-  }
+  readonly getIbcStatus = computedFn(
+    (counterPartyChainID: string): IbcStatus => {
+      const ibcRaw = this.response?.data.data.find(
+        (statusMetric) => statusMetric.channel_id === counterPartyChainID
+      );
+      if (!ibcRaw) {
+        return IbcStatus.Blocked;
+      }
 
-  readonly getIbcStatus = computedFn((counterPartyChainID: string): string => {
-    const ibcRaw = this.response?.data.data.find(
-      (statusMetric) => statusMetric.channel_id === counterPartyChainID
-    );
-    if (!ibcRaw) {
-      return "Congested";
+      // TODO: figure out what metrics constitute congested or not
+      if (ibcRaw.size_queue > 1_000) {
+        return IbcStatus.Congested;
+      } else {
+        return IbcStatus.OK;
+      }
     }
-
-    // TODO: figure out what metrics constitute congested or not
-    if (ibcRaw.size_queue > 1_000) {
-      return "Congested";
-    } else {
-      return "OK";
-    }
-  });
+  );
 }
 
-export class ObservableQueryIbcStatuses extends HasMapStore<ObservableQueryIbcStatus> {
+export class ObservableQueryIbcDepositStatuses extends HasMapStore<ObservableQueryIbcDepositStatus> {
   constructor(
     kvStore: KVStore,
     ibcStatusBaseUrl = "https://api-osmosis-chain.imperator.co"
   ) {
     super(
       (counterPartyChainID) =>
-        new ObservableQueryIbcStatus(
+        new ObservableQueryIbcDepositStatus(
           kvStore,
           ibcStatusBaseUrl,
           counterPartyChainID
@@ -52,6 +51,56 @@ export class ObservableQueryIbcStatuses extends HasMapStore<ObservableQueryIbcSt
   }
 
   get(counterPartyChainID: string) {
-    return super.get(counterPartyChainID) as ObservableQueryIbcStatus;
+    return super.get(counterPartyChainID) as ObservableQueryIbcDepositStatus;
+  }
+}
+
+export class ObservableQueryIbcWithdrawStatus extends ObservableQueryExternalBase<IbcMetrics> {
+  constructor(kvStore: KVStore, baseURL: string, counterPartyChainID: string) {
+    // If we are withdrawing, destination is counterPartyChainID
+    super(
+      kvStore,
+      baseURL,
+      `/ibc/v1/source/osmosis/destination${counterPartyChainID}?minutes_trigger=-1`
+    );
+    makeObservable(this);
+  }
+
+  readonly getIbcStatus = computedFn(
+    (counterPartyChainID: string): IbcStatus => {
+      const ibcRaw = this.response?.data.data.find(
+        (statusMetric) => statusMetric.channel_id === counterPartyChainID
+      );
+      if (!ibcRaw) {
+        return IbcStatus.Blocked;
+      }
+
+      // TODO: figure out what metrics constitute congested or not
+      if (ibcRaw.size_queue > 1_000) {
+        return IbcStatus.Congested;
+      } else {
+        return IbcStatus.OK;
+      }
+    }
+  );
+}
+
+export class ObservableQueryIbcWithdrawStatuses extends HasMapStore<ObservableQueryIbcWithdrawStatus> {
+  constructor(
+    kvStore: KVStore,
+    ibcStatusBaseUrl = "https://api-osmosis-chain.imperator.co"
+  ) {
+    super(
+      (counterPartyChainID) =>
+        new ObservableQueryIbcWithdrawStatus(
+          kvStore,
+          ibcStatusBaseUrl,
+          counterPartyChainID
+        )
+    );
+  }
+
+  get(counterPartyChainID: string) {
+    return super.get(counterPartyChainID) as ObservableQueryIbcWithdrawStatus;
   }
 }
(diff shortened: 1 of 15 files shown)


packages/stores/src/queries-external/store.ts
```diff
@@ -29,3 +38,39 @@ export class ObservableQueryExternalBase<
     super(kvStore, instance, urlPath);
   }
 }
+
+export class ObservableQueryIbcStatus extends ObservableQueryExternalBase<IbcStatus> {
```
jonator, line 42: I actually just added an external query on another branch myself. Did you put the IBC status class defs here because of a circular dependency issue? I'd recommend trying to move it into an index.ts file in the `/ibc-status/` folder.

If you run into an issue, you can leave this here. It can be moved later. My solution was to create a `base.ts` file here and moving the base object there.

packages/web/components/complex/transfer.tsx
```diff
@@ -92,7 +92,8 @@ export const Transfer: FunctionComponent<TransferProps> = ({
       : isMobile
       ? 14
       : 24;
-
+  // Add query to congestion status here
+  const ibcCongestionStatus = useState(false);
```
jonator, line 96: This comes with experience with the codebase- but this component is not just used to display IBC transfers, but also transfers from Ethereum via our Axelar integration (check out USDC or WETH transfer modals).

For that reason, I don't think we should put IBC-specific logic here, but instead we should add more generic capabilities to this component, or update & reuse the `warningMessage` prop we already have. Then, the logic for accessing the IBC status would likely live somewhere else (I'll let you find it :)).

jonator (changes requested):
I'm impressed with your work so far given no FE exp. It's clear you can figure stuff out on your own to some degree. Should you join, you'll likely become very comfortable with the codebase in around 6 months.

packages/stores/src/queries-external/store.ts
etian511: Yup gotcha, that would be consistent with the pattern set by `pool-fees` looks like

packages/stores/src/queries-external/ibc-status/index.ts
```diff
@@ -0,0 +1,49 @@
+import { KVStore } from "@keplr-wallet/common";
+import { HasMapStore } from "@keplr-wallet/stores";
+import { makeObservable } from "mobx";
+import { computedFn } from "mobx-utils";
+import { ObservableQueryExternalBase } from "../base";
+import { IbcStatus } from "./types";
+
+export class ObservableQueryIbcStatus extends ObservableQueryExternalBase<IbcStatus> {
+  constructor(kvStore: KVStore, baseURL: string, _counterPartyChainID: string) {
+    super(kvStore, baseURL, `/ibc/v1/all`);
```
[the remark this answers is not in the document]
jonator: It could get space-inefficient to query all statuses and then store them for every osmosis-counterparty pair

jonator (changes requested):

packages/stores/src/queries-external/ibc-status/index.ts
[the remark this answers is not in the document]
jonator: From dilan:

instead of for each counterPartyChainID , if you want:
from X to osmosis => https://api-osmosis.imperator.co/ibc/v1/source/counterPartyChainID/destinationosmosis-1?minutes_trigger=-1
from osmosis to X => https://api-osmosis.imperator.co/ibc/v1/source/osmosis-1/destinationcounterPartyChainID?minutes_trigger=-1

jonator:
Also seems like there's build errors rn

packages/stores/src/queries-external/ibc-status/index.ts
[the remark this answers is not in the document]
etian511: Yeah that makes sense to use the source<>destination query. At first I thought there was a way to use counterPartyChainID as the only key, but would be inefficient to loop through all the pairs. Will we always be transferring between X and Osmosis? I think there are other pairs right? So we would need to pass in both the source and the destination to the `ObservableQueryExternalBase `constructor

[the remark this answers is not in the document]
jonator: For our purposes, we should only expect to query for Osmosis and the counterparty. Though, we may need to create a map store for each direction, deposit or withdraws.

Perhaps, you could create another store that wraps 2 map stores for each direction. Then has functions that accept a direction (of type `"deposit" | "withdraw") and a counterparty chain ID. Then it calls the appropriate map store depending on direction.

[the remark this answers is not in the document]
etian511: Hmmm ok, for now I made two queries, one for withdrawing and one for depositing, not ideal though as it has some duplicate code

packages/stores/src/queries-external/store.ts
```diff
@@ -24,5 +31,13 @@ export class QueriesExternalStore {
       priceStore,
       poolRewardsBaseUrl
     );
+    this.queryIbcDepositStatuses = new ObservableQueryIbcDepositStatuses(
+      kvStore,
+      ibcStatusBaseUrl
```
jonator, line 36: IMO we should combine these into a single store with the same name you originally had. Then (in the same file) you can have a map store for deposits and withdrawals, each mapping from counterpartyChainId->ibcStatusQuery.

packages/stores/src/queries-external/ibc-status/types.ts
```diff
@@ -0,0 +1,18 @@
...
+    channel_id: string;
+    token_symbol: string;
+    token_name: string;
+    last_tx: string;
+    size_queue: number;
+    duration_minutes: number;
+  }[];
+}
+
+export enum IbcStatus {
```
jonator, line 14: With TypeScript, you can define a `const enum` that will map to 0-indexed integers at compile time.

jonator (changes requested):

jonator:
Each map store would have these as the query URLs:
withdraw => `https://api-osmosis.imperator.co/ibc/v1/source/counterPartyChainID/destinationosmosis-1?minutes_trigger=-1`
deposit => `https://api-osmosis.imperator.co/ibc/v1/source/osmosis-1/destinationcounterPartyChainID?minutes_trigger=-1`

You'd just replace `counterpartyChainID` with whatever the caller passes.

jonator (changes requested):

jonator:
@etian511 

I made a branch with what I was thinking regarding a single store:

https://github.com/osmosis-labs/osmosis-frontend/blob/5533a57672af62301d02abb3085e091d0fa6e7bf/packages/stores/src/queries-external/ibc-status/index.ts#L9-L44

packages/web/components/complex/transfer.tsx
```diff
@@ -36,6 +37,7 @@ export type TransferProps = {
   warningMessage?: string;
   toggleIsMax: () => void;
   transferFee?: CoinPretty;
+  ibcStatus?: IbcStatus;
```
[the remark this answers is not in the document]
etian511: Ah yup should not have added any ibc specific logic here. I will keep it generic

packages/stores/src/queries-external/ibc-status/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    super(
+      kvStore,
+      baseURL,
+      `/ibc/v1/source/osmosis/destination${counterPartyChainID}?minutes_trigger=-1`
+    );
+    makeObservable(this);
+  }
+
+  readonly getIbcStatus = computedFn(
+    (counterPartyChainID: string): IbcStatus => {
```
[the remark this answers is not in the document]
etian511: Ah good point, I'll use `sourceChannelId` to get the specific channel we are looking for. Yeah noticed that was odd when I first did this query: 

`curl -X 'GET' \
  'https://api-osmosis.imperator.co/ibc/v1/source/osmosis/destinationjuno?minutes_trigger=-1' \
  -H 'accept: application/json'`

and got this response:

`[
  {
    "source": "osmosis-1",
    "destination": "juno-1",
    "channel_id": "channel-42",
    "token_symbol": "OSMO",
    "token_name": "Osmosis",
    "last_tx": "",
    "size_queue": 0,
    "duration_minutes": 0
  },
  {
    "source": "osmosis-1",
    "destination": "neta (juno-1)",
    "channel_id": "channel-169",
    "token_symbol": "OSMO",
    "token_name": "Osmosis",
    "last_tx": "",
    "size_queue": 0,
    "duration_minutes": 0
  }
]`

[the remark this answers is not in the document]
etian511: Looks like the sourrce/destination does not have to be a complete string match, so we need channelDs to get the right one

etian511:
> @etian511
> 
> I made a branch with what I was thinking regarding a single store:
> 
> https://github.com/osmosis-labs/osmosis-frontend/blob/5533a57672af62301d02abb3085e091d0fa6e7bf/packages/stores/src/queries-external/ibc-status/index.ts#L9-L44

ah ok this makes much more sense! Thanks Jon!

packages/web/components/complex/transfer.tsx
```diff
@@ -13,6 +13,7 @@ import { BorderButton } from "../buttons";
 import { CheckBox } from "../control";
 import { Disableable, InputProps, LoadingProps } from "../types";
 import { useTranslation } from "react-multi-lang";
+import { IbcStatus } from "@osmosis-labs/stores/types/queries-external/ibc-status/types";
```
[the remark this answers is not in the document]
etian511: I added the line to the end of the `index.ts` file but am still seeing this import error: 

`Module not found: Can't resolve '@osmosis-labs/stores/types/queries-external/ibc-status/index' in '/Users/eugenetian/src/osmosis-frontend/packages/web/modals'`

Should I just be importing `@osmosis-labs/stores`?

packages/stores/src/queries-external/ibc-status/types.ts
etian511: Hmmm weird I am getting this error when setting the enum to const. 

`Type error: Cannot access ambient const enums when the '--isolatedModules' flag is provided.`

packages/web/components/complex/transfer.tsx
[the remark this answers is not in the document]
jonator: Yes just @osmosis-labs/stores.

Also, you may also need to export * from the ibc-status folder in the index file one level up.

The way exporting/importing works on the frontend is likely new to you, especially with the way index files are handled.

packages/stores/src/queries-external/ibc-status/index.ts
```diff
@@ -0,0 +1,143 @@
...
+    );
+    makeObservable(this);
+  }
+
+  readonly getIbcStatus = computedFn((sourceChannelId: string): IbcStatus => {
+    const ibcRaw = this.response?.data.data.find(
+      (statusMetric) => statusMetric.channel_id === sourceChannelId
+    );
+    if (!ibcRaw) {
+      return IbcStatus.Undefined;
```
jonator, line 110: I think it'd be cleaner to return `undefined` instead of this enum type.

packages/web/components/complex/transfer.tsx
```diff
@@ -36,6 +36,7 @@ export type TransferProps = {
   warningMessage?: string;
   toggleIsMax: () => void;
   transferFee?: CoinPretty;
+  transferPathHealth?: "Healthy" | "Congested" | "Blocked" | "Undefined";
```
jonator, line 39: Wouldn't we only need congested or blocked? The question mark next to the member means it doesn't need to be there, which is the undefined case as well. That handles all other cases, including when it's not blocked.

packages/web/components/complex/transfer.tsx
```diff
@@ -281,6 +282,28 @@ export const Transfer: FunctionComponent<TransferProps> = ({
...
+          </GradientView>
+        )}
+        {transferPathHealth && transferPathHealth == "Blocked" && (
+          <GradientView
+            className="text-center"
+            gradientClassName="bg-gradient-negative"
+            bgClassName="bg-osmoverse-900"
+          >
+            <span className="body2 text-rust-500 md:caption">
+              {"Transfer network currently blocked. Check back later."}
```
jonator, line 303: It's a chore we'll have to deal with now, but we need this to be translated in all languages.

packages/web/modals/ibc-transfer.tsx
```diff
@@ -32,6 +38,22 @@ export const IbcTransferModal: FunctionComponent<ModalBaseProps & IbcTransfer> =
 
     const [didAckWithdrawRisk, setDidAckWithdrawRisk] = useState(false);
 
+    const transferHealth = isWithdraw
+      ? queriesExternalStore.queryIbcStatuses.getIbcStatus(
```
jonator, line 42: Please update this to the new simpler API we made.

jonator (changes requested):

packages/stores/src/queries-external/ibc-status/index.ts
etian511: Got it

packages/web/modals/ibc-transfer.tsx
etian511: Is this not the simpler API we made? We pass in the `direction`, `counterPartyChainID`, and `sourceChannelD` in one query?

packages/web/components/complex/transfer.tsx
etian511: Yeah that makes sense, I'll update to just `Congested` | `Blocked`

packages/web/components/complex/transfer.tsx
etian511: ah ok, I see how localizations are done. I'll add translations to the language` json` files. OK to just use google translate and input what these sentences translate to?

jonator: Yeah that works for now. We'll need to periodically ask language speakers to look over new ones as we accumulate them.

packages/web/modals/ibc-transfer.tsx
```diff
@@ -32,6 +38,21 @@ export const IbcTransferModal: FunctionComponent<ModalBaseProps & IbcTransfer> =
...
+          counterpartyChainId,
+          props.sourceChannelId
+        )
+      : queriesExternalStore.queryIbcStatuses.getIbcStatus(
+          "deposit",
+          counterpartyChainId,
+          props.sourceChannelId
+        );
+    const transferPathHealth =
+      (transferHealth == IbcStatus.Congested && "Congested") ||
```
jonator, line 53: Please use `===` instead of `==` in js. It checks the types as well.

https://www.scaler.com/topics/javascript/difference-between-double-equals-and-triple-equals-in-javascript/

jonator (changes requested):
Looking at the info site (https://info.osmosis.zone/ibc), it seems like the CHEQ withdraw page should appear blocked. I looked in the network tab and it appears the query is returning 404. I'm working with Dilan to make sure the API accepts all of our chain IDs.

Also, I'd work with Dilan to make sure we're using the right thresholds for congested and blocked (if you haven't already).

packages/stores/src/queries-external/index.ts
```diff
@@ -1,8 +1,10 @@
 export * from "./pool-fees";
 export * from "./pool-rewards";
 export * from "./store";
-
+export * from "./ibc-status";
 export const IMPERATOR_HISTORICAL_DATA_BASEURL =
   "https://api-osmosis.imperator.co";
 export const IMPERATOR_TX_REWARD_BASEURL =
   "https://api-osmosis-chain.imperator.co";
+export const IMPERATOR_IBC_STATUS_BASEURL =
```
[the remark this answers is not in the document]
jonator: Actually- I must have been mistaken, but I think the right base URL is https://api-osmosis.imperator.co/

I think that should start providing results that we can see in the frontend, vs the 404s I've been seeing.

PR #944, opened 2022-10-31
added fund external incentives gauges + linting

jo-chainmasters:


packages/web/config/feature-flag.ts
```diff
@@ -1159,6 +1162,23 @@ export const ExternalIncentiveGaugeAllowList: {
         "ibc/80825E8F04B12D914ABEADB1F4D39C04755B12C8402F6876EE3168450C0A90BB",
     },
   ],
+  "830": [
+    {
+      gaugeId: "29850",
```
JeremyParish69, line 1167: All three of these gauges look like the default internal perpetual gauges. This list is for whitelisting external, non-perpetual gauges.

jo-chainmasters: @JeremyParish69 Your are right. Sorry, my wrong. Get the correct gaugeId is hard. But fixed now

JeremyParish69 (approved):
LGTM

PR #956, opened 2022-11-04
Add ALTER, BUT, SHD, SIENNA & stkd-SCRT (Secret SNIP-20 tokens)

assafmo:


packages/web/config/chain-infos.ts
```diff
@@ -136,6 +136,46 @@ const chainInfos = (
...
         },
+        {
+          type: "cw20",
+          contractAddress: "secret12rcvz0umvk875kd6a803txhtlu7y0pnd73kcej",
+          coinDenom: "ALTER",
+          coinMinimalDenom:
+            "cw20:secret12rcvz0umvk875kd6a803txhtlu7y0pnd73kcej:ALTER",
+          coinDecimals: 6,
+          coinGeckoId: "alter",
+          coinImageUrl: "/tokens/alter.jpg",
```
JeremyParish69, line 147: do we have a png or svg(preferred) with transparent corners?

packages/web/config/chain-infos.ts
```diff
@@ -136,6 +136,46 @@ const chainInfos = (
...
+        },
+        {
+          type: "cw20",
+          contractAddress: "secret1qfql357amn448duf5gvp9gr48sxx9tsnhupu3d",
+          coinDenom: "SHD",
+          coinMinimalDenom:
+            "cw20:secret1qfql357amn448duf5gvp9gr48sxx9tsnhupu3d:SHD",
+          coinDecimals: 8,
+          coinGeckoId: "shade-protocol",
+          coinImageUrl: "/tokens/shd.jpg",
```
JeremyParish69, line 167: png or svg is preferred...

packages/web/config/chain-infos.ts
```diff
@@ -136,6 +136,46 @@ const chainInfos = (
           isStakeCurrency: true,
           isFeeCurrency: true,
         },
+        {
+          type: "cw20",
+          contractAddress: "secret12rcvz0umvk875kd6a803txhtlu7y0pnd73kcej",
+          coinDenom: "ALTER",
```
JeremyParish69, line 142: I don't think any of these tokens have been added to the chain registry

packages/web/config/chain-infos.ts
assafmo: Yes, I'll update it

packages/web/config/chain-infos.ts
assafmo: Yes, I'll update soon

packages/web/config/chain-infos.ts
assafmo: Those assets live on Secret Network for years now. I'll add them to the chain registry too.

packages/web/config/chain-infos.ts
JohnnyWyles: The SHD png/svg was added to the registry this morning if you need it.
https://github.com/cosmos/chain-registry/tree/master/secretnetwork/images

mradkov: Sienna waiting to be added to the registry in this PR: https://github.com/cosmos/chain-registry/pull/985

mohammedpatla: @JeremyParish69 https://github.com/cosmos/chain-registry/pull/986

packages/web/config/chain-infos.ts
DrPresident: @JeremyParish69 can you sanity check this PR for me? one thing I'm concerned about is whether or not the `coinMinimalDenom` should be `cw20:address` or `cw20:address:symbol` as I see done both ways in different places
https://github.com/assafmo/osmosis-frontend/pull/2

packages/web/config/chain-infos.ts
JeremyParish69: FYI, The chain registry appears to now have nice images for each of these tokens' logos.

assafmo:
Ready for review now.

packages/web/config/ibc-assets.ts
```diff
@@ -163,6 +163,44 @@ export const IBCAssetInfos: (IBCAsset & {
     coinMinimalDenom: "uscrt",
     isVerified: true,
   },
+  {
+    counterpartyChainId: "secret-4",
+    sourceChannelId: "channel-476",
```
JeremyParish69, line 168: Won't treat this as a blocker, but we should have this channel registered to chain reg as well...

assafmo: Cc @DrPresident

mohammedpatla: @JeremyParish69 This PR is ready https://github.com/cosmos/chain-registry/pull/997

packages/web/config/ibc-assets.ts
```diff
@@ -163,6 +163,44 @@ export const IBCAssetInfos: (IBCAsset & {
...
+    destChannelId: "channel-39",
+    coinMinimalDenom: "cw20:secret1qfql357amn448duf5gvp9gr48sxx9tsnhupu3d:SHD",
+    ics20ContractAddress: "secret1tqmms5awftpuhalcv5h5mg76fa0tkdz4jv9ex4",
+  },
+  {
+    counterpartyChainId: "secret-4",
+    sourceChannelId: "channel-476",
+    destChannelId: "channel-39",
```
JeremyParish69, line 176: This looks incorrect to me. I'm getting a different counterparty(secretnetwork) channel ID...
image

packages/web/config/ibc-assets.ts
JeremyParish69: @mohammedpatla still seems to be some errors... Looks like the connection info submitted here and to Chain Registry isn't quite right...

packages/web/config/ibc-assets.ts
JeremyParish69: looks to be channel-44 in the other file. So ya probably just a repeated typo in this file.

mohammedpatla: Yes looks like another typo @assafmo can you just change the IDs to 44 as is in our document?

assafmo: Fixed

packages/web/config/ibc-assets.ts
assafmo: Should be fixed now on master

packages/web/config/ibc-assets.ts
```diff
@@ -1030,6 +1068,46 @@ export const IBCAssetInfos: (IBCAsset & {
     coinMinimalDenom: "ujkl",
     isUnstable: true,
   },
+  {
+    // ALTER
+    counterpartyChainId: "secret-4",
+    sourceChannelId: "channel-476",
+    destChannelId: "channel-44",
+    coinMinimalDenom: "cw20:secret12rcvz0umvk875kd6a803txhtlu7y0pnd73kcej",
+    depositUrlOverride: "https://wrap.scrt.network",
```
JeremyParish69, line 1077: Please forgive my lack of expertise with Secret. Is wrapping and unwrapping really required for these tokens? The wrapping function appears to be greyed out for all these tokens (and doesn't list stk-scrt at all), and it doesn't look like there is any sTOKEN equivalent for these. It's also interesting the it's only provided for deposit and withdraw...
Also, it seems that this URL is used only for wrapping, but not for transferring between chains.

I'm thinking these deposit URLs should probably be removed....

mohammedpatla: @JeremyParish69 Actually, IBC transfers work fine; you need to connect to Keplr and click on the double arrow next to the icon to IBC asset to Deposit and Withdraw. BAD UX yes this is being worked on to improve with a new page.

Yes, all the other tokens need to be wrapped and unwrapped (the native tokens of other chains). But SHD, ALTR, Sienna, BUTT, and stkd-scrt are tokens (SNIP20s) on the SCRT network, so they are already wrapped and converted directly to IBC (CW20) assets, so NO Wrapping is necessary (and hence greyed out).

stkd-SCRT is missing and needs to be added waiting for an update on that myself.

CC: @assafmo correct me if i am wrong on anything.

assafmo: @mohammedpatla is correct

assafmo:
What do we need to do to add these tokens to the main website too?

packages/web/config/ibc-assets.ts
mohammedpatla: stkd-scrt open to be added to wrap as well https://github.com/scrtlabs/wrap.scrt.network/pull/16

mohammedpatla:
@JeremyParish69 Let us know what the next steps are and how can we get it on the main app as assafmo pointed out. ShadeProtocol is getting ready with a proposal for the pool on Osmosis as well, so any points that will push this forward is a plus.

packages/web/config/ibc-assets.ts
JeremyParish69: Thanks for the explanation.
I have two concerns with this deposit link. The first concern being that, although wrapping and unwrapping might be required for certain functions (such as LPing) for certain tokens, I don't think that this external site is required to IBC these tokens across, as it looks like that can be done directly from the Osmosis site using the normal generated ibc transactions. We want to avoid having to link to external sites wherever possible unless absolutely necessary. So, I'm still thinking that the deposit URL overrides aught to be removed.
The second concern being that if this site truly is required to IBC the tokens around, then I'm thinking it should probably be used for both deposit and withdrawal--not only deposit. (although, I still think the URL override should be removed as long as it's still possible to IBC them around).
If removing the deposit URL override doesn't break the ability to IBC transfer these tokens, then can we please remove them?

mohammedpatla: @assafmo is the only one who can comment on this. I am not entirely sure if this would work without the wrap page

JeremyParish69:
> What do we need to do to add these tokens to the main website too?

To get added to the main site goes by community governance, either if a pool containing the token gets Osmosis liquidity mining incentives or a simple text/"signalling" proposal to add to the main site.

packages/web/config/ibc-assets.ts
```diff
@@ -163,6 +163,44 @@ export const IBCAssetInfos: (IBCAsset & {
     coinMinimalDenom: "uscrt",
     isVerified: true,
   },
+  {
```
JeremyParish69, line 166: Why are the tokens listed in ibc-assets twice each? Please remove these first five. Also I think these ics20 contract addresses are likely needed on the latter five.

packages/web/config/ibc-assets.ts
assafmo: To send from Secret you have to send a secret contract tx which involves encryption and custom code in this UI. That's the reason for the override.

To send back from Osmosis it's just like any other IBC token.

packages/web/config/ibc-assets.ts
assafmo: This looks the tokens to Secret Network. I modeled this PR after how the NETA token is defined since it's the same concept (cw20 token from Juno)

JeremyParish69: yes, but they should not be listed twice each. Please remove the ones from line 166-203, but please add these ics20ContractAddresses to the listings starting at line 1071

mohammedpatla: @JeremyParish69 this should be resolved.

JeremyParish69 (approved):
LGTM

assafmo:
Thanks @JeremyParish69!

Issue #959, opened 2022-11-04
Add Polish translation


pawarm:
According to [this](https://github.com/osmosis-labs/osmosis-frontend/issues/312#issuecomment-1216942779), 4% of osmosis users speak Polish, and it's a 5th largest group. 

I think it makes sense for it to be one of the first available translations on the site after the redesign.

PR #960, opened 2022-11-04
Added Polish translation

pawarm:
[Related issue](https://github.com/osmosis-labs/osmosis-frontend/issues/959)

Polish is my native language, I have been using osmosis and other DEXs for a long time and I have enough experience in frontend. 

So I prepared a Polish translation based on my own knowledge and Polish translations of other dexes.

Let me know if any additional review by another Polish-speaking person will be needed here


packages/web/localizations/fr.json
```diff
@@ -319,6 +319,7 @@
     "titleLanguage": "Langue",
     "titleShowDust": "Afficher les pools/actifs < 0.01{fiatSymbol}",
     "languages": {
```
jonator: Please update your branch from `feat/redesign`. I just removed "languages" key from all the json files since I display the name of each language in that language in the menu.

packages/web/stores/user-settings/language.tsx
```diff
@@ -20,6 +20,11 @@ const SUPPORTED_LANGUAGES: MenuDropdownIconItemProps[] = [
     display: "settings.languages.fr",
     iconUrl: "/images/flag-french.png",
   },
+  {
+    value: "pl",
+    display: "settings.languages.pl",
+    iconUrl: "/images/flag-polish.png",
```
jonator: As mentioned in other comment, the display key here should be hardcoded to "Polish" in polish. Also, we decided to not identify languages with flags of countries. Please update your branch and TypeScript should let you remove `iconUrl` key.

jonator (changes requested):
Thanks again for contributing!

packages/web/localizations/fr.json
pawarm: Yeah I noticed that right after making a PR, already fixed, the other one too

PR #961, opened 2022-11-04
Feat/update dockerfiles

niccoloraspa:
This PR adds the Dockerfile to build osmosis frontends for frontier and testnets and a GH action to automatically push the images on every new commit to master.


niccoloraspa:
Running the workflow in a fork to test it: https://github.com/niccoloraspa/osmosis-frontend/actions/runs/3395603667/jobs/5645700125

docker/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  frontend-testnet:
+    build:
+      context: ../
+      dockerfile: ./Dockerfile.testnet
+    ports:
+      - '3002:3000'
+    profiles:
+      - testnet
+      - all
```
jonator, line 33: What is the compose file used for? What scenario would we run all these services together?

niccoloraspa: They are all mutually exclusive unless you select the profile `all`.

By the default it runs the main frontend but you can select frontier or testnet.

If you want, you could run them all together by selecting the all profile. It could be used in some testing scenarios.

PR #965, opened 2022-11-05
Add Turkish Language

HuginnAcademy:
Turkey has one of the largest crypto communities. We would like to add Turkish language option for Osmosis.


jonator:
Please target `feat/redesign` branch

PR #996, opened 2022-11-11
Feat Display Swap Route Info

JoseRFelix:
## What's Changed

Add Swap route info to trade clipboard

### Before

image

### After 

image


packages/web/components/denom/types.ts
```diff
@@ -0,0 +1,7 @@
+import { AppCurrency, Currency } from "@keplr-wallet/types";
+
+export interface DenomImageProps {
```
jonator, line 3: Opinion: I like to inline props in the component unless they are particularly reusable. Like with `ModalBaseProps`. Generally, that's how I like to approach all type definitions.

jonator: Typescript is great, but I want to maintain and grok code, not types

packages/web/components/denom/denom-image.tsx
```diff
@@ -0,0 +1,29 @@
+import { FunctionComponent } from "react";
+import Image from "next/image";
+
+import { DenomImageProps } from "./types";
```
jonator, line 4: Minor: Could we move this into assets/ folder. That's where we've had asset-related components. Yes, I'm aware it clashes with assets from a web dev perspective (images, icons, etc.)

packages/web/components/denom/denom-image.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const DenomImage: FunctionComponent<DenomImageProps> = ({
+  denom,
+  size = 24,
+}) => {
+  return (
```
jonator, line 10: Nitpick: if there's no content in body of function I try to return a single expression to simplify code (the content of the `return (...)` statement)

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -850,9 +857,22 @@ export const TradeClipboard: FunctionComponent<{
 
           <div
             className={classNames(
-              "relative rounded-lg bg-osmoverse-900 px-4 md:px-3 transition-all ease-inOutBack duration-300 overflow-hidden",
-              showEstimateDetails ? "h-56 py-6" : "h-11 py-[10px]"
+              "relative rounded-lg bg-osmoverse-900 px-4 md:px-3 transition-all ease-inOutBack duration-300 overflow-hidden"
             )}
+            style={{
```
jonator, line 862: I'd prefer to only use style as last result.

You can use an object as a parameter of `classNames` function above to conditionally apply utility classes. See [this](https://tailwindcss.com/docs/padding#arbitrary-values) on how to assign arbitrary values in tailwind.

Also, IMO these const values are unneeded and clutter the file. IMO here it's fine to just hardcode the values. Const defs like these are more for highly-consequential hardcoded values, i.e. how many cards to display in the view.

packages/web/components/trade-clipboard/trade-route.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { Pool } from "@osmosis-labs/pools";
+import { observer } from "mobx-react-lite";
+import { FunctionComponent } from "react";
+import { useStore } from "../../stores";
+import { ChainStore } from "../../stores/chain";
+import { DenomImage } from "../denom";
+import { Tooltip } from "../tooltip";
+import { TradeRouteProps } from "./types";
+
+function getDenomsFromPool(chainStore: ChainStore, pool: Pool) {
```
jonator, line 12: Interesting way to organise code in file-private functions.

packages/web/components/trade-clipboard/trade-route.tsx
```diff
@@ -0,0 +1,111 @@
...
+      getPoolsWithDenomAndFee(chainStore, pool)
+    );
+    const poolsWithReorderedDenoms = reorderPathDenoms(
+      sendCurrency,
+      poolsWithDenomAndFee
+    );
+
+    return (
+      <div className="space-y-2">
+        <h6 className="body2">Auto Router</h6>
```
jonator, line 69: Please add translations via google translate.

jonator (commented):
Great work. Mostly nitpicks. And on nitpicks: to start there will likely be a lot of back and forth between us as we converge on styles/preferences on how we do things. In general though, we should maintain the conventions already established in the code.

packages/web/components/denom/types.ts
JoseRFelix: Thanks for the review! 

> Opinion: I like to inline props in the component unless they are particularly reusable. Like with `ModalBaseProps`. Generally, that's how I like to approach all type definitions.

Good to know. I thought `types.ts` files were the convention for all types in general including components. Will update this.

packages/web/components/trade-clipboard/index.tsx
JoseRFelix: Sounds good. I also don't like to lean into using style given its performance. I used it since can't use variable interpolation in Tailwind, and we need to calculate the height of the content dynamically.  

I'll remove the const values

packages/web/components/trade-clipboard/trade-route.tsx
JoseRFelix: I like creating pure functions to make it easier to break down the code and debug it.

jonator (approved):
Wow. Well done. I'll go ahead and merge soon. We'll prob update the design later but this is a great place to start and would be interesting/useful to users to know what's going on behind the scenes.

jonator:
@JoseRFelix design team is coming up with a Figma design then we could get this to prod. Please DM me on telegram with a good email I could use to share the Figma with you.

jonator:
PR pending on whether we want to add stableswap and OSMO fee discount info to vis when v13 goes live

JoseRFelix:
Closing in favor of #1093 PR to allow preview deployments and simplify git workflow

PR #997, opened 2022-11-11
Some Spanish translations fixed

iam-agf:
The current translation is not accurate because lacks of context in many cases and the text doesn't work as how it should. 
Some references have strong issues that even damages the variables inside the text because the translation seems like has been poorly done via Google translator without any care (sorry if I'm wrong, but it really looks like that) and this requires an urgent fix because makes the website look like a scam site. I'll finish the rest in the weekend.

Some examples:

1) slippage variable here was translated so doesn't work (swap > settings > slippage)
image

2) Uppercase word "MITAD" for "Half" 
image

3) "Bonded" value doesn't match with meaning of "Garantizado" that means "Granted" in Spanish.
image


packages/web/localizations/es.json
```diff
@@ -2,7 +2,7 @@
   "app": {
     "banner": {
       "title": "Estás viendo todos los activos sin permiso",
-      "linkText": "Estás viendo todos los activos sin permiso. <Enlace> Haga clic aquí para volver a la aplicación principal </link>."
+      "linkText": "Estás viendo todos los activos sin permiso. <link> Haga clic aquí para volver a la aplicación principal </link>."
```
jonator, line 5: Do we want to change this? It seems to be markup to be processed later.

jonator (changes requested):

packages/web/localizations/es.json
iam-agf: "Enlace" is the translation of the word "link" in Spanish. If you check the English, French, Korean, etc translations, they have the word "link". It seems to be one of the several reasons why I think it's a google translation.

PR #1004, opened 2022-11-14
Update protobuf for v13

Thunnini:
[osmosis-temp.zip](https://github.com/osmosis-labs/osmosis-frontend/files/10004202/osmosis-temp.zip)

There is a problem because the compile method of protobuf that we are currently using is outdated.
For now, use the above method.

Use
```
yarn install
sh ./scripts/predefine-proto.sh
sh ./scripts/define-proto.sh
```

The output is created in the ./output directory.
In predefine-proto.sh, you can select files to compile.


jonator:
Will run again once queries are completed

PR #1006, opened 2022-11-14
Internal math for stableswap

AlpinYukseloglu:
This PR implements relevant math/calc functions for stableswap. Specifically, it implements the following:
* `calcOutGivenIn`
* `calcInGivenOut`
* `calcSpotPrice`
* `solveCfmm`

To align precision bounds with our Go implementation, we also implement a wider precision decimal type for internal calculations

TODO:
* More calc tests
* Basic spot price tests
* BigDec implementation
* Precision issue fixes

### DIFF
--- a/packages/math/src/pool/__tests__/stable.spec.ts
+++ b/packages/math/src/pool/__tests__/stable.spec.ts
@@ -1,5 +1,5 @@
 import * as StableMath from '../stable';
-import { Coin, Dec, Int } from '@keplr-wallet/unit';
+import { Coin, Dec, Int, DecUtils } from '@keplr-wallet/unit';
 
 describe('Test stableswap math', () => {
   describe('calcOutGivenIn', () => {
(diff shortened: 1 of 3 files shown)


packages/math/src/pool/__tests__/stable.spec.ts
```diff
@@ -1,89 +1,362 @@
...
-  describe("calcOutGivenIn", () => {
-    test("even pool basic trade", () => {
+describe('Test stableswap math', () => {
+  describe('calcOutGivenIn', () => {
+    test('even pool basic trade', () => {
       const poolAssets: StableMath.StableSwapToken[] = [
         {
           amount: new Dec(1_000_000_000),
-          denom: "foo",
+          denom: 'foo',
```
jonator, line 10: Hmm it seems your vscode setup (prob prettier) is changing double quotes to single

My only hunch right now is you're missing eslint? Try installing this extension below and see if that fixes it. If not, DM me.

https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

packages/math/src/pool/stable.ts
```diff
@@ -175,13 +249,14 @@ export function binarySearch(
...
+  throw Error('binary search did not converge');
 }
 
-function compare_checkMultErrorTolerance(
+export function compare_checkMultErrorTolerance(
   expected: Dec,
   actual: Dec,
-  tolerance: Dec
+  tolerance: Dec,
+  roundingMode: string,
```
jonator, line 259: A somewhat unique feature of TypeScript- we can enforce the shape of this string by replacing the `string` type with a mores specific union of literals: `"roundDown" | "roundUp"`.

AlpinYukseloglu: Neat!

PR #1012, opened 2022-11-15
Calculate prices with pool weights

JeremyParish69:
Token prices now use Osmosis spot price instead of CoinGecko.

Changes almost all token prices to instead refer to the spot price of it against OSMO (where such a <TOKEN>/OSMO pool exists, with sufficient liquidity) and then in turn to OSMO's spot price against USDC(.axl).
For some tokens, there isn't any large OSMO pool, but there might be another pool(, e.g., stATOM/ATOM,) so an extra spot price is used. (In this case, stATOM -> ATOM -> OSMO -> USDC.)
CoinGecko Ids are sometimes used where there is no large pool, but there does exist a price feed from CoinGecko.
Alas, some tokens still have no price data.

After spot checking some price conversions, the prices--since they're nearly all based on their OSMO pool--are much better at showing the affect on value by price impact (but not swap fee).


jonator:
I noticed now the frontend makes one much smaller call to prices API

jonator:
Please fix conflicts, updated target to stage

jonator (approved):

PR #1018, opened 2022-11-17
Update ECH IBC Channels

JeremyParish69:
src: channel-403, dst: channel-11

New IBC denom hash can be confirmed by using their transfer website: https://app.ech.network/ibc
SHA256("transfer/channel-403/aechelon") = ibc/47EE...177D
This will make the old denom unofficial. Users' old ECH tokens would then be unlisted in the Assets page (should we change it to "ECH (old)"?

Currently looks like this:
image

Getting odd behaviour from the token search:
image

Edit: The complexity of this seemingly simple change as blown up massively. It might be more easily possible if there were a way to change between channels (but still only seeing 1 version at a time). But at this point it is my recommendation to consider abandoning the idea of displaying 2 version at once, and simply ask users to view pools containing ECH to see, trade, and LP their old-channel ECH (it cannot be withdrawn anyway).

We could simply update the the latest channel, removing the old-channel token. The biggest downsides to this are no longer seeing it in the Asset page, and not finding in the Swap page.

Pool 747 (old-channel ECH/OSMO) would look like this (now with the channel name included):
image

While 848 (new-channel ECH/OSMO) would appear canonical, and is included in the Swap page trade router:
image

### DIFF
--- a/packages/web/config/chain-infos.ts
+++ b/packages/web/config/chain-infos.ts
@@ -1736,42 +1736,14 @@ const chainInfos = (
       rpc: "https://rpc-echelon.whispernode.com/",
       rest: "https://lcd-echelon.whispernode.com/",
       chainId: "echelon_3000-3",
-      chainName: "Echelon (current)",
+      chainName: "Echelon",
       bip44: {
         coinType: 60,
       },
       bech32Config: Bech32Address.defaultBech32Config("echelon"),
       currencies: [
         {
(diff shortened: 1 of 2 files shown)


kittyxxSoftPaws:
Just a note on this. Pool 848 is now up to 1513 OSMO at the time of writing. https://frontier.osmosis.zone/pool/848

PR #1021, opened 2022-11-19
Implement `BigDec` type for Stableswap math

AlpinYukseloglu:
This PR implements the `BigDec` type, which is an internal custom decimal type we use to accommodate our stableswap math without breaking precision standards in our standard `Dec`


packages/math/src/pool/stable.ts
```diff
@@ -1,4 +1,5 @@
 import { Coin, Dec, DecUtils } from "@keplr-wallet/unit";
+// import { BigDec } from "../big-dec";
```
jonator, line 2: Should we go ahead and use it in our stables calculations?

jonator (approved):
LGTM

jonator:
Also may want to run `yarn build` before committing- seems the type def file is missing.

PR #1022, opened 2022-11-19
Update stableswap solver to use new math and convert it to `BigDec`

AlpinYukseloglu:
This PR updates our stableswap solver to use the most up-to-date math described in our spec. It also converts all core logic to use our custom decimal type implemented in #1021 while leaving the external-facing interface still on `Dec`

It also fixes the solver tests to so they ensure that k increases with any core operation (i.e. any slippage/rounding error is in the pool's favor) instead of passing trivially

### DIFF
--- a/packages/math/src/pool/stable.ts
+++ b/packages/math/src/pool/stable.ts
@@ -5,11 +5,6 @@ export const StableSwapMath = {
   calcOutGivenIn,
   calcInGivenOut,
   calcSpotPrice,
-  // secondary
-  solveCfmm,
-  compareDec_checkMultErrorTolerance,
-  cfmmConstantMultiNoV,
-  calcWSumSquares,
 };
 
 const oneBigDec = new BigDec(1);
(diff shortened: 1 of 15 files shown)


packages/math/src/pool/stable.ts
```diff
@@ -1,29 +1,38 @@
...
+import { BigDec } from "../big-dec";
 
 export const StableSwapMath = {
   calcOutGivenIn,
   calcInGivenOut,
   calcSpotPrice,
   // secondary
   solveCfmm,
-  compare_checkMultErrorTolerance,
+  compareDec_checkMultErrorTolerance,
```
jonator, line 10: We should only add these if they need to be exposed to users of the package (the web package, currently), not just tests

jonator:
Would I just merge #1022 and not #1021 ?

packages/math/src/pool/stable.ts
```diff
@@ -1,29 +1,38 @@
...
-const oneDec = new Dec(1);
+const oneBigDec = new BigDec(1);
 
 export type StableSwapToken = {
   amount: Dec;
   denom: string;
   scalingFactor: number;
 };
 
+export type BigDecStableSwapToken = {
```
jonator, line 23: Remove export

AlpinYukseloglu:
> Would I just merge #1022 and not #1021 ?

That works!

PR #1025, opened 2022-11-20
Fix #1011 Non-localized stablecoin type

### LINKED ISSUE #1011: Non-localized stablecoin type
When you add **one** asset for liquidity to a pool that involves a stable coin, the last word is called from the variable `{mechanism}`, forcing it to be in English. 

image

In a same LP view, you can notice the word "Unbounded" isn't affected by the translation.
image

Also when you go to a LP and Bond Liquidity, there are words that aren't translated.
image

sehyunc:
Fix #1011 Non-localized stablecoin type
`add-liquidity.tsx`
* use collateral type to key into localization file

`dayjs-locale-es.js`
* update to Day.js locale object as detailed in [docs](https://day.js.org/docs/en/i18n/i18n)

`dayjs-locale-ko.js`
* same as above

`es.json`
* add stablecoin type translation

`_app.tsx`
* set locale object for day.js localization as detailed in [docs](https://day.js.org/docs/en/customization/customization)

Would like a sanity check that this is the way to go for the above solutions. If so, I can update all the localization json specific to day.js, but will need additional translations for stablecoin types to add to general localization files.

Could not replicate bug of "Unbonded" not being translated, although it's wrapping in certain languages.
Screenshot 2022-11-19 at 6 56 59 PM

### DIFF
--- a/packages/web/modals/keplr-connection-selection.tsx
+++ b/packages/web/modals/keplr-connection-selection.tsx
@@ -41,7 +41,7 @@ export const KeplrConnectionSelectModal: FunctionComponent<
           />
           <div className="flex flex-col text-left ml-5">
             <div className="flex items-center gap-2">
-              <h6>{t("installKeplr")}</h6>
+              <h6>{t("keplr.install")}</h6>
               <Image
                 src="/icons/external-link-white.svg"
                 alt="external link"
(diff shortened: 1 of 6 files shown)


sehyunc:
also included fix for a key error in the Connect Wallet Modal

packages/web/modals/keplr-connection-selection.tsx
```diff
@@ -41,7 +41,7 @@ export const KeplrConnectionSelectModal: FunctionComponent<
           />
           <div className="flex flex-col text-left ml-5">
             <div className="flex items-center gap-2">
-              <h6>{t("installKeplr")}</h6>
+              <h6>{t("keplr.install")}</h6>
```
jonator, line 44: Nice catch!

jonator (approved):
Appreciate the various cleanups. LGTM

PR #1028, opened 2022-11-21
Fix content overflow on zh-tw

PikachuEXE:
Before fix
image
image

After fix
(No image yet I am at work add back later)


PikachuEXE:
Note: Just added a commit to fix content overflow/text wrapping issue in **any** language with unicode/full width chars
Since for English only 1 word is in those sub containers, it works without `whitespace-nowrap`
image

PR #1036, opened 2022-11-22
Memoize pool price calculation

jonator:
It seems faster for less assets, but enabling _all_ assets to be calculated from pools is still a bit too much.


packages/stores/src/price/pool.ts
```diff
@@ -48,80 +49,70 @@ export class PoolFallbackPriceStore
...
-      if (route) {
-        const pool = this.queryPool.getPool(route.poolId);
-        if (!pool) {
-          return;
-        }
+  readonly getPrice = computedFn(
+    (coinId: string, vsCurrency?: string): number | undefined => {
+      if (!vsCurrency) {
+        vsCurrency = this.defaultVsCurrency;
+      }
```
ValarDragon, line 56: So computedFn is just memoizing this, were we calling this function repeatedly a lot before?

Am in support of making it memoized / computed!

packages/web/config/chain-infos.ts
```diff
@@ -873,9 +873,9 @@ const chainInfos = (
           coinDenom: "XKI",
           coinMinimalDenom: "uxki",
           coinDecimals: 6,
-          coinGeckoId: "ki",
-          //coinGeckoId: "pool:uxki",
-          coinImageUrl: "/tokens/xki.svg",
+          // coinGeckoId: "ki",
+          coinGeckoId: "pool:uxki",
+          // coinImageUrl: "/tokens/xki.svg",
```
JeremyParish69, line 878: didn't mean to comment out the img here?

packages/stores/src/price/pool.ts
jonator: Yeah basically everything is memoized, for some reason this wasn't. So if given the same input, it will return a cached input, which is an absolute need with how often components get rendered. However, with _all_ asset prices being computed from pool weights, the one-time (due to memo) recursion still causes quite a delay.

jonator:
@JeremyParish69 thoughts on performance with all assets? We can revert som major ones on coingecko if not satisfactory

packages/stores/src/price/pool.ts
jonator: A global ⌘⇧F on `computedFn` will come up with a _lot_ of results

PR #1042, opened 2022-11-23
 Feat New Token Select in Swap Widget

JoseRFelix:


jonator:
Wow your a beast! This is way better. Like how we simply get more visibility with the swap details drawer expanded

jonator:
Since you're already here improving token select, can we add support for using arrow keys + enter to select tokens? I know tab+enter is working. But, I think arrow keys should skip the recommended tokens if there is a query in the search box

JoseRFelix:
> Since you're already here improving token select, can we add support for using arrow keys + enter to select tokens? I know tab+enter is working. But, I think arrow keys should skip the recommended tokens if there is a query in the search box. Also w/ tab, we should select the whole background as a button instead of the token content on the left side of the row. Making the outer wrapper a button w/ an onClick event instead of the token name and icon should be enough.

I'll take a look. Thanks for the suggestion!

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -388,14 +388,14 @@ export const TradeClipboard: FunctionComponent<{
     return (
       <div
         className={classNames(
-          "relative rounded-[18px] flex flex-col gap-8 md:gap-6 bg-osmoverse-800 px-6 md:px-3 pt-12 md:pt-4 pb-8 md:pb-4",
```
jonator, line 391: Can we clip the bottom of the drawer except on mobile? Would look nice to contain the drawer within the swap widget.

packages/web/localizations/zh-cn.json
```diff
@@ -304,7 +304,10 @@
       "mobileMenu": "排序"
     },
     "MAX": "最大",
-    "searchTokens": "搜索代币"
+    "searchTokens": "搜索",
+    "selectToken": {
+      "title": "选择一个代币"
```
jonator, line 309: Could you merge from master and add Taiwanese? I think a recent PR added Taiwanese to master branch.

packages/web/components/trade-clipboard/index.tsx
JoseRFelix: I wanted to do this, but the height of the swap widget is too small so not much content fits the select. I thought about increasing the `min-height` but that would change the overall design. I'll leave a comment in design to see what we can do

jonator (commented):
UI bug: looks like when you've selected an item at the bottom of the list, it has square corners.

Also, @syed8rahman what do you think about the drawer popping up outside of the tool, instead of clipping inside (like with the bond card drawers). It think it's a bit incongruent to have the top darkened area of the tool be usable for collapsing the tool with the token select popping up from outside the tool, if that makes sense.

Screen Shot 2022-11-25 at 11 35 30 AM

JoseRFelix:
> UI bug: looks like when you've selected an item at the bottom of the list, it has square corners.

Good catch, addressed 👍

packages/web/components/trade-clipboard/index.tsx
jonator: Sounds fair. I sent a message to @syed8rahman to see what he thinks about the popup not being clipped. I think it's a bit odd for it not to clip _and_ for there to be a negative area at the top of the tool that can be used to dismiss it.

jonator (commented):
https://github.com/osmosis-labs/osmosis-frontend/blob/jonator/v13/packages/web/components/utils.ts

I like your idea of a top level utils folder. Could we move these into that folder then update the imports accordingly? IMO we should not have an index file and importers should update directly from the respective file.

packages/web/utils/assertion.ts
```diff
@@ -0,0 +1,3 @@
+export function isFunction(value: any): value is Function {
+  return typeof value === "function";
```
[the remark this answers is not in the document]
JoseRFelix: I like the util since it's easier to read `isFunction` instead of `typeof value === "function"`. Also, it follows a functional paradigm making it easier to use in `pipe` or `compose` functions. I'm open to leave it as `typeof` though. Please let me know your thoughts.

packages/web/pages/pool/[id].tsx
```diff
@@ -392,7 +392,7 @@ const Pool: FunctionComponent = observer(() => {
               </div>
               <div className="flex items-center gap-10 lg:gap-3 xl:place-content-between xl:w-full lg:flex-col lg:w-fit lg:items-start">
                 <div>
-                  <span className="text-osmoverse-400 body2 gap-2">
+                  <span className="gap-2 text-osmoverse-400 body2">
```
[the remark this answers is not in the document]
Comment (JoseRFelix) [reply packages/web/pages/pool/[id].tsx:395]: There's a prettier plugin that comes from the official Tailwind team. I think it could be useful for our code base in order to keep this order.

syed8rahman:
Screen Shot 2022-12-07 at 2 57 37 PM

1. The Subheader should also be Caption 5 - looks too large right now.
2. spacing between quick select items up top should be 8px. Figma reference below:
3. Token font should be "Button" 
Screen Shot 2022-12-07 at 2 59 09 PM

packages/web/components/cards/bond-card.tsx
```diff
@@ -341,22 +350,24 @@ const IncentiveBreakdownRow: FunctionComponent<
 > = ({ dailyPoolReward, apr, numDaysRemaining }) => {
   const t = useTranslation();
   return (
-    <div className="flex items-start place-content-between">
-      <div className="flex items-center shrink-0 gap-2">
-        <h6 className="text-osmoverse-200">+{apr.maxDecimals(0).toString()}</h6>
+    <div className="flex place-content-between items-start">
+      <div className="flex shrink-0 items-center gap-2">
+        <h6 className="subtitle1 text-white">
```
jonator, line 355: Anywhere there's h6 and a top-level subtitle1 class, let's change it to a `span` element.

jonator (changes requested):
I still think we should clip the drawer within the swap tool. Additionally, I've seen the scroll bar glitch breifly on on the swap page as it animates up and briefly overflows the page on vertically small screens.

packages/web/package.json
```diff
@@ -45,6 +47,7 @@
     "mobx-react-lite": "^3.2.2",
     "mobx-utils": "^6.0.4",
     "next": "12.0.7",
+    "prettier-plugin-tailwindcss": "^0.2.0",
```
jonator, line 50: Could this be a dev dependency?

jonator (approved):

jonator:
@JoseRFelix Not really related to this PR, but I noticed for long chain names it causes a wrap and increase in height of swap page.

We should prevent text wrapping as we have plenty of room on the right.

Screen Shot 2022-12-12 at 6 36 31 PM

packages/web/package.json
JoseRFelix: It should be. Let me update this

jonator:
@JoseRFelix please fix merge conflicts

JoseRFelix:
@jonator Merge conflicts have been addressed 👍

jonator (approved):

jonator:
My only other minor comment, which is optional to address, is preventing the drawer from rerendering when a user selects an item. Currently, when they click an item, the list and preferred tokens update to reflect the new selection (which is correct functionality) but doesn't feel clean UI wise.

PR #1044, opened 2022-11-23
Fix Docker GitHub Actions 

niccoloraspa:
This PR removes the support for `arm64` on the docker images as they fail to build on arm.

This should restore the Github Action functionality to build and push new docker images on every commit to `master`.


niccoloraspa:
I have also removed some duplicate Docker-related files from the root repository that go re-added by https://github.com/osmosis-labs/osmosis-frontend/pull/978.

All docker stuff it's in the `docker` folder

jonator (approved):

PR #1048, opened 2022-11-23
Make NETA, BOOT unverified

JeremyParish69:
Removing a couple more assets (NETA, BOOT) from Main due to disqualification.


mrlp4:
> Removing a couple more assets (NETA, BOOT) from Main due to disqualification.

Hi! 
Could you please point to the rules which define the parameters for qualification\disqualification?

jonator:
Assets that are in incentivized pools qualify

thomasWos:
Stupid question...
Why is the attribute not called isIncentivized?

Issue #968, opened 2022-11-07
Bridge Integration: Withdraw: regen deposit address when changing account


jonator:
### Problem
With the Axelar bridge integration, while a user is depositing, changing the account in MetaMask causes no issues since generating a deposit address only takes in the destination address (on Osmosis in this case). So, it doesn't matter which account funds are sent from. However, when withdrawing, switching the account in MetaMask (the destination address) causes the generated deposit address to no longer be valid. Sending funds to it would transfer the funds to the address selected while the modal was first opened.

### Proposed Solution
Instead, we should handle this by regenerating the account address every time and _only_ when `address` in the snippet below changes, causing this `useEffect` hook to re-run and invoke the Axelar SDK deposit address API. I faintly remember having this enabled before, but I added the last check (`!depositAddress`) to prevent the `doGen` from being run when other deps besides just `address` is changed. AKA instead of getting the right deps arrays in this and subsequent memo hooks right, I added this check and introduced this bug. Perhaps an additional if statement should be added to just handle the address changing, rather than the component mounting which is already in the snippet.

https://github.com/osmosis-labs/osmosis-frontend/blob/7ae91cb2e8044b2674e7c2eff333a67b3c249fb5/packages/web/integrations/axelar/hooks/use-deposit-address.ts#L60-L64

jonator:
Alternative/fallback solution if nothing else works. Keep track of the address the Axelar deposit address was generated for, and if the MetaMask account is different, display an error and disable button.

PR #1056, opened 2022-11-25
Regenerates withdraw address when changed

milesz8:
https://github.com/osmosis-labs/osmosis-frontend/issues/968

When a users account is changed the axelar deposit address is regenerated for the new account.


jonator:
Testing
- [ ] Changing Ethereum address when depositing doesn't cause a refresh.

PR #1058, opened 2022-11-25
Feat Scrollbars on windows

JoseRFelix:
## What's Changed

Before

image

After
image

### DIFF
--- a/packages/web/styles/globals.css
+++ b/packages/web/styles/globals.css
@@ -5,6 +5,22 @@
 @tailwind components;
 @tailwind utilities;
 
+@layer base {
+  ::-webkit-scrollbar {
+    width: 8px;
+  }
+  ::-webkit-scrollbar-track {
+    background: #231d4b00;
+  }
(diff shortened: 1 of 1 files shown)


packages/web/styles/globals.css
```diff
@@ -5,6 +5,22 @@
 @tailwind components;
 @tailwind utilities;
 
+@layer base {
+  ::-webkit-scrollbar {
+    width: 8px;
```
jonator, line 10: Actually I think 8 it a bit too small to find for cursor. Let's do 10px.

Also- do you know if there's any way to not apply this style on mac?

JoseRFelix: > Actually I think 8 it a bit too small to find for cursor. Let's do 10px.

👍 

> Also- do you know if there's any way to not apply this style on mac?

There's no native CSS way, so we'd have to create a script that detects the browser and add a class to the body depending on the platform

PR #1059, opened 2022-11-25
Remove 'blocked' IBC xfer Status

JeremyParish69:
Remove 'blocked' status, and just left "congested", because users are confused about whether they can attempt IBC transfers.
Also changes congested timing to >20 mins of congestion, up from >5 mins.

Reason: There is currently an anomalous tx stuck in the Osmosis -> Cosmos Hub channel, which is making the transfer button say 'blocked'. This is confusing many users. The button for such an important function perhaps should have more consideration before telling users that it's blocked. It is not clear that they can still attempt to transfer, (and that it will most likely work, as many users report). 

Was: 
image

Proposed:
image


packages/stores/src/queries-external/ibc/status-by-id.ts
```diff
@@ -42,8 +42,8 @@ class ObservableQueryIbcChainStatus extends ObservableQueryExternalBase<
       );
       if (channelData) {
         if (channelData.size_queue > 0) {
-          if (channelData.duration_minutes > 20) return "blocked";
-          else if (channelData.duration_minutes > 5) return "congested";
+          //if (channelData.duration_minutes > 20) return "blocked";
+          /*else */if (channelData.duration_minutes > 5) return "congested";
```
jonator, line 46: Perhasp we can raise the bar with what is considered blocked and what is considered congested? Perhaps if the queue is larger, and the times are larger

jonator: I think it's less consequential to have https://info.osmosis.zone/ibc have a lower bar

jonator:
I think we should also prevent the button from turning red if a wallet is not connected. I noticed that in the preview.

jonator:
I can push a commit if you're up for it. For the higher threshold as well.

jonator:
https://github.com/osmosis-labs/osmosis-frontend/pull/1065

Just merged. Please update this branch. We could still increase the number of minutes required to return each status.

packages/stores/src/queries-external/ibc/status-by-id.ts
```diff
@@ -42,8 +42,7 @@ class ObservableQueryIbcChainStatus extends ObservableQueryExternalBase<
       );
       if (channelData) {
         if (channelData.size_queue > 15) {
-          if (channelData.duration_minutes > 20) return "blocked";
-          else if (channelData.duration_minutes > 5) return "congested";
+          if (channelData.duration_minutes > 20) return "congested";
```
jonator, line 45: Personally, I'd prefer if we don't change the API and still let it return blocked, instead we just look for contested || blocked (as we do) and update the view accordingly.

jonator (changes requested):

packages/stores/src/queries-external/ibc/status-by-id.ts
JeremyParish69: To be clear, if it returns blocked, you want the button to say congested? Or do you want to button to be updated like 'Withdraw (Channel blocked)' and still say 'blocked'?

jonator: It can return blocked, but we'll show it as congested in view (as we do)

JeremyParish69: Should we go back to the 5 and 20 mins then? Or bump up to something like 20 and 60 mins?

jonator (approved):

PR #1070, opened 2022-11-29
Feat Show User Pool Asset Composition

JoseRFelix:
## Before 

image

## After

image


packages/web/pages/pool/[id].tsx
```diff
@@ -289,6 +290,16 @@ const Pool: FunctionComponent = observer(() => {
   const levelCta = bondLiquidityConfig?.calculateBondLevel(bondableDurations);
   const level2Disabled = bondableDurations.length === 0;
 
+  const highestAPRDailyPeriodicRate =
+    bondableDurations[bondableDurations?.length - 1]?.aggregateApr
```
Comment (jonator) [inline packages/web/pages/pool/[id].tsx:294]: Technically, we should be subtracting the swap fee APR from the aggregate since users do earn swap fees on pool shares. Additional incentives are given (from a gauge) when the user locks the shares in a lock. LMK if you want more of a rundown of how the incentive mech works.

packages/web/pages/pool/[id].tsx
```diff
@@ -289,6 +290,16 @@ const Pool: FunctionComponent = observer(() => {
...
+      .quo(new Dec(365)) // get daily periodic rate
```
Comment (jonator) [inline packages/web/pages/pool/[id].tsx:295]: To be more correct here as well, we should try to pull hardcoded parameters from the chain data instead of making assumptions. In the future, epochs may not necessarily be every day, it could change to every week.

See this code for how to get num epochs per year

https://github.com/osmosis-labs/osmosis-frontend/blob/ae9dcaaafa4040e29f6962b124e70f412bb40ccc/packages/stores/src/queries/pool-incentives/incentivized-pools.ts#L307-L315

jonator (changes requested):
Screen Shot 2022-11-29 at 10 40 36 AM
Seems there's some overflow issues going from desktop to tablet. Perhaps we could leave "Your pool balance in it's own row", and wrap the second 2 below. Then eventually wrapping those last two -> three rows as needed.

In the past, we don't necessarily have Figma designs to mock up tablet/intermediary screen sizes, I've just wrapped as I deemed appropriate, and asked the design team if I was unsure.

packages/web/pages/pool/[id].tsx
Comment (JoseRFelix) [reply packages/web/pages/pool/[id].tsx:294]: But aren't swap fees part of the reward? I thought this number was the result of calculating the potential income from bonding the available value.

packages/web/pages/pool/[id].tsx
Comment (jonator) [reply packages/web/pages/pool/[id].tsx:299]: Disregard, since APRs are annualized

Issue #710, opened 2022-08-02
Specify favorite assets


jonator:
[From discord](https://discord.com/channels/798583171548840026/855908784445849610/1001822923507642430)

Similar to the info site, would be great to specify assets that are a users favorite.

User stories:
* A user commonly withdraws/deposits an asset, but since the balance is currently 0 the user is required filter/find the asset.

Should discuss & research whether user stories justify the added UI & dev complexity.

jonator:
Is a good first issue, but **blocked** on design link @victortokaito

victortokaito:
### Specify Favorites![Specify Favorites](https://user-images.githubusercontent.com/39136683/192857760-8fceb128-72ea-437e-99c6-e2e79e2463f4.png)
### Drag to rearrange ![Specify Favorites - shadow drag to rearrange](https://user-images.githubusercontent.com/39136683/192857769-673108a9-23bf-436d-98c0-49fc076d2e3a.png)

[Figma link](https://www.figma.com/file/TbhltqFzPjWgIdxdcQ23dj/Assets?node-id=114%3A3605) ([frames without the 3D visual assets](https://www.figma.com/file/TbhltqFzPjWgIdxdcQ23dj/Assets?node-id=114%3A3320))

victortokaito:
### More fun and game vibe
![Specify Favorites](https://user-images.githubusercontent.com/39136683/192866496-933ca8e4-8708-4e32-a975-dcb35892dac7.png)
![Specify Favorites - shadow drag to rearrange](https://user-images.githubusercontent.com/39136683/192866493-69863346-4611-40e2-8cb8-e4a720a2afdc.png)

jonator:
@victortokaito how would sorting be affected? Shouldn't the stars column be a new sortable column?

victortokaito:
As part of the development workflow, we need to add tracking to new user-facing features. Whenever we merge features, we can get some automated notification through clickup to update amplitude

PR #1071, opened 2022-11-29
Specify favorite assets

sehyunc:
`icon.tsx`
* proposed new method of handling icons in the app
* requires creating new icon components in `/icons` and copy-pasting the SVG from icon source (`star` is from [here](https://fonts.google.com/icons))

`assets-table.tsx`
* use `useLocalStorageState` hook for favorites list

`asset-name.tsx`
* add favorite state to table row using new icon component


sehyunc:
@jonator Blocked by open question from #710: should stars column be a sortable column or should it be at the top by default?

jonator:
> @jonator Blocked by open question from #710: should stars column be a sortable column or should it be at the top by default?

From meeting: we decided that favorited items will _always_ be at the top regardless of filtering and sorting. We can see how natural that feels in the preview. Perhaps filtering via search is the exception.

jonator:
@sehyunc 

Seems there's a build error: make sure to build locally before committing/opening PR:
```


./components/table/assets-table.tsx:5:3
--
00:44:34.066 | Type error: 'useEffect' is declared but its value is never read.
00:44:34.066 |  
00:44:34.066 | 3 \|   FunctionComponent,
00:44:34.066 | 4 \|   useCallback,
00:44:34.067 | > 5 \|   useEffect,
00:44:34.067 | \|   ^
00:44:34.067 | 6 \|   useMemo,
00:44:34.067 | 7 \|   useState,
00:44:34.067 | 8 \| } from "react";


```

jonator:
> icon.tsx

> proposed new method of handling icons in the app
> requires creating new icon components in /icons and copy-pasting the SVG from icon source (star is from [here]> > ? > (https://fonts.google.com/icons))

> `icon.tsx`
> 
> * proposed new method of handling icons in the app
> * requires creating new icon components in `/icons` and copy-pasting the SVG from icon source (`star` is from [here](https://fonts.google.com/icons))

I can see the benefits for SVGs such as these that are highly functional and needed on first load. However the benefit of Next/image for SVGs is it will time the loading and only load content as needed and after the initial load. They're mostly valuable for pngs/jpegs as they do image optimization as well, however for svgs there's a bit less value. Still, by using Next/image we keep svg markup out of the main bundle, and Vercel and cloudflare are then able to take the SVG content and distribute it across a CDN.

packages/web/components/icon/types.ts
```diff
@@ -0,0 +1,8 @@
+export type Icons = "star";
+
+export interface IconProps {
+  color?: string;
+  name?: Icons;
+  onClick?: () => void;
+  size?: number;
+}
```
jonator, line 8: I've shared this convention with jose as well. IMO it's quicker to read the component when props are embedded directly in the type param of the `FunctionComponent` type (equivalently `React.FC`, but we conventionally use `FunctionComponent`). Example `FunctionComponent<{ color?: string, ... }>`

packages/web/components/table/assets-table.tsx
```diff
@@ -278,6 +282,29 @@ export const AssetsTable: FunctionComponent<Props> = observer(
...
+          } else {
+            coin.isFavorite = false;
+            coin.onToggleFavorite = () => {
+              const newFavorites = [...favoritesList, coin.coinDenom];
+              onSetFavoritesList(newFavorites);
+            };
+          }
+          return coin;
+        })
+        .sort((a, b) => Number(b.isFavorite) - Number(a.isFavorite));
```
jonator, line 305: See video on sorting. Perhaps you could filter the favorites from `tableData`, then return a new list with `favoritesList.concat(favoritesFilteredTableData)` or similar.

IMO:
* Sorting: favorites always at top.
* Search query: favorites always at top but filtered by query.

jonator (commented):
It appears favorites are not maintained at the top when sorting ascending:
https://user-images.githubusercontent.com/4606373/204896569-f10fac91-84f4-4938-92b1-05df0664116e.mov

packages/web/components/icon/icon.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { IconProps } from "./types";
+import Star from "./icons/star";
+
+const Icon: React.FC<IconProps> = ({
+  color = "#fff",
+  name,
+  onClick,
+  size = 24,
+}) => {
+  const Child = useMemo(() => {
```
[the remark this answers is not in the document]
sehyunc: My idea was any time a new icon was added, it would be imported and added to the switch case. This frontloads the work as any time an icon is used thereafter, you can just pass the name instead of having to import and render the svg in addition to the Icon component. Can definitely use `children` as well, but thought this was better for usability.

packages/web/components/icon/icons/star.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+const Star: React.FC<IconProps> = ({ color, size = 24 }) => (
+  <svg height={size} viewBox="0 0 24 24" width={size}>
+    <path d="M0 0h24v24H0z" fill="none" />
+    <path
+      fill={color}
+      d="m12 17.27 4.15 2.51c.76.46 1.69-.22 1.49-1.08l-1.1-4.72 3.67-3.18c.67-.58.31-1.68-.57-1.75l-4.83-.41-1.89-4.46c-.34-.81-1.5-.81-1.84 0L9.19 8.63l-4.83.41c-.88.07-1.24 1.17-.57 1.75l3.67 3.18-1.1 4.72c-.2.86.73 1.54 1.49 1.08l4.15-2.5z"
+    />
+  </svg>
+);
```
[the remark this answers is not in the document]
sehyunc: Did the issue end up getting created? I'd be happy to do a migration of existing icons, and maybe at the same time standardize the icon library we use (if we don't already have a standard library).

packages/web/components/table/assets-table.tsx
sehyunc: Fixed this in most recent commit

jonator (approved):

PR #1083, opened 2022-12-01
OSMO price and buy button in sidebar

JoseRFelix:


syed8rahman:
Screen Shot 2022-12-05 at 4 24 13 PM
Not the exact PR but for side nav, the active section should be a different color.

syed8rahman:
Screen Shot 2022-12-05 at 4 35 01 PM
Screen Shot 2022-12-05 at 4 35 16 PM

Buy button icons are in. Please animate a transition with 300ms Ease in out properties. 

Link to [Figma location here ](https://www.figma.com/file/a1FycJWJ7VARpWMfzoCQF1/UI-Refresh-Aug-2022?node-id=1%3A3&t=xGS9ronKniM3H2RZ-1)

syed8rahman:
Screen Shot 2022-12-07 at 2 53 16 PM
There should be an inner padding of 8 px around the Osmo price + graph. 
We should also stick to 2 decimal places for consistency so it doesn't confuse users about rounding + accuracy in case it's .90 flat instead of .91 etc.

syed8rahman:
Also looks like the value is top aligned with the osmo logo. Should be center aligned.

packages/web/components/skeleton.tsx
```diff
@@ -0,0 +1,24 @@
+import classNames from "classnames";
+import { DetailedHTMLProps, FunctionComponent } from "react";
+
+const Skeleton: FunctionComponent<
```
jonator, line 4: Could we name this something like `LoadingSkeletonPlaceholder` or something to clarify it's for loading data?

packages/web/components/navbar-osmo-price.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <Skeleton isLoaded={osmoPrice.isReady}>
+        <Button
+          mode="tertiary"
+          className={classNames(
+            "button group relative !h-11 gap-2 overflow-hidden !rounded-full !border-osmoverse-700 !py-1 font-bold text-osmoverse-100 !transition-all !duration-300 !ease-in-out",
+            "hover:border-none hover:bg-gradient-positive hover:text-osmoverse-1000"
+          )}
+          onClick={() => transferConfig.buyOsmo()}
+        >
+          <CreditCardIcon /> <span className="z-10">Buy tokens</span>{" "}
```
jonator, line 110: Let's be sure to localize the "Buy Tokens" string

jonator (changes requested):
Seems after closing the Transak modal, a weird padding is added to the bottom of the app. We may need to use some hack to prevent this by the Transak SDK. Ideally, the fix should be added to the Transak modal/integration code.

Screen Shot 2022-12-11 at 7 15 13 PM

packages/web/components/skeleton.tsx
JoseRFelix: Renamed it into `SkeletonLoader`

jonator:
It's worth noting that we also have a metric loader component that is similar. IMO we should eventually replace all metric loader usage with this new component.

https://github.com/osmosis-labs/osmosis-frontend/blob/e36f6b13b64bcd47353fb1830a56f2e450c0dca1/packages/web/components/loaders/metric-loader.tsx#L9

packages/web/integrations/transak/use-transak-modal.ts
```diff
@@ -23,6 +23,8 @@ export function useTransakModal(
   const [transak, setTransak] = useState<any | null>(null);
   const [shouldShow, setShouldShow] = useState(false);
 
+  console.log(transak?.EVENTS);
```
jonator, line 26: Please remove log

packages/web/components/navbar-osmo-price.tsx
```diff
@@ -0,0 +1,196 @@
...
+      )}
+
+      {transferConfig?.fiatRampsModal && (
+        <FiatRampsModal {...transferConfig.fiatRampsModal} />
+      )}
+    </div>
+  );
+});
+
+const CreditCardIcon = () => {
```
jonator, line 138: Let's move this into the components/assets folder

jonator (changes requested):

packages/web/components/navbar-osmo-price.tsx
JoseRFelix: I was thinking about it, but the component has some deep CSS classes in the `path`s that only apply to this use case.

packages/web/integrations/transak/use-transak-modal.ts
JoseRFelix: Nice catch, thanks!

PR #1086, opened 2022-12-02
add FURY

cryptofanaticXx1:


packages/web/config/chain-infos.ts
```diff
@@ -630,6 +630,17 @@ const chainInfos = (
...
         },
+        {
+          type: "cw20",
+          contractAddress:
+            "juno1cltgm8v842gu54srmejewghnd6uqa26lzkpa635wzra9m9xuudkqa2gtcz",
+          coinDenom: "FURY",
+          coinMinimalDenom:
+            "cw20:juno1cltgm8v842gu54srmejewghnd6uqa26lzkpa635wzra9m9xuudkqa2gtcz:FURY",
+          coinDecimals: 6,
+          coinGeckoId: "pool:fury",
```
JeremyParish69, line 641: Will this Id be defined in the config/prices.ts file? At the moment this Id would do nothing, but if there is a pool, we could then have a price.

packages/web/config/chain-infos.ts
```diff
@@ -630,6 +630,17 @@ const chainInfos = (
           coinGeckoId: "pool:muse",
           coinImageUrl: "/tokens/muse.svg",
         },
+        {
```
JeremyParish69, line 633: Is a PR to the Chain Registry in the works?

cryptofanaticXx1: done
https://github.com/cosmos/chain-registry/pull/1056

packages/web/config/chain-infos.ts
cryptofanaticXx1: changed the id to https://www.coingecko.com/en/coins/fanfury
"fanfury"
pool is planned, I will define the Id once the pool is created

PR #1091, opened 2022-12-04
Added BZE coin

faneaatiku:


packages/web/config/chain-infos.ts
```diff
@@ -2292,6 +2292,30 @@ const chainInfos = (
...
+      bip44: {
+        coinType: 118,
+      },
+      bech32Config: Bech32Address.defaultBech32Config("bze"),
+      currencies: [
+        {
+          coinDenom: "BZE",
+          coinMinimalDenom: "ubze",
+          coinDecimals: 6,
+          coinGeckoId: "pool:ubze",
```
JeremyParish69, line 2309: This id isn't defined anywhere. Is there a pool we can use to define this?

PR #1093, opened 2022-12-05
Feat Display Swap Route Info

JoseRFelix:


packages/web/utils/ssr.ts
```diff
@@ -0,0 +1 @@
+export const isBrowser = typeof window !== "undefined";
```
jonator, line 1: Are these primitives really needed as separate files? Why not use them directly inline?

packages/web/hooks/use-measure.ts
```diff
@@ -0,0 +1,57 @@
...
+  y: 0,
+  width: 0,
+  height: 0,
+  top: 0,
+  left: 0,
+  bottom: 0,
+  right: 0,
+};
+
+function useMeasure<E extends Element = Element>(): UseMeasureResult<E> {
```
jonator, line 27: Can we add a docs comment to explain this at a high level?

packages/web/utils/ssr.ts
JoseRFelix: I believe it's easier to read and understand thanks to the variable name. What do you think?

packages/web/hooks/use-measure.ts
JoseRFelix: Added 👍

PR #1096, opened 2022-12-05
Remove stargate and no-legacy-stdTx

JeremyParish69:
Remove stargate and no-legacy-stdTx deprecated Keplr suggest chain features.
image


jonator (approved):
LGTM

PR #1098, opened 2022-12-06
Move gasPriceStep to feeCurrency

JeremyParish69:
move gasPriceStep to feeCurrency, as defined for suggest chain: https://docs.keplr.app/api/suggest-chain.html
to get rid of these errors:
image
Also allows for specifying gasPriceSteps for multiple fee currencies. (although this FE just uses the first defined fee Currency--not any) 

Note: did not change for Axelar, since the Keplr Currency types being referenced do not allow for the relocation of the property. (are we on the latest? Edit: we are not; Current ~0.10.14; Latest: 0.11.23)


jonator:
True, may warrant an update to keplr-wallet packages to include changes with the any-fee feature.

jonator (approved):

PR #1102, opened 2022-12-07
Add MEDAS

oxygene76:
Add MEDAS to Frontier


JeremyParish69 (changes requested):
Good start, but insufficient--it also requires the chain be added to chain-infos.ts, and the token logo (svg preferred) should be added to public/tokens/ folder.

packages/web/config/ibc-assets.ts
```diff
@@ -963,6 +963,12 @@ export const IBCAssetInfos: (IBCAsset & {
     sourceChannelId: "channel-326",
     destChannelId: "channel-5",
     coinMinimalDenom: "stustars",
+  },
+   {
```
JeremyParish69, line 967: please move to the bottom of the list

oxygene76: Thank you, just added the chain to chain-infos.ts and added the token logos.

packages/web/config/chain-infos.ts
```diff
@@ -2346,6 +2346,30 @@ const chainInfos = (
...
+        coinType: 118,
+      },
+      bech32Config: Bech32Address.defaultBech32Config("medas"),
+      currencies: [
+        {
+          coinDenom: "MEDAS",
+          coinMinimalDenom: "umedas",
+          coinDecimals: 6,
+          // coinGeckoId: "medasdigital",
+          coinGeckoId: "pool:umedas",
```
JeremyParish69, line 2364: Shall we use pool 859 to define this 'id' in the prices.ts file? With this, the MEDAS token will have a price to display in the frontend.

JeremyParish69: Yes please

oxygene76: Ok, added to price.ts

JeremyParish69 (changes requested):
Can we please add the gasPriceStep?
Also, I'm, getting CORS errors; please ensure the endpoints don't block the site.

oxygene76:
of course, added gasPriceStep.. Had a fast check for cors, will resolve tomorrow. Thank you

PR #1104, opened 2022-12-07
Fix copy when adding stablecoin liquidity

sehyunc:
`en.json`
* add localization for helper text

@jonator what is the process/who is the person to to add this for all languages?

### DIFF
--- a/packages/web/localizations/en.json
+++ b/packages/web/localizations/en.json
@@ -378,5 +378,8 @@
   "Amount is empty": "Amount is empty",
   "Insufficient amount": "Insufficient amount",
   "Amount is zero": "Amount is zero",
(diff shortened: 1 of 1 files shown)


jonator:
@sehyunc it's a bit rough, but localization changes can be done with an online translator (Google Translate, etc.) then we can work with the community or other native speakers to update them as needed.

jonator:
I've actually been in talks with a startup in Berlin that (called inlang) that provides a nice frontend for non-devs to generate PRs with new localizations.

sehyunc:
> I've actually been in talks with a startup in Berlin (called inlang) that provides a nice frontend for non-devs to generate PRs with new localizations.

That would be great, Google Translate is still pretty rudimentary without context. Also, I’ll change base branch to be ‘stage’ so this can be merged in.

PR #1107, opened 2022-12-08
Add arable gauge

gophereth:
- Add arable gauge


gophereth:
Gauge will start rewards distribution on Wednesday, December 14, 2022 4:00:00 PM GMT and we will need to get this working soon. So that we have some time to have enough liquidity bonded before the time to prevent too big inflation at initial. cc. @JeremyParish69

gophereth:
I think we can remove the ibc transfer gas fee part if it's the problem for merging the PR.
Would like to get gauges merged asap. Btw, websocket issue you have shared should be resolved now.

JeremyParish69:
The gauges look good, and would merge just that, but I'm not yet fully understanding the supposed gas issue.

PR #1111, opened 2022-12-08
Redesign Response

JoseRFelix:


jonator:
@JoseRFelix since we removed legacy.osmosis.zone, could we get this completed and deployed so users can at least see the pool asset breakdown?

JoseRFelix:
@jonator Everything should be done except #1083 which has some comments. I think we can start merging after a review.

jonator:
Please remove draft status and request a review once ready! Thanks

packages/stores/src/ui-config/manage-liquidity/bond-liquidity.ts
```diff
@@ -129,6 +136,16 @@ export class ObservableBondLiquidityConfig extends UserConfig {
...
             milliseconds: durationMs,
           });
+          const lockedUserShares = queryLockedCoin.getLockedCoinWithDuration(
+            this.poolDetails.poolShareCurrency,
+            curDuration
+          ).amount;
+
+          const totalShares = this.poolDetails.pool.totalShare;
+          const poolTvl = this.poolDetails.totalValueLocked;
+          const userShareValue = poolTvl.mul(
```
jonator, line 146: It would be more concise to call this `userLockedShareValue`

packages/web/config/chain-infos.ts
```diff
@@ -903,12 +893,7 @@ const chainInfos = (
         average: 0.03,
         high: 0.05,
       },
-      features: [
-        "ibc-transfer",
-        "ibc-go",
-        "wasmd_0.24+",
-        "cosmwasm",
-      ],
+      features: ["ibc-transfer", "ibc-go", "wasmd_0.24+", "cosmwasm"],
```
jonator, line 896: Hmm.. did your prettier fix do this? Is it looking at the right config? I think our config comes from next.js

packages/stores/src/ui-config/manage-liquidity/bond-liquidity.ts
jonator: Also, it doesn't appear to be used anywhere?

JoseRFelix: This var is used on the bond card to display the user's locked share value

packages/web/config/chain-infos.ts
JoseRFelix: Yes, I used the cli directly

jonator (approved):

jonator:
@JoseRFelix 

Before we merge this could you quickly add a fix for this?

telegram-cloud-photo-size-4-5821055853460045430-y

JoseRFelix:
Let me get into this

JoseRFelix:
I'll push this fix into #1042 to avoid merge conflicts

packages/web/localizations/en.json
```diff
@@ -37,7 +37,7 @@
     "myPools": "My Pools",
     "APR": "APR",
     "available": "Available",
-    "liquidity": "Pool Liquidity",
```
JeremyParish69, line 40: This caused Frontend to display the pool's total liquidity as 'My liquidity', which is incorrect

PR #1115, opened 2022-12-08
Add AMBER (Secret SNIP-20 token) to frontier

kent-3:


packages/web/config/chain-infos.ts
```diff
@@ -167,6 +167,16 @@ const chainInfos = (
           coinGeckoId: "alter",
           coinImageUrl: "/tokens/alter.svg",
         },
+        {
+          type: "cw20",
+          contractAddress: "secret1s09x2xvfd2lp2skgzm29w2xtena7s8fq98v852",
+          coinDenom: "ALTER",
```
Sdpierret, line 173: coinDenom must match "AMBER" and not "ALTER"

Sdpierret (changes requested):

packages/web/config/chain-infos.ts
kent-3: fixed, thanks

packages/web/config/chain-infos.ts
```diff
@@ -167,6 +167,16 @@ const chainInfos = (
           coinGeckoId: "alter",
           coinImageUrl: "/tokens/alter.svg",
         },
+        {
+          type: "cw20",
+          contractAddress: "secret1s09x2xvfd2lp2skgzm29w2xtena7s8fq98v852",
+          coinDenom: "AMBER",
```
JeremyParish69, line 173: Can you please register this to the chain registry?

packages/web/config/ibc-assets.ts
```diff
@@ -1061,6 +1061,15 @@ export const IBCAssetInfos: (IBCAsset & {
     depositUrlOverride: "https://wrap.scrt.network",
     ics20ContractAddress: "secret1tqmms5awftpuhalcv5h5mg76fa0tkdz4jv9ex4",
   },
+  {
+    // AMBER
```
JeremyParish69, line 1065: Please move to bottom of list

packages/web/config/chain-infos.ts
kent-3: Yes I have submitted the pull request here: https://github.com/cosmos/chain-registry/pull/1094

PR #1124, opened 2022-12-11
added bze external gauges in web feature-flag

faneaatiku:


faneaatiku:
Sorry for writing here but I wonder if this can be merged ? I half of hour are incentives start. We didn't know that a PR is required to show the incentives in FE, otherwise we would've submit this PR a lot earlier.

JeremyParish69:
@faneaatiku  These gauges were set to run over only 2 and 4 epochs, so they are nearly done already. Will longer gauges be created (ideally at least 30 days)?

faneaatiku:
> @faneaatiku These gauges were set to run over only 2 and 4 epochs, so they are nearly done already. Will longer gauges be created (ideally at least 30 days)?

Seems that I didn't understood the documentation. I started a new gauge. and modified the PR

JeremyParish69 (approved):
LGTM

PR #1129, opened 2022-12-13
Improvement Display Multihop Discount in Trade Router

JoseRFelix:


packages/pools/src/routes.ts
```diff
@@ -270,6 +272,14 @@ export class OptimizedRoutes {
       return {
         ...routes[i],
         amount,
+        isMultihopOsmoFeeDiscount: isOsmoRoutedMultihop(
+          routes[0].pools.map((routePool) => ({
```
jonator, line 276: We shouldn't need to modify this component. We can get `multiHopOsmoDiscount` from the router in the trade clipboard, then IMO it makes sense to just pass that flag into the trade route component

JoseRFelix: You are right, thanks! Let me update this

jonator (approved):

PR #1140, opened 2022-12-14
Update impacthub chainId

JeremyParish69:
Update to ixo-4

Fails validation. IXO token disappears


jonator:
@JeremyParish69 updates on this?

JeremyParish69:
> @JeremyParish69 updates on this?

They updated their chain id, but I think until the IBC connection is revived, user balances will only display correctly with the old chain-id. Remains unstable, but we don't really need to keep the PR since I have no ETA--but would be nice to keep the branch.

jonator:
Closing for now to clean up, can reopen if needed @JeremyParish69

Issue #982, opened 2022-11-09
Use Axelar SDK `getTransferFee`


jonator:
### Problem
Currently, transfer fees are hardcoded with each asset in the Axelar source chain config file. When I initially built the integration, I tried to utilize the `getTransferFee` query from the Axelar SDK, but ran into issues. This means whenever the fee changes for any network<>asset combo, we need to manually update the config (like on Nov 10).

https://github.com/osmosis-labs/osmosis-frontend/blob/0fddb0647993d5effc32fca8b84f9917a998e2fb/packages/web/integrations/axelar/source-chain-config.ts#L20

### Proposed Solution
Perhaps upgrade the AxelarJS SDK, and see if we can get the `getTransferFee` query to work. Reach out to Canh Trinh from Axelar with any issues.

JeremyParish69:
This is really exciting! In the interest of reducing the manual effort (of adding an Axelar asset or withdraw network) even further, would it be difficult to do the same for the erc20 addresses as well( after this)? If so, we'd probably be able to get rid of the source-chain-config.ts file completely, since (network) "id" and "logoUrl" could be moved elsewhere (they describe the network, but don't need to be defined repeatedly for each asset-network permutation), and the entire page would no longer be needed, and that would make 1 fewer file that needs to be modified when adding an Axelar asset.

jonator:
How are you thinking we source the erc20 addresses? Via some ethereum endpoint?

JeremyParish69:
> How are you thinking we source the erc20 addresses? Via some ethereum endpoint?

I saw token addresses within some JSONs, but I don't know if they have an API query just for ERC20 addresses--although it might be better to have all this data (it even includes minimum transfer amounts and mint limits):
https://axelar-mainnet.s3.us-east-2.amazonaws.com/mainnet-asset-config.json

PR #1146, opened 2022-12-15
Gets transfer fee by query instead of by config

milesz8:
https://github.com/osmosis-labs/osmosis-frontend/issues/982

- Uses axelar's AxelarQueryAPI to get fee's
- Removes unneeded config values


packages/web/integrations/axelar/hooks/use-transfer-fee-query.ts
```diff
@@ -0,0 +1,79 @@
...
+        environment,
+      });
+      const amount = Number(
+        new CoinPretty(
+          currency,
+          amountMinDenom === "" ? "0" : amountMinDenom
+        ).toCoin().amount
+      );
+      if (!isNaN(amount)) {
+        console.log("amount " + amount);
```
jonator: Please remove log.

packages/web/integrations/axelar/hooks/use-transfer-fee-query.ts
```diff
@@ -0,0 +1,79 @@
...
+        );
+      } else {
+        throw new Error("Requested fee amount is not a number.");
+      }
+    };
+    let timeout: NodeJS.Timeout | undefined;
+
+    setIsLoading(true);
+    // debounce query on user typing
+    timeout = setTimeout(() => {
```
jonator, line 47: Could you please create a general debounce hook in the /hooks/ folder? Note- you may be able to leverage a `debounce` package we already download into our bundle. It should be able to watch any given value and only run an effect once some debounce period has elapsed.

packages/web/integrations/axelar/transfer.tsx
```diff
@@ -175,6 +174,20 @@ const AxelarTransfer: FunctionComponent<
     /** Amount, with decimals. e.g. 1.2 USDC */
     const amount = isWithdraw ? withdrawAmountConfig.amount : depositAmount;
 
+    console.log("Get transfer fee");
```
jonator: Please remove all logs.

packages/web/integrations/axelar/transfer.tsx
```diff
@@ -175,6 +174,20 @@ const AxelarTransfer: FunctionComponent<
...
+
+    const { transferFee: transferFeeMinAmount } = useTransferFeeQuery(
+      sourceChain,
+      destChain,
+      originCurrency.coinMinimalDenom,
+      amount,
+      originCurrency,
+      isTestNet ? Environment.TESTNET : Environment.MAINNET,
+      1000
```
jonator: Is this too short of a debouce period? Perhaps 3 or 2 seconds?

packages/web/integrations/axelar/transfer.tsx
```diff
@@ -175,6 +174,15 @@ const AxelarTransfer: FunctionComponent<
     /** Amount, with decimals. e.g. 1.2 USDC */
     const amount = isWithdraw ? withdrawAmountConfig.amount : depositAmount;
 
+    const { transferFee: transferFeeMinAmount } = useTransferFeeQuery(
+      sourceChain,
+      destChain,
```
jonator, line 179: I think we want to leave this variable named `transferFee`. The min amount is used to disable UI if the amount isn't at least that. It's related but not the same.

jonator (changes requested):

packages/web/integrations/axelar/hooks/use-transfer-fee-query.ts
```diff
@@ -0,0 +1,78 @@
...
+/** Fetches a new Transfer fee quote when either chain, the amount, or the currency changes.
+ *  `amountMinDenom` is from user input, assumes `=== ""` for no input, and therefore no query.
+ */
+export function useTransferFeeQuery(
+  sourceChain: string,
+  destChain: string,
+  tokenMinDenom: string,
+  amountMinDenom: string,
+  currency: AppCurrency,
+  environment = Environment.MAINNET,
```
JeremyParish69, line 15: should we also allow for the TESTNET environment? The values are not equal, and the TESTNET values appear to be completely erased.

jonator: If we pass in Environment.TESTNET we should have the testnet values

https://github.com/milesz8/osmosis-frontend/blob/5443d2e5d3b525c5309cb4d926dd55e0e3d2c3fd/packages/web/integrations/axelar/transfer.tsx#L183

packages/web/integrations/axelar/transfer.tsx
milesz8: okay

jonator (changes requested):
It appears there's a build error (type error)

PR #1149, opened 2022-12-16
Add IBCX metadata for testnet

junhoyeo:
image


JeremyParish69:
@junhoyeo Will this be added to ibc-assets (so that it's displayed on the Assets page)? Also will this be registered to Chain Registry (if needed, I can take a shot at this)?

junhoyeo:
> @junhoyeo Will this be added to ibc-assets (so that it's displayed on the Assets page)?

@JeremyParish69 There's no need to add this token to ibc-assets, as `/assets` already displays the new token from Osmosis.

image

> Also, will this be registered to the Chain Registry?

Since this token is only for the testnet, it's not necessary to register it in the Chain Registry for now.

> If needed, I can take a shot at this.

But sure! I appreciate the offer.

PR #1157, opened 2022-12-20
Feat Match Button with Design Guidelines

JoseRFelix:


jonator (changes requested):
Please address merge conflicts

packages/web/components/assets/chevron-down-icon.tsx
```diff
@@ -0,0 +1,22 @@
+import classNames from "classnames";
+import { FunctionComponent } from "react";
+
+export const ChevronDownIcon: FunctionComponent<{
```
jonator, line 4: Are you intending for this to replace all uses of `chevron-down.svg` (and it's variants)? If so, I think we should replace them all. Though, it's worth considering this approach will result in a slowly increased main bundle size.

JoseRFelix: After some deliberation, I found this [article](https://benadam.me/thoughts/react-svg-sprites/?utm_source=pocket_reader). It posits that using a sprite SVG is a way to have the bundle benefits of the image while having some of the flexibility of inlining. I believe we could use this for our non-animated SVGs that only require shallow CSS properties. My latest commit introduces an Icon component that leverages this technique. I think this is a nice solution to this problem. What do you think?

packages/web/components/buttons/button.tsx
```diff
@@ -1,59 +1,172 @@
...
+          "border-wosmongton-300",
+          "hover:border-wosmongton-200",
+          "text-wosmongton-300",
+          "hover:text-wosmongton-200",
+          "rounded-md",
+          "disabled:border-osmoverse-600",
+          "disabled:text-osmoverse-400",
+        ],
+        amount: [
+          "border-2",
```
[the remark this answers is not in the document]
JoseRFelix: You're right! Fixed 👍

JoseRFelix:
@jonator Replaced all chevron icons images with the new sprite method. I believe other icons can be migrated incrementally.

jonator (approved):

PR #1170, opened 2022-12-21
improvement: Correct Types for Amplitude Events

JoseRFelix:
## Consideration

Some Amplitude events did not match the expected type from the dashboard. Perhaps, a good idea would be to use the amplitude CLI to generate these types.


jonator:
Oh- was not aware of a CLI

jonator (approved):

Issue #1015, opened 2022-11-16
Optimize pool query


jonator:
### Problem
Our current performance bottleneck on the frontend is currently the [pools](https://lcd-osmosis.keplr.app/osmosis/gamm/v1beta1/pools?pagination.limit=1000) query. If the node is slow to respond, the UI appears bricked, responsive but void of data as it's our primary data dependency. Additionally, the newly-arrived 800+ pools then have to be processed reactively by mobx, firing off 1000s of computedFns and @computed getters, causing lag.

#### Primary
Sometime shortly after v13, @mattverse and others should have released a pools query with parameters for liquidity. Initially, we should only fetch the top ~120 pools by liquidity. Our swap router filters the pools to about 120 based on liquidity already. Then, we should add logic to show more pagination pages or initiate a full query later as the user searches in the pools table, selects "All pools", and/or paginates right.

#### Secondary
Look into using Next.js server side rendering to fetch and possibly minimize pools data from the node every few minutes. Ideally, we should find a way to have the pools data immediately available to the render function so we can SSR tables/cards with data as well. Then, on the client, we'd need to add a hydration mechanism to take the pools data from the NextPage props and fill the store pre-render, thus preventing another pools fetch. See [old issue that Tony advised against implementing](https://github.com/osmosis-labs/osmosis-frontend/issues/359), and [commit](https://github.com/osmosis-labs/osmosis-frontend/commit/2560af7f748a520980e5c562b32e86fe9906e47c) where I reverted my attempt.

daniel-farina:
A simple example of what a difference this would make: 

<img width="1679" alt="image" src="https://user-images.githubusercontent.com/13665117/202295159-fe7ec842-9fce-4820-b22c-26a8389366ef.png">

jonator:
Currently blocked since there's no support for encoding a list of coins in a query param for protobuf.

PR #1171, opened 2022-12-22
Pools page performance

jonator:
Addresses #1015 

* Created external filtered-pools store that uses imperators sorted pools query
* Created IPoolGetter interface, so the pools query object can easily be swapped

TODO:
- [x] Fix tokens appearing properly in TradeClipboard in swap page
- [x] Fix tokens appearing properly in TradeClipboard in pool detail page (stableswap pool 877)
- [x] Fix 1/3 percentage display in add liquidity modal, should truncate to whole number
     Screen Shot 2022-12-23 at 4 33 59 PM
- [x] Balances not appearing in swap tool
- [x] Pools tables on mobile, loading all pools
- [x] Client side exception when first loading swap page on mobile


packages/stores/src/queries-external/filtered-pools/types.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+function floatNumberToStringInt(number: number, exponent: number): string {
+  return new Int(
+    new Dec(number.toString())
+      .mul(DecUtils.getTenExponentN(exponent))
+      .truncate()
+      .toString()
+  ).toString();
+}
```
JoseRFelix, line 115: Maybe we should move these utilities to a `utils.ts` file.

packages/stores/src/queries/pools/types.ts
```diff
@@ -1,6 +1,14 @@
+import { ObservableQuery } from "@keplr-wallet/stores";
 import { WeightedPoolRaw, StablePoolRaw } from "@osmosis-labs/pools";
 import { Duration } from "dayjs/plugin/duration";
 import { CoinPretty } from "@keplr-wallet/unit";
+import { ObservableQueryPool } from "./pool";
+
+export interface IPoolGetter extends ObservableQuery {
```
JoseRFelix, line 7: I'm not a fan of adding I to interfaces. What do you think if we rename this to `PoolGetter`?

JoseRFelix (commented):
LGTM! Left some comments

packages/web/components/complex/all-pools-table-set.tsx
```diff
@@ -245,7 +251,7 @@ export const AllPoolsTableSet: FunctionComponent<{
       sortedAllPoolsWithMetrics,
     ] = useSortedData(tvlFilteredPools, initialKeyPath, initialSortDirection);
 
-    const [query, setQuery, filteredPools] = useFilteredData(
+    const [query, do_setQuery, filteredPools] = useFilteredData(
```
JoseRFelix, line 254: Is there a reason why we add the `do_` prefix?

jonator: We have a wrapper function that performs additional operations before that core function is called. It's somewhat of a convention in our app, we use it a lot to log amplitude events along with the core function.

JoseRFelix: I feel the `do` makes it seem like snake case instead of camel. Maybe we could remove it it and leave the `_` like a private method. What do you think?

jonator: Sure, think we should do it elsewhere in the app?

JoseRFelix (approved):
LGTM 🚀  What a massive difference in performance! Amazing work @jonator

PR #1174, opened 2022-12-23
Xmas Theme

daniel-farina:


gnujoow:
happy Christmas everyone

jonator (changes requested):
Screen Shot 2022-12-23 at 12 38 09 PM

I can still see the old background in the top and bottom of the screen

daniel-farina:
> I can still see the old background in the top and bottom of the screen

I'm adding a very simple login to remove these classes from the div, quick dirty fix for now.

```
<div className={(IS_FRONTIER ? 'absolute h-full w-full bg-home-bg-pattern bg-cover bg-repeat-x' : '')}>
```

jonator:
Eh I'm still not a fan of it not being fit to the page

daniel-farina (commented):
lgtm

PR #1176, opened 2022-12-23
feat: Add Profile Modal

JoseRFelix:


jonator (changes requested):
Let's square out the left padding of the profile logo

Screen Shot 2023-01-04 at 5 12 33 PM

packages/web/public/icons/qr.svg
```diff
@@ -0,0 +1,19 @@
...
+<mask id="path-6-inside-2_195_212" fill="white">
+<rect x="13.5" y="10" width="6" height="6" rx="1"/>
+</mask>
+<rect x="13.5" y="10" width="6" height="6" rx="1" stroke="white" stroke-width="3" mask="url(#path-6-inside-2_195_212)"/>
+<mask id="path-7-inside-3_195_212" fill="white">
+<rect x="5.5" y="13" width="6" height="6" rx="1"/>
+</mask>
+<rect x="5.5" y="13" width="6" height="6" rx="1" stroke="white" stroke-width="3" mask="url(#path-7-inside-3_195_212)"/>
+<path d="M13.5 18H19M14.5 6H18.75C18.8881 6 19 6.11193 19 6.25V8" stroke="white" stroke-width="2" stroke-linecap="round"/>
+</svg>
```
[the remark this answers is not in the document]
JoseRFelix: Thanks for the review 👍 Resolved!

packages/web/utils/string.ts
```diff
@@ -5,3 +5,17 @@ export function truncateString(str: string, num = 8) {
   }
   return str.slice(0, num) + "...";
 }
+
+export function getShortAddress(
+  address: string,
+  opts?: { prefixLength?: number; suffixLength?: number }
```
jonator, line 11: FYI we do have this utility. It struggles at too low of a cutoff, but works nevertheless.

https://github.com/osmosis-labs/osmosis-frontend/blob/28d515b21ea8a999496790d7e82bf29ab90b0115/packages/web/components/complex/transfer.tsx#L123

packages/web/components/navbar/index.tsx
```diff
@@ -150,7 +152,7 @@ const NavBarButton: FunctionComponent<
 };
 
 const SettingsDropdown: FunctionComponent<{
-  userSettings: IUserSetting[];
+  userSettings: UserSetting[];
```
jonator, line 155: Hmm I think this is creating an empty div for the avatar user setting. This is an additional comment to the avatar user setting comment.

packages/web/stores/user-settings/avatar.tsx
```diff
@@ -0,0 +1,34 @@
+import { action, makeObservable, observable, computed } from "mobx";
+import { UserSetting } from ".";
+
+type Avatar = "wosmongton" | "ammelia";
+
+export type AvatarState = { avatar: Avatar };
```
jonator, line 6: Question: considering the future with NFTs etc, are we going to represent that state all within this user setting? I'm wondering if using this for profile state, which will likely grow, is a misuse of this abstraction. Should we create a new UI config store in root store? It will later do the work of querying contracts and loading prefs from localstorage. Curious what your thoughts are.

packages/web/stores/user-settings/index.ts
```diff
@@ -49,14 +57,17 @@ export class UserSettings {
...
   @computed
-  get userSettings(): IUserSetting[] {
+  get userSettings(): UserSetting[] {
     return this._settings;
   }
 
-  getUserSettingById = computedFn((id: string) => {
-    return this._settings.find(({ id: settingId }) => settingId === id);
-  });
+  readonly getUserSettingById = computedFn(
```
jonator, line 64: I like this

packages/web/styles/globals.css
```diff
@@ -188,6 +188,27 @@
...
+  z-index: 9999 !important;
+}
+
+.transak_modal {
+  z-index: 9999 !important;
+}
+
+.ReactModal__Overlay {
+  opacity: 0;
+  transition: opacity 150ms ease-in;
```
jonator, line 201: Nice touch

packages/web/utils/function.ts
```diff
@@ -1 +1,10 @@
+import { isFunction } from "./assertion";
+
 export const noop = () => {};
+
+export function runIfFn<T, U>(
```
jonator, line 5: This is fine, but I sense it's a sign of some over engineering happening somewhere.

packages/web/utils/string.ts
JoseRFelix: Replaced in favor of existing utility, thanks 👍

packages/web/components/navbar/index.tsx
JoseRFelix: Can you explain this a little bit more? I did not understand

packages/web/utils/function.ts
JoseRFelix: Indeed, I only intend to use this on custom headless components as it helps send state values through render props.

packages/web/stores/user-settings/avatar.tsx
JoseRFelix: I agree that this abstraction is simplistic for our future use case. Since this is very lean, we can remove it easily when implementing those features. I want to avoid adding a more complex abstraction for future use cases. What do you think?

packages/web/components/navbar/index.tsx
jonator: In that component, we map over `userSettings`, with the avatar setting containing no component or label, causing a div with no visible content to be created.

packages/web/stores/user-settings/avatar.tsx
jonator: @JoseRFelix think we should create a new store for the user's profile?

JoseRFelix: Yes, just noticed the other issue related to the empty div. I'll create another store

packages/web/components/navbar/index.tsx
JoseRFelix: Oh, got it. Will take a look!

packages/web/stores/user-settings/avatar.tsx
JoseRFelix: @jonator Added a commit with a profile store. Made it as simple as possible.

packages/web/stores/profile.ts
```diff
@@ -0,0 +1,34 @@
...
+      });
+    });
+  }
+
+  @computed
+  get currentAvatar(): Avatar {
+    return this._currentAvatar;
+  }
+
+  setCurrentAvatar(value: Avatar) {
```
jonator, line 30: We need to annotate this function with `@action` since it's updating an observable value.

jonator, line 30: We need to annotate this function with `@action` since it's updating an observable value.

jonator (changes requested):
Let's not forget about mobile. We should have it added to the Figma as well @syed8rahman @victortokaito. It should likely behave similarly to the settings modal (from dropdown menu) on mobile. I'm assuming it should be a pretty straightforward conversion, with items arranged horizontally.

JoseRFelix:
> Let's not forget about mobile. We should have it added to the Figma as well @syed8rahman @victortokaito. It should likely behave similarly to the settings modal (from dropdown menu) on mobile. I'm assuming it should be a pretty straightforward conversion, with items arranged horizontally.

@jonator do you mean the modal itself or the mobile menu item?

packages/web/modals/profile.tsx
```diff
@@ -0,0 +1,442 @@
...
+                    <QRIcon isAnimated />
+                  </ActionButton>
+                </DrawerButton>
+
+                <DrawerContent>
+                  <DrawerOverlay />
+                  <DrawerPanel className="flex h-fit items-center justify-center pt-7 pb-7">
+                    <h6 className="mb-8">Cosmos</h6>
+                    <div className="mb-7 flex items-center justify-center rounded-xl bg-white-high p-3.5">
+                      <QRCode value={address} size={260} />
```
jonator, line 280: We should try to use next/dynamic wherever possible, to keep the main bundle small. QRCode is a perfect example use case of using next/dynamic to explicitly keep it out of the main bundle.

https://github.com/osmosis-labs/osmosis-frontend/blob/c19969322b18a603992af37fd14d0cc749be5dd8/packages/web/modals/wallet-connect.tsx#L75

In this case, simply add 
`const QRCode = dynamic(() => import("qrcode.react"));`
At the top of the file, instead of the import statement.

jonator: Also- it's totally saved us for stuff like lottie animations. We can still take advantage of those big powerful packages, without reducing time-to-interactive. Of course, the tradeoff in this case is the user sees some delay when opening the view. In this case, it's saved us 100+kb.

https://github.com/osmosis-labs/osmosis-frontend/blob/c19969322b18a603992af37fd14d0cc749be5dd8/packages/web/components/animation/bridge/index.tsx#L11

jonator:
> > Let's not forget about mobile. We should have it added to the Figma as well @syed8rahman @victortokaito. It should likely behave similarly to the settings modal (from dropdown menu) on mobile. I'm assuming it should be a pretty straightforward conversion, with items arranged horizontally.
> 
> @jonator do you mean the modal itself or the mobile menu item?

I'm referring to the mobile version of the modal.

jonator:
@JoseRFelix I changed target branch to stage. Please resolve merge conflicts when you get the chance.

packages/web/config/user-analytics-v2.ts
```diff
@@ -142,8 +143,17 @@ export const EventName = {
     depositAssetCompleted: "Deposit asset: Deposit completed",
     withdrawAssetStarted: "Withdraw asset: Withdraw started",
     withdrawAssetCompleted: "Withdraw asset: Withdraw completed",
-    buyOsmoClicked: "Assets: Buy OSMO clicked",
     buyOsmoStarted: "Assets: Buy OSMO started",
     buyOsmoCompleted: "Assets: Buy OSMO completed",
   },
+  // Events in profile modal
+  ProfileModal: {
```
jonator, line 150: @victortokaito we'll need these to be added to amplitude

https://data.amplitude.com/osmosis-zone/Osmosis%20%F0%9F%A7%AA/events/main/latest

jonator (approved):

sunnya97:
Task linked: [CU-38ny2kp Profile Modal](https://app.clickup.com/t/38ny2kp)

jonator:
Screen Shot 2023-01-19 at 9 05 38 PM

Z index issue

JoseRFelix:
> Screen Shot 2023-01-19 at 9 05 38 PM
> 
> Z index issue

Resolved 👍

sunnya97:
Task linked: [CU-8669akrqf Profile Modal - Frontend Dev](https://app.clickup.com/t/8669akrqf)

jonator (approved):

packages/web/hooks/use-controllable-state.ts
```diff
@@ -0,0 +1,49 @@
...
+export interface UseControllableStateProps<T> {
+  value?: T;
+  defaultValue?: T | (() => T);
+  onChange?: (value: T) => void;
+  shouldUpdate?: (prev: T, next: T) => boolean;
+}
+
+/**
+ * The `useControllableState` hook returns the state and function that updates the state, just like React.useState does.
+ *
```
jonator, line 13: Could you add a sentence for what this does, or when it's useful

packages/web/hooks/use-controllable-prop.ts
```diff
@@ -0,0 +1,11 @@
+import { useMemo } from "react";
+
+/**
+ * Given a prop value and state value, the useControllableProp hook is used to determine whether a component is controlled or uncontrolled, and also returns the computed value.
+ *
```
jonator, line 5: Could you also add a sentence here for what it's used for

packages/web/hooks/use-controllable-prop.ts
```diff
@@ -0,0 +1,11 @@
...
+ */
+export function useControllableProp<T>(prop: T | undefined, state: T) {
```
jonator, line 7: Is this unused?

packages/web/hooks/use-controllable-state.ts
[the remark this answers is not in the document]
JoseRFelix: Added it, let me know if it looks good

packages/web/hooks/use-callback-ref.ts
```diff
@@ -0,0 +1,15 @@
+import { useCallback, useEffect, useRef } from "react";
+
+export function useCallbackRef<T extends (...args: any[]) => any>(
```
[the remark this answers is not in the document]
JoseRFelix: Likewise, added docs for this

packages/web/hooks/use-controllable-state.ts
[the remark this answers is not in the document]
jonator: Ah. Makes perfect sense. I could have used this for the tab box component. You can see that as another approach. If there's state they want to control, they can pass it in as an an object containing the current value and an update callback.

PR #1179, opened 2022-12-25
Deploy Stage

### LINKED ISSUE #772: USDC.grav token missing on Frontier under specific conditions
Change the target coin to a US-X and search for US-X again, and notice that USDC.grav is missing.

So, for example: in the below picture, let's say I searched for US-X, selected USDC, but then decided I wanted the other version. Search again, but USDC.grav is missing from the list.

### Steps
1. https://frontier.osmosis.zone/
2. Change the target coin to a USDC
3. Change the target coin again to a USD...notice the USDC.grav disappear from the list 
4. Try to type the full USDC.grav on the search bar and there is no USDC.grav show up to select at all

image

### LINKED ISSUE #896: stAssets hidden from Swap page selection when Asset in same box
stATOM is missing on both main frontend and frontier when ATOM is selected (as is default)
Also occurs with stSTARS, stJUNO and seJUNO when non-liquid assets are selected, but stATOM is the only front end listing.
As a user may go to select stATOM from this box to trade to/from OSMO this impacts trading.

image
image

When the from box is changed it appears in the router.
image

Assuming it is filtered out as being similar to the already selected asset but unsure where to change.

### LINKED ISSUE #1015: Optimize pool query

### LINKED ISSUE #1046: stATOM display bug
The default coins for the swap feature are ATOM / OSMO. For some reason, when ATOM is the upper coin in the selector, if you click ATOM and try to change to stATOM, the coin is not found. Sometimes I can find stATOM, most of the time not (I haven't figured out the pattern yet). It seems like the tokens slowly load into the search modal, and only become selectable once they are loaded.

Screen Shot 2022-11-22 at 11 08 12 PM

### LINKED ISSUE #1160: Polish localisation - very unfortunate translation!
in the polish (pl.json) localization the term "Unbond" has very wrong translation:

144:    "unbond": "Zwiąż",

this translation means exactly opposite to the orginal meaning. "Zwiąż" means in polish: Bond, not Unbond (that is why I was searching for unbonding feature for 3-4 minutes before I switched to english and noticed this bug

Correct wording for that should be:

144: "unbond": "Odwiąż", (means that You can untie something that is hanging from something or is attached to something) - preferred translation
or
144: "unbond": "Rozwiąż", (nearly the same meaning as above but has even better correlation to another polish wordings close to this subject) (this means that You can untie something what is tied: the rope for example)

You can substitute this word also with "Odwołaj" (Cancel) or "Wycofaj" (Withdraw)

Current wording is totally wrong, because it has opposite meaning to the desired one.

jonator:
* Pools performance. Fixes #1015 
* Fix users with bonded GAMM shares not seeing an unbond button
* Fix selecting token that contains same name
  * Fixes #772 
  * Fixes #1046 
  * Fixes #896 
* Fixes #1160 Thanks @sehyunc


arispech (approved):
- Able to unbound
- Swapping works fine

Trading page loads much faster. 


Screenshot 2022-12-27 at 12 17 03 PM

PR #1181, opened 2022-12-27
feat: add Arabic translation

czarcas7ic:
I am not fluent and had to look up quite a few words, but this should (awkwardly at times) get the point across. 

I need to let Vercel produce a test site so I can see what the right to left looks like, and then from there modify a few of these entries to properly show (for example, I wasn't positive how the {} would react when input in the middle of a sentence)


czarcas7ic:
Screenshot 2022-12-27 at 3 07 32 PM
Screenshot 2022-12-27 at 3 01 38 PM

Currently two issues im not sure how to fix

czarcas7ic:
Okay, thats the best I've got lol. The two problems mentioned above still exist, along with the the tab title on the specific pool card page should look like this
Screenshot 2022-12-27 at 4 01 30 PM

czarcas7ic:
I'm not able to run the frontend locally which is why I had to trigger all these Vercel actions, my apologies.

packages/web/localizations/dayjs-locale-ar.js
```diff
@@ -0,0 +1,45 @@
...
+  relativeTime: {
+    future: "بعد %s",
+    past: "منذ %s",
+    s: "ثوان",
+    m: "دقيقة",
+    mm: "%d دقائق",
+    h: "ساعة",
+    hh: "%d ساعات",
+    d: "يوم",
+    dd: "%d أيام",
```
jonator, line 34: The number being displayed in the beginning instead of end of string may be caused here.

https://github.com/osmosis-labs/osmosis-frontend/pull/1181#issuecomment-1366183426

czarcas7ic: Ahhh I will check this out, thanks

jonator:
> I'm not able to run the frontend locally which is why I had to trigger all these Vercel actions, my apologies.

It's pretty easy to run locally, just a few commands after you install the right dev deps. If you want me to help you with this, LMK we could hop on a quick call. @AlpinYukseloglu got up and running pretty quickly.

czarcas7ic:
@jonator  I didn't debug super deep but when I tried first time around with the simple "yarn" command I got the following

```
panic: data-dir is required

goroutine 1 [running]:
main.main()
        github.com/osmosis-labs/osmosis/v13/tests/e2e/initialization/node/main.go:47 +0x5f4
```

jonator:
Can reopen should we get an updated translation.

PR #1184, opened 2022-12-29
Deploy stage

### LINKED ISSUE #982: Use Axelar SDK `getTransferFee`

jonator:
* Queries for Axelar transfer instead of using hardcoded config. Fixes #982


arispech (approved):
Axelar works fine from the front-end, and other DEX functionality working as expected

daniel-farina (approved):

PR #1186, opened 2022-12-29
Fix token select in pool detail page

jonator:
* Initial tokens constructed with the trade token in config should only be used if they're both in sendable currencies

Areas to test:
* Selecting tokens in the swap tool, both on the swap page and in various pool detail pages (2 token, 3+token)


arispech (approved):
The swap tool on the front-end works as expected

PR #1187, opened 2022-12-31
Hotfix

jonator:
* Token query params now update swap tool
* Calc ACRE price from pool weights (more consistent then CG)
* Miscellaneous code readability improvements


JeremyParish69 (approved):
Price change checks out, appears to function correctly

Issue #1180, opened 2022-12-27
Data store improvements


jonator:
### Problem
The organization of our query and data stores was somewhat emergent from the organization of the chain queries and not planned at a high level. We have a lot of query stores that fetch and do highly-coupled compute on low-level chain data/primitives (notably gauges, locks, etc), then we have some data stores that wrap those query stores and do additional computations. Those data stores are not in the root store, but are created and used as needed, which may or may not be inefficient as the user navigates the pages, and sometimes causes us to repeat/copy&paste the computations (I've spotted this most noticeably in the various pool card views).

### Proposed Solution
We may be able to remove a lot of repeated (or dead) code by investigating the following:
* **Separate calculations from query stores**. The query stores should simply query the data, and provide computed properties that return that data in a parsed format with clean TypeScript types that include parsed keplr-wallet/unit objects, `Duration`s, `Date`s, etc.
  * We should create a document or FigJam that identifies all the view-dependent data models that we currently use, and we should collaborate on finding co-dependencies and then organizing them into logical data-oriented stores. Our focus should be on only performing calculations as needed for the user as they navigate the various views. Make heavy use of map stores. Some of this work already emerged in the [pool details](https://github.com/osmosis-labs/osmosis-frontend/blob/4656e74dd887311f2767c68352dc3bd6b3a3c4a0/packages/stores/src/queries/pools/pool-details.ts#L25) store and [superfluid pool](https://github.com/osmosis-labs/osmosis-frontend/blob/4656e74dd887311f2767c68352dc3bd6b3a3c4a0/packages/stores/src/queries/superfluid-pools/pool.ts#L26) store.
  * **We should move data stores into map stores in the root store**. For example, the aforementioned pool detail store and superfluid pool store should be a map store from poolId->store in the root. If we do that alone, we could save computation as a user navigates to/from pool detail pages and pools/assets pages (for the pool cards on those pages). Furthermore, the pool cards could reference those stores for calculating the metrics on the various flavors of pool cards we have.
[...]

PR #1190, opened 2023-01-02
Optimized gauge query + add derived data stores

### LINKED ISSUE #1182: Optimized gauge query
### Problem
Currently, we maintain a hardcoded map of pool IDs => gauge IDs, which is currently way out of date. In our current app, this results in many pools and gauges being queried that consist of all expired gauges.

### Proposed Solution
There exists an [external incentive gauge query](https://lcd.dev-osmosis.zone/swagger/#/Query/ExternalIncentiveGauges) that returns a list of all external gauges (3k+ gauge objects). However, since it includes expired gauges, it is much too expensive to query on the frontend. @JeremyParish69 and I agree that a viable solution would be to create a highly-cached endpoint on our Next.js webserver that runs that large query and filters for nonexpired gauges. It could also be a viable place to vet gauges (as described in #853). Furthermore, rather than just returning gauge IDs, we could have it return actual gauge objects which could then be used to create a new gauge store that maps pool IDs => [`ObservableQueryGauge`](https://github.com/osmosis-labs/osmosis-frontend/blob/1268b9a7ff1aabdc77473b4232f5bd071f32a8f1/packages/stores/src/queries/incentives/gauge-by-id.ts#L16) objects. This would collate hundreds of individual gauge queries into a single highly-cached query provided by our Next.js server.

jonator:
* Add new API endpoint (on web server) that computes active external gauges from all gauge query every 15 minutes. Fixes #1182 
* Added a derived data map store, that computes higher-level metrics on top of lower-level query stores at the root level store (maps poolID=>data store). Added use of these in the various card views (assets page, pools page), as well as pool detail page. Results in less repeated computation, less repeated/similar code, etc. Going to consider this enough to address thoughts in #1180, which could be reopened if needed (as more improvements/refactoring could be made). Eventually, I'd like these stores to contain all computations on lower-level pools/gauge/lock data.

Testing
* Pool detail page, displaying bond cards
* Pools page, external


JeremyParish69 (commented):
Missing coins for gauges for pool 790. Expect page to show both GLTO and GKEY under each duration, but only the GLTO coin and APR is displayed. (Note: even now, the current site erroneously shows GLTO twice under each duration, rather than showing GLTO once and GKEY once under each duration.)

jonator:
I definitely notice a performance improvement, especially when navigating across pages (_especially_ navigating to the pools page).

JeremyParish69 (approved):
correctly pulls brand new gauges in a timely manner. Correctly lists multiple distributed coins, even when under the same gaugeId. Appears to have correct number of external incentives under Pools page and appears accurate on any particular pool's page.

PR #1191, opened 2023-01-02
add STRD gauge to stOSMO / OSMO pool

asalzmann:
replace old stOSMO / OSMO gauge id with new STRD gauge id

### DIFF
--- a/packages/web/config/feature-flag.ts
+++ b/packages/web/config/feature-flag.ts
@@ -707,7 +707,7 @@ const MainnetExternalIncentiveGaugeAllowList: {
   ],
   "833": [
     {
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Note @jonator this will probably be the last gauge to add manually--just didn't want them to miss out on advertised distribution epochs in case your PR will still be a while longer.

FYI @asalzmann we're switching to automated gauge listing via query, so the front end will display these gauges without needing these PRs. I can already see your new gauge displayed in his build that will implement this change.

jonator:
So exciting

asalzmann:
Nice!!

PR #1196, opened 2023-01-06
fix: Swap tool enter key not selecting asset

JoseRFelix:


jonator (changes requested):
Hmm it seems like it works for the top token select, but not the bottom..

JoseRFelix:
> Hmm it seems like it works for the top token select, but not the bottom..

Right, forgot the `tokenOut`. Added it in latest commit

jonator (approved):

PR #1201, opened 2023-01-11
feat: Match design guidelines for topbar and transaction settings popover

JoseRFelix:


jonator (commented):
Seems this is using the old pools query that is no longer supported. I'd update from stage to fix.

packages/web/components/popover.tsx
```diff
@@ -0,0 +1 @@
+export { Popover } from "@headlessui/react";
```
jonator, line 1: What's the purpose of this file?

jonator (changes requested):
Can we make sure the background of the settings popover and the drawer are the same color?

Screen Shot 2023-01-22 at 7 16 48 PM
Screen Shot 2023-01-22 at 7 17 20 PM

packages/web/components/popover.tsx
JoseRFelix: Most likely, we'll extend the popover functionality in the future by adding animations. So I think it's worth using our own import to not have to replace all imports.

jonator (approved):

jonator (approved):

jonator:
@JoseRFelix please fix fix merge conflicts once you get the chance, then I can include this in the next staged bundle

PR #1203, opened 2023-01-12
Redesign pools table

sehyunc:
[Design](https://www.figma.com/file/ddlsIdRm0p30bFC5Lyt5OA/Pools?node-id=115%3A3777&t=rZ9tYg1w5q255aX5-4)


packages/web/components/complex/all-pools-table-set.tsx
```diff
@@ -1,38 +1,41 @@
...
+import { POOLS_PER_PAGE } from ".";
+import { EventName, IS_FRONTIER } from "../../config";
 import {
-  useFilteredData,
-  usePaginatedData,
-  useSortedData,
-  useWindowSize,
   useAmplitudeAnalytics,
+  useFilteredData,
+  // useWindowSize,
```
jonator, line 29: Please remove comment

packages/web/styles/globals.css
```diff
@@ -100,15 +100,15 @@
   }
 
   thead > tr > th:first-child {
-    @apply rounded-l-full py-[18px] pl-10 pr-3 md:rounded-none;
+    @apply rounded-tl-lg py-[18px] pl-10 pr-3 md:rounded-none;
```
jonator, line 103: This will affect all pools in the app. We should confirm with design if this is intended.

sunnya97:
Task linked: [CU-38nrteq Merge Tables Pools page](https://app.clickup.com/t/38nrteq)

jonator (changes requested):
Seems there were some build errors, related to unused imports. It should be visible in your dev env.

JoseRFelix:
This is looking good! Btw let's add some infinite loading with [react virtual](https://tanstack.com/virtual/v3, [react window](https://github.com/bvaughn/react-window) or [react virtuoso](https://virtuoso.dev/). If not, we'll still have issues loading all these pool items.

jonator:
Agree with Jose, though with one table I can already notice substantial performance improvements.

jonator (changes requested):
A few things I noticed out of the gate:
* Rows should be href linkable- meaning I can cmd+click or right click and open in new tab
* The pool quick actions don't seem to work anymore

jonator:
@sehyunc can we store the tab state in the URL instead as well?

jonator (changes requested):
Looks like there's some z-index issues with token icons
Screen Shot 2023-02-08 at 8 32 39 AM

packages/web/components/complex/all-pools-table-set.tsx
```diff
@@ -51,70 +42,74 @@ type PoolWithMetrics = {
...
+    onLockTokens?: () => void;
+  }
+];
+
+type Filter = "superfluid" | "stable" | "concentrated" | "weighted";
+
+const Filters: Record<Filter, string> = {
+  superfluid: "Superfluid",
+  stable: "Stableswap",
+  concentrated: "Concentrated Liquidity",
```
jonator, line 80: Let's comment this out or remove it until CL is released.

packages/web/components/complex/all-pools-table-set.tsx
```diff
@@ -51,70 +42,74 @@ type PoolWithMetrics = {
...
+  {
+    poolId: string;
+    cellGroupEventEmitter: EventEmitter<string | symbol, any>;
+    onAddLiquidity?: () => void;
+    onRemoveLiquidity?: () => void;
+    onLockTokens?: () => void;
+  }
+];
+
+type Filter = "superfluid" | "stable" | "concentrated" | "weighted";
```
jonator, line 75: IMO if this type isn't needed elsewhere, it can just be removed. Generally, we should only add types if they're beneficial elsewhere, as a way to minimize code &/ complexity.

jonator (changes requested):
- [x] For the quick actions column, it should be the minimal width of the ... button as we had before.

jonator (changes requested):
For some reason some of the APRs are not appearing. May be a memoization issue.
Screen Shot 2023-02-09 at 9 12 18 AM

We should have the liquidity sort descending by default:
Screen Shot 2023-02-09 at 9 13 38 AM

Also I've noticed a weird jitter when scrolling on the table:
https://user-images.githubusercontent.com/4606373/217852810-00f73af0-8d06-4177-b13b-4344237b3f69.mov

These controls should be the same height, it may have to do with the way you're using flexbox:
Screen Shot 2023-02-09 at 9 15 25 AM

jonator:
Should also close #1248

jonator:
@syed8rahman is making updates to the design
* It seems the entire table is being loaded initially, causing the perf issues. @JoseRFelix believes we should be able to get a ref to the entire page to get the scroll position, as we had for the embedded table
  * We should be able to load 10-20 initially and not have render issues
* We should see if there's a way to not jump to the top of the page when the filter settings get changed, even though the query params are being updated

sehyunc:
> @syed8rahman is making updates to the design
> 
> * It seems the entire table is being loaded initially, causing the perf issues. @JoseRFelix believes we should be able to get a ref to the entire page to get the scroll position, as we had for the embedded table
>   
>   * We should be able to load 10-20 initially and not have render issues
> * We should see if there's a way to not jump to the top of the page when the filter settings get changed, even though the query params are being updated

I've enabled scroll restoration as mentioned in [these docs](https://nextjs.org/docs/api-reference/next/link#disable-scrolling-to-the-top-of-the-page), clicking filter button no longer resets scroll to top. I've also decreased the limit for the initial fetch.

sunnya97:
Search functionality seems to be having some issues.

Examples:

1.  Can't find pools for LUNA
CleanShot 2023-02-16 at 21 49 27@2x


2. Pools for BTSG are not at the top
CleanShot 2023-02-16 at 21 50 21@2x

sunnya97:
Clicking the Add Liquidity button via the triple dot menu in the pool table causes an application error

CleanShot 2023-02-16 at 21 56 59@2x

sunnya97:
What does the "Weighted" tab mean?  Just any balancer-based pool?

jonator:
> What does the "Weighted" tab mean? Just any balancer-based pool?

Yes

packages/web/components/complex/paginated-table.tsx
```diff
@@ -0,0 +1,134 @@
...
+import Image from "next/image";
+import Link from "next/link";
+import { useRef } from "react";
+import { IS_FRONTIER } from "../../config";
+import { Pool } from "./all-pools-table-set";
+
+const SIZE = 80;
+
+type Props = {
+  paginate: () => void;
```
jonator, line 12: I don't see this actually get called, does this table actually paginate?

packages/web/components/overview/pools.tsx
```diff
@@ -8,106 +8,121 @@ import { useStore } from "../../stores";
...
+              ? width < Breakpoint.XL
+                ? 280
+                : 380
+              : 425
+          }
+        />
+      </div>
+      <div className="absolute right-7 bottom-7">
+        <Button
+          className="rounded-[24px] text-white-full"
```
jonator, line 111: Missing drop shadow; make sure it matches design

packages/web/hooks/use-on-screen.ts
```diff
@@ -0,0 +1,49 @@
+import { RefObject, useEffect, useState } from "react";
+
+interface Args extends IntersectionObserverInit {
+  freezeOnceVisible?: boolean;
+}
+
+function useOnScreen(
```
jonator, line 7: Is this used? If not, please remove.

packages/web/components/complex/paginated-table.tsx
```diff
@@ -0,0 +1,134 @@
...
+  table: Table<Pool>;
+};
+
+const PaginatedTable = ({ table }: Props) => {
+  const tableContainerRef = useRef<HTMLDivElement | null>(null);
+
+  const { rows } = table.getRowModel();
+  const rowVirtualizer = useVirtualizer({
+    count: rows.length,
+    getScrollElement: () => tableContainerRef.current,
```
jonator, line 22: For the ref, we may need to look into passing a ref as props if the user of PaginatedTable wants to use a different container ref. In our case, we'd use the whole page as they scroll.

jonator (changes requested):
- [ ] It appears the table doesn't sort by liquidity on first load

packages/web/components/complex/paginated-table.tsx
sehyunc: Does not paginate, but instead calls `fetchRemainingPools`. I think I mentioned to you earlier the paginate function call isn't consistent, but `fetchRemainingPools` works well so I thought we were going with that. I do need to re-enable it, which I'm working on.

sehyunc: @jonator Should I rename the function to better reflect that we are fetching remaining pools instead of paginating? Might want to keep it in case other use cases actually use pagination.

jonator: Ah no that makes sense. When are remaining pools fetched when the user is not interacting with the sort/search controls? It should load remaining when they scroll to bottom of viewbox.

jonator: Also for the container ref.. make sure it's a ref to the container actually scrolling the *entire* page. Not the immediate container of the table. What matters is `react-table` has a ref to the element handling the scrolling.. otherwise it doesn't know when the page was scrolled to bottom.

jonator: _Also_ make sure that `containerRef` _itself_ is optional, then only use `getScrollElement` when needed

sunnya97:
I'm still getting a significant number of repeated pools.  This screenshot is with the superfluid filter activated.

CleanShot 2023-02-23 at 19 38 17

sunnya97:
@sehyunc any update on this?

jonator (changes requested):
If no filters are selected, there should not be any query params:
`/pools?pool=&incentive=`

Also, we still need to implement #1349

packages/web/components/complex/all-pools-table.tsx
```diff
@@ -0,0 +1,707 @@
...
+        }),
+      [
+        allPools,
+        priceStore,
+        queriesOsmosis.queryGammPoolShare,
+        queriesOsmosis.queryIncentivizedPools,
+        queriesOsmosis.querySuperfluidPools,
+        queriesOsmosis.querySuperfluidOsmoEquivalent,
+        account.bech32Address,
+        queriesExternalStore.queryGammPoolFeeMetrics,
```
jonator, line 202: Not sure if you removed them, but we do actually want the response members of these query stores, since that is what is updated when queries land.

packages/web/components/complex/all-pools-table.tsx
```diff
@@ -0,0 +1,707 @@
...
+          const pool: Pool = [
+            {
+              poolId,
+              poolAssets,
+              stableswapPool: poolWithMetrics.pool.type === "stable",
+            },
+            { value: poolWithMetrics.liquidity },
+            {
+              value: poolWithMetrics.volume24h,
+              isLoading: !queriesExternalStore.queryGammPoolFeeMetrics.response,
```
jonator, line 304: `queriesExternalStore.queryGammPoolFeeMetrics` has an `isFetching` member that would be more idiomatic to use for figuring out if something is fetching.

packages/web/components/drawers/drawer.tsx
```diff
@@ -57,7 +57,7 @@ export const Drawer: FunctionComponent<{
         props.onClose?.();
       },
     }),
-    [isOpen, isAnimationComplete, setIsOpen]
+    [isOpen, isAnimationComplete, setIsOpen, props]
   );
```
jonator, line 61: Won't this memo run every render if the props object is passed? My thought is that if React triggers a re-render due to props changing, then the memo would run every render since props is in deps... LMK if I'm off here.

packages/web/components/input/search-box.tsx
```diff
@@ -20,9 +20,10 @@ const searchBoxClasses = cva(
        * - letter spacing
        */
       size: {
-        small: "h-10 px-5 w-max [&_input]:text-body2 [&_input]:font-body2",
-        medium: "h-12 px-5 w-max [&_input]:text-body2 [&_input]:font-body2",
-        large: "h-14 px-5 w-max [&_input]:text-body1 [&_input]:font-body2",
+        small: "h-10 pl-5 pr-1 w-max [&_input]:text-body2 [&_input]:font-body2",
```
jonator, line 23: It's very nitpicky: but it seems now the search box changes size when it's focused now, causing neighboring elements to shift. Don't let it block you, but if we could avoid that it'd be idea.

packages/web/.eslintrc.json
```diff
@@ -15,7 +15,7 @@
       }
     ],
     "@next/next/no-img-element": "off",
-    "react-hooks/exhaustive-deps": "off",
+    // "react-hooks/exhaustive-deps": "off",
```
jonator, line 18: So should this setting be on?

jonator (commented):
To make it clear that a filter is currently active, could we highlight the text with the same color we use for selected items in the dropdown?:
Screenshot 2023-03-03 at 9 14 42 AM

packages/web/.eslintrc.json
sehyunc: The time is on by default, I will delete the comment.

packages/web/components/drawers/drawer.tsx
sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

sehyunc: You're right, since React uses referential equality in dependency arrays and I believe props is not the same referential object across re-renders. I'll destructure props and pass them in instead.

packages/web/components/input/search-box.tsx
sehyunc: I wasn't able to find a quick fix for this, I thought `box-sizing: border-box;` would work but that's not the case. Will need more time to fix it.

packages/web/components/complex/all-pools-table.tsx
sehyunc: I put back the dependencies that were there before, minus some objects that weren't referenced at all within the useMemo hook.

packages/web/components/input/search-box.tsx
sehyunc: Fixed by making border width consistent across states (default, onFocus).

packages/web/components/complex/paginated-table.tsx
```diff
@@ -0,0 +1,184 @@
...
+                        )}
+                      </a>
+                    </Link>
+                  </td>
+                );
+              })}
+            </tr>
+          );
+        })}
+        <div ref={loaderRef} />
```
jonator, line 173: What is this used for? I got a warning that div cannot be used as child of tbody.

jonator (changes requested):
Generally, it seems _only_ when my wallet is connected, the table rows disappear or don't render:
Screenshot 2023-03-06 at 9 23 21 AM

sehyunc:
> Generally, it seems _only_ when my wallet is connected, the table rows disappear or don't render: Screenshot 2023-03-06 at 9 23 21 AM

Is this issue happening with Keplr wallet or a different wallet? I'm not able to replicate the issue on my end. I have seen this happen when CORS was disabled, could this be a CORS related issue?

packages/web/components/complex/paginated-table.tsx
sehyunc: This is used to trigger pagination, I use the `useOnScreen` hook to observe when the ref appears on screen, and run `fetchRemainingPools` when that happens since this is the last div in the table. I've fixed the console errors by moving this ref into the table row instead.

jonator:
No I'm not seeing any issues in the network tab..

jonator (approved):

PR #1206, opened 2023-01-12
Update Axelar testnet network and channels

daniel-farina:
As documented here: https://docs.axelar.dev/resources/testnet#ibc-channels

The network id for osmosis is `osmosis-5` and there are new channels  for the testnet. This is part of fixing the network configuration for the testnet. 

@jonator  Please let me know if there are any other places we might need to update. 

Please don't merge yet, I will do some test with the preview link first.


JeremyParish69:
@daniel-farina some testnet transfers have worked in the past. Have they stopped working? Usually these shouldn't need changing. Governance can usually restore them. In the case the new channel-ids are to be used, 19XX is a really high number--perhaps confused with connection number?

packages/web/config/ibc-assets.ts
```diff
@@ -27,8 +27,8 @@ export const IBCAssetInfos: (IBCAsset & {
     counterpartyChainId: IS_TESTNET
       ? "axelar-testnet-lisbon-3"
       : "axelar-dojo-1",
-    sourceChannelId: IS_TESTNET ? "channel-312" : "channel-208",
-    destChannelId: IS_TESTNET ? "channel-22" : "channel-3",
+    sourceChannelId: IS_TESTNET ? "channel-135" : "channel-1946",
```
JeremyParish69, line 30: leave the prod channel id as channel-208

packages/web/config/ibc-assets.ts
```diff
@@ -27,8 +27,8 @@ export const IBCAssetInfos: (IBCAsset & {
...
+    destChannelId: IS_TESTNET ? "channel-1946" : "channel-3",
```
JeremyParish69, line 31: the testnet channel ids should be flipped for all 4 of these assets--source (osmosis side) as 1946 and dest (axelar side) as 135

jonator:
@JeremyParish69 should be able to guide you on the config changes

packages/web/config/ibc-assets.ts
daniel-farina: > leave the prod channel id as channel-208
My bad. Good catch! ty

packages/web/config/ibc-assets.ts
JeremyParish69: should be corrected now

daniel-farina: ok, I'll test it!

daniel-farina: @JeremyParish69  can we also add Axelar on the testnet? We should probably have a testnet for frontier as well @jonator .

jonator: If anything, the testnet should be frontier, since it includes all assets from main. This would be simpler

JeremyParish69 (commented):
With the addition of gasPriceStep for Testnet, we should finally be good for this to work. But I believe @daniel-farina is doing transfer validation, is that right?

JeremyParish69 (approved):

PR #1219, opened 2023-01-16
Lower the 'w' in wrapped token symbols

JeremyParish69:
.axl. and .grv
e.g., wbtc, weth, wbnb, wmatic


jonator:
Please update from master to fix build error

jonator (approved):
LGTM. Out of curiosity, why not WBTC?

jonator:
@JeremyParish69 with addition of WAVAX, please update this branch from master

sunnya97:
Cause WBTC is like a very different thing than wETH, wAVAX, etc

Cause the latter are more just smart contract wrapped, and are basically just a different representation of ETH

But the former is like centralized bitgo custodied, so I'd argue it's actually a fundamentally different asset than BTC

sunnya97:
Should we maybe just remove the W's altogether?  Especially given that we will soon be able to deposit native assets?

Is it worth lowercasing the w, just to remove it in a little bit?

sunnya97 (approved):

sunnya97:
@JeremyParish69 can you lowercase w in wAVAX too

JeremyParish69:
> @JeremyParish69 can you lowercase w in wAVAX too

Yep, on it: https://github.com/osmosis-labs/osmosis-frontend/pull/1233

PR #1222, opened 2023-01-18
feat: Add ICNS to Profile Modal

JoseRFelix:
## Considerations

Keplr's new versions (>=0.11.0) increase our bundle size from 1.64 MB to 2.79 MB (65% increase).


packages/web/components/navbar/index.tsx
```diff
@@ -161,6 +161,23 @@ const SettingsDropdown: FunctionComponent<{
   );
 });
 
+export const formatICNSName = (name?: string, maxLength = 28) => {
```
jonator, line 164: Thoughts on moving this to string utils file?

packages/stores/src/queries-external/icns/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  get names(): string[] {
+    if (!this.response || !this.response.data) {
+      return [];
+    }
+
+    return this.response.data.names;
+  }
+}
+
+export class ObservableQueryICNSNames extends ObservableChainQueryMap<ICNSNames> {
```
jonator, line 44: Could we try to update our keplr packages and use the one from Keplr wallet instead of copying and maintaining our own?

sunnya97:
Task linked: [CU-8669b8m20 Integrate ICNS - FE Dev](https://app.clickup.com/t/8669b8m20)

sunnya97:
Task linked: [CU-8669bgure ICNS Frontend - FE Dev](https://app.clickup.com/t/8669bgure)

packages/stores/src/queries-external/icns/index.ts
JoseRFelix: Done! Latest commit updates and uses keplr packages for ICNS

packages/web/hooks/ui-config/use-trade-token-in-config.ts
```diff
@@ -20,7 +21,9 @@ export function useTradeTokenInConfig(
   chainGetter: ChainGetter,
   osmosisChainId: string,
   bech32Address: string,
-  queriesStore: QueriesStore<[CosmosQueries, CosmwasmQueries, OsmosisQueries]>,
+  queriesStore: QueriesStore<
```
jonator, line 24: I actually found out this can be simplified if we use the IQueriesStore interface. Then, only add the queries in that list that are actually needed. (It's more generic).

https://github.com/osmosis-labs/osmosis-frontend/blob/a3b18b99dd7026ed7acd65ec3553abf0d4486cfe/packages/stores/src/account/index.ts#L31-L32

Should fix the build errors elsewhere as well.

packages/web/config/feature-flag.ts
```diff
@@ -65,3 +65,8 @@ export const PromotedLBPPoolIds: {
       },*/
     ]
   : [];
+
+export const ICNSInfo = {
```
jonator, line 69: May be helpful to link this as well

https://www.mintscan.io/osmosis/wasm/contract/osmo1xk0s8xgktn9x5vwcgtjdxqzadg88fgn33p8u9cnpdxwemvxscvast52cdd

jonator (changes requested):

packages/web/hooks/ui-config/use-trade-token-in-config.ts
JoseRFelix: Nice catch 🙌  

Updated

jonator (approved):

PR #1227, opened 2023-01-18
Add: Marsprotocol testnet

linkielink:
- Fixed a minor type issue
- Added ares-1 information
- Added MARS on osmo-test-4 to the chain-infos.ts


packages/web/config/chain-infos.ts
```diff
@@ -2484,7 +2484,37 @@ const chainInfos = (
...
+      bip44: {
+        coinType: 118,
+      },
+      bech32Config: Bech32Address.defaultBech32Config("mars"),
+      currencies: [
+        {
+          coinDenom: "MARS",
+          coinMinimalDenom: "umars",
+          coinDecimals: 6,
+          coinGeckoId: "mars",
```
JeremyParish69, line 2504: It looks like this is referring to a different token on CoinGecko and thus should be removed.

packages/web/config/ibc-assets.ts
```diff
@@ -1188,6 +1188,13 @@ export const IBCAssetInfos: (IBCAsset & {
     destChannelId: "channel-2",
     coinMinimalDenom: "dys",
   },
+  {
+    counterpartyChainId: "ares-1",
+    sourceChannelId: "channel-2083",
+    destChannelId: "channel-2",
+    coinMinimalDenom: "umars",
+    isVerified: true,
```
JeremyParish69, line 1196: I think this object can just be preceded by `IS_TESTNET ?`

linkielink: I tried several versions to make this object conditional, but the only way it allowed me to build the repository with was 
```
    counterpartyChainId: IS_TESTNET ? "ares-1" : "",
    sourceChannelId: IS_TESTNET ? "channel-2083" : "",
    destChannelId: IS_TESTNET ? "channel-2" : "",
    coinMinimalDenom: IS_TESTNET ? "umars" : "",
    isVerified: IS_TESTNET ? true : false,
```

The problem I see with this is that it results in an object looking like this on mainnet: 
```
    counterpartyChainId: "",
    sourceChannelId:  "",
    destChannelId: "",
    coinMinimalDenom: "",
    isVerified: false,
```
I didn't check if empty strings are handled by the `status-by-id.ts` for example., but I could not determine any issues with it. As soon as mars-1 launches I will update the values on the right side of the condition and the "empty" object will be gone.

packages/web/config/chain-infos.ts
linkielink: You are right, Mars Protocols coinGeckoId is `marsprotocol` but it's currently disabled, until mainnet launch

packages/web/config/ibc-assets.ts
JeremyParish69: I decided to give another idea a try, which seems to work--you can check it out here: https://github.com/osmosis-labs/osmosis-frontend/pull/1238

linkielink:
PR Closed as #1238 takes care of the changes.

PR #1236, opened 2023-01-19
Re-order assets by MarketCap on CoinGecko

JeremyParish69:
Tokens without a market cap are by enlisting order.
Axelar bridged assets are sorted by their underlying (and unwrapped) token values.
GBridge tokens are listed immediately below GRAV.
WBTC.axl is listed where WBTC is (~ rank 21), not where BTC is (rank 1)
OSMO and ION are still at the top since they are not ordered in the same file. Would need significant redevelopment to mix their ordering with the others (I recommending letting this pass until the next major refactor)


sunnya97:
Moved WBTC to rank 1.  Pre-favorited ATOM.

sunnya97 (approved):

PR #1239, opened 2023-01-21
Hotfix

### LINKED ISSUE #843: Swap fees from two OSMO/LUNC pools
There are 2 OSMO/LUNC pools, one has a swap fee of 8% and another has a swap fee of 0.2%. The pool that has 8% fee always be picked when trying to swap between Osmo and Lunc

A user reported massage :
"Make swap on swap-tab from LUNC to asset 2. Most optimal swap is misrepresented; screenshots attached. My explanation is as follows > The swap router is optimizing pre-Cost_of_execution, leading to incorrect optimization of swap pathing. 1000 LUNC > OSMO is more often than note being routed through pool 800, an 8% swap fee lunc/osmo pool because the router 'thinks' it's cheaper; until swap fee is factored in it most usually isn't cheaper than directly swapping through pool 561 (.3% lunc/osmo pool)
Below's images do not fully exemplify this 'bug' because the swap tab found the most optimal route; this is not always the case however it is quite a large spread because of the 8% swap fee's skew."

image
image
image

### LINKED ISSUE #1226: Intermittent Front-End Display bug (swap tab & pools creation)
https://app.osmosis.zone/?from=ATOM&to=stATOM

User report:
Bug presents with specific assets like stATOM and WETH.g where the front-end bugs displaying the asset as another asset (For example stATOM=ATOM or WETH.g = WETH)

This mainly is a visual bug but conflicts with user flow (such as creating pools, for example I experienced this bug trying to make the WETH.g:WETH stable swap pool where it actually showed both assets as WETH. It was a front-end problem and underneath both assets were actually unique)
Hard to force, clearing cache seems to help. 

As seen in the images it shows ATOM, not stATOM 

image
image

### LINKED ISSUE #1240: Low market cap tokens not available in swap page
### Problem
Our pools query, to maintain high performance in frontend, only loads top ~120 pools. For the swap page, the tokens available to trade are those in the pools given to the swap router, which is most of those 120 pools that meet the [conditions](https://github.com/osmosis-labs/osmosis-frontend/blob/30278e25eef9a4438afed86901dbd9936f684210/packages/web/pages/index.tsx#L25) in the swap page component.

### Proposed
We should somehow get all the assets in the assets page, then get the highest liquidity pools for those assets, instead of just the top liquidity pools. I'm not sure how that would work given our current query architecture yet.

### Example
LUNA is now available on main site, but its only pool is one with ~$4k currently. Since it meets the [filter](https://github.com/osmosis-labs/osmosis-frontend/blob/30278e25eef9a4438afed86901dbd9936f684210/packages/web/pages/index.tsx#L25) conditions in the swap page (1k OSMO), it can only become available to swap if the user navigates the the pools page, selects all pools (to request remaining pools), then returns to the swap page.

### LINKED ISSUE #1244: Pool Creation Modal's Asset Selection Faulty
Users reported

https://frontier.osmosis.zone/pools

Pools > Create Pool > (either Weighted or Stable) > Add New Token > ATOM (alas stATOM is selected instead)

https://user-images.githubusercontent.com/56182049/214003536-104f520a-79b1-420f-9b6e-4c2b42424158.mov

jonator:
* Use correct (same) endpoint for filtered pools
* Remove high fee LUNA pool - Closes #843 

Later
* Fix issue with some tokens not being in swap tool (mostly frontier) by loading remaining pools when the drawer is opened
  * Fixes #1240 
* Fix token select token being similar name but not same token (`.includes()` => `===` for Denom)
  * Fixes #1244 
  * Fixes #1226 
* Fix translation in completed tx toast

### DIFF
--- a/packages/web/components/trade-clipboard/index.tsx
+++ b/packages/web/components/trade-clipboard/index.tsx
@@ -142,10 +142,31 @@ export const TradeClipboard: FunctionComponent<{
     }, [isSettingOpen]);
 
     // token select dropdown
-    const [showFromTokenSelectDropdown, setFromTokenSelectDropdownLocal] =
+    const fetchedRemainingPoolsRef = useRef(false);
+    const fetchRemainingPoolsOnce = useCallback(() => {
+      if (!fetchedRemainingPoolsRef.current) {
+        fetchedRemainingPoolsRef.current = true;
+        queries.osmosis?.queryGammPools.fetchRemainingPools();
+      }
+    }, [fetchedRemainingPoolsRef]);
+    const [showFromTokenSelectDropdown, _setFromTokenSelectDropdownLocal] =
       useState(false);
-    const [showToTokenSelectDropdown, setToTokenSelectDropdownLocal] =
+    const setFromTokenSelectDropdownLocal = useCallback(
+      (val: boolean) => {
+        fetchRemainingPoolsOnce();
+        _setFromTokenSelectDropdownLocal(val);
+      },
+      [fetchRemainingPoolsOnce]
+    );
+    const [showToTokenSelectDropdown, _setToTokenSelectDropdownLocal] =
       useState(false);
+    const setToTokenSelectDropdownLocal = useCallback(
+      (val: boolean) => {
+        fetchRemainingPoolsOnce();
+        _setToTokenSelectDropdownLocal(val);
+      },
+      [fetchRemainingPoolsOnce]
+    );
     const setOneTokenSelectOpen = (dropdown: "to" | "from") => {
       if (dropdown === "to") {
         setToTokenSelectDropdownLocal(true);
--- a/packages/web/components/trade-clipboard/index.tsx
+++ b/packages/web/components/trade-clipboard/index.tsx
@@ -148,7 +148,7 @@ export const TradeClipboard: FunctionComponent<{
         fetchedRemainingPoolsRef.current = true;
         queries.osmosis?.queryGammPools.fetchRemainingPools();
       }
-    }, [fetchedRemainingPoolsRef]);
+    }, []);
     const [showFromTokenSelectDropdown, _setFromTokenSelectDropdownLocal] =
       useState(false);
     const setFromTokenSelectDropdownLocal = useCallback(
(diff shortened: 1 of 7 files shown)


sunnya97:
@jonator It doesn't seem to be able to "remember" the assets it finds using the pools query.  See video with CRE:

https://user-images.githubusercontent.com/979241/213840956-93564271-ee79-42fe-a66a-1c4fdf552254.mov

arispech:
> @jonator It doesn't seem to be able to "remember" the assets it finds using the pools query. See video with CRE:
> 
>  cre.mov

I can confirm the same behavior. 

1. Start here https://osmosis-frontier-git-jonator-hotfix-osmo-labs.vercel.app/?from=ACRE&to=OSMO
2. Search for Crecent, not showing
image

3. Go to pools, search for "CRE", note ACRE will show up first, and after a few seconds, CRECENT/OSMO will show up.
4. Go back to the swap page
5. Search for Cre, now Crescent shows
image
6. Click on it Crescent and it's not selectable

arispech:
> * Use correct (same) endpoint for filtered pools
> * Remove high fee LUNA pool - Closes [Swap fees from two OSMO/LUNC pools #843](https://github.com/osmosis-labs/osmosis-frontend/issues/843)

The pools page takes a few seconds to load on my end, seems slower than usual.

jonator:
@arispech yes the search takes some time since the page needs to request remaining pools >> process them >> filter with search.

jonator:
@arispech ready for review (btw, you can click "add your review" in the top right banner of page, then add your comment in the review box. That way, I can clearly see it's a review and not a comment. Also, I can re-request a review from you later)

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -142,10 +142,31 @@ export const TradeClipboard: FunctionComponent<{
...
 
     // token select dropdown
-    const [showFromTokenSelectDropdown, setFromTokenSelectDropdownLocal] =
+    const fetchedRemainingPoolsRef = useRef(false);
+    const fetchRemainingPoolsOnce = useCallback(() => {
+      if (!fetchedRemainingPoolsRef.current) {
+        fetchedRemainingPoolsRef.current = true;
+        queries.osmosis?.queryGammPools.fetchRemainingPools();
+      }
+    }, [fetchedRemainingPoolsRef]);
```
JoseRFelix, line 151: We don't need to add the ref as a callback since it will always be stable.

arispech (approved):
The issue where the token is not loading seems fixed.  When doing a search for "cre" for example both tokens load successfully from the pools page and swap page. 
👍

PR #1257, opened 2023-01-26
Update Canto fee

JeremyParish69:
Estimated Fee based on what I see in a Metamask tx from Canto.io


sunnya97:
Switched the gas to use the amounts suggested to Keplr by Canto frontend

CleanShot 2023-01-26 at 16 51 41@2x

sunnya97 (approved):

PR #1268, opened 2023-01-27
Show 3 token icons

jonator:
With addition of stableswap, 3 pools will be more important. Let's show all 3 token icons instead of 1 icon and "+ 2". After 3 tokens, we show "+x".

### DIFF
--- a/packages/web/components/assets/pool-assets-icon.tsx
+++ b/packages/web/components/assets/pool-assets-icon.tsx
@@ -57,6 +57,7 @@ export const PoolAssetsIcon: FunctionComponent<Props> = ({
                 height={size === "md" ? 50 : 40}
               />
             )}
+            <span className="absolute h-full w-full" title={coinDenom} />
           </div>
         )
       )}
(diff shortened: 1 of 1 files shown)


packages/web/components/assets/pool-assets-icon.tsx
```diff
@@ -14,61 +14,53 @@ export const PoolAssetsIcon: FunctionComponent<Props> = ({
...
+                height={size === "md" ? 50 : 40}
+              />
+            ) : (
+              <Image
+                src="/icons/question-mark.svg"
+                alt="no token icon"
+                width={size === "md" ? 50 : 40}
+                height={size === "md" ? 50 : 40}
+              />
+            )}
```
JoseRFelix, line 59: We could use the `denom-image` component here. Also, we could modify to fallback to use the question mark

packages/web/components/assets/pool-assets-icon.tsx
```diff
@@ -14,61 +14,53 @@ export const PoolAssetsIcon: FunctionComponent<Props> = ({
...
-      </div>
+    <div
+      style={{ width: `${assets.slice(undefined, 4).length * 30}px` }}
+      className={classNames(
+        "relative flex h-fit items-center",
+        size === "md" ? "h-[50px]" : "h-[40px]"
+      )}
+    >
+      {assets.map(({ coinDenom, coinImageUrl }, index) =>
+        index > 3 ? null : (
```
JoseRFelix, line 25: We could slice the array instead of checking for a bigger index

```
assets.slice(0, 3)
```

Issue #1260, opened 2023-01-26
Swap Tool: show error state to users


jonator:
### Problem
Currently there are two confusing states the swap tool can get into without any indicator to the user as to what is going on:

* No route for currently selected pair of tokens
* Not enough liquidity for entered "in amount"

### Proposed Solution
Add a message to the disabled buttons explaining why swapping is not possible with given inputs.

PR #1269, opened 2023-01-27
fix: Show Not Enough Pool Liquidity on Huge Swaps

JoseRFelix:
Addresses #1260

### DIFF
--- a/packages/stores/src/ui-config/trade-token-in-config.ts
+++ b/packages/stores/src/ui-config/trade-token-in-config.ts
@@ -23,7 +23,11 @@ import {
   Pool,
   RoutePathWithAmount,
 } from "@osmosis-labs/pools";
-import { NoSendCurrencyError, InsufficientBalanceError } from "./errors";
+import {
+  NoSendCurrencyError,
+  InsufficientBalanceError,
+  NoRouteError,
+} from "./errors";
 
 export class ObservableTradeTokenInConfig extends AmountConfig {
   @observable.ref
@@ -301,6 +305,11 @@ export class ObservableTradeTokenInConfig extends AmountConfig {
       isMultihopOsmoFeeDiscount: false,
     };
 
+    if (paths.length === 0 && this.amount !== "" && this.amount !== "0") {
+      this.setError(new NoRouteError("No route found"));
+      return zero;
+    }
+
     if (paths.length === 0 || this.amount === "" || this.amount === "0") {
       return zero;
     }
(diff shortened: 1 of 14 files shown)


sunnya97:
Task linked: [CU-860pp6kn5 USDC USDT max swap](https://app.clickup.com/t/860pp6kn5)

packages/web/localizations/en.json
```diff
@@ -371,6 +371,8 @@
     "invalidSlippage": "Invalid slippage",
     "noSendCurrency": "Currency to send not set",
     "insufficientBal": "Insufficient balance",
+    "insufficientLiquidity": "Insufficient liquidity",
+    "noRoute": "No Route for this trade",
```
jonator, line 375: Actually can we please not capitalize "Route"

packages/stores/src/ui-config/trade-token-in-config.ts
```diff
@@ -410,6 +437,9 @@ export class ObservableTradeTokenInConfig extends AmountConfig {
     }
 
     if (this.amount) {
+      if (this._notEnoughLiquidity) return new NotEnoughLiquidityError();
```
jonator, line 440: I think the no balance error should take precedence over the liquidity error. I see other DEXs do that, and IMO it's better optics and more accurate.

jonator (changes requested):

jonator:
Ehh I'm not really a fan of the UX of creating branches from ClickUp, the names are illegible. We should focus on linking to issues and secondarily linking to ClickUp (from ClickUp).

packages/stores/src/ui-config/trade-token-in-config.ts
```diff
@@ -410,6 +437,9 @@ export class ObservableTradeTokenInConfig extends AmountConfig {
     }
 
     if (this.amount) {
+      if (this._error instanceof NoRouteError) return this._error;
+      if (this._notEnoughLiquidity) return new NotEnoughLiquidityError();
+
```
jonator, line 442: I think this needs to be moved below the balance code below (below that last if statement).

jonator (changes requested):

packages/stores/src/ui-config/trade-token-in-config.ts
JoseRFelix: If I moved it down there, `Insufficient Balance` would take precedence. Users who like to experiment can get confused when the `out` token stops providing estimates, just like it happened to Jim. I think leaving it like it's right now will benefit both users. What do you think?

jonator: I guess that's true given that we don't calculate past the pool's liquidity as other DEXs do. However, it's not ideal to be advertising that we don't have enough liquidity. Both are true to some extent.

SGTM

jonator (approved):

Issue #1231, opened 2023-01-19
Unable to swap stATOM form/to multiple tokens


arispech:
As seen in the video the swap table shows the approximate amount swap to, when swap from ATOM and OSMO but not from USDC EVMOS STRD stOSMO for example

https://user-images.githubusercontent.com/56182049/213376278-753ce0cb-985c-463b-b983-a5340d412d05.mov

jonator:
When we implement the optimized router, it should allow for this since our router doesn't route through more than 2 pools currently :(

jonator:
This is due to our naive router implementation. Once we get a full router, this should be handled.

Issue #1148, opened 2022-12-16
Match Design Guidelines and Headless Components 


JoseRFelix:
## Problem

As the app grows we must start relying on design guidelines to maintain consistent styling and user experience. Further, components must be customizable, and accessible for screen readers and keyboard users.

## Proposed Solution

We must start creating components that adhere to our design guidelines and maintain accessibility throughout the app. Since we use Tailwind CSS, it makes sense to use headless components like [Headless UI](https://headlessui.com/) and [Radix](https://www.radix-ui.com/) that integrate beautifully and are accessible. Furthermore, following modern practices standardized by libraries like Material UI and Chakra UI, we'll use the library [CVA](https://github.com/joe-bell/cva) to build components with variants to avoid the arduous task of class matching. 

I've identified the following components that must be migrated/integrated:

### Button
- [x] General Component (#1157)
- [x] Icon Button ([inspiration](https://chakra-ui.com/docs/components/icon-button/usage)) (#1199)
### Popover
- [x] Topbar Settings (#1201)
- [x] Transaction Settings (#1201)
### Modal
- [ ] New Pool
- [ ] Deposit 
- [ ] Withdraw
- [ ] Connect Wallet
### Input
- [x] Search input (#1212)
### Select
- [ ] Deposit (Network select)
### Drawer
- [x] General Component (#1176)
- [ ] Token select
- [ ] Bond Card
### Icons
- [ ] Add Accessible [Icon Component](https://www.radix-ui.com/docs/primitives/utilities/accessible-icon)

syed8rahman:
Adjust Swap settings to match Figma UI https://www.figma.com/file/om93lu8iXOhxwGHSUpAXEP/Trade-UI?node-id=93%3A1765&t=3ZhgpLpgcHzq2Pia-1
<img width="278" alt="Screen Shot 2022-12-16 at 11 55 56 AM" src="https://user-images.githubusercontent.com/22246254/208149434-4f043358-6320-4ca8-b5ea-fc04956316f2.png">

jonator:
Closing in favor of tracking on ClickUp

PR #1271, opened 2023-01-27
Publish Stage

### LINKED ISSUE #1142: Profile modal
* Placeholder profile pic
* Osmo name
* Balance
* Wallet

[Design](https://www.figma.com/file/6iFXxwaZIlPm5y7IuEJpiL/Profile?node-id=111%3A5130&t=zCHchz9LxD3kpjd1-4)

### LINKED ISSUE #1255: Typo "1 day" not "1 days"
Screenshot 2023-01-25 at 3 08 07 AM

### LINKED ISSUE #1260: Swap Tool: show error state to users

### LINKED ISSUE #1280: Profile can't scroll on mobile view
iOS - can't scroll on mobile view 
image

Android - all elements fit on one screen but space/padding seem a little off 
image

jonator:
* Settings modal closes #1142 
* Progress on headless #1148 
* Fix: Closes #1255 
* Fix: Closes #1260
* Fix: Closes #1280 
* Fix: transfer status not updating
* Update inlang config (translations frontend)
* Filters completed gauges
* Reorder assets
* ICNS names in transfer modal
* Warning when adding liquidity to pool with scaling factor controller

Test Guide
* Settings Modal
* Mobile Profile Modal
* Swap tool
  * Large amounts
  * Token pairs that don't have routes (see non-routes here: https://github.com/osmosis-labs/osmosis-frontend/issues/1231)
* Make sure no pages fail with changes to gauge query (assets page)
* Transfer both assets via IBC and Axelar. To test failure for Axelar, pay with insufficient fee. Transfers should not remain pending in the table.
  * Notice ICNS names for both osmosis and counterparty chains (cosmos, stars)


jonator:
> Large amounts have not been fixed and the button shows no route for this trade before showing an insufficient balance
@JoseRFelix and I agreed we should show that there's no route before showing insufficient balance.

> Token pairs that don't have routes (see non-routes here: Unable to swap stATOM form/to multiple tokens https://github.com/osmosis-labs/osmosis-frontend/issues/1231) have not been fixed and the button shows no route for this trade before show swap
This will be fixed when we implement a fully optimized router. Not in this set of changes.

@arispech

jonator:
Hmm it appears that stargaze particularly doesn't have websocet enabled on their endpoint. I tried our Akash node and it worked. I think it's not a bug on our frontend but a misconfiguration of the stargaze node we use. However, I do think we could handle it better and reflect this on our frontend. Perhaps if their websocket connection is not accepted we show "Unknown" or "In transit" instead of "Pending" forever?

arispech:
> Hmm it appears that stargaze particularly doesn't have websocet enabled on their endpoint. I tried our Akash node and it worked. I think it's not a bug on our frontend but a misconfiguration of the stargaze node we use. However, I do think we could handle it better and reflect this on our frontend. Perhaps if their websocket connection is not accepted we show "Unknown" or "In transit" instead of "Pending" forever?

I think showing something that says automatic confirmation unavailable for network X is better because it might push that network to support what is needed to work correctly on Osmosis but your suggestion is a good idea too.

arispech (approved):

PR #1276, opened 2023-01-27
Fix errors in docker CI

niccoloraspa:
This PR fixes the errors related to the docker image build:

- It downgrades node to use `16.19.0` due to the following error when building:

```bash
#12 2.488 [2/4] Fetching packages...
#12 100.0 error @axelar-network/axelar-cgp-solidity@4.3.1: The engine "node" is incompatible with this module. Expected version ">=16 <17". Got "18.12.1"
#12 100.1 error Found incompatible module.
#12 100.1 info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

- Fixes problems in the workflow when using the date in the image tag


jonator (approved):

daniel-farina (approved):
Lgtm

PR #1279, opened 2023-01-29
fix: Bridge Transfer Status UI Instability

JoseRFelix:
Axelar changed its API schema: https://docs.axelarscan.io/cross-chain/transfers-status.

Notable changes: 

- `data.source` No longer exists. It's instead called `data.send`
- We must use `data.id`  instead. 
- Id includes the source chain. So to get the id, we have to: `data.id.split("_")[0]`
- insufficient_fee Is now located in `data.send`

### DIFF
--- a/packages/web/integrations/axelar/transfer-status-source.ts
+++ b/packages/web/integrations/axelar/transfer-status-source.ts
@@ -66,31 +66,32 @@ export class AxelarTransferStatusSource implements ITxStatusSource {
     }
 
     const [data] = transferStatus;
+    const idWithoutSourceChain = data?.id.split("_")[0].toLowerCase();
 
     // insufficient fee
-    if (data.source && data.source.insufficient_fee) {
+    if (data.send && data.send.insufficient_fee) {
       return {
-        id: data.source.id.toLowerCase(),
+        id: idWithoutSourceChain,
         status: "failed",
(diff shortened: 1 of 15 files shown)


jonator:
Hmm actually I may have noticed that these statuses are not translated

packages/web/integrations/axelar/transfer-status-source.ts
```diff
@@ -66,31 +66,32 @@ export class AxelarTransferStatusSource implements ITxStatusSource {
...
+    const idWithoutSourceChain = data?.id.split("_")[0].toLowerCase();
 
     // insufficient fee
-    if (data.source && data.source.insufficient_fee) {
+    if (data.send && data.send.insufficient_fee) {
       return {
-        id: data.source.id.toLowerCase(),
+        id: idWithoutSourceChain,
         status: "failed",
         reason: "Insufficient fee",
```
jonator, line 76: RE i18n: This is where the failure reason key should be passed. I'm thinking we can create a type with const strings for all the possible failure reasons, then they get mapped to a localization key (and translation) in the view layer. Similar to what we do for `TxStatus`.

JoseRFelix: Nice catch! Added in latest commit

jonator (approved):

PR #1286, opened 2023-01-31
fix: Can't Scroll Profile Modal

JoseRFelix:


jonator (changes requested):
It appears the modal overflows on mobile. iPhone 13 Pro.
IMG_81873FE5A0AE-1

Also appears to overflow on my mobile chrome tools:
Screen Shot 2023-01-31 at 3 37 09 PM

jonator (approved):

PR #1287, opened 2023-01-31
Sunny/unwrapped axelar

sunnya97:


JeremyParish69:
@sunnya97 Do we want this changed for all the wrapped Assets (not just BNB)?
Not sure how tricky it might be to update in all languages....

sunnya97:
Yeah I was just testing things out.  Was a bit more complicated than expected.  Will probably just punt this and wait for proper unwrapped support

PR #1302, opened 2023-02-05
add Brazilian Portuguese translation

virtugroweey:
I'm native brazilian portuguese speaker. I tried adapt the translation to something that can be approximately understandable. I think is good atm.


jonator:
We're about to include a lot of changes from stage branch

virtugroweey:
> We're about to include a lot of changes from stage branch

Oh, ok. i will update the translation when the release the next update.

jonator (approved):

PR #1307, opened 2023-02-06
Re-enable IXO

JeremyParish69:
Re-enable IXO. now that the ibc connection is restored.
Updatechain id from impacthub-3 to ixo-4
https://www.mintscan.io/osmosis/proposals/412
Validated that transfers work.


sunnya97 (approved):

sunnya97:
utAck

PR #1315, opened 2023-02-08
add autonomy stop limits

omnitired:


jonator:
I think @syed8rahman agreed that we should have a separate frontend for these new order types for now until we can reassess our current UX.

PR #1317, opened 2023-02-09
Fallback to node pools query if main query store fails

jonator:
This is an automated approach to switching to the slower pools query when the optimized query object fails to fetch. This should allow our frontend to still be usable without imperator (or other) data services. This only recently became a concern once a core and important query, the pools query, was moved to data services instead of directly from the node due to performance issues.

### DIFF
--- a/packages/stores/src/queries/fallback-query-store.ts
+++ b/packages/stores/src/queries/fallback-query-store.ts
@@ -1,0 +1,34 @@
+import { observable, autorun, makeObservable, runInAction } from "mobx";
+import { ObservableQuery } from "@keplr-wallet/stores";
+
+/** Wraps an ordered list of similar query stores and falls back to the first store that returns a valid response. */
+export class FallbackStore<TStore extends ObservableQuery> {
+  @observable.ref
+  protected _responsiveStore: TStore;
+
+  constructor(readonly orderedStores: TStore[]) {
+    this._responsiveStore = orderedStores[0];
+
+    // Runs any time error changes in current store.
+    autorun(() => {
+      if (this._responsiveStore.error) {
+        // If the current store has an error, fallback to the next store.
+        const index = orderedStores.indexOf(this._responsiveStore);
+        if (index < orderedStores.length - 1) {
+          runInAction(() => (this._responsiveStore = orderedStores[index + 1]));
+        } else {
+          console.warn(
+            "FallbackStore ran out of fallback stores. Last store error:",
+            this._responsiveStore.error
+          );
+        }
+      }
+    });
+
+    makeObservable(this);
+  }
+
+  get responsiveStore(): TStore {
+    return this._responsiveStore;
+  }
+}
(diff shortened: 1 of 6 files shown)


packages/stores/src/queries/fallback-query-store.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (this._responsiveStore.error) {
+        // If the current store has an error, fallback to the next store.
+        const index = orderedStores.indexOf(this._responsiveStore);
+        if (index < orderedStores.length - 1) {
+          runInAction(() => (this._responsiveStore = orderedStores[index + 1]));
+        } else {
+          console.warn(
+            "FallbackStore ran out of fallback stores. Last store error:",
+            this._responsiveStore.error
+          );
```
JoseRFelix, line 23: Maybe we should restart the loop by setting the responsive store to index 0, so we don't end up with a dead end.

jonator: Wouldn't that cause an infinite loop? Or do we add a flag to remember that we arrived at the end, and are defaulting to the first?

JoseRFelix: I assume the query will keep retrying between all the stores until one is successful. Do you think that would crash the app?

jonator: Eh I think it would possibly make the page unresponsive. I think we can't assume the prev failed services will come back online as the user is sitting with the window open. We should assume the last fallback store won't fail, that should be the edge case.

packages/stores/src/queries/pools/pools.ts
```diff
@@ -98,4 +99,13 @@ export class ObservableQueryPools
...
   });
+
+  /** TODO: implement pagination when we hit the limit of pools, for now, the url will be set to the max number of pools in the autorun above */
+  paginate() {
+    /** do nothing */
+  }
+
+  fetchRemainingPools() {
+    /** do nothing since all pools get fetched. */
+  }
```
JoseRFelix, line 110: We need to handle this edge case when we have the redesigns pools page @sehyunc. cc\ @jonator

jonator: Actually I'm thinking the paginate method should be removed across the board since it's not actually used. I added it prematurely to be used in the future, but the way the filtered query works is it fetches a subset of pools, then all of the pools. Pagination is not used or tested. I can remove it.

JoseRFelix: If that's the case let's remove it

jonator: https://github.com/osmosis-labs/osmosis-frontend/blob/8547c2e0ae5c72d563f13f16095f35c13a48f8e0/packages/web/components/complex/all-pools-table-set.tsx#L525

Oh wait I checked he did use it. I'll keep it around then. For the fallback store, paginate is a noop since all pools should be loaded immediately.

sehyunc: The latest version of the branch doesn't have pagination implemented, and performance seems to be very good because of virtualization. I think we could use the query the way @jonator described it and remove pagination for less complexity.

JoseRFelix (approved):
LGTM

PR #1318, opened 2023-02-09
Re-enable Cerberus IBC

JeremyParish69:
Apparently working again, but also apparently shutting down soon.
https://www.mintscan.io/cerberus/proposals/31

Tried all working nodes, but they all either have CORS blocking or don't have wss enabled.


JeremyParish69:
Chain is dead.

PR #1319, opened 2023-02-10
Add token name

JeremyParish69:
So far this only adds the name property to the type and name values to the assets. It does not show the name under the asset in the Assets page.


jonator:
Can we actually put this on hold until we work on the token info page? Unless it's super high prio?

I think we're going to add a lot more info per token soon.

sunnya97:
@jonator why do we need to block this though on the token info page, instead of just doing this first?

PR #1321, opened 2023-02-10
add stEVMOS

riley-stride:
Waiting until pool is created to add Pool ID to packages/web/config/price.ts


packages/web/config/price.ts
```diff
@@ -874,6 +874,19 @@ export const PoolPriceRoutes: IntermediateRoute[] = [
     ),
     destCoinId: "pool:uluna",
   },
+  {
+    alternativeCoinId: "pool:staevmos",
+    poolId: "", // <= fill this out once pool is live
+    spotPriceSourceDenom: DenomHelper.ibcDenom(
+      [{ portId: "transfer", channelId: "channel-326" }],
+      "stuluna"
```
JeremyParish69, line 882: when you update the pool id, don't forget to correct this denom

riley-stride: Nice catch @JeremyParish69 . Will do, thank you!

riley-stride:
@JeremyParish69 I realized we'll need the vercel build to create the pool, so I'm commenting out the price.ts changes for now until we have a vercel build and pool id

JeremyParish69 (approved):

PR #1326, opened 2023-02-12
delist crbrus

doggystylez:
cerberus-chain-1 [shutting down](https://www.mintscan.io/cerberus/proposals/31)


doggystylez:
supercedes https://github.com/osmosis-labs/osmosis-frontend/pull/1318

JeremyParish69:
We will keep dead tokens listed (but only on Frontier with IBC transfers disabled).

PR #1341, opened 2023-02-15
Add NRIDE Token

JeremyParish69:
Adds NRIDE token, with pool 924 for its spot price


arrivets:
Hi @JeremyParish69, I create a PR in cosmos chain-registry to update the images:
https://github.com/cosmos/chain-registry/pull/1428
Would it be possible to use those images here too?
Thanks

jonator (approved):

PR #1343, opened 2023-02-15
Hotfix: small visual issues + remove Mars promo > feedback banner

jonator:
* See [prod pool 888](https://frontier.osmosis.zone/pool/888), pool composition:
  * Fix display of small amounts (<0 whole number, or rounded up) in pool composition (BTC)
  * Show more characters of token denom on desktop (prevent ellipsis for WBTC.grv
* Remove MARS bg and banner promo
  * Replace banner with ask for feedback to https://tally.so/r/wQ1KR7 (@syed8rahman )


packages/web/config/feature-flag.ts
```diff
@@ -56,17 +56,19 @@ export const Announcement:
       },
     }
   : {
-      localStorageKey: "show_mars_banner",
-      enTextOrLocalizationPath: "mars.banner.title",
+      localStorageKey: "feedback_wQ1KR7",
+      enTextOrLocalizationPath: "Help us shape the future of Osmosis.",
```
JoseRFelix, line 60: We need to add localization to this

JoseRFelix (changes requested):
Looks good! We only need to add localization

sunnya97 (approved):

packages/web/config/feature-flag.ts
jonator: It's only going to be up for about a week. Since the form is in english, syed and I agreed it's fine to just use english for now.

JoseRFelix (approved):

Issue #1274, opened 2023-01-27
Swap: different quote and keplr min out amounts


jonator:
### Problem
Seems there's a difference on the min out amounts on the swap tool, and those that Keplr receives. One user (linked above) was particularly disadvantaged because of this. It seems the imperator query only updates from the node periodically (some number if minute intervals), causing the details drawer to become quite a bit different from the instantly-refreshed (directly from the node) sign pop up . We rerun the pools query every 30 seconds from the frontend while the user has the window open, but it was intended for the node pools query which was always up to date. Now, the query could run 30 seconds later and still receive the same data.

### Proposed Solution
* Make imperator query update more often (seconds)
and/or
* Iterate and query the pools in the route the user is currently looking at, with a debounce function to prevent over-querying

jonator:
Will reopen if encountered

PR #1347, opened 2023-02-20
Swap tool fixes

### LINKED ISSUE #1346: EVMOS decimals on the swap page break UI
image

jonator:
* Re-query pools in route when details are expanded (helps address #1274)
* Limit available balance to 8 decimals. Closes #1346 
* Only show swap route vis in home page, not in pool detail page

TODO
- [x] Max 8 decimals on token select

### DIFF
--- a/packages/web/components/trade-clipboard/index.tsx
+++ b/packages/web/components/trade-clipboard/index.tsx
@@ -121,7 +121,7 @@ export const TradeClipboard: FunctionComponent<{
     const setShowEstimateDetails = useCallback(
       (value: boolean) => {
         // refresh current route's pools
-        if (value === true) {
+        if (value) {
           tradeTokenInConfig.optimizedRoutePaths.forEach((route) => {
             route.pools.forEach((pool) => {
               console.log("refresh pool", pool.id);
--- a/packages/web/components/trade-clipboard/index.tsx
+++ b/packages/web/components/trade-clipboard/index.tsx
@@ -1100,20 +1100,21 @@ export const TradeClipboard: FunctionComponent<{
                   </span>
                 </div>
               </div>
-              {tradeTokenInConfig.optimizedRoutePaths
-                .slice(0, 1)
-                .map((route, index) => (
-                  <TradeRoute
-                    key={index}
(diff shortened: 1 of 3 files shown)


packages/web/components/trade-clipboard/index.tsx
```diff
@@ -114,10 +114,28 @@ export const TradeClipboard: FunctionComponent<{
...
     // show details
-    const [showEstimateDetails, setShowEstimateDetails] = useState(false);
+    const [showEstimateDetails, _setShowEstimateDetails] = useState(false);
     const isEstimateDetailRelevant = !(
       tradeTokenInConfig.amount === "" || tradeTokenInConfig.amount === "0"
     );
+    const setShowEstimateDetails = useCallback(
+      (value: boolean) => {
+        // refresh current route's pools
+        if (value === true) {
```
JoseRFelix, line 124:
```suggestion
        if (value) {
```

packages/web/components/trade-clipboard/index.tsx
```diff
@@ -146,7 +164,7 @@ export const TradeClipboard: FunctionComponent<{
         fetchedRemainingPoolsRef.current = true;
         queries.osmosis?.queryGammPools.fetchRemainingPools();
       }
-    }, []);
+    }, [queries.osmosis?.queryGammPools]);
```
JoseRFelix, line 167: @jonator What do you think of enabling `react-hooks/exhaustive-deps` eslint rule? If we want to opt out, we should add a eslint-disable comment. Exhaustive deps are important to avoid unexpected bugs coming from `useEffect`.

JoseRFelix (changes requested):
Looks good, left some minor comments

packages/web/components/trade-clipboard/index.tsx
jonator: https://github.com/osmosis-labs/osmosis-frontend/issues/1349

I agree, @sehyunc pointed this out and will include it in his pools page redesign PR.

JoseRFelix (approved):

PR #1368, opened 2023-03-02
(CL) WIP - Add CL pool simulation

jonator:


packages/math/src/pool/concentrated/one-for-zero.ts
```diff
@@ -0,0 +1,111 @@
...
+export type OneForZero = {
+  isOutGivenIn: boolean;
+  sqrtPriceLimit: Dec;
+  swapFee: Dec;
+};
+
+export class OneForZeroStrategy implements SwapStrategy {
+  constructor(private readonly oneForZero: OneForZero) {}
+
+  computeSwapStep(
```
p0mvn, line 22: Note: this function is going to be updated chain-side after refactor. We can work together next week on the updates

packages/math/src/pool/concentrated/one-for-zero.ts
```diff
@@ -0,0 +1,111 @@
...
+      this.oneForZero.swapFee,
+      this.oneForZero.isOutGivenIn
+    );
+
+    let nextSqrtPrice: Dec;
+    let hasReachedTarget = false;
+
+    if (amountRemainingLessFee.gte(amountOne)) {
+      nextSqrtPrice = sqrtPriceTarget;
+      hasReachedTarget = true;
```
p0mvn, line 48: I don't think we should be setting this to true here. While unlikely, it is possible that on the following line:
```
 nextSqrtPrice = getNextSqrtPriceFromAmount1RoundDown(
        liquidity,
        curSqrtPrice,
        amountRemainingLessFee
      );
``` 

`nextSqrtPrice` will reach target. Therefore, I suggest doing a comparison check similar to: https://github.com/osmosis-labs/osmosis/blob/6014a7c6f7f330869469f6ffc15f62cff3b854c1/x/concentrated-liquidity/internal/swapstrategy/one_for_zero.go#L65

packages/math/src/pool/concentrated/zero-for-one.ts
```diff
@@ -0,0 +1,116 @@
...
+      this.zeroForOne.swapFee,
+      this.zeroForOne.isOutGivenIn
+    );
+
+    let nextSqrtPrice: Dec;
+    let hasReachedTarget = false;
+
+    if (amountRemainingLessFee.gte(amountZero)) {
+      nextSqrtPrice = sqrtPriceTarget;
+      hasReachedTarget = true;
```
p0mvn, line 60: Similar comment about reaching target

packages/math/src/pool/concentrated/one-for-zero.ts
```diff
@@ -0,0 +1,111 @@
...
+        hasReachedTarget,
+        amountOne,
+        amountRemaining,
+        this.oneForZero.swapFee
+      );
+    } else {
+      feeChargeTotal = amountZero
+        .mul(this.oneForZero.swapFee)
+        .quo(new Dec(1).sub(this.oneForZero.swapFee));
+    }
```
p0mvn, line 80: Note: working on the refactor of this component chain side. Will be ready next week.

packages/math/src/pool/concentrated/quotes.ts
```diff
@@ -0,0 +1,112 @@
...
+
+import { maxSpotPrice, minSpotPrice, smallestDec } from "./const";
+import { TickOverflowError } from "./errors";
+import { makeSwapStrategy } from "./swap-strategy";
+import { sqrtPriceToTick, tickToSqrtPrice } from "./tick";
+import { addLiquidity, approxSqrt } from "./utils";
+
+export const ConcentratedLiquidityMath = {
+  calcOutGivenIn,
+  // calcInGivenOut,
```
p0mvn, line 11: Please don't start on this one right now. I think it will be much easier to port once chain-side refactor is complete

packages/math/src/pool/concentrated/swap-strategy.ts
```diff
@@ -0,0 +1,34 @@
...
+    feeChargeTotal: Dec;
+  };
+  initTickValue(curTick: Int): Int;
+  validatePriceLimit(sqrtPriceLimit: Dec, curSqrtPrice: Dec): boolean;
+  setLiquidityDeltaSign(liquidityDelta: Dec): Dec;
+}
+
+export function makeSwapStrategy(
+  izZeroForOne: boolean,
+  isOutGivenIn: boolean,
```
p0mvn, line 26: Note: post-refactor there will be no need for this field

packages/math/src/pool/concentrated/utils.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+  const diff = sqrtPriceB.sub(sqrtPriceA);
+
+  if (roundUp) {
+    return liquidity.mul(diff).roundUpDec();
+  }
+  return liquidity.mul(diff);
+}
+
+export function getNextSqrtPriceFromAmount0RoundUp(
```
p0mvn, line 40: Note: to support swapping in given out, there will be more helpers for computing next sqrt price. PRin progress: https://github.com/osmosis-labs/osmosis/pull/4477

packages/math/src/pool/concentrated/one-for-zero.ts
jonator: I changed this since it seemed to be a reference equality check? Not a value equality check?

packages/math/src/pool/concentrated/quotes.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  let sqrtPriceStart: Dec;
+  let i = curTickIndex;
+  while (
+    swapState.amountRemaining.gt(smallestDec) &&
+    !swapState.sqrtPrice.equals(sqrtPriceLimit)
+  ) {
+    sqrtPriceStart = swapState.sqrtPrice;
+
+    const nextTick: Int | undefined = tickDepths?.[i];
```
jonator, line 71: @p0mvn thoughts on these `Int` depth values for ticks? Obv since we don't deal with stores directly on frontend, we just use an array of depths (originating from this query https://app.clickup.com/t/860pq0mn3). Does this check out? Then, later I throw an exception (`TickOverflowError`) if we ran out of ticks, indicating we need to query again for more ticks.

p0mvn: Discussing this on Slack but, basically, considering changing this from "tick depth" semantics to "liquidity net" model that is aligend with swap computations chain side.

The query is still tracked by that issue and the suggestion to utilize "liquidity net' does not affect the progress we've already made on it. 

The use of arrays makes sense to me.

Let's just get an agreement on "tick depth" vs "liquidity net"

packages/math/src/pool/concentrated/quotes.ts
```diff
@@ -0,0 +1,215 @@
...
+    priceLimit = maxSpotPrice;
+  }
+
+  const sqrtPriceLimit = approxSqrt(priceLimit);
+  const swapStrategy = makeSwapStrategy(isZeroForOne, sqrtPriceLimit, swapFee);
+  const tokenInAmountSpecified = new Dec(tokenIn.amount);
+
+  const curTickWithNetLiq = inittedTicksWithNetLiquidity.find(({ tickIndex }) =>
+    tickIndex.equals(curTick)
+  );
```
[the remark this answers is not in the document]
jonator: So regardless we're iterating over indexes in positive direction? In other words, the net depths query alwasy returns current price tick as 0 index?

packages/math/src/pool/concentrated/quotes.ts
```diff
@@ -0,0 +1,215 @@
...
+    );
+    swapState.amountCalculated = swapState.amountCalculated.add(amountZeroOut);
+
+    if (nextTickSqrtPrice.equals(sqrtPriceNext)) {
+      const liquidityNet = swapStrategy.setLiquidityDeltaSign(
+        new Dec(nextTick.netLiquidity.toString())
+      );
+
+      swapState.liquidity = addLiquidity(swapState.liquidity, liquidityNet);
+    }
```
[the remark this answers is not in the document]
jonator: Is that initting a new tick, or creating a new tickIndex? Do we insert some value into the `inittedTicksWithNetLiquidity` array? Not sure how to handle the update to `inittedTickIndex`, which is handled differently on the FE since we can't query or init ticks on our end. May still be missing knowledge of how we handle stopping an iteration mid-tick.

Perhaps we take our PriceToTick computed value and find() it in `inittedTicksWithNetLiquidity` array?

packages/math/src/pool/concentrated/tick.ts
```diff
@@ -0,0 +1,185 @@
...
+  );
+  if (tickIndex.lt(minTick) || tickIndex.gt(maxTick)) {
+    throw new Error(
+      `tickIndex is out of range: ${tickIndex.toString()}, min: ${minTick.toString()}, max: ${maxTick.toString()}`
+    );
+  }
+
+  const geometricExponentDelta = new Dec(tickIndex).quoTruncate(
+    geometricExponentIncrementDistanceInTicks.truncateDec()
+  );
```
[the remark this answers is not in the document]
jonator: I believe it achieves the same thing, but maintains the `Dec` type.

See:
[`truncateDec()`](https://github.com/chainapsis/keplr-wallet/blob/7a4db487b74170d968ecc2c7979d39579251277c/packages/unit/src/decimal.ts#L369)

Vs;

[`truncate()`](https://github.com/chainapsis/keplr-wallet/blob/7a4db487b74170d968ecc2c7979d39579251277c/packages/unit/src/decimal.ts#L357)

packages/math/src/pool/concentrated/quotes.ts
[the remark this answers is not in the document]
p0mvn: Yeah we can either do it at index 0 or as a separate return from the tick query:

My latest suggestion was to have it as a separate return but we can change that if you prefer index 0 approach:
https://github.com/osmosis-labs/osmosis/pull/4714#discussion_r1147764952

packages/math/src/pool/concentrated/tick.ts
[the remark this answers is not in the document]
p0mvn: Makes sense but what about the final result?

`quoTruncate` only truncates at precision end. On the other hand, `TruncateInt` truncates the full fractional part

packages/math/src/pool/concentrated/quotes.ts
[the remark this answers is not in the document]
p0mvn: Upon thinking more about how FE differs from chain, I suspect we might not need to port it here. But let me try breaking down the logic.

That branch is executed when a swap does not reach the next initialized tick.

As a result, we stop mid-way between one initialized tick and the next initialized tick.

At that point, we only have the knowledge of the cur sqrt price. Therefore, we need to compute the cur tick from the sqrt price. The tick itself is not initialized.

Once we have the mid-non-initialized tick, we compute the next initialized tick at the next iteration of the loop:
https://github.com/osmosis-labs/osmosis/blob/d61c0eda66f7c94110307e2bca97a461ee809afd/x/concentrated-liquidity/swaps.go#L339

On-chain, we compute the next tick by reading it from the state. Here, I think we should be able to just move forward to the next element in the list returned by the chain query.

p0mvn, line 115: I suspect that the solution to the other two threads might be just moving the call to `swapStrategy.nextInitializedTickIndex` that is discussed in [this thread](https://github.com/osmosis-labs/osmosis-frontend/pull/1368/files/5863e51afdb574c1d231c4c1bdf5c149e6775207#r1147747087) to be inside this branch because this is the time when we cross the tick.

When we have `!nextTickSqrtPrice.equals(sqrtPriceNext)` (notice nehation), the next initialized tick does not change even though the factual (uninitialized) tick might differ from the previous iteration

p0mvn (commented):
Replied to all threads, please ping me once ready for re-review

packages/math/src/pool/concentrated/quotes.ts
jonator: Done for both swap types

packages/math/src/pool/concentrated/quotes.ts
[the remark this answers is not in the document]
p0mvn: I replied in that thread with further thoughts. @jonator could you please check the linked comment again? Do you agree with returning current tick liquidity separately?

packages/math/src/pool/concentrated/quotes.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    swapState.inittedTickIndex++;
```
p0mvn, line 117: This should be inside and at the end of this branch:
```
    if (nextTickSqrtPrice.equals(sqrtPriceNext)) {
      const liquidityNet = swapStrategy.setLiquidityDeltaSign(
        new Dec(nextTick.netLiquidity.toString())
      );

      swapState.liquidity = addLiquidity(swapState.liquidity, liquidityNet);
    }
```

The reason is that when `!nextTickSqrtPrice.equals(sqrtPriceNext)`, we are not reaching the next initialized tick yet so its value should stay the same

packages/math/src/pool/concentrated/quotes.ts
p0mvn: Had a follow-up : https://github.com/osmosis-labs/osmosis-frontend/pull/1368/files#r1148474486

but going to resolve this thread to continue discussions in the new one only

packages/math/src/pool/concentrated/quotes.ts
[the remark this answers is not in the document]
jonator: LGTM. Also please see slack DM re net liquidity

packages/math/src/pool/concentrated/tick.ts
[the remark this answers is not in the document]
jonator: Converted math to Ints to match chain code. LMK what you think. Also I intend to add test cases so I'll make sure it works to spec.

packages/math/src/pool/concentrated/quotes.ts
p0mvn: I might have misinterpreted it as being outside. On another look now, the location looks good to me. Apologies for any confusion

packages/math/src/pool/concentrated/quotes.ts
```diff
@@ -0,0 +1,210 @@
...
+  if (curTickArrIndex < 0) {
+    throw new Error("curTickWithNetLiq not found in inittedTicks");
+  }
+
+  const swapState: SwapState = {
+    amountRemaining: tokenInAmountSpecified, // tokenIn
+    amountCalculated: new Dec(0), // tokenOut
+    inittedTickIndex: swapStrategy.initTickValue(curTickArrIndex),
+    sqrtPrice: curSqrtPrice,
+    liquidity: poolLiquidity,
```
p0mvn, line 67: Suggest calling this `current_tick_liquidity`

packages/math/src/pool/concentrated/quotes.ts
```diff
@@ -0,0 +1,210 @@
...
+  const curTickWithNetLiq = inittedTicks.find(({ tickIndex }) =>
+    tickIndex.equals(curTick)
+  );
+  if (!curTickWithNetLiq) {
+    throw new Error("curTickNet not found in inittedTicks");
+  }
+  const curTickArrIndex = inittedTicks.indexOf(curTickWithNetLiq);
+  if (curTickArrIndex < 0) {
+    throw new Error("curTickWithNetLiq not found in inittedTicks");
+  }
```
p0mvn, line 60: I'm suggesting returning cur tick index separately from the array of next ticks with their liquidity net amounts.

Please see this for context: https://github.com/osmosis-labs/osmosis/pull/4714/files#r1149769486

p0mvn (commented):
out given in LGTM overall. 

I have some suggestions around naming and returns of chain side query which affect how we initialize ticks here

Haven't reviewed in given out yet.

packages/math/src/pool/concentrated/quotes.ts
jonator: I removed use of `curTick`

Does that mean we should still init the index with `initTickValue`? Or should the next tick always be 0 regardless of strategy?

packages/pools/src/router/routes.ts
```diff
@@ -0,0 +1,366 @@
...
+        )
+      ) {
+        isMultihopOsmoFeeDiscount = true;
+        const { maxSwapFee, swapFeeSum } = getOsmoRoutedMultihopTotalSwapFee(
+          route.pools
+        );
+        poolSwapFee = maxSwapFee.mul(poolSwapFee.quo(swapFeeSum));
+      }
+
+      // less fee
```
JoseRFelix, line 189: I think we should expand on this comment a little bit more

packages/stores/src/queries/concentrated-liquidity/liquidity-net-in-direction.ts
```diff
@@ -0,0 +1,133 @@
...
+    );
+  }
+
+  @computed
+  get currentTick() {
+    return new Int(this.response?.data.current_tick ?? 0);
+  }
+
+  @computed
+  get currentLiquidity() {
```
JoseRFelix, line 44: Same as above

packages/math/src/pool/concentrated/math.ts
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * This function approximates the square root of a given decimal.
+ * It uses Newton's method to approximate the square root.
+ * It does this by iterating through the formula:
+ * x_{n+1} = x_n - (x_n^2 - a) / (2 * x_n)
+ * where x_0 is the initial approximation, a is the number whose
+ * square root we want to find, and x_n is the current approximation.
+ * The number of iterations is controlled by maxIters.
+ *
+ * TODO: move to decimal object see: https://github.com/chainapsis/keplr-wallet/pull/674
```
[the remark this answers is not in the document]
jonator: I tried just updating the decimal package, but I ran into problems with not having the other keplr packages also get updated..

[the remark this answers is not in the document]
JoseRFelix: That's unfortunate. I think it would be better to leave this function in our code base for now

packages/stores/src/queries/concentrated-liquidity/liquidity-net-in-direction.ts
```diff
@@ -0,0 +1,133 @@
...
+          netLiquidity: new Dec(depth.liquidity_net),
+        };
+      }) ?? []
+    );
+  }
+
+  @computed
+  get currentTick() {
+    return new Int(this.response?.data.current_tick ?? 0);
+  }
```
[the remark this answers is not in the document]
jonator: No opposite direction. It will return this same reference until any observable in `this.response.data.current_tick` changes.

References are a problem when passed to memoization primitives such as useMemo/use* deps and computedFn params.

JoseRFelix:
@jonator It seems build is failing

packages/pools/src/router/__tests__/route.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+      };
+      const lowLiqWeight = await calculateWeightForRoute(
+        routeLowLiq,
+        makeTvlFunc(lowLiqPools)
+      );
+      const highLiqWeight = await calculateWeightForRoute(
+        routeHighLiq,
+        makeTvlFunc(highLiqPools)
+      );
+      expect(lowLiqWeight.lt(highLiqWeight)).toBeTruthy();
```
chikeichan, line 45: this test is failing locally - should it be `gt` instead since lower weight is better?

chikeichan, line 45: this fails locally - should this return lower weight for route with higher liquidity instead since lower weight is better?

packages/pools/src/router/__tests__/route.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+      };
+      const lowLiqWeight = await calculateWeightForRoute(
+        routeLowLiq,
+        makeTvlFunc(lowLiqPools)
+      );
+      const lowDepthWeight = await calculateWeightForRoute(
+        lowDepth,
+        makeTvlFunc(lowDepthPools)
+      );
+      expect(lowLiqWeight.lt(lowDepthWeight)).toBeTruthy();
```
chikeichan, line 68: same here - should this be `gt` instead?

packages/pools/src/router/__tests__/routes.spec.ts
```diff
@@ -0,0 +1,412 @@
...
+      const routes = await router.getOptimizedRoutesByTokenIn(
+        {
+          denom: "uion",
+          amount: new Int("100"),
+        },
+        "uosmo"
+      );
+
+      expect(routes.length).toBe(1);
+      expect(routes[0].pools[0].id).toBe("1");
```
chikeichan, line 37: this fails locally and picking pool `2` instead of `1`. perhaps a bug in `getOptimizedRoutesByTokenIn` when sorting routes by weight?

jonator:
@JoseRFelix @chikeichan build should pass now, including tests

jonator:
Merging. Will make smaller PRs going forward to aid in complete review

PR #1364, opened 2023-02-28
Publish Stage

jonator:
* Closes #920 
* Swap Tool: Limit available balance to 8 decimals. Closes https://github.com/osmosis-labs/osmosis-frontend/issues/1346
* LayerSwap: use Coinbase to deposit USDC instead of Kado 
* **New Pools Page!** includes infinite scroll. Closes #1349 #1248 #1175 #1057
* Closes #1388 Closes #1393
* Closes #1390
 

### Testing
* Axelar deposit/withdraw ETH, GLMR, AVAX, MATIC, FTM, BNB
  * Depositing: since these are gas tokens being transferred, make sure the gas cost is calculated and properly deducted from the transfer amount when clicking max. Make sure there's enough gas to click confirm, to prevent metamask from blocking us since it claims we don't have enough gas token to pay gas fees.
* Axelar deposit withdraw non-wrapped asset
* Connecting/disconnecting Metamask
* Ensure links to satellite work properly; for deposit/withdraw, and the proper assets/networks are present on satellite
* Only show swap route vis in home page, not in pool detail page
* Deposit USDC from Coinbase via LayerSwap
* statom deeplink

PR #1371, opened 2023-03-02
Remove w for auto-wrapped assets

JeremyParish69:
bnb, eth, glmr, avax, matic, ftm


JeremyParish69:
> Hey I think this is already done as a part of the native integration currently in stage: #1364

Yes, you are correct. My mistake

PR #1374, opened 2023-03-06
Conliq: Add Liquidity Modal

chikeichan:
TODO:

- [ ] add translations
- [ ] fix linters errors
- [ ] add loading state when fetching historical price data
- [ ] refactor to use `AddLiquidityConfig` observable store instead of `useState`
- [ ] integrate with add liquidity on localosmosis

Screen Shot 2023-03-02 at 1 14 21 PM


packages/web/modals/add-liquidity.tsx
```diff
@@ -50,12 +51,13 @@ export const AddLiquidityModal: FunctionComponent<
 
   return (
     <ModalBase
-      title={t("addLiquidity.title")}
+      // title={t("addLiquidity.title")}
       {...props}
       isOpen={props.isOpen && showModalBase}
+      hideCloseButton
+      className="!max-w-[57.8125rem]"
```
jonator, line 58: Obviously, since we're given the pool ID, we'll have to eventually query the pool for whether it's a CL pool or balancer/stableswap then show the right view.

jonator (commented):
Some things I noticed out of the gate:
* Make sure you use our prettier config and linter. You may have noticed we have a pre-commit linter hook. I recommend setting it up to work with VSCode or whatever you use.
* We'll likely re-use the depth chart, perhaps in horizontal layout. It should be added to the `components/chart` folder to be reused.
* We should add the pair token price query to our observable query stores (in the store package). Here's an example for individual [token](https://github.com/osmosis-labs/osmosis-frontend/blob/1d88eaa6d41418f08ae670cdf41de40386c21845/packages/stores/src/queries-external/token-historical-chart/index.ts#L30) prices, which should be nearly identical. We'll definitely be using it in the future. So it would be nice to have it be reusable in the future. If you're unfamiliar w/ Mobx, I'd be down to hop on a call and discuss it.
* I can address the other comments in a later review.

chikeichan:
closing this and replacing with #1378 after rebasing on `stage`

PR #1378, opened 2023-03-09
Conliq: Add Liquidity Modal

chikeichan:
TODO:

- [x] add translations
- [x] fix linters errors
- [ ] add loading state when fetching historical price data
- [ ] refactor to use `AddLiquidityConfig` observable store instead of `useState`
- [ ] integrate with add liquidity on localosmosis

Screen Shot 2023-03-02 at 1 14 21 PM


chikeichan:
re-adding comment from @jonator on #1374 

Some things I noticed out of the gate:

- Make sure you use our prettier config and linter. You may have noticed we have a pre-commit linter hook. I recommend setting it up to work with VSCode or whatever you use.
- We'll likely re-use the depth chart, perhaps in horizontal layout. It should be added to the components/chart folder to be reused.
- We should add the pair token price query to our observable query stores (in the store package). Here's an example for individual [token](https://github.com/osmosis-labs/osmosis-frontend/blob/1d88eaa6d41418f08ae670cdf41de40386c21845/packages/stores/src/queries-external/token-historical-chart/index.ts#L30) prices, which should be nearly identical. We'll definitely be using it in the future. So it would be nice to have it be reusable in the future. If you're unfamiliar w/ Mobx, I'd be down to hop on a call and discuss it.
- I can address the other comments in a later review.

packages/web/modals/add-liquidity.tsx
```diff
@@ -50,12 +52,13 @@ export const AddLiquidityModal: FunctionComponent<
 
   return (
     <ModalBase
-      title={t("addLiquidity.title")}
+      // title={t("addLiquidity.title")}
       {...props}
       isOpen={props.isOpen && showModalBase}
+      hideCloseButton
+      className="!max-w-[57.8125rem]"
```
chikeichan, line 59: adding comment from @jonator from #1374 

Obviously, since we're given the pool ID, we'll have to eventually query the pool for whether it's a CL pool or balancer/stableswap then show the right view.

packages/stores/src/queries-external/token-pair-historical-chart/index.ts
```diff
@@ -0,0 +1,92 @@
...
+        priceStore,
+        poolId,
+        baseDenom,
+        quoteDenom,
+        String(tf) as Tf
+      );
+    });
+  }
+
+  get(poolId: string, tf: Tf = "7d", baseDenom = "", quoteDenom = "") {
```
jonator, line 87: NIT: For readability at caller I'd recommend params not have defaults. I can't imagine a scenario where there'd be defautls.

packages/stores/src/queries-external/token-pair-historical-chart/index.ts
```diff
@@ -0,0 +1,92 @@
...
+import { HasMapStore } from "@keplr-wallet/stores";
+import { Dec } from "@keplr-wallet/unit";
+import { computed, makeObservable } from "mobx";
+import { IPriceStore } from "src/price";
+
+import { IMPERATOR_HISTORICAL_DATA_BASEURL } from "..";
+import { ObservableQueryExternalBase } from "../base";
+import { TokenPairHistoricalPrice } from "./types";
+
+const AvailableRangeValues = ["7d", "1mo", "1y"] as const;
```
jonator, line 11: Consider a union type like `"7d" | "1mo" | "1y"`, and make it exportable from types.ts.

packages/web/modals/add-liquidity.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { ObservableAddLiquidityConfig } from "@osmosis-labs/stores";
-import { useStore } from "../stores";
-import { useConnectWalletModalRedirect, useAddLiquidityConfig } from "../hooks";
-import { AddLiquidity } from "../components/complex/add-liquidity";
-import { ModalBase, ModalBaseProps } from "./base";
+import { observer } from "mobx-react-lite";
+import { FunctionComponent } from "react";
 import { useTranslation } from "react-multi-lang";
+
+import { AddConcLiquidity } from "../components/complex/add-conc-liquidity";
```
jonator, line 6: We could use `next/dynamic` to do a dynamic import of this component, to keep visx and other code out of the main bundle.

Example: https://github.com/osmosis-labs/osmosis-frontend/blob/825873b112d261332704b9c04bb4fad41f38c5c1/packages/web/modals/bridge-transfer.tsx#L14

packages/web/package.json
```diff
@@ -52,6 +57,7 @@
     "mobx-react-lite": "^3.2.2",
     "mobx-utils": "^6.0.4",
     "next": "12.0.10",
+    "next-transpile-modules": "9",
```
jonator, line 60: What is this used for?

chikeichan: ah it was used to make visx work with next.js - no longer needed as i've pinned visx version to one that's supported.

packages/web/modals/add-liquidity.tsx
jonator: Yeah this is something I'll add as part of this task https://app.clickup.com/t/864dur3k1

Sorry for late reply

packages/stores/src/ui-config/manage-liquidity/add-liquidity.ts
```diff
@@ -576,4 +608,73 @@ export class ObservableAddLiquidityConfig extends ManageLiquidityConfigBase {
...
+      typeof amount === "number" ? new Dec(amount) : amount;
+  };
+
+  get conliqBaseDepositAmountIn(): Dec {
+    return this._conliqBaseDepositAmountIn;
+  }
+
+  get conliqQuoteDepositAmountIn(): Dec {
+    return this._conliqQuoteDepositAmountIn;
+  }
```
jonator, line 679: It seems this could belong in it's own add liquiity config; something like `ObservableAddConcLiquidityConfig` like you had with the `AddConcLiquidityModal`, perhaps it can also extend `ManageLiquidityBase`???

Would help separat the logic a big, as it seems to not be dependent on this `ObservableAddLiquidityConfig` class in this file at all

jonator (commented):
Left some comments, also note some older comments from a few weeks ago

jonator:
Closing in favor of #1498

PR #1379, opened 2023-03-09
Update Gas Prices

JeremyParish69:
as per Chain Registry, some updates originating from Keplr Chain Registry.

-Deleted one of GenesisL1's duplicate gasPrice specification.
-Carbon will be converted to only use Carbon's Frontend, not Keplr, since our Frontend cannot price the gas properly (need to adjust to specific amounts per tx, not generalized gas prices).

Recommended testing:

Probably should be tested:

-for cases where the low amount is reduced from earlier, that we test a transaction using the low setting:
MEDAS, EBL

-for cases where gas fees are added and weren't there before, and where they aren't in the Keplr Chain Registry at all, we should probably test this since they amounts might be different than the default:
LAMB

Probably doesn't need any testing:

-for cases where the low amount is increased, and is in the Keplr Chain Registry with identical 'low' fee, then we probably don't need to test these, since the Keplr Chain Registry would have acted as the default pricing:
LUNC, INJ, PLQ, AXL, EVMOS

-for cases where to low amount is reduced, but it's in the Keplr Chain Registry, then we probably don't need to test these since Keplr forces the default gas prices:
LUNA

-for cases where gas fees are added and weren't there before, and where they match the Keplr Chain Registry, we likely don't need to test them, since they would have defaulted to what Keplr has:
ATOM, USTC, SCRT, REGEN, DVPN, XPRT, IRIS, CRO, IOV, EUR, JUNO, LIKE, IXO, BCNA, BTSG (KCR-PR), BOOT, CMDX, LUM, DSM, ROWAN, UMEE, USDC.grv, USDT.grv, MNTL, KAVA, BLD, IST, JKL

-for cases where gas fees are added and weren't there before, and where they don't match the Keplr Chain Registry:
(none)

-for cases where the average and/or high price changes but the low fee doesn't, we likely don't need to test those, since the low amount theoretically already has proven to work (we shouldn't have any low fee prices that never work, but it is possible they are defined in some cases):
OSMO, MEME, CRE, MARS, CANTO

-for cases where the low amount is increased, and isn't in or doesn't match the Keplr Chain Registry, that we go to the chain's primary interface and see how a chain suggestion looks like (might not be possible for Chains natively added to Keplr:
ACRE (their site uses Metamask, not Keplr--but it should still at least work since the low amount isn't being lowered)


packages/web/config/chain-infos.ts
```diff
@@ -44,8 +44,8 @@ const chainInfos = (
           isFeeCurrency: true,
           gasPriceStep: {
             low: 0,
```
delivan, line 46:
```suggestion
            low: 0.0025,
```
the minimum gas price should be 0.0025 after v15, so low gas price is also should be 0.0025

JeremyParish69: what's the ETA on the upgrade?

jonator (approved):

PR #1395, opened 2023-03-16
feat: Add Onramp Selection Modal

JoseRFelix:


jonator (commented):
Would it be possible to reuse this config as the source of truth of available ramps?

https://github.com/osmosis-labs/osmosis-frontend/blob/71f9704b48dd472056e0d1083b80d94cfa608233/packages/web/integrations/bridge-info.ts#L56-L72

jonator (approved):

Issue #1391, opened 2023-03-15
Migrate deposit address logic from hooks to a Mobx store


jonator:
### Problem
When I initially built the Axelar integration, like with most things, it grew to be more complex then I anticipated. I started by building the logic directly with hooks, but I'd like to contain the deposit address generation logic within a MobX store.

### Background
[What is Axelar?](https://docs.axelar.dev/)
For us, Axelar provides a secure bridge for EVM assets originating from EVM chains. Examples: ETH/WETH and USDC from Ethereum, MATIC/WMATIC from Polygon, etc.

How Axelar integration works:
* We generate "deposit addresses" using the Axelar JS SDK. We pass it the source chain, destination chain, and asset key. Optionally, when bridging to an EVM chain, we can pass a flag to request that Axelar unwrap the asset into it's native form (ERC20 to Native gas token)
* To kick off the transfer, we simply have users send funds to the deposit address. When transferring to Osmosis, this comes in the form of a ERC20 transfer or send to a EVM address via MetaMask. When withdrawing, this comes in the form of an IBC transfer from Osmosis to an address on Axelar chain (a Cosmos chain).

### Proposed Solution
This new `AxelarDepositAddressUIConfig` should live in the integrations/axelar folder, to keep Axelar-related code adjacent.

We can either maintain the instance in the AxelarTransfer modal component, or within the root.ts store (in the React context). This comes with it's own tradeoffs of how long we can hold onto data. For an example of a Mobx store in a non-root React component, see the [ObservableTransferUiConfig](https://github.com/osmosis-labs/osmosis-frontend/blob/50374547a4d50ee7c5288d93400f0c24d94a05f2/packages/web/stores/assets/transfer-ui-config.ts#L35) store, who's reference is maintained in a useState [wrapper hook](https://github.com/osmosis-labs/osmosis-frontend/blob/50374547a4d50ee7c5288d93400f0c24d94a05f2/packages/web/hooks/ui-config/use-transfer-config.ts#L7) in the [assets page](https://github.com/osmosis-labs/osmosis-frontend/blob/50374547a4d50ee7c5288d93400f0c24d94a05f2/packages/web/pages/assets/index.tsx#L50).

[...]

PR #1405, opened 2023-03-20
Update useDepositAddress to use fromPromise for caching Axelar calls

spryor:
Issue: https://github.com/osmosis-labs/osmosis-frontend/issues/1391

I considered a couple possible solutions to this. First, I followed the suggestion from the issue of making a new observable class instance that was saved in a `useState`. This approach worked for the most part, however, there were issues with computedFn not correctly memoizing based on the input args. Based on other warnings in my console, this issue appears to be happening elsewhere, though I did not have time to verify.  While the class instance could be made to work it didn't, in opinion, return better ergonomics or expanded functionality so long as it wasn't in the root store. While I could have put it in the root store there didn't seem to be a benefit of doing so since it is currently only used in one place in the app. In fact, adding it to the store would add extra complexity regarding data lifespan, as noted in the issue above. In the end, I opted for a simpler solution that simplifies the existing hook, removes dead code, and lets mobx handle the async state by wrapping each deposit address request promise in an observable and caching each observable.

### DIFF
--- a/packages/web/config/price.ts
+++ b/packages/web/config/price.ts
@@ -1345,8 +1345,8 @@ export const PoolPriceRoutes: IntermediateRoute[] = [
     alternativeCoinId: "pool:fox",
     poolId: "949",
     spotPriceSourceDenom: DenomHelper.ibcDenom(
-        [{ portId: "transfer", channelId: "channel-169" }],
-        "cw20:juno1u8cr3hcjvfkzxcaacv9q75uw9hwjmn8pucc93pmy6yvkzz79kh3qncca8x"
+      [{ portId: "transfer", channelId: "channel-169" }],
+      "cw20:juno1u8cr3hcjvfkzxcaacv9q75uw9hwjmn8pucc93pmy6yvkzz79kh3qncca8x"
     ),
     spotPriceDestDenom: "uosmo",
     destCoinId: "pool:uosmo",
--- a/packages/web/integrations/axelar/hooks/use-deposit-address.ts
+++ b/packages/web/integrations/axelar/hooks/use-deposit-address.ts
@@ -14,7 +14,7 @@ function useCache<T>() {
 }
 
 /**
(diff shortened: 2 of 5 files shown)


packages/web/integrations/axelar/hooks/use-deposit-address.ts
```diff
@@ -15,96 +28,66 @@ export function useDepositAddress(
...
-    destinationAddress,
+    axelarAssetTransfer,
     sourceChain,
     destChain,
-    coinMinimalDenom,
-    autoUnwrapIntoNative,
-    setIsLoading,
-  ]);
-
-  const doGen = useCallback(
```
spryor, line 76: I removed `doGen` as it wasn't required. `doGen` wraps `generateAddress` which already returns a promise. Moreover, `generateAddress` already handles throw errors and doesn't return anything for successful calls. So line 82 here is unreachable since `address` is always going to be falsy.

jonator:
I appreciate the rationale, and even agree. But I'm curious to see what you came up with, regardless of whether it's better or worse then the use of the hook. 

RE: the `computedFn`, are you perhaps passing in a reference? If we initialize a reference to an object or function in the body of the render function of the calling component without memoization (useCallback, useMemo), it'd be different every time and would prevent memoization.

packages/web/config/price.ts
```diff
@@ -1345,8 +1345,8 @@ export const PoolPriceRoutes: IntermediateRoute[] = [
     alternativeCoinId: "pool:fox",
     poolId: "949",
     spotPriceSourceDenom: DenomHelper.ibcDenom(
-        [{ portId: "transfer", channelId: "channel-169" }],
-        "cw20:juno1u8cr3hcjvfkzxcaacv9q75uw9hwjmn8pucc93pmy6yvkzz79kh3qncca8x"
+      [{ portId: "transfer", channelId: "channel-169" }],
```
spryor, line 1348: This is just a `yarn lint:fix` whitespace change

PR #1407, opened 2023-03-21
Optimize arUSD, CNTO Images

JeremyParish69:


packages/web/public/tokens/arusd.svg
jonator, line 1: Hey this is 3k lines, are we sure it's not an image hiding within an svg?

JeremyParish69: Yes, it wasn't an embedded image, but I was able to open and re-export and bring it down to 5 KB

PR #1427, opened 2023-03-23
feat: Add Observable Pools with Metrics

JoseRFelix:
## Description
 
To fix the pools page table's slow performance and responsiveness, this PR creates an observable pool class to handle pools directly from Mobx instead of down-streaming multiple steps during render.


packages/stores/src/derived-data/pools/metrics.ts
```diff
@@ -0,0 +1,167 @@
...
+    makeObservable(this);
+  }
+
+  @computed
+  get allPools() {
+    const allPools = this.queriesStore
+      .get(this.chainId)
+      .osmosis?.queryGammPools.getAllPools();
+
+    const pools = allPools?.map(
```
jonator, line 121: This may be really good since it may only map over all pools when the pools query re/loads, instead of any underlying metric changing.

Ideally, we should have a list of ObservablePoolWithMetric references that are maintained (and therefore observed), and only iterated over when the actual list of pools changes.

packages/stores/src/derived-data/pools/metrics.ts
```diff
@@ -0,0 +1,183 @@
...
+import { ObservablePoolsBonding } from "../pool/bonding";
+import { ObservablePoolDetails } from "../pool/details";
+
+class ObservablePoolWithMetric {
+  @observable
+  pool: ObservableQueryPool;
+
+  constructor(
+    pool: ObservableQueryPool,
+    protected readonly poolDetails: ObservablePoolDetails,
```
jonator, line 26: Also note: there's an `ObservableQueryPool` within `ObservablePoolDetails`

packages/stores/src/derived-data/pools/metrics.ts
```diff
@@ -74,28 +85,26 @@ class ObservablePoolWithMetric {
...
-  @computed
   get feesSpent7d() {
     return this.feePoolMetrics.feesSpent7d;
   }
 }
 
 /** Fetches all pools directly from node in order of pool creation. */
 export class ObservablePoolsWithMetric {
-  protected _pools = observable.array();
+  @observable
```
jonator, line 99: Why does this need to be observable?

JoseRFelix: You're right! Removed it accordingly

jonator:
Please provide an overview of changes in description going forward

jonator (approved):

PR #1438, opened 2023-03-27
feat: Remove Zero Fee for Swaps

JoseRFelix:
## Description

Previously, zero fee swaps were possible whenever the user had less than `0.001 osmo` on their wallet.  This pull request makes fees always required for swaps.


jonator (approved):

arispech:
This looks good. 

Currently on app.ososis.zone, The UI allows zero fees as seen here
image


Working as expected on this PR, with the introduction of this, fees will now be required for all swaps, regardless of the amount of osmo in the user's wallet. 
image 👍

PR #1449, opened 2023-03-29
Update LUMEN Logo

JeremyParish69:
png to svg, new design


jonator (commented):
That svg is waaay too big IMO

JeremyParish69:
> That svg is waaay too big IMO

The first PNG of the redesigned logo was over 5MB, and then the generated SVG for that was over 2.5MB. But then we wrestled with it pretty vigorously to finally get it down below 500KB (thinking that was the allowed maximum). I don't think we can get it much lower without severely degrading the quality. If we can't use this SVG due to its size, do we have to compress the PNG version?

jonator:
Yeah generally generating SVGs from PNGs defeats the purpose of SVGs. Normally the image data (which can be more complex) is directly encoded in the SVG has byte data in an attribute on an image tag in the SVG.

It's possible for SVGs to not be appropriate for some images if they're complex enough.

jonator (approved):

Issue #603, opened 2022-06-07
Add new trade routing logic


dogemos:
Add the improved router implementation for the Osmosis trade frontend

jonator:
@dogemos any updates on progress here?

jonator:
[Original slack thread](https://osmosis-network.slack.com/archives/C027ELA2ZM1/p1667576069713229?thread_ts=1667575819.348319&cid=C027ELA2ZM1)

Perhaps, we can add a query param to the pools query to only fetch pools with a min amount of OSMO liquidity. Then, we'd only get a response with pools worth calculating routes through, with the added benefit of a more space+bandwidth efficient query.

jonator:
- [ ] Add swap fees, including OSMO incentives, in calculation of optimized routes

PR #1451, opened 2023-03-30
Support routes with more pools

### LINKED ISSUE #1231: Unable to swap stATOM form/to multiple tokens

jonator:
Used a new recursive (vs iterative) algorithm for finding top 3 routes of select pools to route through

Helps address early portions of #603 
Fixes #1231

TODO:
- [x] Address any new performance issues, especially when opening the token select drawer for the first time (remaining pools are queried and routes are recalculated; the recursive algorithm causes more iterations than previous naive approach)
- [x] Increase hop limit to 4 to manage stX to stX

Screenshot 2023-04-01 at 10 14 36 AM


JoseRFelix:
@jonator I think we need to merge/rebase changes from `stage`

jonator:
@JoseRFelix I think it's because I branched from master, which had a lot of changes, and am now trying to merge to stage

JoseRFelix (approved):
LGTM 🚀

PR #1455, opened 2023-03-31
fix: capture errors

jonator:
* Prevent quoting errors from crashing frontend, display routing error instead.


jonator:
Confirmed fix by support member

PR #1458, opened 2023-04-01
Update Gauge Display Time by it's first Epoch

JeremyParish69:
instead of by start time, which isn't as useful as its corresponding epoch time

### DIFF
--- a/packages/web/pages/api/active-gauges.ts
+++ b/packages/web/pages/api/active-gauges.ts
@@ -23,7 +23,7 @@ export default async function activeGauges(
   const resp = await fetch(endpoint);
   const { data } = (await resp.json()) as ExternalIncentiveGaugesResponse;
   
-  const epochs_endpoint = `${ChainInfos[0].rest}osmosis/epochs/v1beta1/epoch-infos`;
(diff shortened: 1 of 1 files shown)


jonator (changes requested):
It seems this is the format of the response; seems `epoch_data` is not correct:
Screenshot 2023-04-03 at 3 09 23 PM

doggystylez:
@JeremyParish69 I was just looking at epoch time today and it seems the start time has been static since Sept 2021.

```epochs:
- current_epoch: "97"
  current_epoch_start_height: "0"
  current_epoch_start_time: "2021-09-23T17:16:09.898160996Z"
  duration: 86400s
  epoch_counting_started: true
  identifier: day
  start_time: "2021-06-18T17:00:00Z"
```

```
- current_epoch: "690"
  current_epoch_start_height: "9550620"
  current_epoch_start_time: "2023-05-09T17:16:09.898160996Z"
  duration: 86400s
  epoch_counting_started: true
  identifier: day
  start_time: "2021-06-18T17:00:00Z"
```

JeremyParish69:
@doggystylez that's good to know--I didn't know that. This might save us the query, although I'm not sure if we can reliable count on it to never shift again.

jonator (approved):

PR #1462, opened 2023-04-04
Add GRDN

JeremyParish69:
Add Guardian CW20 from Juno.
Deposit/Withdraw has been validated.


jonator (approved):
We've tested, and confirmed it works @JeremyParish69

PR #1467, opened 2023-04-06
Kleomedes Image

sumandobanzar:


JeremyParish69:
Thanks @sumandobanzar ,

Could you please register KLEO to the Chain Registry https://github.com/cosmos/chain-registry/blob/master/juno/assetlist.json and also combine the changes from both PRs you've created to this repo into just one?

sumandobanzar:
.

PR #1476, opened 2023-04-10
feat: Display Warning When Pool APR is Too High

JoseRFelix:
## Description

Add a warning for users in the pools table to make them aware if a pool could be experiencing high inflation. The warning is displayed if the staking APR is `5` times larger than the Osmosis chain staking APR and the user is not in frontier.


packages/web/components/complex/all-pools-table.tsx
```diff
@@ -354,8 +356,42 @@ export const AllPoolsTable: FunctionComponent<{
...
+                .gt(
+                  inflation.inflation.toDec().quo(new Dec(100)).mul(new Dec(5))
+                );
+
               return (
-                <MetricLoaderCell value={props.getValue().apr.toString()} />
+                <MetricLoaderCell
+                  isLoading={
+                    queriesOsmosis.queryIncentivizedPools.isAprFetching ||
+                    inflation.isFetching
```
jonator, line 376: There may be a debounce issue with this, causing the animation to restart repeatedly and stutter...

jonator (approved):

packages/web/components/cards/bond-card.tsx
```diff
@@ -189,135 +193,164 @@ const Drawer: FunctionComponent<{
...
         }
-      )}
-    >
+      });
+      return Array.from(imgSrcDenomMap.values());
+    }, [incentivesBreakdown]);
+    const t = useTranslation();
+
+    const queriesCosmos = queriesStore.get(chainId).cosmos;
+    const inflation = queriesCosmos.queryInflation;
```
jonator, line 226: Don't we want to get the inflation of the counterparty token? LMK if I'm misunderstanding.. 

I think we can get the chain ID from the coin denom with this function;
https://github.com/osmosis-labs/osmosis-frontend/blob/efc03405f304615602ca199917c7ec2bcbde1b39/packages/stores/src/chain.ts#L60

packages/web/components/complex/all-pools-table.tsx
JoseRFelix: yea, adding `inflation.isFetching` caused a weird state transition. Removed it as there was no point in hiding apr while inflation loads.

packages/web/components/cards/bond-card.tsx
JoseRFelix: Calculating inflation accurately for each token chain can be challenging, as some tokens, such as `WBTC`, `USDC`, and `FOX`, are not directly associated with the chain's inflation. Johnny's method uses the Osmosis chain's inflation as a reference point. Here's a graph he made showcasing that high inflation pools would be identified with this baseline. The red line represents Osmosis Staking APR times 5. 
image

JoseRFelix:
Ready for a re-review. Added new design changes

jonator (approved):

PR #1479, opened 2023-04-12
Shade Protocol migration

DrPresident:
Added new shade denomination, changed old SHD to an icon indicating it is "stale"

### DIFF
--- a/packages/web/config/chain-infos.ts
+++ b/packages/web/config/chain-infos.ts
@@ -216,9 +216,20 @@ const chainInfos = (
         {
           type: "cw20",
           contractAddress: "secret1qfql357amn448duf5gvp9gr48sxx9tsnhupu3d",
-          coinDenom: "SHD",
+          coinDenom: "SHDold",
           coinMinimalDenom:
             "cw20:secret1qfql357amn448duf5gvp9gr48sxx9tsnhupu3d:SHD",
+          coinDecimals: 8,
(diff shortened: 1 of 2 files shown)


JeremyParish69:
Thanks, @DrPresident . What is the pool id for the new SHD?

DrPresident:
> Thanks, @DrPresident . What is the pool id for the new SHD?

I haven't done anything with new pools, I was thinking we would need to do this first and then create one? I may need some guidance I'm not entirely confident on what steps are needed

JeremyParish69:
> I haven't done anything with new pools, I was thinking we would need to do this first and then create one? I may need some guidance I'm not entirely confident on what steps are needed

You can use this generated preview link to transfer new SHD and create a pool: https://osmosis-frontier-git-jp69-add-shd-new-osmo-labs.vercel.app/

DrPresident:
> > I haven't done anything with new pools, I was thinking we would need to do this first and then create one? I may need some guidance I'm not entirely confident on what steps are needed
> 
> You can use this generated preview link to transfer new SHD and create a pool: https://osmosis-frontier-git-jp69-add-shd-new-osmo-labs.vercel.app/

I sent you a DM on telegram, I'm running into some small issues with pool creation. One is that when I search for SHD it seems to have a "collision" between old & new shd. I think they probably have the same "key" for how they are referenced through the UX. I also added SILK to this PR (can move to its own if thats preferred).

PR #1487, opened 2023-04-15
Go turbo 🚀

jonator:
https://turbo.build/repo

Makes our builds 10x faster.

To try it out.

Go to root of repo and install deps, including turbo:
`yarn`

Run:
`yarn build-lint-fast`

Run it again to see the caching in action.

This is the whole repo cached, with build and linting:
Screenshot 2023-04-14 at 10 58 04 PM

## @daniel-farina - Using on Vercel Builds
This is a vercel project. It supports remote caching (to get the times in the screenshot), and is built with mono-repos as first class.

That means, perhaps we can run this for the build command in Vercel (instead of yarn build) to prevent from having to rebuild the slow (3 min) stores package if there's no changes there, and instead use the cached package on Vercel servers? Thoughts?

## TODO
- [ ] Get `npx turbo dev` to work, so we can use caching when running a watch dev server.


JoseRFelix:
Nice, this is a massive improvement! 🚀 . IMO, if we are moving into Turborepo, it would be best to discontinue using Lerna and replace it completely.

jonator:
> Nice, this is a massive improvement! 🚀 . IMO, if we are moving into Turborepo, it would be best to discontinue using Lerna and replace it completely.

Does it have the ability to do parallel common command execution?

daniel-farina:
If this is based on caching, I'm not sure how it would work on Vercel's deployment machines but it's worth the try.  I created a new Vercel project to test and configured it with this PR branch only.

daniel-farina:
It's failing on Vercel for now with:
https://vercel.com/osmo-labs/osmosis-frontend-turbo-build-test/6mph4K3f8Z9JP4A7UucbgCxyKWnv
```

I tried a couple of different things but `yarn build-lint-fast` does not seem to work.  I'll keep playing with it.

**Comment by JoseRFelix** [comment]
> > Nice, this is a massive improvement! 🚀 . IMO, if we are moving into Turborepo, it would be best to discontinue using Lerna and replace it completely.
> 
> Does it have the ability to do parallel common command execution?
> 
> Reading: [turbo.build/repo/docs/getting-started/existing-monorepo](https://turbo.build/repo/docs/getting-started/existing-monorepo)
> 
> Though is it missing publishing automation that lerna has? Seems it is: [turbo.build/repo/docs/handbook/publishing-packages](https://turbo.build/repo/docs/handbook/publishing-packages)
> 
> Can we just use it for building and linting? Lerna could be for publishing?
> 
> It suggests this for managing publishing: [changesets/changesets](https://github.com/changesets/changesets)

It does by default! You can opt-out by adding the `dependsOn` key that will wait for the dependencies to build first.

Yea, on its own it doesn't handle publishing like Lerna. As you mentioned Vercel recommends using changesets. I've used it before, and it makes handling changelogs very easy. However, I don't know if it's worth doing that jump right now. It could take some time to tweak and set up. 

Since lerna and turbo are development tools, I don't think it's a big issue to use them both, but perhaps we could also use turbo for development.

**Comment by jonator** [comment]
@JoseRFelix 

I'm getting this when I run `npx turbo dev`:

```bash
ERROR  run failed: error preparing engine: Could not find the following tasks in project: dev
Turbo error: error preparing engine: Could not find the following tasks in project: dev
```

JoseRFelix:
> @JoseRFelix 
> 
> 
> 
> I'm getting this when I run `npx turbo dev`:
> 
> 
> 
> ```bash
> 
> ERROR  run failed: error preparing engine: Could not find the following tasks in project: dev
> 
> Turbo error: error preparing engine: Could not find the following tasks in project: dev
> 
> ```

We need to add 'dev' to turbo.json. Like we did for build and lint

daniel-farina (approved):
Quick as a hiccup 🚀

jonator:
@JoseRFelix I get it now. So I updated all workspace commands to use turbo. Across the board, they should simply be faster, but reruns should be cached and nearly instant now. Also, I set up a dependency pipeline so you can simply run the command you want to run without having to first run any other commands (except for `yarn`). In other words, if you run `yarn dev` and the project hasn't been built yet, it will get built first. If it's been built first before running `yarn dev` it will hit the cache and be instant. Same applies to tests, linting, etc.


@daniel-farina you shouldn't need to make _any_ changes to our deployments; I removed `build-lint-fast` to keep our current commands. These change the `yarn X` commands to use turbo instead of lerna to do the same thing; but now we have a faster runtime (rust) and fine-grained caching enabled 🚴‍♂️

README.md
```diff
@@ -21,6 +21,8 @@ Start web server
 yarn && yarn build && yarn start
 ```
 
+If you'd like to build with caching with Turbo repo (4x faster, 10x+ faster cached), replace `yarn build` with `yarn build-lint-fast`. To leverage our shared/remote cache, follow [these](https://turbo.build/repo/docs/core-concepts/remote-caching#for-local-development) instructions.
+
```
JoseRFelix, line 25: I think that with the recent changes this could be removed

jonator: Ah meant to update readme. Forgot.

jonator:
> If this is based on caching, I'm not sure how it would work on Vercel's deployment machines but it's worth the try. I created a new Vercel project to test and configured it with this PR branch only.

It's built by Vercel, so similar to the out of the box features we get with Next.js, if we build projects with turbo repo we get things with 0 config, such as remote caching.

RE remote caching; if we push a change to just the web package to Vercel, instead of building the other packages, particularly the stores package which can be slow, it re-uses cached versions from previous builds since no source code/input changes would be present. This should cut build times from 3m to seconds in some cases. Also, if you sign into turbo (`turbo login`), then link the repo (`turbo link`) you can download from that shared cache as well, so the first time you clone our repo you could use cached builds instead of building from scratch 🙂. Overkill for us right now, but great for larger repos I'm sure.

PR #1495, opened 2023-04-18
add catmos to frontier

ApesOnCosmos:


JeremyParish69:
The token image must go in the config/public/tokens/ directory.

JeremyParish69 (approved):

PR #1498, opened 2023-04-18
(CL) Add liquidity

chikeichan:


jonator (commented):
I think my liquidity_for_range query on my localosmosis is returning an error right now.. but that should not throw exceptions still. Both exceptions were this:

Screenshot 2023-04-27 at 1 54 14 PM

packages/stores/src/queries-external/token-pair-historical-chart/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      this.poolId !== "" &&
+      AvailableRangeValues.includes(this.tf) &&
+      this.tf != null &&
+      this.baseDenom != null &&
+      this.quoteDenom != null
+    );
+  }
+
+  @computed
+  get getChartPrices(): TokenPairHistoricalPrice[] | undefined {
```
jonator, line 52: IMO it'd be cleaner to return a `[]` instead of `undefined`.

We can check if data is loading by using the base class' `isFetching` member

packages/stores/src/queries-external/token-pair-historical-chart/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  protected canFetch(): boolean {
+    return (
+      this.poolId !== "" &&
+      AvailableRangeValues.includes(this.tf) &&
+      this.tf != null &&
+      this.baseDenom != null &&
+      this.quoteDenom != null
+    );
+  }
```
jonator, line 49: When would these all be null? In the getter for the map store they should be required.

packages/web/components/chart/concentrated-liquidity-depth.tsx
```diff
@@ -0,0 +1,229 @@
...
+              },
+            })}
+            horizontal={horizontal}
+          >
+            {/* Uncomment when testing alignment */}
+            {/*<AnimatedAxis*/}
+            {/*  orientation="right"*/}
+            {/*  numTicks={5}*/}
+            {/*  strokeWidth={0}*/}
+            {/*/>*/}
```
jonator, line 120:
```suggestion
```

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -0,0 +1,908 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+      <VolitilitySelectorGroup
+        updateInputAndRangeMinMax={updateInputAndRangeMinMax}
+        addLiquidityConfig={addLiquidityConfig}
+      />
+      <div className="flex flex-col">
+        <div className="px-2 py-1 text-sm">Amount to deposit</div>
```
jonator, line 532: Need localization here

packages/web/modals/add-liquidity.tsx
```diff
@@ -32,14 +45,33 @@ export const AddLiquidityModal: FunctionComponent<
...
+      chainStore,
+      chainId,
+      poolId,
+      queriesStore
+    );
+
+  // initialize pool data stores once root pool store is loaded
+  const { poolDetail } = derivedDataStore.getForPool(poolId as string);
+  const pool = poolDetail?.pool?.pool;
+  const isConcLiq = pool instanceof ConcentratedLiquidityPool;
```
jonator, line 59: Nitpick: I'd prefer if we do `poolDetail?.pool.type === "concentrated"`

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -0,0 +1,908 @@
...
+          {t("addConcentratedLiquidity.volatilityDescription")}
+        </div>
+        <a
+          className="flex flex-row items-center text-caption font-caption text-wosmongton-300"
+          href="#"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          {t("addConcentratedLiquidity.superchargedLearnMore")}
+          <Image src="/icons/arrow-right.svg" height={12} width={12} />
```
jonator, line 603: I'm getting a warning of a missing alt prop here.

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -0,0 +1,908 @@
...
+
+  const { nativeBalances, ibcBalances } = assetsStore;
+
+  const fiatPer = coin?.currency.coinGeckoId
+    ? priceStore.getPrice(coin.currency.coinGeckoId, undefined)
+    : 0;
+
+  const [walletBalance] = nativeBalances
+    .concat(ibcBalances)
+    .filter((balance) => balance.balance.denom === coin?.denom);
```
jonator, line 704: This is one way to get the balance. IMO a simplr way is to use the cosmos SDK balances query, passing in the user bech32 address and the pool denoms:

```
const balance = this._queryBalances
      .getQueryBech32Address("osmo23123123123123")
      .getBalanceFromCurrency("uosmo");
```

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -0,0 +1,908 @@
...
+    addLiquidityConfig,
+    pool,
+    poolName,
+    superfluidPoolDetail,
+    poolDetail,
+    onRequestClose,
+  }) => {
+    const { priceStore, queriesExternalStore } = useStore();
+    const router = useRouter();
+    const { id: poolId } = router.query as { id: string };
```
jonator, line 143: IMO we should get the poolId from the `addLiquidityConfig`, as we may not have a query route to rely on in some use cases. Please update this everywhere the router.query is used.

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -0,0 +1,908 @@
...
+              imgSrc="/images/conliq_mock_range.png"
+            />
+          </div>
+        </div>
+        <div className="flex w-full items-center justify-center">
+          <Button
+            className="w-[25rem]"
+            onClick={() => addLiquidityConfig.setModalView(selected)}
+          >
+            Next
```
jonator, line 246: This needs to be localized. There's prob an existing key for this.

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
```diff
@@ -0,0 +1,313 @@
...
+  @observable
+  protected _baseDepositAmountIn: Dec = new Dec(0);
+
+  @observable
+  protected _quoteDepositAmountIn: Dec = new Dec(0);
+
+  constructor(
+    chainGetter: ChainGetter,
+    initialChainId: string,
+    poolId: string,
```
jonator, line 82: Everything that is expected to remain the same for the life of this UI config, like the pool, should not be observable with getters/setters. (This applies to the members above). For this poolId specifically, we could remove a lot of code by using `protected readonly` to overload it as a member.

packages/web/localizations/en.json
```diff
@@ -17,6 +17,25 @@
...
+    "priceRange": "Price Range",
+    "currentPrice": "current price",
+    "basePerQuote": "{base} per {quote}",
+    "selectVolatilityRange": "Select volatility range",
+    "volatilityDescription": "Tight ranges earn more fees per dollar, but earn no fees when price is out of range.",
+    "superchargedLearnMore": "Learn about supercharged liquidity",
+    "fullRange": "Full range",
+    "amountToDeposit": "Amount to deposit",
+    "addLiquidity": "Add Liquidity"
+  },
```
jonator, line 38: These will need to be added to other langs at some point. Tedious, I know. We use google translate in the mean time until a user makes a comment about it's correctness.

jonator (changes requested):
Could you work with Syed at getting a higher quality graphic for these buttons. They look very cheap.

[Link to figma frame](https://www.figma.com/file/RxNOvfvzmgtsgCPUetQvcG/Concentrated-Liquidity?node-id=1173-48250&t=BRysHB3EUgdhEdRc-4)
Screenshot 2023-05-04 at 9 17 55 AM

packages/math/src/pool/concentrated/math.ts
```diff
@@ -238,3 +239,37 @@ export function convertTokenInGivenOutToTokenOutGivenIn(
   // is in given out -- token1 for token0
   return new Dec(specifiedTokenOut.amount).quo(currentPrice).truncate();
 }
+
+export function calculateDepositAmountForQuote(
+  curPrice: Dec,
+  lowerTick: Int,
+  upperTick: Int,
+  baseDeposit: Dec,
+  exponentAtPriceOne: number
```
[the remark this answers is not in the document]
jonator: Merged. Please update branch and you should get access to `exponentAtPriceOne` in `@osmosis-labs/math` package

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
```diff
@@ -0,0 +1,313 @@
...
+    ];
+  };
+
+  get range(): [Dec, Dec] {
+    return this._range;
+  }
+
+  get tickRange(): [Int, Int] {
+    return [
+      priceToTick(this._range[0], this.pool.exponentAtPriceOne),
```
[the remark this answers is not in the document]
jonator: Merged. Please update branch and you should get access to exponentAtPriceOne in @osmosis-labs/math package

packages/stores/src/queries-external/token-pair-historical-chart/index.ts
chikeichan: It is required by type, by in runtime these params are derived using a `split`

```
      const [poolId, tf, baseDenom, quoteDenom] = symbolTfBaseAndQuote.split(",");
```

although the format for `symbolTfBaseAndQuote` might be enforced, i think it's safer to do a null check here (same pattern as `token-historical-chart/index.ts`

packages/stores/src/queries-external/token-pair-historical-chart/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    protected readonly tf: Tf
+  ) {
+    super(
+      kvStore,
+      "https://api.osmosis.zone/",
+      // baseURL,
+      `/pairs/v1/historical/2/chart?asset_in=${baseDenom}&asset_out=${quoteDenom}&range=${tf}&asset_type=symbol`
+      // `/pairs/v1/historical/${poolId}/chart?asset_in=${baseDenom}&asset_out=${quoteDenom}&range=${tf}&asset_type=symbol`
+    );
+    console.log(baseURL);
```
jonator, line 37:
```suggestion
```

jonator (changes requested):
Some things I noticed, adding as review before I forget:

### Visual tweaks

This is 20px padding, should be 12px:
Screenshot 2023-05-12 at 10 24 38 AM

Padding seems a bit tight here, with button sizes not quite right
Screenshot 2023-05-12 at 10 27 13 AM

PR #1505, opened 2023-04-21
Publish Stage

jonator:
* Fix issue with token select in create pool workflow
* Swap router: filter by TVL instead of specific tokens

## Testing
* Make sure create pool token select works
* Check for performance concerns in swap tool, drawers. Especially on frontier.
* Double check swap tool on pool detail pages


arispech (approved):
I was able to select tokens when creating a pool (selected ATOM shows ATOM not stATOM)👍
The swap tool's drawers performance is looking good. Especially on frontier👍
The swap tool on pool detail pages works as expected👍

PR #1510, opened 2023-04-21
Add FIL

JeremyParish69:
axelar-bridged Wrapped FIL
transfers validated.
now using pool 1006


jonator (approved):
Assuming bridging has been tested in previews?

PR #1517, opened 2023-04-25
add VOID to frontend

Jason0xOg:


packages/web/public/tokens/void.png
jonator, line 1: Seems the corners are clipped or something, can you confirm if this is intentional?

jonator (approved):

packages/web/public/tokens/void.png
Jason0xOg: It's fine @jonator

PR #1526, opened 2023-04-27
Add BLZ token from bluzelle

Galadrin:


JeremyParish69:
Need to register IBC connection to Cosmos Chain Registry https://github.com/cosmos/chain-registry/tree/master/_IBC

packages/web/config/chain-infos.ts
```diff
@@ -3242,6 +3242,34 @@ const chainInfos = (
...
+        coinType: 483,
+      },
+      bech32Config: Bech32Address.defaultBech32Config("bluzelle"),
+      currencies: [
+        {
+          coinDenom: "BLZ",
+          coinMinimalDenom: "ubnt",
+          coinDecimals: 6,
+          coinImageUrl: "/tokens/bluzelle.svg",
+          coinGeckoId: "bluzelle",
```
JeremyParish69, line 3260: please define the price using a pool's spot price (will require that a pool be created)

JeremyParish69 (approved):

PR #1527, opened 2023-04-27
feat: Create `proto-codecs` package

JoseRFelix:
## Motivation

Allow adding encoders for new features like CL and omit unused proto to reduce bundle size. 

## Relevant files to review
- `scripts/codegen.ts`
- `scripts/get-proto.sh`
- `scripts/gen-package-types.ts`


packages/README.md
```diff
@@ -4,6 +4,8 @@
 
 @osmosis-labs/stores - contains Mobx stores for accounts and queries, price data, historical data (imperator), common UI config (creating pool, swapping, managing liquidity) and the GAMM currency registrar
 
+@osmosis-labs/osmojs - Contains libraries for interacting with the Osmosis chain
```
jonator, line 7: Is `osmojs` a fitting name? Isn't this like `types` or something since it has the proto encoders/decoders?

Regardless- IMO this description leaves a lot of room for improvement.

JoseRFelix: > Is osmojs a fitting name? Isn't this like types or something since it has the proto encoders/decoders?

I haven't decided on the name. Instead of osmojs, I'll name it `proto-encoder` as a placeholder. For sure, there must be a better name for this. I'm open to any suggestions.

This is a library to generate the blockchain encoders/decoders plus amino types. 

> Regardless- IMO this description leaves a lot of room for improvement.

For sure, I'll be improving this

packages/proto-encoder/scripts/get-proto.sh
```diff
@@ -0,0 +1,99 @@
...
+
+PROTO_DIR="./chain-proto"
+COSMOS_SDK_VERSION="v0.47.1"
+
+OSMOSIS_COMMIT_HASH="ac133fdd73dbd27b8bc7b6d49049f5d594e771cb"
+COSMOS_COMMIT_HASH="6ed904cdb75fe08a9310e7df49eaf9c5b2b64692"
+IBC_COMMIT_HASH="10324f5bf6840892cd7bba8a40231fc52a8b1745"
+WASMD_COMMIT_HASH="4c906d5a53a255c551d6ed981a548cffe47ae9f0"
+ICS23_COMMIT_HASH="f4deb054b697458e7f0aa353c2f45a365361e895"
+
```
jonator, line 13: For the osmosis-proto-tool I made: using ChatGPT (I don't know bash) I was actually able to extract the repo tags/versions Osmosis depends on in it's go.mod file in the 1-get-proto.sh script. By default, it gets the latest version of Osmosis on main branch, then all it's dependencies. If you want, you can supply the Osmosis commit hash as a param to get a specific version of Osmosis.

Is there a reason you used all commit hashes?

See:
https://github.com/osmosis-labs/osmosis-proto-tool/blob/0630cd5541b8d1227c2c350239d204f8bef6823d/1-get-proto.sh#L61

packages/proto-encoder/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@osmosis-labs/proto-encoder",
```
jonator, line 2: Actually, sorry, after knowing what this package is for (coming in and running a command to generate a new coding library for Osmosis. aka Telescope just for Osmosis). I think `proto-encoder` is not a great name. Open to suggestions.

Thoghts on just `protos` ? Or is that missing `encoder` that implies that it has encoding/decoding runtime functions as well? In that case, `proto-coders` (coder is generic across encoder and decoder)?

packages/proto-encoder/README.md
```diff
@@ -0,0 +1,63 @@
...
+```bash
+yarn generate:packages
+```
+
+## Generating Proto Encoders
+
+To generate the proto encoders, you can run the following command:
+
+```bash
+yarn generate:proto-encoders
```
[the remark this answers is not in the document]
JoseRFelix: Oops, I changed the name. Resolved in my new commit 👍

packages/proto-encoder/scripts/get-proto.sh
JoseRFelix: Oh didn't notice that. I'll use this for cosmos, ibc and wasm versions, thanks!

packages/proto-encoder/package.json
JoseRFelix: I also thought about using `protos`, but I feel it does not give a good idea of what the package contains since it also includes encoders and decoders.  I like `proto-coders` sounds catchy. However, what about `proto-codecs`? Codecs is a general term for encoder and decoder. wdyt?

jonator: SGTM

packages/README.md
```diff
@@ -4,7 +4,7 @@
 
 @osmosis-labs/stores - contains Mobx stores for accounts and queries, price data, historical data (imperator), common UI config (creating pool, swapping, managing liquidity) and the GAMM currency registrar
 
-@osmosis-labs/proto-encoders - This package includes generated TypeScript libraries for encoding and decoding blockchain data, such as transactions. It facilitates the conversion of Protobufs into TypeScript libraries for seamless data handling.
+@osmosis-labs/proto-codecss - This package includes generated TypeScript libraries for encoding and decoding blockchain data, such as transactions. It facilitates the conversion of Protobufs into TypeScript libraries for seamless data handling.
```
jonator, line 7:
```suggestion
@osmosis-labs/proto-codecs
```

jonator (changes requested):

PR #1529, opened 2023-04-28
Update feature-flag.ts

staking-hub:


jonator:
@staking-hub we no longer use this file for tracking gauges. Gauges are filtered and returned in an API in `active-gauges-ts`

staking-hub:
Ok

PR #1530, opened 2023-04-28
(CL) MVP route splitting

jonator:
https://app.clickup.com/t/866a13ram


arispech (approved):
working fine on my end.

packages/pools/src/router/__tests__/routes.spec.ts
```diff
@@ -1,412 +1,1137 @@
...
-        }),
-        makeWeightedPool(),
-      ];
-
-      const router = makeDefaultTestRouterParams({
-        pools,
-        incentivizedPoolIds: [],
-        stakeCurrencyMinDenom: "ufoo",
+  describe("getOptimizedRoutesByTokenIn", () => {
+    describe("handles", () => {
```
JoseRFelix, line 16: I believe we should use `describe` sparingly. Nested code tends to be more difficult to read and maintain. It seems to me that `OptimizedRoutes` and handles do not offer much information, so we could possibly remove these `describe` blocks. Perhaps we could incorporate 'handles' directly within `it`.

packages/pools/src/router/__tests__/routes.spec.ts
```diff
@@ -1,412 +1,1137 @@
...
+        const router = makeDefaultTestRouterParams({ pools: [] });
+        const promisedSplit = router.getOptimizedRoutesByTokenIn(
+          { denom: "uion", amount: new Int("100") },
+          "uosmo"
+        );
+        expect(promisedSplit).resolves.toEqual([]);
       });
-
-      const routes = await router.getOptimizedRoutesByTokenIn(
+      it("invalid tokens", () => {
```
JoseRFelix, line 25: I also think that `it` statements should be read as full sentences to understand what the test is doing without having to run it within jest.

packages/pools/src/router/__tests__/routes.spec.ts
```diff
@@ -1,412 +1,1137 @@
...
-        }),
-      ];
-
-      const router = makeDefaultTestRouterParams({
-        pools,
-        incentivizedPoolIds: [],
-        stakeCurrencyMinDenom: "ufoo",
+
+    describe("favors high liquidity", () => {
+      test("splits into similar liquidity pools", async () => {
```
JoseRFelix, line 88: I think we should decide to use either `test` or `it` for consistency. wdyt?

packages/pools/src/router/route.ts
```diff
@@ -13,43 +13,76 @@ export interface Route {
   tokenInDenom: string;
 }
 
-/** Throws if route is invalid */
-export function validateRoute(route: Route) {
+/** Single path through pools, with the initial amount calculated. */
```
JoseRFelix, line 16: Could you expand this comment a little bit more 🙏 , I'm confused as to whether this refers to a method or the interface

packages/stores/src/account/msg/amino.ts
```diff
@@ -0,0 +1,115 @@
+import { Msg } from "@cosmjs/launchpad";
```
JoseRFelix, line 1: We should keep in mind that we'll have to refactor this file before merging CL or multiple wallets. These amino files are no longer required thanks to the `proto-codecs` package

packages/stores/src/account/types.ts
```diff
@@ -48,14 +49,18 @@ export const defaultMsgOpts: OsmosisMsgOpts = {
...
-  },
-  swapExactAmountOut: {
+    gas: 25_0000 * numPools,
+  }),
+  swapExactAmountOut: (numPools: number) => ({
     type: "osmosis/gamm/swap-exact-amount-out",
-    gas: 250000,
-  },
+    gas: 25_0000 * numPools,
+  }),
```
JoseRFelix, line 63: This too

packages/web/hooks/ui-config/use-trade-token-in-config.ts
```diff
@@ -59,119 +61,109 @@ export function useTradeTokenInConfig(
...
-            resetConfigUserInput();
-            return "exact-in";
-          }
-        );
+      const tokenOutMinAmount = config
+        .outAmountLessSlippage(maxSlippage)
+        .toCoin().amount;
+
+      ////////
+      // Send messages to account
```
JoseRFelix, line 121:
```suggestion
      /**
      * Send messages to account
      */
```

packages/web/hooks/use-overridable-state.ts
```diff
@@ -0,0 +1,110 @@
+import { useState } from "react";
```
JoseRFelix, line 1: I'm somewhat hesitant about renaming this hook to overridable, as it doesn't precisely convey its functionality. In my opinion, the term 'overridable' is quite broad; for instance, `useState` could technically be considered overridable as well. I believe keeping it as `controllable` is a more suitable name, given that the concept of controllable/uncontrollable is well recognized within the React ecosystem. This term shows that the hook's behavior can be switched between uncontrolled and controlled modes as required

JoseRFelix (changes requested):
It's looking good! I've added a few comments for consideration

packages/pools/src/router/__tests__/routes.spec.ts
jonator: Generally I agree in code, but in this scenario IMO it helps keep the types of tests organized given how many there are. It creates a nice tree structure that helps surface the more important tests first, and helps categorize the test ("handles" tests are for testing common error cases/defaults). This tree structure is especially powerful in the jest VSCode extension.

jonator: Screenshot 2023-05-15 at 12 57 48 PM

packages/stores/src/account/msg/amino.ts
jonator: Yeah I refactored the swap messages (added one, removed one, refactored 2). I imagine we'll have some nasty conflicts there.

packages/stores/src/account/types.ts
jonator: How does it handle gas?

packages/pools/src/router/__tests__/routes.spec.ts
JoseRFelix: IMO inlining handles like `handles no pools`, `handles invalid tokens`, and `handled not enough liquidity` makes it easier to read for me and also makes me avoid looking consistently up and down. Likewise, if tests fill the list, we would have to constantly scroll up. I guess we could also rely on memory, but I believe inlining it would make it less cumbersome. wdyt?

packages/stores/src/account/types.ts
JoseRFelix: It will be the same object as this one. We'd only have a MessageComposer instead of a type to replace the amino classes. On the other hand, if the dev omits the gas, it's possible to simulate and get an estimated fee

JoseRFelix (approved):
LGTM! 🚀

PR #1531, opened 2023-04-28
update PR template to ask for clickup task link

MaxMillington:
There is no way for GH to automatically have one PR template for open source contributors and another for org members. We can have multiple templates, but you would have to manually tell GH which one to use every time you opened a PR. I think combining them together is less work, so I just added a line about clickup to the current one. I believe the current one was in the wrong folder, which is why it wasn't doing anything. Feedback appreciated if we think anything else should be added. I personally just care about everyone including the clickup link.


.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
jonator, line 1: Could we also add a section for "Key Workflows to Test" for our QA team.

MaxMillington: Do we have a list of key workflows somewhere?

jonator: @arispech keeps one for QA (I believe in Airtable) for her workflow testing.

But rather I was thinking just a bulletpoint list that devs are expected to populate

MaxMillington: Updated

PR #1534, opened 2023-05-01
Hotfix: swap tool / tx differences

jonator:
* When the user clicks "swap" all the pools involved are refreshed, and we re-calculate the swap with updated pool values. However, this can be misleading as, especially with multihop swaps, large pool activity could occur between inputting into the swap tool and clicking swap. Some users have clicked swap, and signed transactions that were, in some more rare cases, quite a bit less favorable. Instead of refreshing, we should maintain the values the users see and let the tx fail on chain instead.

### DIFF
--- a/packages/stores/src/account/index.ts
+++ b/packages/stores/src/account/index.ts
@@ -613,7 +613,6 @@ export class OsmosisAccountImpl {
       "swapExactAmountIn",
       async () => {
         // refresh data and get pools
-        await queries.queryIncentivizedPools.waitFreshResponse();
         const pools: Pool[] = [];
         for (const route of routes) {
           const queryPool = queries.queryGammPools.getPool(route.poolId);
@@ -624,14 +623,7 @@ export class OsmosisAccountImpl {
             );
           }
 
-          await queryPool.waitFreshResponse();
-
-          const pool = queryPool.pool;
-          if (!pool) {
-            throw new Error("Unknown pool");
-          }
-
-          pools.push(pool);
+          pools.push(queryPool.pool);
         }
 
         // make message with estimated min out amounts
@@ -794,11 +786,7 @@ export class OsmosisAccountImpl {
         if (!queryPool) {
           throw new Error(`Pool #${poolId} not found`);
         }
-        await queryPool.waitFreshResponse();
         const pool = queryPool.pool;
-        if (!pool) {
-          throw new Error("Unknown pool");
-        }
 
         // reconcile weighted and stable pool asset data
         const inPoolAsset = queryPool.getPoolAsset(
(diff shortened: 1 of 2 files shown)


JoseRFelix (commented):
The tx `token_out_min_amount` is different from the front-end minimum received. Is this expected?

image

arispech:
@JoseRFelix  As far as I know the `token_out_min_amount` on Keplr singing messages that is created when click the swap button on the UI will not be the same as the`front-end minimum received` as it's constantly updating when the page refreshed. 

I try to screen shot this as fast as possible before the `front-end minimum received` gets updated and the number is quite close
Screenshot 2023-05-01 at 2 03 29 PM

JoseRFelix:
@arispech I thought the purpose of this PR was to use the same data displayed in the FE

jonator:
@JoseRFelix I use the same formula now for the out-amount-less-slippage as is used in the message creation math. That's why it was slightly different. I know it's not the cleanest for this fix, but I've already cleaned up our swap logic quite a bit for the CL release. Also, for CL I'm planning on not re-simulating the swaps in the makeMsg functions and instead just trusting the router and passing those values directly to the message creation. This should reduce the amount of code and thus complexity.

@arispech the minOutAmount value in keplr should be the same and should stay the same when you view the tool and click swap.

MaxMillington:
@jonator Have we run this by @syed8rahman to get his thoughts? 

Looking at this from the outside, it seems like this is just as much a UX/messaging issue as anything else. 

The code as been like this for a long time correct? And this is the first real complaint we have had? If the fix we are applying solves this issue, but potentially could make another issue more prevalent, I would think that we should have a deeper discussion before deploying this. Again, it may just be we need to build a better UX or have more clear messaging. 

Just my two cents from what I can understand so far.

JoseRFelix (approved):
LGTM!

jonator:
> @jonator Have we run this by @syed8rahman to get his thoughts?
> 
> Looking at this from the outside, it seems like this is just as much a UX/messaging issue as anything else.
> 
> The code as been like this for a long time correct? And this is the first real complaint we have had? If the fix we are applying solves this issue, but potentially could make another issue more prevalent, I would think that we should have a deeper discussion before deploying this. Again, it may just be we need to build a better UX or have more clear messaging.
> 
> Just my two cents from what I can understand so far.

We encountered this many months ago as well. It should remain just as rare, except with this instead of the user accidentally signing an unfavorable trade (if they didn't inspect the tx sign popup) the trade just fails. That would result in a less misleading UX that doesn't seem like a bug or we're taking advantage of users. For most retail trades, they should be able to swap within their slippage tolerance.

With the CL release, I'm thinking that I'll refresh the relevant pools more often if the user has >0 input in the swap tool. This should lessen the chance of slippage failures while at the same time having the quotes match in the swap tool and popup.

PR #1537, opened 2023-05-01
Update testnet assets to osmo-test-5

JeremyParish69:
osmo, ion, atom, mars, usdc, juno (+ other axelar tokens).
Update osmosis testnet chain info for osmo-test-5. (rpc, rest, explorer)

### DIFF
--- a/packages/web/config/chain-infos.ts
+++ b/packages/web/config/chain-infos.ts
@@ -17,12 +17,12 @@ const chainInfos = (
       rpc:
         OSMOSIS_RPC_OVERWRITE ??
         (IS_TESTNET
-          ? "https://rpc.testnet.osmosis.zone/"
+          ? "https://rpc.osmotest5.osmosis.zone/"
           : "https://rpc-osmosis.keplr.app/"),
       rest:
         OSMOSIS_REST_OVERWRITE ??
         (IS_TESTNET
-          ? "https://lcd.testnet.osmosis.zone/"
+          ? "https://lcd.osmotest5.osmosis.zone/"
           : "https://lcd-osmosis.keplr.app/"),
       chainId:
         OSMOSIS_CHAIN_ID_OVERWRITE ??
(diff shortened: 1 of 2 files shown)


daniel-farina:
@JeremyParish69 Is this ready to be merged in the meantime. At least one pool shows and USDC as well.

jonator (changes requested):
Hey it seems we have a blend of testnet and mainnet chain infos, which to me is not very readable.

Can we instead separate them into different lists entirely, and we can use a top-level variable to switch to the right list?

We do something similar for ibc-assets, but it would be more readable there too if we can separate those into separate variables as well.

JeremyParish69:
> @JeremyParish69 Is this ready to be merged in the meantime. At least one pool shows and USDC as well.

Not quite, there's been some changes that should be tested first.

jonator (approved):

packages/web/integrations/axelar/transfer.tsx
```diff
@@ -226,7 +226,7 @@ const AxelarTransfer: FunctionComponent<
     };
 
     /** Osmosis chain ID accepted by Axelar APIs. */
-    const osmosisAxelarChainId = isTestNet ? "osmosis-5" : "osmosis";
+    const osmosisAxelarChainId = isTestNet ? "osmosis-6" : "osmosis";
```
jonator, line 229: So osmo test 5's chain ID is `osmosis-6`?

jonator (approved):

packages/web/integrations/axelar/transfer.tsx
JeremyParish69: Only according to Axelar's API, yes
image

jonator: Oh this is Axelar. My bad. Thanks, wasn't paying complete attention.

PR #1543, opened 2023-05-03
(CL) simplify: remove exponentAtPriceOne param, rounding changes

jonator:
Inspired by the changes in these chain PRs, in order:
* https://github.com/osmosis-labs/osmosis/pull/4957 (first commit)
* https://github.com/osmosis-labs/osmosis/pull/5022 (second commit)
* https://github.com/osmosis-labs/osmosis/pull/5019 (no changes on frontend)

TODO:
- [x] quotes.spec.ts: update the `inittedTicks` members of the test cases to the new net tick values (as would be returned from chain query) derived from the new go test pools with a different `exponentAtPriceOne` values

### DIFF
--- a/packages/math/src/rounding.ts
+++ b/packages/math/src/rounding.ts
@@ -1,53 +1,16 @@
-import { Dec } from "@keplr-wallet/unit";
-
-import { BigDec } from "./big-dec";
-
-export function compareBigDec_checkMultErrorTolerance(
-  expected: BigDec,
-  actual: BigDec,
-  tolerance: BigDec,
-  roundingMode: string
-) {
-  let comparison = 0;
-  if (expected.gt(actual)) {
-    comparison = 1;
-  } else {
-    comparison = -1;
+export function checkMultiplicativeErrorTolerance<
+  Decimal extends {
+    gt(other: Decimal): boolean;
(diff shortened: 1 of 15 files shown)


jonator:
@p0mvn 

Add you you as reviewer to scan over the rounding changes:
https://github.com/osmosis-labs/osmosis-frontend/pull/1543/commits/91d83066e9ea1b4b5ae3ec5e7fd3aef1e31ed7eb

packages/math/src/rounding.ts
```diff
@@ -0,0 +1,87 @@
+import { Dec } from "@keplr-wallet/unit";
+
+import { BigDec } from "./big-dec";
+
+export function compareBigDec_checkMultErrorTolerance(
```
JoseRFelix, line 5: What do you think about having it in camelCase instead of a mix of snake case and camel? 

```suggestion
export function compareBigDecCheckMultErrorTolerance(
```

jonator: Why not make it generic to the type of decimal being used?

p0mvn (approved):
LGTM on quick scan

jonator:
I'm going to merge this to unblock Jacky. All tests pass.

PR #1545, opened 2023-05-03
Add ARB

JeremyParish69:
validation successful


packages/web/public/networks/arbitrum.svg
jonator, line 1: Is there a reason we have 2 of these svg files?

JeremyParish69: One for the token in the /tokens/ folder, and one the network in the /networks/ folder. I suppose we might be able to get rid of the /networks/ folder if we just use the corresponding gas token for the network, although I'm not certain whether that may ever need to be different.

JeremyParish69: Are we OK to proceed as is? or do you want it removed first?

jonator: We can leave it given the difference

PR #1549, opened 2023-05-05
feat: Improve Multiple Wallets Bridge Experience

JoseRFelix:


jonator (commented):
#### Axelar Integration Bug
If I:

1. Open the modal with Keplr _disconnected_
2. Disconnect MetaMask in extension
3. I reach this invalid state where it says to reconnect MetaMask but the MetaMask connect button is unreachable since the whole form is disabled

Screenshot 2023-05-09 at 6 42 22 PM

#### UX Concern
It seems strange that I'm prompted with two "Connect wallet" buttons:
Screenshot 2023-05-09 at 6 45 18 PM

jonator:
Also @JoseRFelix how should turbo repo be configured with the code generation step? Shouldn't it included in the output globs list so that when output is replayed, those generated files are first inserted into the source folder for the `next build` process to work?

turbo.json
```diff
@@ -3,7 +3,7 @@
   "pipeline": {
     "build": {
       "outputs": ["build/**", ".next/**", "!.next/cache/**"],
```
jonator, line 5: Thought don't generated files need to be specified in the outputs list so they are zipped into the cache and unzipped to be replayed?

JoseRFelix: I've included the output files in the [`generate`](https://github.com/osmosis-labs/osmosis-frontend/blob/493c199aecd3b139298c5b06aa82fddefe20f592/turbo.json#L15-L23) key. As it executes before the build process, it should retrieve the files stored in the cache

jonator: Ah missed that. Nice 👍

turbo.json
```diff
@@ -3,14 +3,23 @@
...
+    "analyze": {},
+    "generate": {
+      "outputs": [
+        "src/codegen/**",
+        "config/wallet-registry.ts",
+        "config/wallet-assets.ts",
+        "config/chain-infos.ts"
+      ],
+      "dependsOn": ["^generate"]
+    }
```
jonator, line 23: Call me nitpicky, but could we sort these `pipeline` members in rough order of execution? Would prob be a bit more readable for new devs.

packages/proto-codecs/scripts/package-types.ts
```diff
@@ -4,21 +4,20 @@
  */
 export type AvailablePackages =
```
jonator, line 5: If this file is generated, it seems weird to have it in a scripts folder?

IMO

jonator (changes requested):
### **IBC** Transfer - **Withdraw** - UI Bug

1. Open IBC withdraw menu with Keplr connected (or any cosmos wallet too)
2. Click switch wallet under akash address
3. Select a wallet that isn't installed (Leap)
4. Close that modal and see that Keplr (original wallet) is no longer connected, and the modal is in this disabled state: (FYI I see `ERROR - Error: Client Not Exist!
 ` in the console, appearing from cosmos-kit?)

Screenshot 2023-05-10 at 9 40 00 PM

Furthermore, from the above state, if I:
1. Click "Connect wallet" at the bottom
2. The prompt to install the last selected wallet appears (Leap)
3. However when I go back and click an installed wallet (Keplr) the IBC transfer modal closes, and I can't click withdraw on Akash again. Clicking withdraw on another asset however, opens the connect wallet modal from cosmos-kit again.

Perhaps, the UX could be to prompt the user to select another wallet that supports the IBC counterparty, _before_ the transfer modal appears, exactly as we do with EVM wallets for Axelar assets? May be more consistent? Though, being able to switch interchain wallets from the transfer modal as we can with Axelar is convenient, too.

Also, IMO it's a bad UX principle to have multiple wallets layer on top of eachother. This should be a push/pop based modal workflow for selecting wallets, like it it for Axelar.

The sequence, like with Axelar:
Select Wallet (skip if already selected) <-> Transfer Modal

This complexity is currently encapsulated in TransferUIConfig, perhaps we can add this / piggyback there as well for IBC transfers?

https://github.com/osmosis-labs/osmosis-frontend/blob/db299d24e893e7b0b3f5fc7bdac81b3fc6e0c7fc/packages/web/stores/assets/transfer-ui-config.ts#L35

Also _ideally_ we share the same UI for selecting a wallet whether it's an EVM chain wallet or Cosmos wallet. Bonus if you can make them consistent. (The most consistency would be putting EVM wallets in the same UI we use for these cosmos-kit wallets).

### React Issue

It may be caused by next/image?? But I'm getting this in my console in any transfer modal. I've cleaned and rebuilt completely

Screenshot 2023-05-10 at 9 37 02 PM

packages/proto-codecs/scripts/package-types.ts
JoseRFelix: I think it makes sense to have it inside `scripts` since it's intended to be used there. However, I understand the concern about its placement alongside non-generated files. I think we can place it inside a folder called `scripts/codegen`. wdyt?

JoseRFelix:
@jonator An update on this feature, I'm currently waiting for my [PR](https://github.com/cosmology-tech/cosmos-kit/pull/196) in Cosmos-Kit to be merged. It will enable catching errors in a promise when connecting to a wallet, which is required for the push/pop workflow.

JoseRFelix:
@jonator I was able to implement a workaround for now. Cosmos-Kit's team will take more time revising the changes to ensure they are not breaking. This workaround is straightforward and will avoid us from getting blocked. I believe it's ready for a re-review.

packages/web/modals/wallet-select.tsx
```diff
@@ -197,11 +204,19 @@ const ModalContent: FunctionComponent<
           .then(() => {
             setLazyWalletInfo(undefined);
             onConnectProp?.();
+          })
+          .catch(() => {
+            /** DO NOTHING */
```
jonator, line 209: Why do nothing? Couldn't this reveal useful unexpected errors?

turbo.json
```diff
@@ -2,12 +2,21 @@
...
-      "outputs": ["build/**", ".next/**", "!.next/cache/**"],
+      "outputs": [
+        "build/**",
+        ".next/**",
+        "!.next/cache/**",
+        "src/codegen/**",
+        "scripts/generated/**",
+        "config/wallet-registry.ts",
+        "config/wallet-assets.ts",
+        "config/chain-infos.ts"
```
jonator, line 13: Why not put these in a generated folder in config folder? Would also help inform from import paths that it's using generated stuff?

IMO, generated stuff makes things super confusing for new devs.. let's isolate it as much as possible, as we do with integrations folder.

Bonus: we can refer to it in a single blob if everything generated is in a generated folder (including codegen): 

```
["src/generated/**", "src/**/generated/**"]
```

jonator (changes requested):
### Client not Exist! Bug

With Keplr connected:
1. Open IBC deposit modal (ATOM)
2. Click switch wallet
3. Select a wallet that's not installed (it should prompt to install, no?)

The modal showed "initializing" but I noticed Next.js throw an error in the background of the modal reading "Client not Exist!"

Screenshot 2023-05-16 at 4 40 34 PM

### UX

IMO the arrangement of these controls, causing the height to increase, is not good. Maybe we can have a dropdown menu behind the "..." menu found on the pools table rows?

But regardless, the switch wallet button should not be visible if they're editing.
Screenshot 2023-05-16 at 4 41 56 PM


### Bundle size
```
Page                                       Size     First Load JS
┌ ○ /                                      1.48 kB        1.95 MB
├   /_app                                  0 B            1.95 MB
├ ○ /404                                   411 B          1.95 MB
├ ○ /500                                   411 B          1.95 MB
├ λ /api/active-gauges                     0 B            1.95 MB
├ ○ /assets                                7.32 kB        1.96 MB
├ ○ /bootstrap                             1.88 kB        1.95 MB
├ ○ /disclaimer                            2.21 kB        1.95 MB
├ ○ /pixels                                10.4 kB        1.96 MB
├ ○ /pool/[id]                             5.9 kB         1.96 MB
└ ○ /pools                                 26.5 kB        1.98 MB
+ First Load JS shared by all              1.95 MB
  ├ chunks/framework-d124cb1da80cd7e9.js   42.2 kB
  ├ chunks/main-f174e1ca2a2fbe25.js        27 kB
  ├ chunks/pages/_app-c6f5f1f77e8c13b4.js  1.88 MB
  ├ chunks/webpack-7764dc04ba24c27e.js     2.39 kB
  └ css/f6864125b2533403.css               14.2 kB
```

It appears the main bundle has grown by almost .5 kb? Is there anything else you can think of that we could do to reduce this? Any more dynamic loading? **What if we dynamically load the message composers in the account store for the first time a message gets sent as they're all async functions anyways?** If it takes too much refactoring, forget it, but Next.js makes this kind of thing super easy.

Or is this mostly cosmos-kit or proto-codecs? It's not the end of the world, but may become a growing problem like the pools list was.

See:
Screenshot 2023-05-16 at 4 36 21 PM
Screenshot 2023-05-16 at 4 36 52 PM

Also found out this is bigger then keplr-wallet/stores 😱: by far the biggest offender. It's imported by many of the keplr-wallet packages. It seems to contain everything needed to drive the actual wallet extension signing logic + ethereum mnemonic support :(
Screenshot 2023-05-16 at 5 15 12 PM


### Question
https://github.com/osmosis-labs/osmosis-frontend/blob/c051f7d313ee6f7f28d0462fd2f2058c62bb6951/packages/stores/src/account/base.ts#L113
Why `any`?


Getting close!

JoseRFelix:
> ### Client not Exist! Bug
> With Keplr connected:
> 
> 1. Open IBC deposit modal (ATOM)
> 2. Click switch wallet
> 3. Select a wallet that's not installed (it should prompt to install, no?)
> 
> The modal showed "initializing" but I noticed Next.js throw an error in the background of the modal reading "Client not Exist!"
> 
> Screenshot 2023-05-16 at 4 40 34 PM
> ### UX
> IMO the arrangement of these controls, causing the height to increase, is not good. Maybe we can have a dropdown menu behind the "..." menu found on the pools table rows?
> 
> But regardless, the switch wallet button should not be visible if they're editing. Screenshot 2023-05-16 at 4 41 56 PM
> 
> ### Bundle size
> ```
> Page                                       Size     First Load JS
> ┌ ○ /                                      1.48 kB        1.95 MB
> ├   /_app                                  0 B            1.95 MB
> ├ ○ /404                                   411 B          1.95 MB
> ├ ○ /500                                   411 B          1.95 MB
> ├ λ /api/active-gauges                     0 B            1.95 MB
> ├ ○ /assets                                7.32 kB        1.96 MB
> ├ ○ /bootstrap                             1.88 kB        1.95 MB
> ├ ○ /disclaimer                            2.21 kB        1.95 MB
> ├ ○ /pixels                                10.4 kB        1.96 MB
> ├ ○ /pool/[id]                             5.9 kB         1.96 MB
> └ ○ /pools                                 26.5 kB        1.98 MB
> + First Load JS shared by all              1.95 MB
>   ├ chunks/framework-d124cb1da80cd7e9.js   42.2 kB
>   ├ chunks/main-f174e1ca2a2fbe25.js        27 kB
>   ├ chunks/pages/_app-c6f5f1f77e8c13b4.js  1.88 MB
>   ├ chunks/webpack-7764dc04ba24c27e.js     2.39 kB
>   └ css/f6864125b2533403.css               14.2 kB
> ```
> 
> It appears the main bundle has grown by almost .5 kb? Is there anything else you can think of that we could do to reduce this? Any more dynamic loading? **What if we dynamically load the message composers in the account store for the first time a message gets sent as they're all async functions anyways?** If it takes too much refactoring, forget it, but Next.js makes this kind of thing super easy.
> 
> Or is this mostly cosmos-kit or proto-codecs? It's not the end of the world, but may become a growing problem like the pools list was.
> 
> See: Screenshot 2023-05-16 at 4 36 21 PM Screenshot 2023-05-16 at 4 36 52 PM
> 
> Also found out this is bigger then keplr-wallet/stores 😱: by far the biggest offender. It's imported by many of the keplr-wallet packages. It seems to contain everything needed to drive the actual wallet extension signing logic + ethereum mnemonic support :( Screenshot 2023-05-16 at 5 15 12 PM
> 
> ### Question
> https://github.com/osmosis-labs/osmosis-frontend/blob/c051f7d313ee6f7f28d0462fd2f2058c62bb6951/packages/stores/src/account/base.ts#L113
> 
> 
> Why `any`?
> Getting close!

### Client Not Exist

After some investigation, the `client not exist` error is coming from a side effect within Cosmoskit. This error is thrown globally, so it won't cause the application to crash in a production environment, though it will appear in the console. I plan to address this by opening a ticket with the Cosmoskit team. However, this issue should not prevent us from proceeding with the merge.

### UX

I agree. I'll put these options in a dropdown if both need to be visible. 

### Bundle Size

I agree that we don't want our bundle size to grow linearly as we add more features. I believe we could experiment with how to lazily add these messages. However, that will take some effort as I'll have to do a lot of refactoring. Maybe, I could tackle this once I have more bandwidth after merging main/frontier, and finished staking. wdyt?

### Any

I'll take a look 👍

packages/web/public/icons/sprite.svg
```diff
@@ -360,5 +360,23 @@
...
....5C11.9887 10.5 12.2501 10.2386 12.2501 9.91665ZM9.91675 9.91665C9.91675 9.59465 9.65541 9.33332 9.33341 9.33332C9.01141 9.33332 8.75008 9.59465 8.75008 9.91665C8.75008 10.2386 9.01141 10.5 9.33341 10.5C9.65541 10.5 9.91675 10.2386 9.91675 9.91665ZM7.59139 9.90753C7.58897 9.75302 7.52536 9.60578 7.4145 9.49813C7.30364 9.39047 7.1546 9.3312 7.00008 9.33332L4.08341 9.33332L4.08341 7.84764C4.08341 7.52564 3.69987 7.35829 3.46362 7.57762L1.30916 9.57485C1.11958 9.76444 1.11958 10.0723 1.30916 10.2619L3.46362 12.2591C3.69987 12.479 4.08341 12.3117 4.08341 11.9891L4.08341 10.5L7.00008 10.5C7.07813 10.5011 7.15559 10.4864 7.22789 10.457C7.30019 10.4276 7.36585 10.384 7.42099 10.3288C7.47612 10.2735 7.51962 10.2078 7.54889 10.1354C7.57816 10.0631 7.59261 9.98558 7.59139 9.90753ZM5.25008 4.08332C5.25008 3.76132 4.98875 3.49998 4.66675 3.49998C4.34475 3.49998 4.08341 3.76132 4.08341 4.08332C4.08341 4.40532 4.34475 4.66665 4.66675 4.66665C4.98875 4.66665 5.25008 4.40532 5.25008 4.08332ZM2.91675 4.08332C2.91675 3.76132 2.65541 3.49998 2.33341 3.49998C2.01142 3.49998 1.75008 3.76132 1.75008 4.08332C1.75008 4.40532 2.01142 4.66665 2.33341 4.66665C2.65541 4.66665 2.91675 4.40532 2.91675 4.08332Z"
```
jonator, line 377: Small reminder to delete any icons that are here but also in icons folder, to keep things clean

jonator (approved):
Approved, feel free to merge

PR #1554, opened 2023-05-05
Max/app store

MaxMillington:
PR for the [App Store](https://app.clickup.com/t/864duuqb9).

Using getStaticProps here since we don't expect much of the data to change. Included translations and amplitude events. 

Screen Shot 2023-05-19 at 12 12 43 PM


packages/web/components/cards/app-display-card.tsx
```diff
@@ -0,0 +1,90 @@
...
+interface AppDisplayCardProps {
+  title?: string;
+  subtitle?: string;
+  imageUrl?: string;
+  twitterUrl?: string;
+  githubUrl?: string;
+  externalUrl?: string;
+  mediumUrl?: string;
+  index: number;
+}
```
jonator, line 18: This is a nit: but as you may notice across the app, we typically define function components as closures with inline prop type defs, like:

`FunctionComponent<{ prop1: string }>`

FunctionComponent: more readable
Inlining: less code, helps draw attention to actual type defs that are meant to be reused and used as annotation for APIs.

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+        title={featuredApp.title}
+        subtitle={featuredApp.subtitle}
+        imageUrl={featuredApp.hero_image_URL}
+        githubUrl={featuredApp.github_URL}
+        twitterUrl={featuredApp.twitter_URL}
+        externalUrl={featuredApp.external_URL}
+        mediumUrl={featuredApp.medium_URL}
+      />
+      <div className="body2 mb-2 pt-7 pl-6 font-bold text-osmoverse-200">
+        All apps
```
jonator, line 108: Should be translated

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    setSearchValue(value);
+    setFuzzySearchResults(appDataResults);
+  };
+
+  const iterableData = searchValue ? fuzzySearchResults : nonFeaturedApps;
+
+  return (
+    <main className="m-auto max-w-container bg-osmoverse-900 py-3 md:px-3">
+      <div className="flex justify-between pl-6">
+        <div style={{ flexBasis: "55%" }}>
```
jonator, line 83: This style doesn't have to be inlined:
https://tailwindcss.com/docs/flex-basis

packages/web/components/cards/hero-card.tsx
```diff
@@ -0,0 +1,131 @@
...
+          .content {
+            position: relative;
+            z-index: 2;
+          }
+        `}</style>
+      </a>
+    </div>
+  );
+};
+``;
```
jonator, line 131:
```suggestion
```

packages/web/components/cards/hero-card.tsx
```diff
@@ -0,0 +1,131 @@
...
+            left: 0;
+            width: 100%;
+            height: 100%;
+            background: linear-gradient(transparent, rgba(0, 0, 0, 1));
+            z-index: 1;
+          }
+          .content {
+            position: relative;
+            z-index: 2;
+          }
```
jonator, line 125: We should be able to accomplish everything with tailwind, for maintainability/readability/consistency.

We have already used some background images in our tailwind config:
https://github.com/osmosis-labs/osmosis-frontend/blob/115004828f3295d17130a0141688e1a4221783b8/packages/web/tailwind.config.js#L156

For ref:
https://tailwindcss.com/docs/background-image

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  let featuredApp = applications?.find((app) => app.featured === true);
+  const nonFeaturedApps = applications?.filter((app) => !app.featured);
+
+  featuredApp = featuredApp ? featuredApp : nonFeaturedApps[0];
+
+  const options = {
+    keys: ["title"],
+  };
+  const fuse = new Fuse(applications, options);
```
jonator, line 60: Heads up we have a useFilteredData hook in use-filtered-data.ts that will manage all of the state and will maintain an instance of Fuse as long as possible so we can get its caching benefits as long as the data doesn't change. Overkill for the amount of data, but still worth mentioning.

Periodically, I recommend looking in the hooks folder for utils that may be useful.
Also, we import `react-use` that also has some helpful links.

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { useTranslation } from "react-multi-lang";
+
+import { Button } from "~/components/buttons";
+import { HeroCard } from "~/components/cards";
+import AppDisplayCard from "~/components/cards/app-display-card";
+import { SearchBox } from "~/components/input";
+import { EventName } from "~/config";
+import { useAmplitudeAnalytics } from "~/hooks";
+
+type appDataType = {
```
jonator, line 12: Across the board, types are not lowercased.

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    setFuzzySearchResults(appDataResults);
+  };
+
+  const iterableData = searchValue ? fuzzySearchResults : nonFeaturedApps;
+
+  return (
+    <main className="m-auto max-w-container bg-osmoverse-900 py-3 md:px-3">
+      <div className="flex justify-between pl-6">
+        <div style={{ flexBasis: "55%" }}>
+          <h4 className="pb-2 text-h4 font-h4 text-wosmongton-100">
```
jonator, line 84: FYI in globals.css we apply some of these tailwind font/text styles to the h4 tag automatically:
https://github.com/osmosis-labs/osmosis-frontend/blob/115004828f3295d17130a0141688e1a4221783b8/packages/web/styles/globals.css#L33

For the most part, you shouldn't need to apply both the text- and the font- tag as they're available as a single class in the link above. Saves text.

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+      <div className="flex w-full items-center overflow-x-auto rounded-lg bg-osmoverse-1000 px-8 py-6 text-osmoverse-400 2xl:gap-4 xl:gap-3 1.5lg:px-4 md:flex-col md:items-start md:gap-3 md:px-5 md:py-5">
+        <span className="text-xs">{t("store.storeDisclaimer")}</span>
+      </div>
+    </main>
+  );
+};
+
+export default AppStore;
+
+export async function getStaticProps() {
```
jonator, line 150: FYI this is exactly what Next.js SSR rendering via static props was designed for. We get SEO benefits through SSR-generated HTML (with proper a tags, spans, etc.) while still being able to use React to allow the user to filter the page in the client, in this example.

jonator (changes requested):
## Responsiveness
Should be responsive. Given that we have low design bandwidth, and this is simple, I'd say it should slowly compress down to an individual column as the screen shrinks.

Screenshot 2023-05-19 at 3 16 14 PM

It seems like the container is having some problems with a min padding between the sidebar.
Screenshot 2023-05-19 at 3 22 05 PM

---

Extra nit picky since this is your first PR and there's some conventions we have in place to help make everything look the same/familiar.

packages/web/components/cards/hero-card.tsx
MaxMillington: couldn't get the animations to work in tailwind

packages/web/pages/apps/index.tsx
MaxMillington: Good catch

packages/web/components/cards/app-display-card.tsx
```diff
@@ -0,0 +1,100 @@
...
+    return (
+      <>
+        <style jsx>{`
+          .app-display-card:hover .card-image {
+            transform: scale(1.15);
+          }
+          .card-image {
+            background-image: url(${imageUrl});
+            transition: transform 0.3s ease-in-out;
+          }
```
jonator, line 45: Are these still needed? I see you added some tailwind transitinos below. 

Above this `.app-display-card` could also be implemented with tailwind.

packages/web/tailwind.config.js
```diff
@@ -223,6 +226,7 @@ module.exports = {
           5: "4.75rem",
           6: "5rem",
         },
+        "300px": "300px",
```
jonator, line 229: With tailwind, you can almost always. use arbitrary values as needed, inlined: https://tailwindcss.com/docs/padding#arbitrary-values.

Stuff should only be added to the tailwind config if it's reused or expected to be reused.

jonator (changes requested):
Seems like there's still a container issue on smaller screens:
Screenshot 2023-05-19 at 3 22 05 PM

packages/web/components/cards/app-display-card.tsx
MaxMillington: I'm not sure tailwind alone can effect a transform on a child div when hovering over the parent. I had to add this specifically due to the product requirement of the img zooming in when you hover anywhere on the card. 

The extra transform is redundant, but as far as I can tell the transform on `.app-display-card` is needed

MaxMillington:
> Seems like there's still a container issue on smaller screens: Screenshot 2023-05-19 at 3 22 05 PM

Can you be more specific? I'm not seeing the problem here. @jonator

packages/web/components/cards/app-display-card.tsx
jonator: I think the group utility classes are meant to fill this need. But it may be a plugin and may require some changes in the tailwind config. Given how uncommon this use case has been so far, let's leave this as is for now and we can re-enable group tags later as needed.

https://tailwindcss.com/docs/plugins#parent-and-sibling-states

packages/web/components/cards/hero-card.tsx
```diff
@@ -0,0 +1,110 @@
...
+              {mediumUrl && (
+                <IconLink url={mediumUrl} ariaLabel="Medium">
+                  <Icon id="medium" height="16px" width="16px" fill="white" />
+                </IconLink>
+              )}
+              {githubUrl && (
+                <IconLink url={githubUrl} ariaLabel="GitHub">
+                  <Icon id="github" height="16px" width="16px" fill="white" />
+                </IconLink>
+              )}
```
JoseRFelix, line 88: Same as above

packages/web/components/cards/hero-card.tsx
```diff
@@ -0,0 +1,110 @@
...
+            .backgroundImage {
+              background-size: cover;
+              background-position: center;
+              background-repeat: no-repeat;
+              transition: transform 0.5s ease-in-out;
+            }
+            .heroImage:hover .backgroundImage {
+              transform: scale(1.15);
+            }
+          `}</style>
```
JoseRFelix, line 105: These styles could be included within Tailwind. If needed, let's use [arbitrary values](https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values)

packages/web/components/cards/icon-link.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+
+interface IconLinkProps {
```
JoseRFelix, line 3:
```suggestion
import { HTMLProps } from "react"
interface IconLinkProps extends  HTMLProps<HTMLAnchorElement>  {
  url: string;
 }
```

packages/web/components/cards/icon-link.tsx
```diff
@@ -0,0 +1,29 @@
...
+  className?: string;
+  ariaLabel?: string;
+}
+
+export const IconLink: React.FC<IconLinkProps> = ({
+  children,
+  url,
+  className,
+  ariaLabel,
+}) => {
```
JoseRFelix, line 15:
```suggestion
export const IconLink: React.FC<IconLinkProps> = (props) => {
```

packages/web/components/cards/icon-link.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <a
+      href={url}
+      target="_blank"
+      rel="noopener noreferrer"
+      className={className}
+      aria-label={ariaLabel}
+    >
+      {children}
+    </a>
```
JoseRFelix, line 25:
```suggestion
    <a
      target="_blank"
      rel="noopener noreferrer"
      {...props}
   />
```

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,168 @@
+import Fuse from "fuse.js";
```
JoseRFelix, line 1: Instead of nesting the `index` route, we could flatten it by creating `apps.tsx` within `packages/web/pages/apps`. This makes it more discoverable by VS Code file search

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+    applications: App[];
+  };
+};
+
+export const AppStore: React.FC<AppStoreProps> = ({ apps }) => {
+  const [searchValue, setSearchValue] = React.useState<string>("");
+  const [fuzzySearchResults, setFuzzySearchResults] = React.useState<
+    AppDataType[]
+  >([]);
+  const [fuse, setFuse] = React.useState<Fuse<App> | null>(null);
```
JoseRFelix, line 46: Nit: I'd prefer if we used `useState` instead of `React.useState` to keep consistency within the code base.

packages/web/components/cards/app-display-card.tsx
```diff
@@ -0,0 +1,100 @@
...
+        `}</style>
```
JoseRFelix, line 46: This could also be added with tailwind arbitrary values. I'm not a fan of mixing styling solutions. It adds more complexity

packages/web/components/cards/hero-card.tsx
```diff
@@ -0,0 +1,110 @@
...
+  title: string;
+  subtitle: string;
+  imageUrl: string;
+  fallbackImageUrl?: string;
+  label?: string;
+  githubUrl?: string;
+  twitterUrl?: string;
+  externalUrl: string;
+  mediumUrl?: string;
+}> = memo(
```
JoseRFelix, line 20: Is it necessary for us to use memoization here? Sometimes, preemptive optimizations can end up being costlier than not optimizing at all.

JoseRFelix (changes requested):

packages/web/components/cards/app-display-card.tsx
```diff
@@ -0,0 +1,100 @@
...
+                {githubUrl && (
+                  <IconLink url={githubUrl} ariaLabel="GitHub">
+                    <Icon
+                      id="github"
+                      height="14px"
+                      width="14px"
+                      fill="#958FC0"
+                    />
+                  </IconLink>
+                )}
```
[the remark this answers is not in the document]
MaxMillington: Not sure I'm following why this is recommended? The `&&` operator does the same as a truthiness check, i.e. the equivalent of adding `!!` or `Boolean()`. There is even a commonly used eslint rule disallowing it: https://eslint.org/docs/latest/rules/no-extra-boolean-cast

Are we using some style guide that recommends it?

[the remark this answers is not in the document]
JoseRFelix: Javascript has its quirks around truthy and falsy values. It's always a good idea to cast a value if it's expected to be a boolean. You can read more about why [here](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx).

packages/web/pages/apps/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+        <div className="flex flex-col pr-2" style={{ flexBasis: "35%" }}>
+          <h6 className="font-semibold">{t("store.getFeatured")}</h6>
+          <p className="pt-4 text-xs text-osmoverse-200">
+            {t("store.featuredDescription")}
+          </p>
+        </div>
+        <div className="flex w-48 items-center justify-center pl-6">
+          <Button mode="secondary" size="sm" onClick={handleApplyClick}>
+            {t("store.applyButton")} &rarr;
+          </Button>
```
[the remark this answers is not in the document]
MaxMillington: Do we have a ButtonLink? Or do you mean wrap it in an `a` tag?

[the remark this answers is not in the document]
JoseRFelix: We share the button CVA [styles](https://github.com/osmosis-labs/osmosis-frontend/blob/a3627d66598ede67efbb269846ddb6627f990496/packages/web/components/navbar/index.tsx#L385-L396)

packages/web/components/cards/hero-card.tsx
MaxMillington: memoization is never necessary, but when you don't expect the props to change is generally considered a good use case for the `memo` optimization: https://react.dev/reference/react/memo

packages/web/pages/apps/index.tsx
JoseRFelix: Yes, Next will work with both structures. This is just a suggestion to improve VSCode search results. If you search for pools, the editor has trouble finding the `pages` index file that is nested within `pools`

packages/web/components/cards/app-display-card.tsx
MaxMillington: Tailwind doesn't appear to be able to support this use case. I couldn't get the animations to work correctly in its engine.

packages/web/components/cards/hero-card.tsx
JoseRFelix: I don't see where the React docs recommend using `memo` for all possible use cases. As with all performance optimization methods within React, it introduces a small overhead. In some cases, the benefits outweigh the cost. However, I don't believe `memo` will bring significant benefits in this situation.

packages/web/components/cards/icon-link.tsx
```diff
@@ -0,0 +1,21 @@
...
+      rel="noopener noreferrer"
+      className={props.className}
+      aria-label={props.ariaLabel}
+    >
+      {props.children}
+    </a>
+  );
+};
+
+export default IconLink;
```
jonator, line 21: Please remove default export, and inline props type.

jonator (approved):
Approved, with minor nit comment

PR #1565, opened 2023-05-16
feat: Merge Main and Frontier

JoseRFelix:


JoseRFelix:
> It should also filter assets in the assets page

#1576 should handle this

jonator:
@JoseRFelix thoughts on an optional query param to override the setting? Something like `unverified_assets=true` or the like? 

Could be useful for redirecting frontier.osmosis.zone to the frontend with them visible. Thoughts?

JoseRFelix:
> @JoseRFelix thoughts on an optional query param to override the setting? Something like `unverified_assets=true` or the like?
> 
> Could be useful for redirecting frontier.osmosis.zone to the frontend with them visible. Thoughts?

I think it's a good idea! It shouldn't be too hard to implement. Let me get on this

jonator:
> > @JoseRFelix thoughts on an optional query param to override the setting? Something like `unverified_assets=true` or the like?
> > Could be useful for redirecting frontier.osmosis.zone to the frontend with them visible. Thoughts?
> 
> I think it's a good idea! It shouldn't be too hard to implement. Let me get on this

Notice:
https://app.clickup.com/t/863guc4ez

jonator:
On hold for now until we get CL and multiple wallets out

jonator (approved):

jonator:
@JoseRFelix https://app.clickup.com/t/86781jr1w

Left some comments

jonator:
@JoseRFelix hey can you add the banner only on frontier in this PR? Great work!

JoseRFelix:
> @JoseRFelix hey can you add the banner only on frontier in this PR? Great work! I know you removed the env var, but perhaps we can make one last reference to it for the banner while the frontier deployment is up for a week or to.

@jonator This PR effectively removes the commands to build frontier and all its theme values. If we want to do this we have two options:

1. Revert all style changes to still support Frontier and remove it in another PR once the deployment is down. 
2. Branch off master, add the banner, and change the Vercel deployment branch. I believe this one would be the easiest but will become outdated that maybe is not an issue since it's deprecated.  

I would prefer `2.`, but would appreciate to hear your thoughts

jonator:
I think we decided on a 3rd option in slack:
* Redirect fronter.osmosis.zone to app.osmosis.zone, but use a setting query param and a referrer query param (from frontier.osmosis.zone) to display a modal explaining what happened. Then, we can completely remove everything related to frontier, and have the redirect live at the infra layer.

packages/web/components/navbar/index.tsx
```diff
@@ -509,3 +521,70 @@ const ExternalLinkModal: FunctionComponent<
...
+const FrontierMigrationModal: FunctionComponent<
+  ModalBaseProps & { onOpenSettings: () => void }
+> = (props) => {
+  const t = useTranslation();
+
+  return (
+    <ModalBase
+      {...props}
+      className="!max-w-lg bg-[#332133]"
+      title="Introducing Unverified Assets"
```
jonator, line 534: Should we localize this?

jonator (approved):

PR #1568, opened 2023-05-17
Update chain-infos.ts to add IBCX

gnujoow:
related with https://github.com/osmosis-labs/osmosis-frontend/pull/1162


jonator:
@gnujoow is it ready for launch? LMK

JeremyParish69:
I think it still needs to be added in ibc-assets, and defining a pool:spot in prices.ts would be ideal, if we can get a pool set up.

JeremyParish69:
Preview looks good. No errors. Just not sure if this can be added now, or if this should be displayed with the banner and background (which need additional review, and thus could make it have to wait).

PR #1572, opened 2023-05-18
update volatility description after changing strategy

chikeichan:


packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
```diff
@@ -446,6 +446,27 @@ export class ObservableAddConcentratedLiquidityConfig extends TxChainSetter {
...
+      this.tickRange[1].equals(this.moderateTickRange[1]);
+    const isRangeCustom =
+      !isRangeAggressive && !isRangeModerate && !isRangePassive;
+
+    if (isRangePassive) return "passive";
+    if (isRangeModerate) return "moderate";
+    if (isRangeAggressive) return "aggressive";
+    if (isRangeCustom) return "custom";
+    return "custom";
+  }
```
jonator, line 468: This is a perfect example of something that should be in a UI config. 

Thoughts on changing name to `strategyPresets` ? Low prio

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -623,6 +623,27 @@ const StrategySelectorGroup: FunctionComponent<
...
+    descriptionText = t(
+      "addConcentratedLiquidity.volatilityAggressiveDescription"
+    );
+  } else if (volatilityType === "moderate") {
+    descriptionText = t(
+      "addConcentratedLiquidity.volatilityModerateDescription"
+    );
+  } else if (volatilityType === "custom") {
+    descriptionText = t("addConcentratedLiquidity.volatilityCustomDescription");
+  }
```
jonator, line 646: Given how it's used here, what if we change the `volatilityType` to be "passive" | "aggressive" | "moderate" | null

Then we don't do the last check for "custom" and just leave it as the initialized value.

jonator (approved):

PR #1573, opened 2023-05-18
set min and max tick on passive

chikeichan:


packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
```diff
@@ -271,11 +272,13 @@ export class ObservableAddConcentratedLiquidityConfig extends TxChainSetter {
   };
 
   get range(): [Dec, Dec] {
```
jonator, line 274: This can be decorated with `@computed`

Tip: my general rule is anything that isn't a direct access to a observable, or involves any computation at all, should be a `@computed` property

jonator (approved):

PR #1575, opened 2023-05-19
add inlang cli for machine translations

samuelstroschein:
This PR adds the possibility to run `npx inlang machine translate` to translate all missing translations in one go. 

Additional changes: 

- the inlang.config.js has been updated to the new plugin API
- the contributing translations section in the README has been improved


samuelstroschein:
@jonator I leave the auto-translation up to you! Running `npx inlang machine translate` is all it takes :)

jonator (approved):

jonator (commented):
Thoughts on adding a "translate" or "auto-translate" command to the scripts member of the root package.json manifest?

samuelstroschein:
@jonator ah node 16 doesn't have `fetch`. I thought node 16 reaches the end of life this month but that has been postponed. Thus, I'll fix this issue. Give me 10 min. 

> Thoughts on adding a "translate" or "auto-translate" command to the scripts member of the root package.json manifest? 

There seems to be little/no benefit. 

You can abstract that command your own npm script but that opens the (minimal) need for you to explain to others "run this (custom) command". We had a similar debate if we should have a "release" script which would abstract `npx changeset`. But, developers familiar with changeset will run `npx changeset` regardless of a custom `release` script. 

Ultimately, up to you. Worth mentioning: You can embed `npx inlang machine translate` into any other npm script like "release" e.g. `release: npx inlang machine translate && npm run build`.

Uh actually, embedding the machine translate command into the build step is great because: 

- You get a fallback if some translations don't exist
- Without committing machine translations and thereby losing visibility into what should be human-reviewed and what not. Translations would still show as missing, nudging contributors to contribute "proper translations". Machine translations are often out of context. For you as a maintainer, there is no way to know if a machine translation is good or not.

jonator:
IMO it's worth adding to the command because otherwise we rely on devs to have the npx dependency to translate. Otherwise, they can just run `yarn translate`. IMO the scripts section greatly improves dev UX because they don't have to know what tool is being used or is needed to accomplish a thing. They can look in a central place for everything that can be done in a repo, even if it's not documented.

samuelstroschein:
The fetch bug is fixed now (on my machine with Node 16.20) by updating `@inlang/cli` to `0.9.6`. 

> IMO it's worth adding to the command because otherwise we rely on devs to have the npx dependency to translate 

NPM gets auto-installed with Node and NPM auto-installs NPX :D 

> IMO the scripts section greatly improves dev UX because they don't have to know what tool is being used or is needed to accomplish a thing. They can look in a central place for everything that can be done in a repo, even if it's not documented. 

That is a good argument. 

I propose to merge this PR. You can add a script after merger. I would definitely refrain from committing machine translations to the repo. In other words, use the translate command right before the build step ensuring that Osmosis is fully translated but also ensuring that contributors see that translations are missing and that they can contribute. Otherwise, you are running into the risk of having a translated osmosis front-end with bad machine translations.

jonator:
What API are you using. I saw mentioned elsewhere- are you embedding a google API key you are paying for? Wouldn't mind adding our own key in the inlang config

jonator:
Also minor feedback: it says translating to 11 languages, but just lists 3. Not sure why that could be.
Screenshot 2023-05-19 at 9 41 04 AM

packages/web/localizations/fa.json
```diff
@@ -175,7 +175,13 @@
...
     },
     "table": {
-      "poolId": "استخر #{id}"
+      "poolId": "استخر #{id}",
+      "stable": "مبادله پایدار",
+      "weighted": "وزن دار",
+      "internal": "مشوق های داخلی",
+      "external": "مشوق های خارجی",
+      "superfluid": "ابر سیال",
+      "noIncentives": "بدون انگیزه"
```
samuelstroschein, line 184: @jonator just double checking if this was your intention. by committing the machine translations, contributors have a lower incentive to contribute because they don't see that translations are missing. 

From https://github.com/osmosis-labs/osmosis-frontend/pull/1575#issuecomment-1554687897

> I would definitely refrain from committing machine translations to the repo. In other words, use the translate command right before the build step ensuring that Osmosis is fully translated but also ensuring that contributors see that translations are missing and that they can contribute. Otherwise, you are running into the risk of having a translated osmosis front-end with bad machine translations.

jonator: Yeah I'd rather have translations be present and not be perfect, then for them to be missing and the user would see the JSON key path instead.

samuelstroschein:
> What API are you using. I saw mentioned elsewhere- are you embedding a google API key you are paying for? Wouldn't mind adding our own key in the inlang config 

Correct. We use a paid google translate API key. Users exposing secret env variables to inlang is not possible so far (stuff like process.env doesn't work, etc.). Thus, we are willing to pay for the machine translations till we can't stem the bills any longer (will take a few months if not years). 

> Also minor feedback: it says translating to 11 languages, but just lists 3. Not sure why that could be.
 
Opened an issue https://github.com/inlang/inlang/issues/807. Thanks for reporting!

jonator:
Thanks for throwing this together so quickly. 💪

I'm sure Osmosis foundation would be willing to pay a subscription. Do you take USDC 😅 ?

samuelstroschein:
> I'm sure Osmosis foundation would be willing to pay a subscription. Do you take USDC 😅 ? 

Haha no. We'll get to monetization in a year or so

packages/web/localizations/fa.json
samuelstroschein: > Yeah I'd rather have translations be present and not be perfect, then for them to be missing and the user would see the JSON key path instead. 

If you run machine translate right before the build step (in the build script), users would never face missing translations.

jonator (approved):

packages/web/localizations/fa.json
jonator: IMO it should be run separately as a separation of concerns. I'm not against adding pre build steps, but it should be necessary for _every_ build, and should semantically be "building" the app.

It's for the same reasons we don't run tests before every build. Though I _would_ be a fan of running an `inlang test translations` command to do a check of translations to see if it's necessary to run `yarn translate`.

felixhaeberle:
@jonator We introduced a new "--force" flag which needs to be passed now in order to not get a warning when machine translating. 

This is due to improving the translation quality and warn people about not translating everything automatically because this could lead to out-of-context translations. In a build step however, this is fine as a fallback to have "something in place".

PR #1576, opened 2023-05-20
feat: Display Unverified Assets According to Toggle State

JoseRFelix:


packages/web/stores/derived-data/pools/verified.ts
```diff
@@ -22,11 +22,11 @@ export class ObservableVerifiedPoolsStore
   ) {}
 
   getPool(id: string): ObservableQueryPool | undefined {
-    return this._pools.get(id);
+    return this._allPools.get(id) ?? this._verifiedPools.get(id);
   }
 
   poolExists(id: string) {
-    return this._pools.has(id);
+    return this._allPools.has(id) ?? this._verifiedPools.has(id);
```
jonator, line 29: Nit: I'd prefer `||` for boolean types

packages/web/stores/derived-data/pools/verified.ts
```diff
@@ -46,13 +46,12 @@ export class ObservableVerifiedPoolsStore
```
jonator, line 46: Note, the computedFn a few lines above can't have optional params

jonator (approved):
LGTM

PR #1579, opened 2023-05-20
Add ETH LSTs

JohnnyWyles:
@JeremyParish69 for review thanks, will be adding pools on the staging link here


JeremyParish69:
II added in spot pricing. Looks good to me now @JohnnyWyles

PR #1580, opened 2023-05-21
Add stUmee to Frontend

shellvish:
stUmee is live on the Stride backend, this adds it to the Osmosis frontend.

Please let me know if there's anything else I should add to this PR!


packages/web/config/chain-infos.ts
```diff
@@ -2901,6 +2901,13 @@ const chainInfos = IS_TESTNET
               coinGeckoId: "pool:staevmos",
               coinImageUrl: "/tokens/stevmos.svg",
             },
+            {
+              coinDenom: "stUMEE",
+              coinMinimalDenom: "stuumee",
+              coinDecimals: 6,
+              coinGeckoId: "pool:stuumee",
```
JeremyParish69, line 2908: Thanks for this. The last thing required is defining this price for 'pool:stuumee' in the prices.ts file. If you don't know how, I don't mind updating to your branch if you can let me know the pool id.

jonator:
Draft- waiting for pool

packages/web/config/chain-infos.ts
JeremyParish69: Now that we know pool 1035, can we please define the price (named 'pool:stuumee') in the config/prices.ts file?

shellvish:
Apologies for the delay here, just added the pool!

PR #1582, opened 2023-05-22
refactor chain infos to check type correctly

delivan:
when assigning variable, it needs to be with a type declaration, not a type assertion for the type checking correctly


packages/web/config/utils.ts
```diff
@@ -1,23 +1,23 @@
...
-> & {
+export interface SimplifiedChainInfo
+  extends Omit<ChainInfoWithExplorer, "stakeCurrency" | "feeCurrencies"> {
   currencies: Array<
     AppCurrency &
+      FeeCurrency &
       PeggedCurrency & {
         isStakeCurrency?: boolean;
         isFeeCurrency?: boolean;
+        pegMechanism?: "algorithmic" | "collateralized" | "hybrid";
```
jonator, line 17: `pegMechanism` be found in `PeggedCurrency`? `pegMechanism` is supposed to be on the `originCurrency`.

jonator (approved):

PR #1586, opened 2023-05-23
Publish Stage

jonator:
Minor release so far

Main changes
* Fix asset table sort
* Add type checking to config/chain-infos.ts


arispech (approved):
Front-end*(assets page) works as expected

PR #1592, opened 2023-05-24
fix popup bug and height bug

MaxMillington:
Fixing the bugs I pushed 🙃 

The popup blocker was being triggered because we were essentially opening the cosmos eco form in two places, once via the link, and once via `window.open`. I had forgotten to remove `window.open` when a reviewer asked to have the button wrapped in an `<a>` tag. 

@jonator also removed the default export per your request


jonator:
Thanks for updating the icon component

jonator (approved):

PR #1597, opened 2023-05-25
fix gradient. Add pill. Add mounting event. Address aspect ratio.

MaxMillington:
Addressing some QA fixes. Including adding a 'New' pill to the side bar. Doing what I could with aspect ratio. Adding an amplitude event, and darkening the gradient. 

Screen Shot 2023-05-25 at 2 32 00 PM


packages/web/components/cards/hero-card.tsx
```diff
@@ -95,6 +95,13 @@ export const HeroCard: React.FunctionComponent<{
           .heroImage:hover .backgroundImage {
             transform: scale(1.15);
           }
+          .gradient {
```
MaxMillington, line 98: It didn't work correctly with the gradient in tailwind. If anyone knows the correct tailwind classes I'm all ears.

packages/web/hooks/use-amplitude-analytics.ts
```diff
@@ -1,10 +1,11 @@
-import { useCallback, useEffect } from "react";
 import {
-  init as amplitudeInit,
-  identify,
   Identify,
+  identify,
```
MaxMillington, line 3: just alphabetizing

packages/web/pages/apps/index.tsx
```diff
@@ -80,6 +86,25 @@ export const AppStore: React.FC<AppStoreProps> = ({ apps }) => {
 
   const iterableData = searchValue ? fuzzySearchResults : nonFeaturedApps;
 
+  const getSize = useCallback(() => {
```
MaxMillington, line 89: This handles making sure the search bar isn't too big

packages/web/pages/apps/index.tsx
```diff
@@ -47,15 +48,20 @@ export const AppStore: React.FC<AppStoreProps> = ({ apps }) => {
...
   useEffect(() => {
     const options = {
       keys: ["title"],
     };
     setFuse(new Fuse(applications, options));
   }, [applications]);
 
-  const t = useTranslation();
-  const { logEvent } = useAmplitudeAnalytics();
+  useEffect(() => {
```
MaxMillington, line 62: Log event is memoized so this should still only run once

packages/web/components/cards/hero-card.tsx
jonator: They're added as extensions to the config:

https://github.com/osmosis-labs/osmosis-frontend/blob/7ffa25d8ff0bd2c738a9a75914e69493392529a0/packages/web/tailwind.config.js#L167-L175

packages/web/components/indicators/pill.tsx
```diff
@@ -0,0 +1,7 @@
+import React, { FC } from "react";
+
+export const Pill: FC<{ children: React.ReactNode }> = ({ children }) => (
+  <div className="inline-block flex w-16 items-center justify-center rounded-full bg-gradient-positive px-2 py-1 uppercase text-black">
+    {children}
+  </div>
```
jonator, line 6: This pill doesn't appear to be the same size as figma.

I think the y padding is 2 px too large on top & bottom.

packages/web/components/main-menu.tsx
```diff
@@ -82,17 +94,27 @@ export const MainMenu: FunctionComponent<{
...
                           !selected,
+                        "w-full": isNew,
                       }
                     )}
                   >
-                    {label}
-                  </p>
+                    {isNew ? (
+                      <div className="flex flex-row items-center justify-between">
+                        {label}{" "}
```
jonator, line 109: I prefer a flex gap vs a space character

packages/web/components/main-menu.tsx
```diff
@@ -82,17 +94,27 @@ export const MainMenu: FunctionComponent<{
...
+                        <Pill>
+                          <span>{t("new")}</span>
```
jonator, line 111: Font is "button", just add "button" class

packages/web/hooks/use-amplitude-analytics.ts
jonator: Yes we have a lint that should consistently sort the order of the import statements and the order of the terms within each.

packages/web/components/indicators/pill.tsx
```diff
@@ -0,0 +1,7 @@
+import React, { FC } from "react";
+
+export const Pill: FC<{ children: React.ReactNode }> = ({ children }) => (
+  <div className="inline-block flex w-16 items-center justify-center rounded-full bg-gradient-positive px-2 py-1 uppercase text-black">
```
jonator, line 4: I'm getting a lint error that:

`'inline-block' applies the same CSS properties as 'flex'.(cssConflict)`

Do you have prettier and/or eslint extension in VSCode?

packages/web/pages/apps/index.tsx
jonator: Actually on mount events should be passed as params to `useAmplitudeAnalytics`:

https://github.com/osmosis-labs/osmosis-frontend/blob/7ffa25d8ff0bd2c738a9a75914e69493392529a0/packages/web/hooks/use-amplitude-analytics.ts#L13

jonator (changes requested):
Few things I noticed

packages/web/components/indicators/pill.tsx
MaxMillington: lint runs as a pre-commit hook, so it must be broken

packages/web/pages/apps/index.tsx
jonator: https://github.com/osmosis-labs/osmosis-frontend/blob/2af65d6456e7baca144f70bc2562309e4ac7616b/packages/web/components/types.ts#L51

We have these breakpoints, which align with our tailwind breakpoints. Using those allows all breakpoints to occur at the same point (from CSS and JS), instead of having them confusingly occur at different points given these hardcoded values.

Generally across this entire review, the Apps page is fully isolated so really anything works. But in case other devs come here for reference, we should follow established conventions.

packages/web/pages/apps/index.tsx
```diff
@@ -80,6 +84,25 @@ export const AppStore: React.FC<AppStoreProps> = ({ apps }) => {
...
+      return "small";
+    } else if (width <= 900) {
+      return "medium";
+    } else if (width <= 1500) {
+      return "long";
+    }
+    return "large";
+  }, [width]);
+
+  const [searchBoxSize, setSearchBoxSize] = useState<
```
jonator, line 98: This is not a functional use of state and useEffect. Not idiomatic React. 

We should simply have the size be a function of the current width for each render. (What's in the useCallback).

jonator (changes requested):
A nit and a comment on the use of useState and useCallback

jonator (changes requested):
I know you didn't write this, but it seems like our handling of the banner on mobile is not prepending the page properly. Please see what you can do.
Screenshot 2023-05-26 at 11 19 10 AM

jonator (approved):

PR #1602, opened 2023-05-26
(CL) Concentrated liquidity feature flags

jonator:
* Filter CL pools from router
* Redirect pool detail page to `pools/` if CL pool (same behavior if pool is not found)

We'll be ready to use the useFlags hook for the other CL items as they are added.


MaxMillington (approved):
lgtm

PR #1609, opened 2023-05-30
Final app store fixes

MaxMillington:
Add hover state to social links. 

Change margin based on feedback
Screen Shot 2023-05-30 at 2 09 48 PM
Screen Shot 2023-05-30 at 2 10 12 PM


packages/web/components/cards/app-display-card.tsx
```diff
@@ -46,7 +46,11 @@ export const AppDisplayCard: FunctionComponent<{
       <div className="cursor-pointer" onClick={handleAppClicked}>
         <div className="app-display-card bg-white h-[280px] overflow-hidden rounded-2xl bg-osmoverse-800 shadow-md lg:h-[300px]">
           <div className="overflow-hidden">
-            <div className="card-image inset-0 h-40 overflow-hidden rounded-2xl bg-cover bg-center transition-transform duration-300 ease-in"></div>
+            <img
```
jonator, line 49: Is there a reason we're not using Next.js `Image` components? 

https://nextjs.org/docs/app/building-your-application/optimizing/images

MaxMillington: @jonator Yes. It's because we would have to add `raw.githubusercontent.com` (since all our images are hosted here) to our next config in order to get it to work. This means that if we ever changed where we hosted our images (e.g. not host them in GH but some other CDN and just keep the link in GH), we would have to make a config change.

jonator (approved):

PR #1611, opened 2023-05-31
(CL) Intro UI

jonator:
* Shows intro on every page once (state save in local storage, clear cache to reset or open in incognito)
* Links to instagram reel with Lottie Images

[Funnels UI Figma](https://www.figma.com/file/RxNOvfvzmgtsgCPUetQvcG/Supercharged-Liquidity?type=design&node-id=1173-38015&t=fJsSHIoi3BU9Sryb-4)
[Learn UI Figma](https://www.figma.com/file/RxNOvfvzmgtsgCPUetQvcG/Supercharged-Liquidity?type=design&node-id=1173-47297&t=fJsSHIoi3BU9Sryb-4)


packages/web/components/stepper/index.tsx
```diff
@@ -0,0 +1,323 @@
...
+ *  <Step>
+ *   <p>test 2</p>
+ *  </Step>
+ *  <StepsIndicator />
+ * </Stepper>
+ */
+const Stepper: FunctionComponent<StepsProps> = (props) => {
+  const { children, autoplay } = props;
+
+  const timerTimeInMs = useRef(0);
```
MaxMillington, line 222: Why use `useRef` instead of `useState` here? Don't you want a re-render to occur when the value changes?

packages/web/components/stepper/index.tsx
```diff
@@ -0,0 +1,323 @@
...
+  const stepsContext = useSteps({ count: stepElements.length });
+  const [isHovering, setIsHovering] = useState(false);
+
+  const isStopped = useMemo(
+    () => (autoplay?.stopOnHover && isHovering) || Boolean(autoplay?.isStopped),
+    [autoplay?.isStopped, autoplay?.stopOnHover, isHovering]
+  );
+
+  useEffect(() => {
+    if (autoplay && Boolean(autoplay?.delayInMs)) {
```
MaxMillington, line 243: This makes sense to me, but we could also use a setTimeout which might be less complex than using a counter.

Something like: 

```
useEffect(() => {
  if (autoplay && !isStopped) {
    // Schedule the next step to occur after the specified delay.
    const timeoutId = setTimeout(stepsContext.nextStep, autoplay.delayInMs);

    // The cleanup function cancels the timeout, if it hasn't already fired.
    return () => clearTimeout(timeoutId);
  }
}, [autoplay, stepsContext, isStopped]);
```

Just a suggestion

MaxMillington:
@jonator can you point to specific sections you want reviewed? This is obv a huge PR...

packages/web/components/animation/dynamic-lottie-animation.tsx
```diff
@@ -0,0 +1,25 @@
...
+> = (props) => {
+  // JSON contents
+  const [lottie, setLottie] = useState<any | null>(null);
+
+  // dynamic load JSON animation data
+  useEffect(() => {
+    if (!lottie) {
+      props.futureLottieContents.then(setLottie);
+    }
+  }, [lottie, props.futureLottieContents]);
```
MaxMillington, line 22: I think if you remove the `lottie`, this effect will run less often (small efficiency bump)

packages/web/components/animation/dynamic-lottie-animation.tsx
```diff
@@ -0,0 +1,25 @@
...
+    futureLottieContents: Promise<any>;
+  } & Omit<ComponentProps<typeof Lottie>, "animationData">
+> = (props) => {
+  // JSON contents
+  const [lottie, setLottie] = useState<any | null>(null);
+
+  // dynamic load JSON animation data
+  useEffect(() => {
+    if (!lottie) {
+      props.futureLottieContents.then(setLottie);
```
MaxMillington, line 20: is it possible that there would be an error here?

packages/web/components/animation/dynamic-lottie-animation.tsx
```diff
@@ -0,0 +1,25 @@
...
+    }
+  }, [lottie, props.futureLottieContents]);
+
+  return <Lottie {...props} animationData={lottie}></Lottie>;
```
MaxMillington, line 24: what happens if `null` is passed to `animationData`? `lottie` could be `null`

packages/web/components/funnels/concentrated-liquidity/learn-more.tsx
```diff
@@ -0,0 +1,79 @@
...
+const InfoSlide: FunctionComponent<
+  {
+    bodyText: string;
+  } & ComponentProps<typeof DynamicLottieAnimation>
+> = ({ bodyText, futureLottieContents }) => {
+  return (
+    <section>
+      <p className="body2 h-[50px]">{bodyText}</p>
+      <DynamicLottieAnimation
+        className="!h-[340px]"
```
MaxMillington, line 74: Why is the `!` needed?

packages/web/components/layouts/main.tsx
```diff
@@ -50,6 +56,12 @@ export const MainLayout: FunctionComponent<{
         title={selectedMenuItem?.label ?? ""}
         menus={menus}
       />
+      {featureFlags.concentratedLiquidity && (
```
MaxMillington, line 59: 👍

packages/web/components/funnels/concentrated-liquidity/supercharge-dai-osmo-pool.tsx
```diff
@@ -0,0 +1,84 @@
...
+  title,
+  caption,
+  primaryCta,
+  secondaryCta,
+  onCtaClick,
+  onSecondaryClick,
+}) => {
+  const { isMobile } = useWindowSize();
+
+  const Buttons = () => (
```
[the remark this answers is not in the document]
jonator: The reason I placed them here is just semantics, but since they share so many of the parent props and variables, I just decided to inline it in the body and still avoid repeating the markup below where a different structure is used for mobile.

If you still disagree, I'm down to extract it in the same file and pass it it's own props.

packages/web/components/funnels/concentrated-liquidity/learn-more.tsx
jonator: It's equivalent to `!important`. But it appears to not be needed here, removed.

packages/web/components/animation/dynamic-lottie-animation.tsx
```diff
@@ -0,0 +1,25 @@
...
+/** Loads the lottie libarary component, and it's
+ *  animation data dynamically (outside of the main bundle). */
+export const DynamicLottieAnimation: FunctionComponent<
+  {
+    /** JSON file source, likely a dynamic import: `import("./lottie.json")`. */
+    futureLottieContents: Promise<any>;
+  } & Omit<ComponentProps<typeof Lottie>, "animationData">
+> = (props) => {
+  // JSON contents
+  const [lottie, setLottie] = useState<any | null>(null);
```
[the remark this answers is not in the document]
jonator: It appears the `animationData` prop is `unkown`.

packages/web/components/animation/dynamic-lottie-animation.tsx
jonator: I hardcoded it to test it out. It appears to fail gracefully and display nothing. Do you have an idea of how this should be handled?

packages/web/components/animation/dynamic-lottie-animation.tsx
jonator: There may be an error. Do you have an idea of how it should be handled? Errors are not expected and therefore can bubble up as an uncaught exception (in promise).

packages/web/components/animation/dynamic-lottie-animation.tsx
jonator: OK I removed it, but also disabled the lint warning for exhaustive deps.

packages/web/components/stepper/index.tsx
jonator: @JoseRFelix can you address this please.

packages/web/components/stepper/index.tsx
jonator: @JoseRFelix can you address this please.

packages/web/components/animation/dynamic-lottie-animation.tsx
MaxMillington: Good. I guess `Lottie` takes care of that for us. I was just going to suggest something like `{lottie} && <Lottie />` just to make sure we don't accidentally blow anything up.

packages/web/components/funnels/concentrated-liquidity/supercharge-dai-osmo-pool.tsx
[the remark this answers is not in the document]
MaxMillington: It's definitely more performant to define them outside of the component. In general, there are cases where it's not necessary to optimize for performance and it's worth the hit if it helps devEx, but I'm not sure that's a case here ( I'm not sure how often this rerenders)? I think it makes sense to define outside the component so we don't have to redefine on every render.

packages/web/components/stepper/index.tsx
JoseRFelix: We don't need to. This value is within the interval to maintain progress even when the timer is paused or stopped. Adding a `useState` would cause an excess of re-renders

PR #1619, opened 2023-06-01
Jose/update amplitude events

MaxMillington:


packages/web/stores/root.ts
```diff
@@ -13,6 +13,7 @@ import { KeplrWalletConnectV1 } from "@keplr-wallet/wc-client";
 import {
   ChainInfoWithExplorer,
   ChainStore,
+  DerivedDataStore,
```
MaxMillington, line 16: @jonator @JoseRFelix This change causes a type error. Calling `derivedDataStore.poolsWithMetrics` in another file we get something like "poolsWithMetrics does not exists on type." If we keep the import the same, it works fine without the type error. Both imports ultimately refer to the same file. Can anyone think of a reason for the TS failure?

MaxMillington: So what do you think makes the most sense here? To add the type to the stores? Or to keep this import in the web package?

MaxMillington:
Closing this as it's not a priority

PR #1622, opened 2023-06-02
Add error boundary

mattupham:
## Purpose? (Required)

Sometimes the screen would go completely white when an error was thrown, add a [React Error Boundary ](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) to handle uncaught exceptions

## What was modified? (Required)

- Added an error boundary
- Added an error fallback component
- Added bug / support links
- Added i18n

## How was this tested? (Required)

- Manually threw an error within a child component

## Screenshots (Optional)

Screenshot 2023-06-01 at 17 30 58
Screenshot 2023-06-01 at 17 30 50

## Notes for Reviewers (Optional)

- Sentry will be added in a follow up PR

### DIFF
--- a/packages/web/components/error/error-boundary.tsx
+++ b/packages/web/components/error/error-boundary.tsx
@@ -24,7 +24,7 @@ class ErrorBoundary extends React.Component<
 
   componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
     console.log({ error, errorInfo });
-    // TODO log error to Amplitude in follow up PR
+    // TODO log error to Sentry in follow up PR
   }
 
   render() {


daniel-farina (approved):
lgtm, is it possible to test an error message from the previews?

packages/web/components/error/error-boundary.tsx
```diff
@@ -0,0 +1,39 @@
...
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError(_: Error): ErrorBoundaryState {
+    return { hasError: true };
+  }
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
+    console.log({ error, errorInfo });
+    // TODO log error to Amplitude in follow up PR
```
MaxMillington, line 27: and sentry

packages/web/components/error/error-boundary.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface ErrorBoundaryProps {
+  children: ReactNode;
+  fallback: ReactNode;
+}
+
+interface ErrorBoundaryState {
+  hasError: boolean;
+}
+
+class ErrorBoundary extends React.Component<
```
MaxMillington, line 12: Can we turn this into a function component with hooks instead?

packages/web/components/error/error-boundary.tsx
```diff
@@ -0,0 +1,39 @@
...
+  ErrorBoundaryProps,
+  ErrorBoundaryState
+> {
+  constructor(props: ErrorBoundaryProps) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError(_: Error): ErrorBoundaryState {
```
MaxMillington, line 21: side note but it's classic 'react team' MO that they make this huge deal about hooks and then don't build an equivalent for this

MaxMillington (commented):
Looks good to me. We just need to add sentry. Getting everything in place for that today

packages/web/pages/_app.tsx
```diff
@@ -147,7 +150,9 @@ function MyApp({ Component, pageProps }: AppProps) {
           transition={Bounce}
         />
         <MainLayout menus={menus}>
-          {Component && <Component {...pageProps} />}
+          <ErrorBoundary fallback={ErrorFallback}>
+            {Component && <Component {...pageProps} />}
+          </ErrorBoundary>
```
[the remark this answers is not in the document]
MaxMillington: I actually think this is better anyway. Theoretically this wont crash the main layout then (of course we don't have anything wrapping the main layout though)

MaxMillington (approved):
I think that this is probably ok to merge and then as a followup we can add sentry and change the styling of the error screen. Right now, this will prevent the screen from going white, which is a win

packages/web/components/error/error-boundary.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface ErrorBoundaryProps {
+  children: ReactNode;
+  fallback: ReactNode;
+}
+
+interface ErrorBoundaryState {
+  hasError: boolean;
+}
+
+class ErrorBoundary extends React.Component<
```
jonator, line 12: Is there a rationale for using a class component? IMO we should only us FunctionComponent.

packages/web/localizations/es.json
```diff
@@ -461,5 +461,10 @@
   "autonomy": {
     "description": "Explora Órdenes de Límite y Stop presentadas por Autonomy: ",
     "link": "osmosis.autoswap.trade"
-  }
-}
+  },
+  "errors.fallbackText1": "Ocurrió un error",
```
jonator, line 465: Oh interesting. Seems like inlang doesn't know how to embed a child member to the JSON object

jonator: I reported this FYI

MaxMillington:
@mattupham in the future let's make sure to only merge to the 'stage' branch unless it's a hotfix. Especially on a friday

PR #1629, opened 2023-06-03
fix: asset0/asset1 calculation

czarcas7ic:
Closes: #XXX

Fixes:
- Inputting asset0 properly calculates asset1
- Percent values for asset0 and asset1

Still broken:
- Off by 1 when inputting asset1
 - still looking into this but pushing what I have so far

### DIFF
--- a/packages/math/src/pool/concentrated/math.ts
+++ b/packages/math/src/pool/concentrated/math.ts
@@ -296,43 +296,84 @@ export function convertTokenInGivenOutToTokenOutGivenIn(
 // add liquidity
 // docs ref: https://github.com/osmosis-labs/osmosis/blob/b764323ce7702185d2089b9e76a0115c7058f37e/x/concentrated-liquidity/README.md#L573
 
+// calcAmount0
 export function calculateDepositAmountForQuote(
-  curSqrtPrice: Dec,
-  lowerTick: Int,
   upperTick: Int,
-  baseDeposit: Int
-): Int {
-  const sqrtPu = tickToSqrtPrice(upperTick);
-  const sqrtPl = tickToSqrtPrice(lowerTick);
-  const sqrtPc = curSqrtPrice;
+  currentSqrtPrice: Dec,
+  baseDeposit: Dec
+): Dec {
+  const upperTickSqrt = tickToSqrtPrice(upperTick);
 
-  // calculate liquidity needed for token0
-  const L = new Dec(baseDeposit)
-    .mul(sqrtPu)
-    .mul(sqrtPl)
-    .quo(sqrtPu.sub(sqrtPl));
+  let sqrtPriceA = currentSqrtPrice;
+  let sqrtPriceB = upperTickSqrt;
 
-  // calculate delta y
-  const deltaY = L.mul(sqrtPc.sub(sqrtPl));
+  if (sqrtPriceA.gt(sqrtPriceB)) {
+    sqrtPriceA = upperTickSqrt;
+    sqrtPriceB = currentSqrtPrice;
+  }
 
-  return deltaY.truncate();
+  const numerator = baseDeposit.mul(sqrtPriceB.sub(sqrtPriceA));
+  const denominator = sqrtPriceB.mul(sqrtPriceA);
+
+  return numerator.quo(denominator);
 }
 
+// calcAmount1
 export function calculateDepositAmountForBase(
-  curSqrtPrice: Dec,
   lowerTick: Int,
-  upperTick: Int,
-  quoteDeposit: Int
-): Int {
-  const sqrtPu = tickToSqrtPrice(upperTick);
-  const sqrtPl = tickToSqrtPrice(lowerTick);
-  const sqrtPc = curSqrtPrice;
+  currentSqrtPrice: Dec,
+  quoteDeposit: Dec
+): Dec {
+  const lowerTickSqrt = tickToSqrtPrice(lowerTick);
 
-  // calculate liquidity needed for token1
-  const L = new Dec(quoteDeposit).quo(sqrtPu.sub(sqrtPl));
+  let sqrtPriceA = currentSqrtPrice;
+  let sqrtPriceB = lowerTickSqrt;
 
-  // calculate delta x
-  const deltaX = L.mul(sqrtPu.sub(sqrtPc)).quo(sqrtPu.mul(sqrtPc));
+  if (sqrtPriceA.gt(sqrtPriceB)) {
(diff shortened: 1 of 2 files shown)


packages/math/src/pool/concentrated/math.ts
```diff
@@ -296,43 +296,84 @@ export function convertTokenInGivenOutToTokenOutGivenIn(
...
-    .mul(sqrtPl)
-    .quo(sqrtPu.sub(sqrtPl));
-
-  // calculate delta y
-  const deltaY = L.mul(sqrtPc.sub(sqrtPl));
-
-  return deltaY.truncate();
+  currentSqrtPrice: Dec,
+  baseDeposit: Dec
+): Dec {
```
jonator, line 304: Could we try to use Dec for all amounts in the function signature (here: baseDeposit, and the return value)

It's fine to convert to dec then truncate to convert back. It's just standard to calculate on amounts with Ints.

czarcas7ic: Did you mean try to use Ints for all amounts?

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
```diff
@@ -136,28 +137,46 @@ export class ObservableAddConcentratedLiquidityConfig extends TxChainSetter {
...
+      if (sqrtPriceA.gt(sqrtPriceB)) {
+        sqrtPriceA = upperTickSqrt;
+        sqrtPriceB = this.pool.currentSqrtPrice;
+      }
+
+      const liquidity0 = amount1
+        .toDec()
+        .mul(sqrtPriceA.mul(sqrtPriceB))
+        .quo(sqrtPriceB.sub(sqrtPriceA));
+
```
jonator, line 169: Could we try to move this liquidity0 (and liquidity1 calc below) into the body of the math functions?

jonator (changes requested):

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
czarcas7ic: Agreed, done here https://github.com/osmosis-labs/osmosis-frontend/pull/1629/commits/b7a16a9266f79f95ae71a5192c7f72eb4fa38ab0

packages/math/src/pool/concentrated/math.ts
czarcas7ic: Done here https://github.com/osmosis-labs/osmosis-frontend/pull/1629/commits/3bcfbff73aca5e0e02304b253bbbe44d3132d21f

It also fixed the 49.999 percent issue!

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
```diff
@@ -275,20 +295,26 @@ export class ObservableAddConcentratedLiquidityConfig extends TxChainSetter {
     if (this.baseDepositOnly) return [new RatePretty(1), new RatePretty(0)];
     if (this.quoteDepositOnly) return [new RatePretty(0), new RatePretty(1)];
 
-    const quoteDeposit = new Int(1);
+    let amount1 = new Int(1000000);
```
jonator, line 298: This appears to be assuming the amount has 6 decimals. Would that matter if say, amount1 is ETH with 18 decimals?

jonator (approved):

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
```diff
@@ -53,22 +53,28 @@ export function useAddConcentratedLiquidityConfig(
...
+            currency: config.quoteDepositAmountIn.sendCurrency,
+            amount: config.quoteDepositAmountIn.amount,
+          };
+        }
+        if (!config.baseDepositOnly) {
+          baseDepositValue = {
+            currency: config.baseDepositAmountIn.sendCurrency,
+            amount: config.baseDepositAmountIn.amount,
+          };
+        }
```
jonator, line 70: Nit: all of this code could be distilled into 2 ternary expressions... 

```
config.baseDepositOnly ? undefined : {
  currency: config.baseDepositAmountIn.sendCurrency,
  amount: config.baseDepositAmountIn.amount,
}
```

But not worth blocking this PR. Just FYI

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
jonator: If we want the 1 OSMO (with decimals, not uosmo), we could use something like `DecUtils.getTenExponentNInPrecisionRange(numDecimals)` which would yield `1000000` for OSMO.

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
czarcas7ic: Sorry, I am shooting from the hip in terms of getting through ts syntax 😅

czarcas7ic: https://github.com/osmosis-labs/osmosis-frontend/pull/1629/commits/71ed486f2badd135dd2a7ffec649b610095df2de

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
czarcas7ic: Asked in slack, but will also ask here: is there a frontend method that returns decimal amount given a denom?

jonator: Yes those objects are of type "Currency" (or "AppCurrency") and here, you can get the currencies for the deposit assets like:
`this._baseDepositAmountIn.sendCurrency.coinDecimals`
`this._quoteDepositAmountIn.sendCurrency.coinDecimals`

jonator (approved):

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
czarcas7ic: Addressed here, good catch https://github.com/osmosis-labs/osmosis-frontend/pull/1629/commits/ad1367242d115c31c7f38bb2e97c2174481e77e4

PR #1633, opened 2023-06-04
Add Token - ROAR

JeremyParish69:

To add a token ROAR from Terra 2.0

## Testing and Verifying

Uses TFM external as transfer site.


jonator (approved):
That png is quite large at 54 k. Are there any smaller alternatives?

PR #1639, opened 2023-06-06
ignore env local

mattupham:

- add a gitignore for .env.local

thoughts on adding others?
- .env
- .env.example
- .env.local
- .env.staging


jonator:
Oh it should be git ignored within the packages/web folder, which is the only place it should be used (we only use env vars in web package)

PR #1640, opened 2023-06-06
fix: Account Sequencing and Accurate IBC Transfer Amount

JoseRFelix:

Fix the following bugs: 

- [x] Get 404 when depositing CW tokens
- [x] Transfer status not updated until the page is refreshed for ETH tokens
- [x] Deposit status of native ETH(Axelar) tokens get stuck at  pending

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/866ad98b0)

### DIFF
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -31,6 +31,7 @@ import {
   WalletManager,
   WalletStatus,
 } from "@cosmos-kit/core";
+import { BaseAccount } from "@keplr-wallet/cosmos";
 import {
   ChainedFunctionifyTuple,
   ChainGetter,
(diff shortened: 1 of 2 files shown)


packages/stores/src/account/base.ts
```diff
@@ -713,8 +713,8 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
     }
 
     return {
-      accountNumber: Number(account.account_number),
-      sequence: Number(account.sequence),
+      accountNumber: Number(account.accountNumber.toString()),
```
MaxMillington, line 716: What's going on here? Why use `toString()` only to convert to a `Number`?

JoseRFelix: For signing we need to convert these values into numbers. Since The `account` values are both `Int` objects and it lacks a `toNumber()` method, we first need to convert them to strings.

MaxMillington: Do you mean that `accountNumber` and `sequence` are integers? If so, integer is a subtype of `Number` is it not? And I think JS has no specific `Int` type?

jonator: We use `Dec` and `Int` objects for handling amounts and mathematical operations on those amounts. They use bignumber (which uses BigInt) under the hood, since raw JS number types have precision issues especially when doing math.

PR #1641, opened 2023-06-07
(CL) Collect fees tx

jonator:

> Add a description of the overall background and high level changes that this PR introduces

_(E.g.: This pull request improves area A by adding ...._

### ClickUp Task

Please provide a link to the ClickUp task for this pull request:

[ClickUp Task URL](https://app.clickup.com/t/863gxjjbb)

## Brief Changelog

- In position expanded view, the collect fees button show now open with some combination of collect spread rewards and/or collect incentive reward messages
- If there are no rewards for that position, the button should be disabled
- The sign dialog should only contain messages for incentives that are available in that position (from spread or incentives), this is done to save gas since I found each message to cost ~100k in gas
- The send message function re-queries individual positions to ensure there are indeed rewards to collect, and if there are no rewards, it returns a rejected promise
- Added e2e tests to test standard spread fee collection, and the no rewards fail case

## Key Workflows to test

- Expanding positions and collecting rewards, once a swap has occured

## Testing and Verifying

1. Either open an existing pool or create a new test pool in the create pool modal (on testnet)
2. Navigate to pool detail page for target CL pool
3. Create a position on your account if needed
4. Swap in the CL pool
5. Ensure the collect fees button becomes available to click
6. Click "Collect fees" on your in-range position, ensure sign dialog in wallet just shows collect spread rewards message


packages/stores/src/queries/concentrated-liquidity/position-by-id.ts
```diff
@@ -135,16 +143,24 @@ export class ObservableQueryLiquidityPositionById extends ObservableChainQuery<L
...
 
-  protected setResponse(response: Readonly<QueryResponse<LiquidityPosition>>) {
+  protected setResponse(
+    response: Readonly<QueryResponse<{ position: LiquidityPosition }>>
+  ) {
     super.setResponse(response);
-    this.setRaw(response.data);
+    this.setRaw(response.data.position);
+    const rewardDenoms = Array.from(
+      new Set(
```
MaxMillington, line 152: Good use of set so long as the array is not too big

packages/stores/src/account/index.ts
```diff
@@ -816,6 +815,165 @@ export class OsmosisAccountImpl {
...
+            osmosis.concentratedliquidity.v1beta1.MsgCollectIncentives.encode({
+              sender: this.base.bech32Address,
+              positionIds: positionIdsWithIncentiveRewards.map((id) =>
+                Long.fromString(id)
+              ),
+            }).finish(),
+        };
+
+        // only accumulate collection msgs that have rewards
+        const aminoMsgs: any[] = [];
```
JoseRFelix, line 926:
```suggestion
        const aminoMsgs: { type: string, value: Record<string, any> }[] = [];
```

packages/stores/src/account/index.ts
```diff
@@ -816,6 +815,165 @@ export class OsmosisAccountImpl {
...
+        const protoMsgs: any[] = [];
```
JoseRFelix, line 927:
```suggestion
        const protoMsgs: { typeUrl: string, value: any }[] = [];
```

JoseRFelix (changes requested):
LGTM, just left a small nit

JoseRFelix (changes requested):
LGTM, left a small nit

packages/stores/src/queries/concentrated-liquidity/position-by-id.ts
jonator: Yes this array should never exceed 10

PR #1644, opened 2023-06-08
(CL) My pos cleanup

jonator:

Improve and polish the my position card.
Screenshot 2023-06-08 at 10 53 28 AM

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/863gxjv4v)

## Brief Changelog

- Improves position component header, more closely matching Figma
- For assets in the expanded position view, if there's more than 2 they will be organized in a uniform 2x2 table, with the USD value in the 3rd column
- Add placeholder text if there are no assets, like no rewards
- Show current price in depth chart of pool overview

## Testing and Verifying

Visual testing

### DIFF
--- a/packages/stores/src/queries-external/concentrated-liquidity/position-range.ts
+++ b/packages/stores/src/queries-external/concentrated-liquidity/position-range.ts
@@ -1,0 +1,79 @@
+import { KVStore } from "@keplr-wallet/common";
+import { HasMapStore } from "@keplr-wallet/stores";
+import { RatePretty } from "@keplr-wallet/unit";
+import { maxTick, minTick } from "@osmosis-labs/math";
+import { computed, makeObservable } from "mobx";
+
+import { IMPERATOR_TX_REWARD_BASEURL } from "..";
+import { ObservableQueryExternalBase } from "../base";
+
+/** Queries Imperator for extrapolated APR of a given position's tick range. */
+export class ObservableQueryPositionRangeApr extends ObservableQueryExternalBase<{
+  spreadFactorApr: string;
+  incentivesApr: string;
+}> {
+  constructor(
+    kvStore: KVStore,
+    baseURL: string,
+    protected readonly poolId: string,
+    protected readonly lowerTickIndex: number,
+    protected readonly upperTickIndex: number
+  ) {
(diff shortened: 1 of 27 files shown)


packages/stores/src/queries-external/concentrated-liquidity/position-range.ts
```diff
@@ -0,0 +1,79 @@
...
+) {
+  return `${poolId}:${lowerTickIndex}:${upperTickIndex}`;
+}
+
+function parseKey(key: string) {
+  const [poolId, lowerTickIndex, upperTickIndex] = key.split(":");
+
+  return {
+    poolId,
+    lowerTickIndex: parseInt(lowerTickIndex),
```
MaxMillington, line 76: Is there a guarantee that the split values will be parsed into integers? If not, this could lead to a `nan` here

packages/stores/src/queries/concentrated-liquidity/position-by-id.ts
```diff
@@ -144,6 +146,24 @@ export class ObservableQueryLiquidityPositionById extends ObservableChainQuery<{
     );
   }
 
+  /** Aggregation of claimable rewards by coin denom. */
+  @computed
+  get totalClaimableRewards(): CoinPretty[] {
+    return Array.from(
+      [...this.claimableSpreadRewards, ...this.claimableIncentiveRewards]
+        .reduce<Map<string, CoinPretty>>((sumByDenoms, coin) => {
+          const current = sumByDenoms.get(coin.currency.coinMinimalDenom);
```
MaxMillington, line 155: I think you could do 
`if (sumByDenoms.has(coin.currency.coinMinimalDenom))`

Might be clearer

packages/stores/src/queries-external/concentrated-liquidity/position-range.ts
jonator: I will handle this case

packages/stores/src/queries/concentrated-liquidity/position-by-id.ts
jonator: Yes but I'm going to need to get it to add to it on line 157 regardless?

PR #1647, opened 2023-06-08
Add recommended vscode extensions

mattupham:

We should add our list of recommended extensions to our vscode workspace config:
from here https://app.clickup.com/37420681/v/dc/13nzm9-7280/13nzm9-11340

syntax is ({publisherName}.{extensionName})
https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86783rfaa)

## Brief Changelog

Add vscode extensions file

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -9,10 +9,7 @@
     "GitHub.vscode-pull-request-github",
     "ms-vsliveshare.vsliveshare",
     "jock.svg",
-    "wix.vscode-import-cost",
     "Orta.vscode-jest",
     "dbaeumer.vscode-eslint"
-  ],
-  // List of extensions recommended by VS Code that should not be recommended for users of this workspace.
-  "unwantedRecommendations": []
+  ]
 }


.vscode/extensions.json
```diff
@@ -0,0 +1,18 @@
...
+  // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+  // List of extensions which should be recommended for users of this workspace.
+  "recommendations": [
+    "esbenp.prettier-vscode",
+    "eamodio.gitlens",
+    "GitHub.vscode-pull-request-github",
+    "ms-vsliveshare.vsliveshare",
+    "jock.svg",
+    "wix.vscode-import-cost",
```
jonator, line 12: Actually, this has in the past caused annoying hangup issues for me (even on an m1 mac) when saving files, let's remove it and leave it in the doc so people can add it on their own if they want.

.vscode/extensions.json
```diff
@@ -0,0 +1,18 @@
...
+    "Orta.vscode-jest",
+    "dbaeumer.vscode-eslint"
+  ],
+  // List of extensions recommended by VS Code that should not be recommended for users of this workspace.
+  "unwantedRecommendations": []
```
jonator, line 17: Nit: If this is empty, does it need to be there?

MaxMillington (approved):
LGTM but we should probably merge into stage always even for small things like this

PR #1650, opened 2023-06-09
(CL) Migrate to increase liquidity message

jonator:

Since we're no longer aggregating positions by price range and don't have charging periods, we should use the AddToPosition message instead of creating a new position with the same range. This is the functionality behind the increase liquidity modal.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/863gxjgk4)

## Brief Changelog

- Adds the `MsgAddToPosition` message
  - e2e tests
- Wires the message up to the increase liquidity modal
- Fixes for rounding to nearest tick
- Sorts position card list by join time

## Testing and Verifying

- E2e tests work
- Add to position increases the liquidity of a position

### DIFF
--- a/packages/stores/src/account/__tests_e2e__/positions.spec.ts
+++ b/packages/stores/src/account/__tests_e2e__/positions.spec.ts
@@ -83,6 +83,7 @@ describe("Create CL Positions Txs", () => {
     ).resolves.toBeDefined();
 
     // old position is replaced
+    // we can't rely on position IDs as other positions may have been globally created first in future tests
     await expect(
       getUserPositionsIdsForPool(queryPool!.id)
     ).resolves.toHaveLength(2);
--- a/packages/stores/src/account/__tests_e2e__/positions.spec.ts
+++ b/packages/stores/src/account/__tests_e2e__/positions.spec.ts
@@ -1,0 +1,171 @@
+/* eslint-disable */
+import { ObservableQueryPool } from "src/queries";
+import {
+  chainId,
+  getLatestQueryPool,
+  RootStore,
+  waitAccountLoaded,
+} from "../../__tests_e2e__/test-env";
+import { maxTick, minTick } from "@osmosis-labs/math";
+// import { Int } from "@keplr-wallet/unit";
+
+describe("Create CL Positions Txs", () => {
+  const { accountStore, queriesStore, chainStore } = new RootStore();
+  let queryPool: ObservableQueryPool | undefined;
+
+  beforeAll(async () => {
+    const account = accountStore.getAccount(chainId);
+    account.cosmos.broadcastMode = "sync";
+    await waitAccountLoaded(account);
+  });
+
+  beforeEach(async () => {
+    const account = accountStore.getAccount(chainId);
+
+    // prepare CL pool
+    await new Promise((resolve, reject) =>
+      account.osmosis.sendCreateConcentratedPoolMsg(
+        "uion",
+        "uosmo",
+        1,
+        0.001, // must have spread factor to generate fees
+        undefined,
+        (tx) => {
+          if (tx.code) reject(tx.log);
+          else resolve(tx);
+        }
+      )
+    );
+
+    queryPool = await getLatestQueryPool(chainId, queriesStore);
+  });
+
+  it("should be able to be created", async () => {
+    await expect(createFullRangePosition(queryPool!.id)).resolves.toBeDefined();
+    await expect(
+      getUserPositionsIdsForPool(queryPool!.id)
+    ).resolves.toHaveLength(1);
+    await expect(createFullRangePosition(queryPool!.id)).resolves.toBeDefined();
+    await expect(
+      getUserPositionsIdsForPool(queryPool!.id)
+    ).resolves.toHaveLength(2);
+  });
+
+  it("can have liquidity be added to it", async () => {
+    const account = accountStore.getAccount(chainId);
+
+    // create 2 positions, since there need to be at least 2 to be able to add liquidity
+    await createFullRangePosition(queryPool!.id);
+    await createFullRangePosition(queryPool!.id);
+    const userPositionIds = await getUserPositionsIdsForPool(queryPool!.id);
+    const lastPositionId = userPositionIds[userPositionIds.length - 1];
+
+    // add liquidity to position
+    const specifiedAmount0 = "1000";
+    const calculatedAmount1 = "1000";
+
+    await expect(
+      new Promise((resolve, reject) =>
+        account.osmosis
+          .sendAddToConcentratedLiquidityPositionMsg(
+            lastPositionId,
+            specifiedAmount0,
+            calculatedAmount1,
+            undefined,
+            undefined,
+            (tx) => {
+              if (tx.code) reject(tx.log);
+              else resolve(tx);
+            }
+          )
+          .catch(reject)
+      )
+    ).resolves.toBeDefined();
+
+    // old position is replaced
+    await expect(
+      getUserPositionsIdsForPool(queryPool!.id)
+    ).resolves.toHaveLength(2);
+  });
+
+  it("rejects adding liquidity to position if last position in pool", async () => {
+    const account = accountStore.getAccount(chainId);
+
+    // create initial position
+    await createFullRangePosition(queryPool!.id);
+    const userPositionIds = await getUserPositionsIdsForPool(queryPool!.id);
+    const lastPositionId = userPositionIds[userPositionIds.length - 1];
+
+    // add liquidity to position
+    const specifiedAmount0 = "1000";
+    const calculatedAmount1 = "1000";
+
+    await expect(
+      new Promise((resolve, reject) =>
+        account.osmosis
+          .sendAddToConcentratedLiquidityPositionMsg(
+            lastPositionId,
+            specifiedAmount0,
+            calculatedAmount1,
+            undefined,
+            undefined,
+            (tx) => {
+              if (tx.code) reject(tx.log);
+              else resolve(tx);
+            }
+          )
+          .catch(reject)
+      )
+    ).rejects.toBeDefined();
+  });
+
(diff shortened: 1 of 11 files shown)


packages/stores/src/account/__tests_e2e__/positions.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+              else resolve(tx);
+            }
+          )
+          .catch(reject)
+      )
+    ).rejects.toBeDefined();
+  });
+
+  /** Leave `poolId` undefined to get all position IDs. */
+  async function getUserPositionsIdsForPool(poolId?: string) {
```
MaxMillington, line 123: A general thought on this test file. The poolId is optional. Is it possible that the pool passed in here or the below function won't exist? Should we write a test for that case?

packages/stores/src/account/__tests_e2e__/positions.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+            }
+          )
+          .catch(reject)
+      )
+    ).resolves.toBeDefined();
+
+    // old position is replaced
+    await expect(
+      getUserPositionsIdsForPool(queryPool!.id)
+    ).resolves.toHaveLength(2);
```
MaxMillington, line 88: is it worth checking what the actual values are?

packages/stores/src/account/index.ts
```diff
@@ -815,6 +815,91 @@ export class OsmosisAccountImpl {
...
+    amount0: string,
+    amount1: string,
+    maxSlippage = "2.5",
+    memo: string = "",
+    onFulfill?: (tx: any) => void
+  ) {
+    // refresh position
+    const queryPosition =
+      this.queries.queryLiquidityPositionsById.getForPositionId(positionId);
+    await queryPosition.waitFreshResponse();
```
MaxMillington, line 829: do we need to worry about errors in these functions? We aren't handling error states

packages/stores/src/account/index.ts
```diff
@@ -815,6 +815,91 @@ export class OsmosisAccountImpl {
     );
   }
 
+  async sendAddToConcentratedLiquidityPositionMsg(
+    positionId: string,
+    amount0: string,
+    amount1: string,
+    maxSlippage = "2.5",
```
MaxMillington, line 822: just curious where this number comes from? We should probably put it in a constant with a description

packages/stores/src/account/index.ts
```diff
@@ -815,6 +815,91 @@ export class OsmosisAccountImpl {
...
+        sender: this.base.bech32Address,
+        amount0: amount0,
+        amount1: amount1,
+        token_min_amount0: amount0WithSlippage,
+        token_min_amount1: amount1WithSlippage,
+      },
+    };
+
+    const protoMsg = {
+      typeUrl: "/osmosis.concentratedliquidity.v1beta1.MsgAddToPosition",
```
MaxMillington, line 853: do we have a constants file for endpoints like this?

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
```diff
@@ -98,5 +101,33 @@ export function useAddConcentratedLiquidityConfig(
...
+          amount1,
+          undefined,
+          undefined,
+          (tx) => {
+            if (tx.code) reject(tx.log);
+            resolve();
+          }
+        );
+
+        try {
```
MaxMillington, line 122: What's going on here? the try-catch block is empty

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
```diff
@@ -98,5 +101,33 @@ export function useAddConcentratedLiquidityConfig(
...
+        } catch (e: any) {
+          console.error(e);
```
MaxMillington, line 124: Did you mean to leave this in?

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
```diff
@@ -98,5 +101,33 @@ export function useAddConcentratedLiquidityConfig(
     config.poolId,
   ]);
 
-  return { config, addLiquidity };
+  const increaseLiquidity = useCallback(
+    (positionId: string) => {
+      return new Promise<void>(async (resolve, reject) => {
```
MaxMillington, line 106: the structure of the promise executor is a little confusing to me. I might be missing something, but would something like this work?

```
const increaseLiquidity = useCallback(
    (positionId: string): Promise<void> => {
      const amount0 = config.baseDepositAmountIn.getAmountPrimitive().amount;
      const amount1 = config.quoteDepositAmountIn.getAmountPrimitive().amount;

      return account.osmosis
        .sendAddToConcentratedLiquidityPositionMsg(
          positionId,
          amount0,
          amount1
        )
        .then((tx) => {
          if (tx.code) throw new Error(tx.log);
        })
        .catch((e: any) => {
          throw new Error(e.message);
        });
    },
    [config.baseDepositAmountIn, config.quoteDepositAmountIn, account.osmosis]
  );
  ```

packages/stores/src/account/__tests_e2e__/positions.spec.ts
jonator: That's moreso a utility for letting a test either get all user positions (undefined) or all user positions filtered for a pool (poolId="2"). I can remove optionality to make things less variable in this test env?

packages/stores/src/account/__tests_e2e__/positions.spec.ts
jonator: The positions IDs may not be deterministically created depending on other tests in the env, so I don't think we can reliably specify the expected IDs. If other tests create positions and run before these tests, the IDs would change and break these tests.

packages/stores/src/account/index.ts
jonator: Closest thing may be msg-opts.ts? But I think those are just used for Amino URLs and gas settings. Open to answering Qs for what those are.

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
jonator: Yeah I think this API could be improved, it originates from keplr-wallet libs. I believe the only place the tx contents are returned is in the onFullfilled callback passed into the send* funtion.

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
jonator: Good catch

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
jonator: The console log? Another option would be to remove that and have the caller of `increaseLiquidity` handle the error as they please? Fine with either.

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
jonator: I believe the `send*` function(s) (ours) which calls `sendMsgs` (keplr-wallet libs) on cosmos base account resolves to `undefined` when the tx is _broadcast_ (i.e. they click sign). However, it sets up a [websocket listener](https://github.com/chainapsis/keplr-wallet/blob/7a4db487b74170d968ecc2c7979d39579251277c/packages/stores/src/account/cosmos.ts#L441) to receive push updates as the tx gets processed, then returns the committed tx object in the `onFullfilled` callback.

See def for `sendMsgs`:
https://github.com/chainapsis/keplr-wallet/blob/7a4db487b74170d968ecc2c7979d39579251277c/packages/stores/src/account/cosmos.ts#L363

With @JoseRFelix 's work to pull send logic into our code, we can make changes to this if we wish. We discussed this a bit but I believe Jose chose to stick with this API (resolve top level fn on broadcast, pass callback for fullfill).

packages/stores/src/account/index.ts
jonator: Eh yes that's possible, if the endpoint fails. But generally we assume queries will work as expected, and errors logged in those base query classes, otherwise we'd have to write a lot of boilerplate to handle errors across every use.

packages/stores/src/account/index.ts
jonator: It's arbitrary, a normal slippage % value. I can add docs to fn sig.

jonator: More background; wherever you see slippage being specified, typically token amounts are being transformed from one form to another (like L values with concentrated liquidity, or gamm shares for balancer and stable pools, or in and out of a constantly changing AMM pool). Slippage generally refers to the changes to chain state between signing the tx and the tx being committed in the block.

Here, liquidity values could change between signing and the position being created. So we subtract 2.5% from the amount calculated and use that as minimum amounts the user is willing to accept as addition to the position. It's a user protection mechanism.

packages/stores/src/account/__tests_e2e__/positions.spec.ts
MaxMillington: Yea I think that makes sense 👍

packages/stores/src/account/__tests_e2e__/positions.spec.ts
MaxMillington: 👍 Maybe just a comment would be good then?

packages/stores/src/account/index.ts
MaxMillington: Makes sense. I still think we should put in a constant though

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
MaxMillington: The latter makes more sense to me, but I don't have the full context

PR #1651, opened 2023-06-10
(CL) Make Increase Liquidity Modal Mobile Responsive

JoseRFelix:


packages/web/modals/increase-concentrated-liquidity.tsx
```diff
@@ -120,12 +122,35 @@ export const IncreaseConcentratedLiquidityModal: FunctionComponent<
...
+                id="close-thin"
+                className="text-wosmongton-400 hover:text-wosmongton-100"
+                height={24}
+                width={24}
+              />
+            }
+            onClick={props.onRequestClose}
+          />
+        </div>
+      </div>
```
jonator, line 150: i think the basemodal provides a close button for use if we simply set hideclosebutton=false

packages/web/components/cards/my-position/status.tsx
```diff
@@ -13,6 +13,7 @@ export const MyPositionStatus: FunctionComponent<{
   lowerPrice: Dec;
   upperPrice: Dec;
   negative?: boolean;
+  className?: string;
```
jonator, line 16: giga nit: we have `CustomClasses` type that adds this

jonator (commented):
Nits

jonator (changes requested):
See some issues with arrangement of +/- buttons on iPhone
Screenshot 2023-06-10 at 12 28 50 PM

JoseRFelix:
> See some issues with arrangement of +/- buttons on iPhone

Interesting I'm getting the correct size on the first load. It could be an issue with `onresize` event. Can you confirm if it appears correctly when you load it for the first time on your side?

image

jonator (approved):

PR #1662, opened 2023-06-14
(CL) Rename Fee to Spread Factor and Add Translations

JoseRFelix:


jonator:
@JoseRFelix FYI we may be adding additional translations for CL, in that case do you think it would be easiest to wait until we've added those to update translations?

jonator (approved):

PR #1671, opened 2023-06-14
mattupham/yarn lint fix errors master

mattupham:

- fixes  errors when running `yarn run lin:fix`

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/867837y54)

## Brief Changelog

- ignore .webp files
- ignore .env.local
- fix formatting

## Testing and Verifying

before:
Screenshot 2023-06-14 at 15 39 15

after:
Screenshot 2023-06-14 at 15 57 12

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1623,6 +1623,11 @@
     "@types/yargs" "^16.0.0"
     chalk "^4.0.0"
 
+"@jridgewell/sourcemap-codec@^1.4.13":
+  version "1.4.15"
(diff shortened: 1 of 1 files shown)


packages/web/.prettierignore
```diff
@@ -8,7 +8,10 @@
 *.svg
 *.png
 *.jpg
+*.webp
+
 *.env
+*.env.local
```
jonator, line 14: I think there can be _many_ flavors of env files so you may want to do `*.env.*` if that works (not sure)

See
https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#environment-variable-load-order

yarn.lock
jonator, line 1: What is the reason for these yarn lock changes? Was it out of date?

jonator (approved):

PR #1674, opened 2023-06-15
Add Nolus mainnet

ivivanov:
This PR aims to add Nolus mainnet configurations. 
All IBC channels are tested and work as expected
Pool: 1041 is created usdc/unls

### DIFF
--- a/packages/web/config/chain-infos.ts
+++ b/packages/web/config/chain-infos.ts
@@ -239,7 +239,7 @@ const testnetChainInfos: SimplifiedChainInfo[] = [
       },
     ],
     features: ["ibc-transfer", "ibc-go"],
-    explorerUrlToTx: "https://explorer-rila.nolus.io/nolus-rila/tx/{txHash}",
+    explorerUrlToTx: "https://explorer-rila.nolus.io/rila-1/tx/{txHash}",
   },
   {
     rpc: "https://rpc-eu-1.kaon.kyve.network",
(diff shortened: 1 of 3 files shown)


packages/web/config/chain-infos.ts
```diff
@@ -3598,6 +3598,34 @@ const mainnetChainInfos: SimplifiedChainInfo[] = [
...
+    bip44: {
+      coinType: 118,
+    },
+    bech32Config: Bech32Address.defaultBech32Config("nolus"),
+    currencies: [
+      {
+        coinDenom: "NLS",
+        coinMinimalDenom: "unls",
+        coinDecimals: 6,
+        coinGeckoId: "usd-coin",
```
JeremyParish69, line 3615: Please correct to "pool:unls"

packages/web/config/chain-infos.ts
```diff
@@ -3598,6 +3598,34 @@ const mainnetChainInfos: SimplifiedChainInfo[] = [
...
+        coinImageUrl: "/tokens/nolus.svg",
```
JeremyParish69, line 3616: I don't see this image uploaded anywhere

packages/web/config/ibc-assets.ts
```diff
@@ -1813,6 +1813,13 @@ export const IBCAssetInfos: (IBCAsset & {
         destChannelId: "channel-0",
         coinMinimalDenom: "ulore",
       },
+      {
+        counterpartyChainId: "pirin-1",
+        sourceChannelId: "channel-783",
+        destChannelId: "channel-0",
+        coinMinimalDenom: "unls",
+        isVerified: true,
```
JeremyParish69, line 1821: I don't recall this ever being approved. Please remove isVerified

packages/web/config/chain-infos.ts
ivivanov: It is already there form an old PR
[packages/web/public/tokens/nolus.svg](https://github.com/osmosis-labs/osmosis-frontend/blob/master/packages/web/public/tokens/nolus.svg)

packages/web/config/ibc-assets.ts
ivivanov: removed

packages/web/config/chain-infos.ts
ivivanov: Thanks, its fixed now

PR #1677, opened 2023-06-17
use inlang ide extension

samuelstroschein:
Updated inlang to include the ide extension.


jonator (approved):

jonator:
We need to remember to remove our custom tests in the localization folder that accomplished what inlang linter does now

PR #1691, opened 2023-06-21
Force injective/stride to use the direct signing for ibc transfer.

Thunnini:

In ibc-go v7, there is a bug which MsgTransfer not implement `LegacyMsg`.
And, it makes error like below
```
recovered: expected *legacytx.LegacyMsg when using amino JSON stack: goroutine 8036803 [running]: runtime/debug.Stack() runtime/debug/stack.go:24 +0x65 github.com/cosmos/cosmos-sdk/baseapp.newDefaultRecoveryMiddleware.func1({0x23894a0, 0xc0e4f9dff0}) github.com/cosmos/cosmos-sdk@v0.47.3/baseapp/recovery.go:71 +0x27 github.com/cosmos/cosmos-sdk/baseapp.newRecoveryMiddleware.func1({0x23894a0?, 0xc0e4f9dff0?}) github.com/cosmos/cosmos-sdk@v0.47.3/baseapp/recovery.go:39 +0x30 github.com/cosmos/cosmos-sdk/baseapp.processRecovery({0x23894a0, 0xc0e4f9dff0}, 0xc0984e42c0?) github.com/cosmos/cosmos-sdk@v0.47.3/baseapp/recovery.go:28 +0x37 github.com/cosmos/cosmos-sdk/baseapp.processRecovery({0x23894a0, 0xc0e4f9dff0}, 0x46b5e60?) github.com/cosmos/cosmos-sdk@v0.47.3/baseapp/recovery.go:33 +0x5e github.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).runTx.func1() github.com/cosmos/cosmos-sdk@v0.47.3/baseapp/baseapp.go:641 +0xf0 panic({0x23894a0, 0xc0e4f9dff0}) runtime/panic.go:890 +0x262 github.com/cosmos/cosmos-sdk/x/auth/ante.SetUpContextDecorator.AnteHandle.func1() github.com/cosmos/cosmos-sdk@v0.47.3/x/aut
```
There is no way to solve this problem because it is bug on chain itself.
Temporarily, force stride to use the direct signing for ibc transfer.
However, it is still impossible to use ledger for ibc transfer because ledger only supports amino signing.

Specifically, in the case of injective/stride, direct signing is used for ibc transfer. However, in the case of a ledger, amino is used as it is. The above error occurs on the ledger, but it resolves itself naturally once the corresponding chains resolve the issue and update the chain.
After the problems with ibc-go v7 are resolved, we can revert this commit.

The actual changes are [here](https://github.com/chainapsis/keplr-wallet/commit/116d66ccb50019fd30f843ded898c6b4af4243e5)

## Brief Changelog

- Bump "@keplr-wallet/*" to "0.10.24-ibc.go.v7.hot.fix". "0.10.24-ibc.go.v7.hot.fix" version has ability to request the direct signing and force to use the direct signing if chain is injective/stride and the msg in tx is MsgTransfer and key is not from ledger nano

## Testing and Verifying

- IBC transfer should be requested as direct signing if injective/stride and key is not from ledger nano


delivan (approved):
LGTM. this needs to be resolved quickly

jonator:
@Thunnini do we need to add `ibc-go-v7-hot-fix` feature to stride and inj (cc @JeremyParish69 if we need to, please add it)

delivan:
@jonator It doesn't need. already exception handled for those chains. cc @JeremyParish69

PR #1694, opened 2023-06-21
(MW) Use Direct Signing for Chains affected by ibc-go-v7

JoseRFelix:

Currently, there is a bug in ibc-go-v7 (#1691), where transactions signed with 'signAmino' result in an error on chains such as Stride and Injective. PR intends to manage these errors by opting for 'signDirect' when required. However, it's important to note that the Cosmostation wallet doesn't support 'signDirect'. As a result, deposits to Injective and Stride have been disabled.

Please note that injective will not be disabled in the assets table since its deposit and withdrawal redirect to an external page.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/866agfwwq)

## Testing and Verifying

- [ ] Do deposit transactions to Stride with Leap and Keplr. 
- [ ] Verify that deposit link is disabled when connected to Comostation wallet

### DIFF
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -522,6 +522,10 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
 
     const walletWindowName = getWalletWindowName(wallet.walletName);
 
+    /**
+     * Remove once ibc-go-v7 fix is released.
+     * @see https://github.com/osmosis-labs/osmosis-frontend/pull/1691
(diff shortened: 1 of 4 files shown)


jonator (approved):

jonator:
Let's make sure to squash this so we can easily revert in future

Stride is expecting a resolution in the next few days:

https://osmosis-network.slack.com/archives/C03E4D0T7T5/p1687357832863789

PR #1699, opened 2023-06-22
Hot Fix: Handle Leap and Keplr App Web Connection

JoseRFelix:
## Brief Description 

If on mobile and `leap` or `keplr` is in `window`, it means that the user enters the front end from that particular wallet app in the app browser. So, there is no need to use wallet connect, as it resembles the browser's extension's usage.

## Testing

- [ ] Use Leap mobile browser to access this preview link, and test its connection
- [ ] Check that in the normal browser only WalletConnect is visible and no extension wallet

### DIFF
--- a/packages/web/modals/wallet-select.tsx
+++ b/packages/web/modals/wallet-select.tsx
@@ -373,7 +373,54 @@ const ModalContent: FunctionComponent<
 
     const wallets = [...WalletRegistry]
       // If mobile, filter out browser wallets
-      .filter((w) => (isMobile ? !w.mobileDisabled : true))
+      .reduce((acc, w, _index, array) => {
+        if (isMobile) {
+          /**
+           * If an extension wallet is found in mobile, this means that we are inside an app browser.
+           * Therefore, we should only show that compatible extension wallet.
+           * */
+          if (acc.length > 0 && acc[0].name.endsWith("-extension")) {
+            return acc;
+          }
+
+          const windowAsRecord = window as Record<string, any>;
+
+          /**
+           * If on mobile and `leap` is in `window`, it means that user enters
+           * frontend from Leap's app in app browser. So, there is no need
+           * to use wallet connect, and it resembles extension's usages.
+           */
+          if (
+            windowAsRecord?.leap &&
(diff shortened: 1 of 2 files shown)


packages/web/modals/wallet-select.tsx
```diff
@@ -373,7 +373,54 @@ const ModalContent: FunctionComponent<
...
+          }
+
+          /**
+           * If on mobile and `keplr` is in `window`, it means that the user enters
+           * the frontend from Keplr's app in app browser. So, there is no need
+           * to use wallet connect, as it resembles the extension's usage.
+           */
+          if (
+            windowAsRecord?.keplr &&
+            windowAsRecord?.keplr?.mode === "mobile-wallet"
```
Thunnini, line 409: This should be "mobile-web"?
https://github.com/chainapsis/keplr-wallet/blob/master/packages/types/src/wallet/keplr.ts#L58

JoseRFelix: Nice catch! Updated 👍

JoseRFelix: Nice catch! Updated 👍

jonator (approved):

packages/web/modals/wallet-select.tsx
```diff
@@ -373,7 +373,55 @@ const ModalContent: FunctionComponent<
...
+           * If on mobile and `leap` is in `window`, it means that the user enters
+           * the frontend from Leap's app in app browser. So, there is no need
+           * to use wallet connect, as it resembles the extension's usage.
+           */
+          if (
+            windowAsRecord?.leap &&
+            windowAsRecord?.leap?.mode === mobileWebModeName
+          ) {
+            return array
+              .filter((w) => w.name === "leap-extension")
```
MaxMillington, line 399: we should probably keep these strings e.g. 'leap-extension' 'keplr-extension' etc in constants somewhere

JoseRFelix: Nice suggestion; I added an enum in my latest commit to search for the available wallet

PR #1708, opened 2023-06-25
Use a Stargaze NFT as a profile picture

MightOfOaks:
The PR adds the option to select an NFT that the user owns on Stargaze to be used as a profile picture. 

## Brief Changelog
- The logic is self-contained on profile.tsx
- Uses profileStore to store the image URI for the NFT selected by the user
- Adds "stargaze-pfp" as an Avatar type
- Adds "profile-stargaze.png" to /public/images
- Adds "ipfs-gw.stargaze-apis.com" & "res.cloudinary.com" as viable image sources
## Requirements
- The following environment variable needs to be defined:
NEXT_PUBLIC_STARGAZE_GRAPHQL_API="https://graphql.mainnet.stargaze-apis.com/graphql"
## Testing and Verifying
[Screencast from 25-06-2023 21:47:18.webm](https://github.com/osmosis-labs/osmosis-frontend/assets/11138024/33901105-8641-4e06-9f44-f65d315b5abc)

This change has been tested locally by rebuilding the website and verified.

- Does this pull request introduce a new feature or user-facing behavior changes? **Yes**
- How is the feature or change documented? **Not documented at the moment**


packages/web/components/navbar/index.tsx
```diff
@@ -247,21 +247,35 @@ const WalletInfo: FunctionComponent<
           className="group flex place-content-between items-center gap-[13px] rounded-xl border border-osmoverse-700 px-1.5 py-1 hover:border-[1.3px] hover:border-wosmongton-300 hover:bg-osmoverse-800 md:w-full"
         >
           <div className="h-8 w-8 shrink-0 overflow-hidden rounded-[7px] bg-osmoverse-700 group-hover:bg-gradient-positive">
-            {profileStore.currentAvatar === "ammelia" ? (
+            {profileStore.currentAvatar === "ammelia" && (
```
luke-ae, line 250: Can keep this the same way it was using `?` and have`: null` at the end

packages/web/modals/profile.tsx
```diff
@@ -77,17 +80,48 @@ export const ProfileModal: FunctionComponent<
...
   } = useDisclosure();
+  const {
+    isOpen: isStargazePfpSelectOpen,
+    onClose: onCloseStargazePfpSelect,
+    onOpen: onOpenStargazePfpSelect,
+  } = useDisclosure();
 
   const wallet = accountStore.getWallet(chainId);
+  const address = wallet?.address ?? "";
+  const endpoint = "https://graphql.mainnet.stargaze-apis.com/graphql";
```
luke-ae, line 91: Ideally this is in an `.env` and not public

packages/web/modals/profile.tsx
```diff
@@ -103,6 +137,133 @@ export const ProfileModal: FunctionComponent<
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  useEffect(() => {
```
luke-ae, line 140: Try not to use a `useEffect` to set state, this could just be written as a variable 

```
const stargazeAddress =
    address.length > 0
      ? Bech32Address.fromBech32(address, "osmo").toBech32("stars")
      : "";

File: packages/web/modals/profile.tsx [hunk 4]
```diff
@@ -103,6 +137,133 @@ export const ProfileModal: FunctionComponent<
...
+    if (address.length > 0) {
+      setStargazeAddress(
+        Bech32Address.fromBech32(address, "osmo").toBech32("stars")
+      );
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [address]);
+
+  const fetchStargazeNFTs = async () => {
```
luke-ae, line 149: Wrap these as callbacks so we don't need to omit deps from useEffects.

Could also pull all these queries into their own hook to clean up the profile a bit

packages/web/modals/profile.tsx
```diff
@@ -77,17 +80,48 @@ export const ProfileModal: FunctionComponent<
...
+  const endpoint = "https://graphql.mainnet.stargaze-apis.com/graphql";
 
   const [hasCopied, setHasCopied] = useState(false);
   const [_state, copyToClipboard] = useCopyToClipboard();
   const [_isReady, _cancel, reset] = useTimeoutFn(
     () => setHasCopied(false),
     2000
   );
+  const [stargazeAddress, setStargazeAddress] = useState("");
+  const [stargazeNFTs, setStargazeNFTs] = useState<any[]>([]);
```
luke-ae, line 100: Let's not use `any` here, can copy what you need from `Token` and `Collection` types from our schema in Galaxy

packages/web/modals/profile.tsx
```diff
@@ -77,17 +80,48 @@ export const ProfileModal: FunctionComponent<
...
+  const [stargazeCollections, setStargazeCollections] = useState<any[]>([]);
+  const [selectedCollection, setSelectedCollection] = useState<any>(null);
+  const [searchCollectionValue, setSearchCollectionValue] = useState("");
+  const [searchTokenValue, setSearchTokenValue] = useState("");
```
luke-ae, line 104: Could we get a loading state as well for a better UX

packages/web/modals/profile.tsx
```diff
@@ -103,6 +137,133 @@ export const ProfileModal: FunctionComponent<
...
+    if (stargazeAddress.length > 0) {
+      fetchStargazeCollections();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [stargazeAddress]);
+
+  useEffect(() => {
+    if (stargazeAddress.length > 0) {
+      if (selectedCollection?.collection?.name === "Badges")
+        fetchStargazeBadges();
```
luke-ae, line 261: wrap these in `{ }`

packages/web/modals/profile.tsx
```diff
@@ -174,10 +348,213 @@ export const ProfileModal: FunctionComponent<
...
+                                    <Icon
+                                      className="flex shrink-0 items-center text-white-full"
+                                      id={"search"}
+                                      height={12}
+                                      width={12}
+                                    />
+                                  </span>
+                                </div>
+                              )}
+                              {filteredTokens?.length > 0 &&
```
luke-ae, line 501: Could pull this into it's own variables since it's been used 3 times

const tokensWithImage = filteredTokens?.filter((nft) => nft?.media?.type === "image")

luke-ae (changes requested):

packages/web/modals/profile.tsx
MightOfOaks: Updated the first part.

Reluctant to add a hook as it may not find use anywhere else in the code.

jonator:
Few product-level feedback items:

* Could we use this logo instead of the circular one? Also I'm assuming we need to request access to their stargaze account?
Screenshot 2023-06-26 at 9 06 06 AM

* For the actual selection in the drawer, what are your thoughts of just making it be a list of all of their NFTs? Sorted by floor price or popularity (if possible). It think only super users will have too many NFTs for a poor UX, and if they're sorted by popularity, it should help address that as we can safely assume they'd prefer the more valuable NFTs as PFPs

packages/web/modals/profile.tsx
jonator: I agree with your take @MightOfOaks. We will likely not reuse any of this anytime soon, so it should be as isolated as possible in the codebase

luke-ae:
> 

- There's no access needed to Stargaze, we can just query our API with the STARS address. 
- We definitely could however we've got 2,671 addresses with more than 50 NFTs, so a list becomes a bit useless especially when most are named as collection + token ID e.g. "Bad Kid #2451" 

image

jonator:
Wow this is super cool, and very interchain. 

More thoughts:
* The dropdown workflow seems clunky and unpolished. No worthy of shipping IMO. What if we just have a scrollable list on the left then the images on the right once a collection is selected? The dropdown doesn't seem useful.

packages/web/stores/profile.ts
```diff
@@ -8,13 +8,17 @@ import {
...
+type Avatar = "wosmongton" | "ammelia" | "stargaze-pfp";
 
 export class ProfileStore {
   @observable
   private _currentAvatar: Avatar;
+  @observable
+  private _stargazeAvatarUri: string | undefined;
 
   private readonly _avatarStorageKey = "profile_store_current_avatar";
+  private readonly _stargazeAvatarUriStorageKey =
```
jonator, line 20: Can we add their stargaze bech 32 address to this key? That way, when the account is switched it doesn't show an NFT for a different account.

jonator (changes requested):

jonator:
Sorry, I've talked to the team and we've decided that our NFT strategy with wosmoNFTs should come first before we support stargaze NFTs (which we plan to do)

We'd like to be a bit more intentional with our rollout of NFT functionality

PR #1723, opened 2023-06-27
mattupham/ad-banner-add-get-static-props

mattupham:

- clean up the logic of rendering ads
- add to getStaticProps so we don't have to use dynamic import

### ClickUp Task

## Brief Changelog

- ad adds for getStaticProps
- only shuffle ads if the "ads" prop changes
- 
## Testing and Verifying

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/34e04f6f-4db9-45b9-82fd-7584f7c224b7

### DIFF
--- a/packages/web/pages/index.tsx
+++ b/packages/web/pages/index.tsx
@@ -1,8 +1,7 @@
 import { Dec } from "@keplr-wallet/unit";
 import { useFlags } from "launchdarkly-react-client-sdk";
 import { observer } from "mobx-react-lite";
-import type { NextPage } from "next";
-import type { GetStaticProps } from "next";
+import type { GetStaticProps, InferGetServerSidePropsType } from "next";
 import { useMemo } from "react";
 
 import { AdBanner } from "~/components/ad-banner/ad-banner";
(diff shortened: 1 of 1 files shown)


packages/web/pages/index.tsx
```diff
@@ -107,3 +107,8 @@ const Home: NextPage = observer(function () {
 });
 
 export default Home;
+
+export const getServerSideProps: GetServerSideProps<HomeProps> = async () => {
```
MaxMillington, line 111: getServerSideProps or getStaticProps? I think we expect infrequent changes for the ad banners. If we decide to use getStaticProps, make sure to include a cache refresh

packages/web/components/ad-banner/ad-banner-content.tsx
```diff
@@ -2,22 +2,12 @@ import classNames from "classnames";
...
-  iconImageAlt: string;
-  gradient: string;
-  fontColor: string;
-  arrowColor: string;
-  featured: boolean;
-}
+type AdBannerContentProps = Ad;
 
-export const AdBannerContent: React.FC<AdBannerProps> = memo(
+export const AdBannerContent: React.FC<AdBannerContentProps> = memo(
```
jonator, line 10: Let's try to stick with `FunctionComponent` over `React.FC`, as it is elsewhere in app

packages/web/components/ad-banner/ad-banner-types.tsx
jonator, line 1: Why does this necessitate a separate file? If so, why tsx and why not types.ts?

jonator (approved):
NIT, LGTM

packages/web/components/ad-banner/ad-banner-types.tsx
jonator: I disagree, they can be shared and still co located. Counterintuitively, the _less_ important a type is (like we do for raw query responses, the more permissible it is to isolate it in its own types file.

I agree with cory house here:
https://twitter.com/housecor/status/1655188355569659908

PR #1734, opened 2023-06-28
Add KSM

JeremyParish69:

To add the PICA and KSM tokens from Composable

## Brief Changelog

-adds PICA and KSM to chain-infos.ts, ibc-assets.ts, and defines prices for each, and adds token logos (both under 100 KB).

## Testing and Verifying

## Pending
KSM does not work. Will rebuild previews once Stage (which will have the 'Unsafe IBC Registrar') is merged to master brange.


packages/web/config/generate-chain-infos/source-chain-infos.ts
```diff
@@ -3689,6 +3689,42 @@ const mainnetChainInfos: SimplifiedChainInfo[] = [
...
+    },
+    bech32Config: Bech32Address.defaultBech32Config("centauri"),
+    currencies: [
+      {
+        coinDenom: "PICA",
+        coinMinimalDenom: "ppica",
+        coinDecimals: 12,
+        coinGeckoId: "pool:ppica",
+        coinImageUrl: "/tokens/composable.png",
+        isStakeCurrency: true,
```
JeremyParish69, line 3708: I'm getting the error 'stakeCurrency" is required, even though there is a stakeCurrency defined....

JoseRFelix: This should be fixed by using #1735 assets generation.

JoseRFelix:
@JeremyParish69 Fix has been merged to master. This should be unblocked

packages/web/config/generate-chain-infos/source-chain-infos.ts
JeremyParish69: Yes, it does--thanks!

jonator (approved):

JeremyParish69:
No longer needed as KSM was added to stage

PR #1742, opened 2023-06-30
fix: Unauthorized Error While Signing some IBC Asset Transactions

JoseRFelix:

This PR solves the unauthorized error users received when doing IBC transfers to chains like Gitopia. The issue lay in the `revision_number` being "0". If a chain lacks a version number, this value should be omitted from the transaction message, otherwise, it will fail.  

This PR also resolves an issue when transactions are executed before the `TxTracer`'s detection. To maintain accuracy, we'll now utilize the values from tx_result as needed.

## Brief Changelog

- Use `tx_result` when needed if transaction is executed before being found by tracer.
- Remove revision from message if value is "0"

## Testing and Verifying

- [ ] Should allow to withdraw Lore to Gitopia

### DIFF
--- a/packages/stores/src/account/cosmos/index.ts
+++ b/packages/stores/src/account/cosmos/index.ts
@@ -124,6 +124,10 @@ export class CosmosAccountImpl {
       );
     }
 
+    const revisionNumber = ChainIdHelper.parse(
+      destinationInfo.network
+    ).version.toString();
+
     const msg = this.msgOpts.ibcTransfer.messageComposer({
       sourcePort: channel.portId,
       sourceChannel: channel.channelId,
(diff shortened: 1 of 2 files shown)


packages/stores/src/account/cosmos/index.ts
```diff
@@ -124,6 +124,10 @@ export class CosmosAccountImpl {
       );
     }
 
+    const revisionNumber = ChainIdHelper.parse(
+      destinationInfo.network
+    ).version.toString();
+
     const msg = this.msgOpts.ibcTransfer.messageComposer({
```
jonator, line 131: Did you remove our bespoke IBC transfer logic/files?

jonator (approved):

packages/stores/src/account/cosmos/index.ts
```diff
@@ -134,9 +138,14 @@ export class CosmosAccountImpl {
...
-          ChainIdHelper.parse(destinationInfo.network).version.toString()
-        ),
+        /**
+         * Omit the revision_number if the chain's version is 0.
+         * Sending the value as 0 will cause the transaction to fail.
+         */
+        revisionNumber:
+          revisionNumber !== "0"
+            ? Long.fromString(revisionNumber)
+            : (undefined as any),
```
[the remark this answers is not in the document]
JoseRFelix: Telescope is generating the `revisionNumber` type as a required `Long`. I have to investigate if the proto is malformed or if Telescope has a bug.

packages/stores/src/account/cosmos/index.ts
JoseRFelix: This is a port from Keplr's `sendIbcMsg` that we used in our previous `AccountStore`. All other IBC transfer files are the same.

PR #1747, opened 2023-07-03
(CL) Display superfluid positions

jonator:

In the pool detail page and pools page, positions are listed and superfluid staked positions are displayed with their own status badge. Superfluid positions can also be unstaked. These positions show info about the validator being staked to, as well as the time the position will become available if it's being unstaked.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/863gqk6j2)

## Testing and Verifying

For a staked position
- Show it's status
- Show the amount of OSMO staked
- Show the validator image, commission
- Show APR (may need to be fixed)
- Show how long it will take to become available once unstaked
- User can increase liquidity (a different message is used to increase staked position liq)
- "Unstake" instead of "Remove" liquidity button

For an unstaking position
- Status is restored to full range
- Time assets will become available is displayed
- Other actions are unavailable


JoseRFelix (commented):
Left a small nit

PR #1748, opened 2023-07-03
(CL) e2e test CL swap in

JoseRFelix:


packages/stores/src/account/__tests_e2e__/create-pool.spec.ts
```diff
@@ -3,15 +3,18 @@ import {
...
   waitAccountLoaded,
 } from "../../__tests_e2e__/test-env";
 
 describe("Create Pool Tx", () => {
   const { accountStore } = new RootStore();
 
+  let account: ReturnType<(typeof accountStore)["getWallet"]>;
   beforeAll(async () => {
-    const account = accountStore.getWallet(chainId);
+    await initAccount(accountStore, chainId);
```
jonator, line 16: Can we just have initAccount do everything given the params:
* Get the account
* Wait for it to be loaded
* Finally return a _defined_ account

jonator (changes requested):

jonator:
@JoseRFelix taking this over

PR #1749, opened 2023-07-03
(CL) Stake existing full range pos

jonator:

If a user has a full range position in a superfluid enabled pool, this button allows them to stake it. As with balancer pools, it advertises the APR on the button, and includes superfluid styling. If a SF position already exists, it reuses the validator address.

### ClickUp Task

## Brief Changelog

- "Go superfluid" button, with +APR
If not SF staked in this pool before and a full range position, the expanded view includes:
- "Go superfluid" button, with APR
- Opens select SF validator modal (reused from balancer)
- Triggers staking message

Try again with a fresh full range position with the above position staked:
- Validator selection should be skipped and tx should pop up

TODO
- [x] Validator selection modal CSS seemed to have been broken in the past. Fix it


packages/proto-codecs/src/codegen/osmosis/cosmwasmpool/v1beta1/genesis.ts
```diff
@@ -97,3 +153,81 @@ export const GenesisState = {
     };
   },
 };
+export const PoolI_InterfaceDecoder = (
```
MaxMillington, line 156: This section looks good, but one thing I notice is that there isn't really any error handling. Given what we have seen in the sentry, is it worth wrapping in an error handler?

packages/proto-codecs/src/codegen/osmosis/superfluid/tx.ts
```diff
@@ -2452,3 +2495,225 @@ export const MsgAddToConcentratedLiquiditySuperfluidPositionResponse = {
...
+  decode(
+    input: _m0.Reader | Uint8Array,
+    length?: number
+  ): MsgLockExistingFullRangePositionAndSFStake {
+    const reader = input instanceof _m0.Reader ? input : new _m0.Reader(input);
+    const end = length === undefined ? reader.len : reader.pos + length;
+    const message = createBaseMsgLockExistingFullRangePositionAndSFStake();
+    while (reader.pos < end) {
+      const tag = reader.uint32();
+      switch (tag >>> 3) {
```
MaxMillington, line 2531: could the `tag` be negative? If so, you might want to use `>>`

MaxMillington (approved):
Lot's of scary messaging code, otherwise looks fine

packages/proto-codecs/src/codegen/osmosis/superfluid/tx.ts
jonator: This is generated 😅

packages/proto-codecs/src/codegen/osmosis/cosmwasmpool/v1beta1/genesis.ts
jonator: I think errors here would be unexpected

PR #1757, opened 2023-07-05
fix: Pool Creation Broadcast

JoseRFelix:

This PR fixes the pool creation flow, and this message: 

image

The error pertains to a discrepancy between the amino and proto messages. The proto messages are incapable of incorporating decimal numbers because they are converted into a `BigInt` on the chain side. Considering our need for a decimal number in the amino message, we can still maintain the decimals in the `toAmino`. However, we need to adjust the `fromAmino` function. This function represents the body passed to the proto message and must accommodate a decimal format compatible with the proto buffer. 

To clarify our transaction signing process: we begin by obtaining the amino message object via the `toAmino` function, which is then signed using the wallet. This signed object is then transformed into a proto-compatible message using the `fromAmino` function and encoded accordingly with the proto registry. Following this process, our message is prepared and ready for broadcast.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/85ztc3hdp)

## Brief Changelog

- update `fromAmino` To convert amino decimal to compatible proto values

## Testing and Verifying

- [ ] Can create balancer pool and data is consistent
- [ ] Can create stableswap pool and data is consistent

### DIFF
--- a/packages/stores/src/account/amino-converters.ts
+++ b/packages/stores/src/account/amino-converters.ts
@@ -10,6 +10,8 @@ import { MsgLockTokens } from "@osmosis-labs/proto-codecs/build/codegen/osmosis/
 import { MsgTransfer } from "cosmjs-types/ibc/applications/transfer/v1/tx";
 import Long from "long";
 
+import { changeDecStringToProtoBz } from "./utils";
+
 const osmosisAminoConverters: Record<
   keyof typeof originalOsmosisAminoConverters,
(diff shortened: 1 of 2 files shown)


jonator (approved):
This makes me v sad. But LGTM

jonator:
Is there any way this can be automated? This is incredible given the simplicity of the issue.. @pyramation

PR #1771, opened 2023-07-07
fix: Display Promo and Update Codecs With New Amino Names

JoseRFelix:
## Brief Changelog

- Update proto codecs to include new amino names
- Fix promo banner height


jonator (approved):
Yayyyyyyyy 🎉🥳

PR #1775, opened 2023-07-07
fix: invert in over out spot price

jonator:

When calculating the display spot price for a CL pool, we messed up the in over out logic depending on the token being swapped in. This should fix the swap tool for CL pools.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86780cqt2)


jonator:
🤠

PR #1779, opened 2023-07-08
One click migrate to CL

jonator:

We used to display a modal to present a user with options of what shares to migrate to the linked CL pool. Now, when they click migrate, we immediately display a multi message for them to sign that migrates all forms of gamm shares. ALSO, if the user has migrate-able shares, the pools page will display the same CTA with the same one click multi message to immediately migrate those shares.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/862k3kj40)

## Testing and Verifying

Create shares of various forms to migrate, and notice the corresponding multi message that appears immediately for migrating. Both on balancer pool and pools page when shares are owned. Sending the message should redirect you to the new pool with the new migrated positions.


packages/stores/src/account/osmosis/index.ts
```diff
@@ -1524,7 +1542,7 @@ export class OsmosisAccountImpl {
       owner: this.address,
       coins: primitiveTokens,
       duration: {
-        seconds: BigInt(duration),
+        seconds: BigInt(Math.floor(duration / 1_000)),
```
JoseRFelix, line 1545: Maybe we should add a comment to understand why this is needed.

packages/web/components/funnels/concentrated-liquidity/use-cfmm-to-cl-migration.ts
```diff
@@ -23,66 +27,103 @@ export function useCfmmToClMigration(cfmmPoolId: string): {
...
+  );
+
+  const firstClLinkedUserPoolId = userPoolIds.find((poolId) =>
+    Boolean(
+      osmosisQueries.queryCfmmToConcentratedLiquidityPoolLinks.get(poolId)
+        .concentratedLiquidityPoolId
+    )
+  );
+
+  const cfmmPoolId_ = cfmmPoolId ?? firstClLinkedUserPoolId ?? "";
```
JoseRFelix, line 45: What do you think about renaming the param to cfmmPoolIdParam and remove _ from the variable.

JoseRFelix (commented):
Left some small nits

JoseRFelix (approved):

PR #1780, opened 2023-07-09
test/fix: more descriptive error message when failing quotes due to lack of precision

p0mvn:
…

The issue I've experienced is not being unable to generate quotes due to a lack of precision.

Precision must be increased to 36 post-launch for quote generation to work correctly in all cases.

If liquidity is high enough here:
https://github.com/osmosis-labs/osmosis-frontend/blob/c7889f4e4186bdb5862abc9a0275aeb5f3d377d6/packages/math/src/pool/concentrated/math.ts#L145

It will divide `amountRemaining` all the way to zero, leaving the sqrt price with no update. That will lead to the new sqrt price being equal to the current sqrt price. As a result, we failed to calculate the quoted amount and swap.

## Brief Changelog
- added a more descriptive error to all swap steps
- added a repro test

## Testing and Verifying

Covered by test

### DIFF
--- a/packages/math/src/pool/concentrated/__tests__/quotes.spec.ts
+++ b/packages/math/src/pool/concentrated/__tests__/quotes.spec.ts
@@ -1,5 +1,6 @@
 import { Coin, Dec, Int } from "@keplr-wallet/unit";
 
+import { maxTick } from "../const";
 import { calcInGivenOut, ConcentratedLiquidityMath } from "../quotes";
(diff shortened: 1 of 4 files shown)


jonator (changes requested):
Can you repeat for in given out? So we don't miss this in the future when we finally support in given out

p0mvn:
> Can you repeat for in given out? So we don't miss this in the future when we finally support in given out

Yessir, already done for both "out given in" and "in given out"

jonator (approved):

PR #1784, opened 2023-07-10
Fix input price

jonator:

Use Dec to support larger direct price inputs. Note: it gets rounded to the nearest tick, so some values are not possible to directly input (especially larger prices). The original issue was likely caused by the Intl number formatter, so I created a way to get around that for Dec type. I use the IntPretty type to trim decimals still. Now all prices (even when rounded to ticks) are derived from raw user input, instead of being set as an action. You'll notice this when the input box is focused, the raw value is there but the derived value is displayed.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86780cqt2)

## Brief Changelog

- Support large prices with Dec type

TODO
- [x] Support decimal input
- [x] Fix using slider

## Testing and Verifying

Enter large prices in both inputs.

### DIFF
--- a/packages/web/components/complex/add-conc-liquidity.tsx
+++ b/packages/web/components/complex/add-conc-liquidity.tsx
@@ -733,7 +733,7 @@ const PriceInputBox: FunctionComponent<{
             // otherwise, display the nearest tick rounded price
             isFocused
               ? config.rangeRaw[forPriceIndex]
-              : formatPretty(config.range[forPriceIndex])
+              : formatPretty(config.range[forPriceIndex], { maxDecimals: 8 })
           }
           onFocus={() => setIsFocused(true)}
           onInput={(val) =>
--- a/packages/web/components/complex/add-conc-liquidity.tsx
+++ b/packages/web/components/complex/add-conc-liquidity.tsx
@@ -306,8 +306,8 @@ const AddConcLiqView: FunctionComponent<
 
   const updateInputAndRangeMinMax = useCallback(
     (min: number, max: number) => {
-      setMinRange(new Dec(min));
-      setMaxRange(new Dec(max));
+      setMinRange(min.toString());
+      setMaxRange(max.toString());
     },
     [setMinRange, setMaxRange]
   );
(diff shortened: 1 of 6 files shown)


JoseRFelix (approved):

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -710,31 +706,44 @@ const PresetStrategyCard: FunctionComponent<
...
-    {infinity ? (
-      <div className="flex h-[41px] items-center px-2">
-        <Image
-          alt="infinity"
-          src="/icons/infinity.svg"
-          width={16}
-          height={16}
+  forPriceIndex: 0 | 1;
+  addConcLiquidityConfig: ObservableAddConcentratedLiquidityConfig;
+}> = observer(({ label, forPriceIndex, addConcLiquidityConfig: config }) => {
```
JoseRFelix, line 711: I think it's preferable to keep the variable name as 'addConcLiquidityConfig'. While it's lengthier, it has the advantage of being self-explanatory, eliminating the need to read through the parameters.

```suggestion
}> = observer(({ label, forPriceIndex, addConcLiquidityConfig }) => {
```

packages/web/utils/formatter.ts
```diff
@@ -5,21 +5,37 @@ import {
...
 
 /** Formats a pretty object as compact by default. i.e. $7.53M or $265K, or 2K%. Validate handled by pretty object. */
 export function formatPretty(
   prettyValue: PricePretty | CoinPretty | RatePretty | Dec,
-  opts?: Partial<Intl.NumberFormatOptions>
+  opts: FormatOptions = {}
 ) {
-  const { ...formatOpts } = opts || {};
+  const optsWithDefaults: FormatOptionsWithDefaults = {
+    ...{ ...DEFAULT },
```
JoseRFelix, line 28:
```suggestion
    ...DEFAULT,
```

packages/web/utils/formatter.ts
```diff
@@ -28,24 +44,39 @@ export function formatPretty(
...
+  opts: FormatOptionsWithDefaults = DEFAULT
 ): string {
   const options: Intl.NumberFormatOptions = {
     maximumSignificantDigits: 3,
     notation: "compact",
     compactDisplay: "short",
     ...opts,
   };
-  let num = Number(new IntPretty(dec).maxDecimals(2).locale(false).toString());
+  console.log("opts max dec", opts.maxDecimals);
```
JoseRFelix, line 55: Is this console.log intentional?

JoseRFelix (changes requested):

packages/web/utils/formatter.ts
JoseRFelix: We should probably add a comment to prevent future removal of this section. At a glance, it appears to be an accidental duplicate clone.

PR #1785, opened 2023-07-10
(CL) Match Pick Strategy with Figma and Fix Lottie

JoseRFelix:
## Brief Changelog

- Match Pick Strategy with Figma
- Fix Lottie visual issue


packages/web/components/funnels/concentrated-liquidity/learn-more.tsx
```diff
@@ -10,9 +10,24 @@ import {
   StepsIndicator,
 } from "~/components/stepper";
 
+const lotties: Record<string, Record<string, any>> = {};
```
jonator, line 13: This is safe?

jonator (approved):

packages/web/components/funnels/concentrated-liquidity/learn-more.tsx
JoseRFelix: It's pretty safe. There's no need for React to re-render, as we will consistently have the most recent result in the `loadLottie` function.

PR #1786, opened 2023-07-10
mattupham/ad banner remote cms

mattupham:

- link ad banner to remote "CMS"

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/8678aq0a3)

## Brief Changelog

- fetch from github
- don't render ads if empty list
- update types / schema

### DIFF
--- a/packages/web/pages/index.tsx
+++ b/packages/web/pages/index.tsx
@@ -27,7 +27,7 @@ export const getStaticProps: GetStaticProps<HomeProps> = async () => {
     console.error("Error fetching ads:", error);
   }
 
-  return { props: { ads } };
+  return { props: { ads }, revalidate: 3600 };
 };
 
 const Home = ({ ads }: InferGetServerSidePropsType<typeof getStaticProps>) => {


packages/web/pages/index.tsx
```diff
@@ -18,8 +18,19 @@ interface HomeProps {
   ads: Ad[];
 }
 
+const ADS_URL =
```
jonator, line 21: Let's put this somewhere in config folder

packages/web/pages/index.tsx
```diff
@@ -18,8 +18,19 @@ interface HomeProps {
...
+  "https://raw.githubusercontent.com/osmosis-labs/fe-content/main/cms/swap-rotating-banner.json";
+
 export const getStaticProps: GetStaticProps<HomeProps> = async () => {
```
MaxMillington, line 20: I think we will need to use the `revalidate` option in `getStaticProps` here to bust the cache every once in a while. Maybe once an hour or something like that. Otherwise, the cache only updates on every rebuild (i.e. every deploy)

MaxMillington: Sounds good to me. We could always update it if need be.

MaxMillington (approved):

jonator (approved):

PR #1795, opened 2023-07-11
Display position performance metrics

jonator:

Fetch a position ID's performance metrics and display in the position card; principal assets when position was created, and the sum of rewards earned in that position. We use those values and the current assets to calculate the ROI.

### ClickUp Task

## Brief Changelog

- Position metrics query from the Imperator indexer service
- ROI calculation

## Testing and Verifying

Hardcode the position ID to `107910` to see real data in the my-position/index and my-position/expanded components.

See: https://api-osmosis-chain-testnet.imperator.co/cl/v1/position/id/107910


packages/stores/src/queries-external/position-performance/index.ts
```diff
@@ -0,0 +1,203 @@
...
+  @computed
+  get totalPrincipalValue(): PricePretty {
+    if (!this.response) return new PricePretty(this.fiatCurrency, 0);
+    return new PricePretty(
+      this.fiatCurrency,
+      this.response.data?.principal?.value ?? 0
+    );
+  }
+
+  readonly calculateReturnOnInvestment = computedFn(
```
MaxMillington, line 117: seems like there is some similar logic here to whats in `totalEarned`. Can we make it more dry?

PR #1799, opened 2023-07-12
(CL) Move root functions to `utils.ts` and implement naive monotonic sqrt function

AlpinYukseloglu:

This PR implements an alternative square root function that should better approximate monotonicity and hopefully solve the class of FE issues related to square root approximation.

This implementation is drastically simpler than a truly monotonic square root function, which I will implement only if absolutely necessary as it will be more overhead to write & review.

Pending @p0mvn's confirmation that this function solves the issues in his branch, this approach is infinitely better on the time/value tradeoff.

### ClickUp Task

## Brief Changelog

This PR makes two changes:
1. Moves `approxRoot` and related files to higher level `utils.ts` file
2. Implements a _mostly_ monotonic square root function that should be extremely simple to review and merge

## Testing and Verifying

- The `approxRoot` tests were left in the concentrated file as there didn't seem to be a natural place to move them to at the higher level folder. Happy to change this if needed.
- Correctness for the monotonic sqrt function will ultimately be checked in @p0mvn in progress branch increasing precision of CL math

### DIFF
--- a/packages/math/src/pool/concentrated/__tests__/math.spec.ts
+++ b/packages/math/src/pool/concentrated/__tests__/math.spec.ts
@@ -1,9 +1,9 @@
 import { Dec } from "@keplr-wallet/unit";
 
 import { checkMultiplicativeErrorTolerance } from "../../../rounding";
+import { approxRoot } from "../../../utils";
 import { smallestDec } from "../const";
 import {
-  approxRoot,
   calcAmount0Delta,
   calcAmount1Delta,
   getFeeChargePerSwapStepOutGivenIn,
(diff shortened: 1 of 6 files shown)


AlpinYukseloglu:
Converted to draft as it seems we're going to need a more involved solution

jonator (approved):
> The approxRoot tests were left in the concentrated file as there didn't seem to be a natural place to move them to at the higher level folder. Happy to change this if needed.

It's fine to create a `__TESTS__` folder at top level and move them there. They should ideally be placed at the same dir level as the code under test, for navigability.

PR #1808, opened 2023-07-12
fix: redirect pool detail of CL FF is off

jonator:
### ClickUp Task

### DIFF
--- a/packages/web/pages/pool/[id].tsx
+++ b/packages/web/pages/pool/[id].tsx
@@ -68,7 +68,8 @@ const Pool: FunctionComponent = observer(() => {
           memoedPools={[queryPool]}
         />
       )}
-      {queryPool?.type === "concentrated" ? (
+      {featureFlags.concentratedLiquidity &&
+      queryPool?.type === "concentrated" ? (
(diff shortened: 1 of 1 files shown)


packages/web/pages/pool/[id].tsx
```diff
@@ -44,6 +46,16 @@ const Pool: FunctionComponent = observer(() => {
...
+  useEffect(() => {
+    if (
+      queryPool &&
+      !featureFlags.concentratedLiquidity &&
+      queryPool.type === "concentrated"
+    ) {
+      router.push(`/pools`);
+    }
+  }, [queryPool, featureFlags.concentratedLiquidity, router]);
+
```
Comment (JoseRFelix) [inline packages/web/pages/pool/[id].tsx:58]: Maybe we should also add the flag verification in the jsx below like:

```
{featureFlags.concentratedLiquidity && queryPool?.type === "concentrated" ? (
        <ConcentratedLiquidityPool poolId={poolId} />
      ) : (
        queryPool && <SharePool poolId={poolId} />
      )}
```

JoseRFelix (approved):

PR #1825, opened 2023-07-13
update ld config to check for laptoporlarger

MaxMillington:

Adding the 'laptopOrLarger' flag to our Launch Darkly initialization. This should allow us to target whether or not a user is on mobile when LD initializes i.e. and hide certain features from mobile users.

## Brief Changelog

- Add laptopOrLargerFlags
- bunch of linting fixes

## Testing and Verifying


jonator (commented):
Is there a way for this to only affect CL FF?

PR #1828, opened 2023-07-13
(CL) Add 1h Price Range to CL Historical Chart

JoseRFelix:


packages/web/components/chart/token-pair-historical.tsx
```diff
@@ -206,6 +206,11 @@ export const PriceChartHeader: FunctionComponent<{
               classes?.buttons
             )}
           >
+            <ChartButton
+              label="1 hour"
```
jonator, line 210: Can we localize this?

sunnya97:
Do we still want to do this **right now**, if we're going to use the classical pool chart instead?  We still will need it for new assets later

jonator:
@sunnya97 should we use classical chart or add 1h time range?

MaxMillington:
Any updates on this PR? Should we close it if it isn't actively being worked on?

jonator:
It's working for me: 
Screenshot 2023-08-10 at 5 55 09 PM

@JoseRFelix thoughts?

jonator:
Can reference this branch for adding it in the future, but it is quite stale

PR #1829, opened 2023-07-13
Fix price decimals

jonator:
### ClickUp Task


packages/stores/src/queries-external/token-pair-historical-chart/index.ts
```diff
@@ -25,7 +25,7 @@ export class ObservableQueryTokenPairHistoricalChart extends ObservableQueryExte
     super(
       kvStore,
       baseURL,
-      `/pairs/v1/historical/${poolId}/chart?asset_in=${baseDenom}&asset_out=${quoteDenom}&range=${priceRange}&asset_type=symbol`
+      `/pairs/v1/historical/${poolId}/chart?asset_in=${quoteDenom}&asset_out=${baseDenom}&range=${priceRange}&asset_type=symbol`
```
jonator, line 28: We were passing in the wrong denoms as params to the query the whole time

packages/stores/src/queries/concentrated-liquidity/position-by-id.ts
jonator, line 1: Adjust the price per position

packages/stores/src/queries/pools/pool.ts
jonator, line 1: Adjust the current price (not sqrt) per pool

sunnya97:
Is fixed in pool details page, but not pools overview

CleanShot 2023-07-14 at 04 15 28@2x

jonator:
@sunnya97 strange, that is only happening to me in previews

sunnya97:
> @sunnya97 strange, that is only happening to me in previews

wdym?  Its not happening on local machine?

jonator:
> > @sunnya97 strange, that is only happening to me in previews
> 
> wdym? Its not happening on local machine?

No I can't repro locally

sunnya97:
What?  Why would that be??  Does it have to do with some query whitelists?  Or maybe something launch darkly related?

jonator:
Preview:
Screenshot 2023-07-14 at 11 58 12 AM

Local dev:
Screenshot 2023-07-14 at 11 58 24 AM

sunnya97:
wtf

sunnya97 (approved):

PR #1837, opened 2023-07-14
fix: don't show chart unavailable when loading

jonator:

When the user changed the chart time range, the chart would flicker "unavailable" while loading. Now, we check if it's loading and only show if we know there's no data.

### ClickUp Task


JoseRFelix (approved):

sunnya97:
Can we have it show a loading animation?

jonator:
> Can we have it show a loading animation?

Are you thinking a spinner? That would be easiest right now

sunnya97:
Yeah

jonator:
@sunnya97 added loader

PR #1839, opened 2023-07-14
Update learn more copy & animations

jonator:
### ClickUp Task


sunnya97:
I think Syed / the animator will have new copy / animations very soon.  Let's wait for that

jonator:
> I think Syed / the animator will have new copy / animations very soon. Let's wait for that

Yep I just have step 1 so far

sunnya97:
@jonator do you know how to make the lottie animations restart from the beginning every time they come you come to their page?  It feels like they start from a random time (but that might be because they all are running even when they're hidden)

sunnya97:
Merging this and creating a new issue for restarting animations

PR #1843, opened 2023-07-14
(CL) Fix Depth Chart

JoseRFelix:
### Brief fhangelog

- Display depth chart even if historical data is not available
- Show accurate depth data in chart


packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
```diff
@@ -245,14 +245,31 @@ export class ObservableHistoricalAndLiquidityData {
...
         price,
         depth: getLiqFrom(
-          priceToTick(new Dec(spotPrice).quo(this.multiplicationQuoteOverBase)),
+          priceToTick(
+            spotPriceToConvert.gt(maxSpotPrice)
+              ? maxSpotPrice
+              : spotPriceToConvert.lt(minSpotPrice)
+              ? minSpotPrice
+              : spotPriceToConvert
+          ),
```
jonator: Looks like were checking for min and max in two places? Does JS number type not work here? If so, let's just use the Dec type to compare and check for min and max (not normalized)

jonator (approved):

PR #1844, opened 2023-07-14
enforce absolute imports

mattupham:

- note, right now these rules only applies to packages/web

We're using a mix of relative and absolute imports - absolute imports are clear and easier to work with. We should stay consistent

It's messy to have 2 different ways of importing files, and relative imports have a higher risk of causing issues

- vscode automatically detects our tsconfig and import absolute by default as well
- this is something as devs we shouldn't even have to think about, let the settings / config handle this for us


jonator:
Why is this necessary? It adds more clutter in imports. Is there a functional difference?

MaxMillington (approved):
Makes sense to me

PR #1847, opened 2023-07-14
fix: direct swap with price impact broken by router check

p0mvn:

Context: https://osmosis-network.slack.com/archives/C05F9ABFKQX/p1689366314885689

The router check disallows swaps when swapping more than pool's balance in directly from pool's page.

Discussed with @jonator and agreed to only run this check in the router whenever there are multiple pools available.

## Testing and Verifying

- Tested manually on edgenet


packages/pools/src/router/routes.ts
```diff
@@ -146,6 +146,8 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
 
     let routes = this.getCandidateRoutes(tokenIn.denom, tokenOutDenom);
 
+    const candidateRoutesLength = routes.length;
```
jonator, line 149: Let's not bother pulling this var

jonator (changes requested):

packages/pools/src/router/routes.ts
```diff
@@ -161,18 +163,20 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
...
-        route.pools[0].getLimitAmountByTokenIn(tokenIn.denom)
-      )
-    );
-    routes = routes.filter((_, i) =>
-      routesInitialLimitAmounts[i].gte(tokenIn.amount)
-    );
+    // Only perform the filter if there are many routes.
+    // For a direct swap or, if there is only one route
+    // available, use it as is.
+    if (candidateRoutesLength > 1) {
```
jonator, line 169: Remember: routes have multiple pools, should we also check that the one route has one pool?

jonator (changes requested):

packages/pools/src/router/routes.ts
p0mvn: The core of the issue is that we might filter out the only possible route with a filter that is only useful for selecting between multiple routes. 

So I think it might be beneficial to keep it as is. WDYT?

jonator (approved):

jonator (commented):
@p0mvn my only concern now, is since we're iterating over so many ticks with larger quotes it can be quite laggy (on an M1 mac) with larger values. Is this worth it?

jonator:
Going to close for now to prevent browser freezing from iterating over too many ticks.

jonator:
Checked again with router improvements. Still not worth it given the absurdity of price impact at those amounts.

PR #1850, opened 2023-07-15
Zooming

jonator:
### ClickUp Task

### DIFF
--- a/packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
+++ b/packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
@@ -17,8 +17,8 @@ import {
 import { action, computed, makeObservable, observable } from "mobx";
 import { DeepReadonly } from "utility-types";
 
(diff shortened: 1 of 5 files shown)


packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
```diff
@@ -105,7 +108,7 @@ export class ObservableHistoricalAndLiquidityData {
     if (!this.pool || this.pool.type !== "concentrated") return new Dec(0);
     return (
       this.pool.concentratedLiquidityPoolInfo?.multiplicationQuoteOverBase ??
-      new Dec(0)
+      new Dec(1)
```
sunnya97, line 111: Why did this change?

jonator: When the pool is loading, the spot price is 0, causing the multiplicationQuoteOverIn to be 0, causing divide by 0 exception in the depthChart getter

PR #1851, opened 2023-07-15
fix: use market value for asset %s

jonator:
### ClickUp Task


sunnya97:
Always adds up to 99% but that's fine.  We can fix this later.

sunnya97 (approved):

PR #1852, opened 2023-07-15
set final price in graphs to live spot price

sunnya97:
### ClickUp Task

### DIFF
--- a/packages/web/components/complex/add-conc-liquidity.tsx
+++ b/packages/web/components/complex/add-conc-liquidity.tsx
@@ -522,6 +522,7 @@ const Chart: FunctionComponent<{
   addLiquidityConfig: ObservableAddConcentratedLiquidityConfig;
 }> = observer(({ addLiquidityConfig, chartConfig }) => {
   const { fullRange, rangeWithCurrencyDecimals } = addLiquidityConfig;
+
   const { yRange, historicalChartData, lastChartData, setHoverPrice } =
     chartConfig;
 
--- a/packages/web/components/complex/add-conc-liquidity.tsx
+++ b/packages/web/components/complex/add-conc-liquidity.tsx
@@ -522,7 +522,6 @@ const Chart: FunctionComponent<{
   addLiquidityConfig: ObservableAddConcentratedLiquidityConfig;
 }> = observer(({ addLiquidityConfig, chartConfig }) => {
   const { fullRange, rangeWithCurrencyDecimals } = addLiquidityConfig;
-
   const { yRange, historicalChartData, lastChartData, setHoverPrice } =
     chartConfig;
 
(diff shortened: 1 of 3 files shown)


packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -522,6 +522,7 @@ const Chart: FunctionComponent<{
   addLiquidityConfig: ObservableAddConcentratedLiquidityConfig;
 }> = observer(({ addLiquidityConfig, chartConfig }) => {
   const { fullRange, rangeWithCurrencyDecimals } = addLiquidityConfig;
+
```
sunnya97, line 525:
```suggestion
```

packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
```diff
@@ -119,6 +119,16 @@ export class ObservableHistoricalAndLiquidityData {
...
 
+  @action
+  setLastChartData = (setPrice: number) => {
+    this.historicalChartData[this.historicalChartData.length - 1].high =
+      setPrice;
+    this.historicalChartData[this.historicalChartData.length - 1].low =
+      setPrice;
+    this.historicalChartData[this.historicalChartData.length - 1].close =
+      setPrice;
+  };
```
jonator, line 130: Eh this may not be a good way to do it, modifying the query store..

Instead, I'd get the current price from pool and return it in `lastChartData`:
```
  @computed
  get lastChartData(): TokenPairHistoricalPrice | null {
    const price = Number(
      this.queries.queryPools.getPool(this.poolId)
        ?.concentratedLiquidityPoolInfo?.currentPrice || 0
    );
    return {
      close: price,
      high: price,
      low: price,
      open: price,
      time: new Date().getTime(),
    };
  }
```

packages/web/components/pool-detail/concentrated.tsx
```diff
@@ -364,6 +364,12 @@ const ChartHeader: FunctionComponent<{
 const Chart: FunctionComponent<{
   config: ObservableHistoricalAndLiquidityData;
 }> = observer(({ config }) => {
+  if (config?.pool?.concentratedLiquidityPoolInfo) {
+    config.setLastChartData(
+      Number(config.pool.concentratedLiquidityPoolInfo.currentPrice)
+    );
+  }
```
jonator, line 371: We should wrap this in a useEffect to only set it on initial render, and pass in `config?.pool?.concentratedLiquidityPoolInfo` & `config` in the dependency array of useEffect (second param)

jonator (changes requested):

packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
sunnya97: But lastChartData isn't what's used when rendering the price chart

packages/web/components/pool-detail/concentrated.tsx
sunnya97: Uh, idk what useEffect does or means 😅

sunnya97:
I'm just going to merge for now, please fix it if needed.

PR #1854, opened 2023-07-16
fix initial load of add liquidity

sunnya97:
### ClickUp Task


sunnya97:
This is a hacky solution but it will work for now

PR #1857, opened 2023-07-16
fix near bounds calculation

sunnya97:
### ClickUp Task

### DIFF
--- a/packages/stores/src/currency-registrar/index.ts
+++ b/packages/stores/src/currency-registrar/index.ts
@@ -11,6 +11,7 @@ export class LPCurrencyRegistrar<C extends ChainInfo = ChainInfo> {
   protected readonly registerLPCurrency = (
     coinMinimalDenom: string
   ): AppCurrency | [AppCurrency | undefined, boolean] | undefined => {
+    // if (!coinMinimalDenom) throw new Error("Missing coinMinimalDenom");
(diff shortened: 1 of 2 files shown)


packages/stores/src/currency-registrar/index.ts
```diff
@@ -11,6 +11,7 @@ export class LPCurrencyRegistrar<C extends ChainInfo = ChainInfo> {
   protected readonly registerLPCurrency = (
     coinMinimalDenom: string
   ): AppCurrency | [AppCurrency | undefined, boolean] | undefined => {
+    // if (!coinMinimalDenom) throw new Error("Missing coinMinimalDenom");
```
sunnya97, line 14:
```suggestion
```

packages/web/components/cards/my-position/status.tsx
```diff
@@ -47,9 +47,11 @@ export const MyPositionStatus: FunctionComponent<
     )
   );
 
+  const rangeDiff = new Dec(Number(upperPrice.sub(lowerPrice).toString()));
```
jonator, line 50: I think this Number conversion is not necessary. You should be able to directly use the Dec objects.

jonator (changes requested):

packages/web/components/cards/my-position/status.tsx
sunnya97: k

PR #1878, opened 2023-07-17
fix: init hover price to current pool price

jonator:

Also update copy in learn more instagram flow

### ClickUp Task

### DIFF
--- a/packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
+++ b/packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
@@ -14,11 +14,13 @@ import {
   PriceRange,
   TokenPairHistoricalPrice,
 } from "@osmosis-labs/stores";
-import { action, computed, makeObservable, observable } from "mobx";
+import { action, autorun, computed, makeObservable, observable } from "mobx";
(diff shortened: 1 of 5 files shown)


packages/web/stores/derived-data/concentrated-liquidity/historical-and-liquidity-data.ts
```diff
@@ -45,6 +47,11 @@ export class ObservableHistoricalAndLiquidityData {
     protected readonly queryTokenPairHistoricalPrice: DeepReadonly<ObservableQueryTokensPairHistoricalChart>
   ) {
     makeObservable(this);
+
+    // Init last hover price to current price in pool once loaded
+    autorun(() => {
+      if (this.lastChartData) this.setHoverPrice(this.lastChartData.close);
```
JoseRFelix, line 53: Is there a risk for this to override the current hover value when data is refreshed? It could be clunky when it happens.

jonator: Yeah, but the when hovering the the chart's value is rapidly set.

sunnya97:
Just fyi, hover doesn't work on the expanded position graphs (didn't work before this PR as well)

sunnya97 (approved):

PR #1881, opened 2023-07-17
remove tally feedback banner

MaxMillington:
Removing the tally feedback banner per syed's request


MaxMillington:
@sunnya97 We should still probably remove this stale code don't you think?

PR #1913, opened 2023-07-24
Remove local import rule

mattupham:
### ClickUp Task

### DIFF
--- a/packages/web/.eslintrc.json
+++ b/packages/web/.eslintrc.json
@@ -22,7 +22,7 @@
     "no-restricted-imports": [
       "error",
       {
-        "patterns": [".", "..", "./*", "../*"]
+        "patterns": [".", "..", "../*"]
       }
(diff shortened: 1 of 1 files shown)


packages/web/.eslintrc.json
```diff
@@ -22,7 +22,7 @@
     "no-restricted-imports": [
       "error",
       {
-        "patterns": [".", "..", "./*", "../*"]
+        "patterns": [".", "..", "../*"]
```
jonator, line 25:
```suggestion
        "patterns": [".", ".."]
```

Let's remove this as well, as it can be useful to import nearby components in the folder above.

jonator: @mattupham wait I think the recent commit removed the opposite of the suggestion?

jonator:
Reverted import rule

PR #1914, opened 2023-07-24
Subtract gas in add conc liq

jonator:

When creating a position with OSMO and clicking max, subtract amount of OSMO needed to pay max gas.

### ClickUp Task


czarcas7ic (approved):
Tested and worked locally

PR #1920, opened 2023-07-25
Upgrade flow

jonator:

Add a general upgrade flow that will be used to migrate users to all of the new CL pools, as well as other upgrade flows in the future.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/860rcqtct)

## Brief Changelog

* Upgrade store in root store
* Upgrade button in nav bar
* Opens upgrade modal with list of upgrades, including info about each upgrade
* When viewing for first time and if you have upgrades, the upgrades modal pops up. After, a tutorial card appears under upgrades informing the user that they can return to it via the nav bar button. This happens once and is remembered via local storage.
* Updated pools page CTA card to be more general, and to open the upgrades modal instead of the tx directly
* Updated the pool detail page modal for balancer pools to open the upgrades modal with only that pool's upgrade, to allow user to see more info about what is going to happen

## Testing and Verifying

* Use incognito windows to ensure the one time modal state is working. Ensure one time modal state works gracefully with existing modals, like the CL intro modal.
* Ensure pools page CTA uses new modal
* Ensure balancer pool CTA uses new modal, with only that pool's upgrade visible.

This feature is anticipated to be used for all future CL pool migrations, as well as other future migrations such as converting users to the canonical form of assets in the future.


MaxMillington (approved):

sunnya97:
@jonator if APRs aren't fetchable, should make them hidden rather than show 0%

PR #1922, opened 2023-07-25
Unbonding UI, stake/unstake calls

MaxMillington:

Add the unbonding UI along with the real calls for staking and unstaking.

The estimated unstaking time is hardcoded, will need to make dynamic in separate PR

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/8678h93e2)

## Brief Changelog

Implement stake
Implement unstake
Generated types/files using telescope

## Testing and Verifying

Screenshot 2023-07-31 at 5 08 11 PM


packages/stores/src/account/osmosis/index.ts
```diff
@@ -2040,6 +2039,130 @@ export class OsmosisAccountImpl {
...
+  /**
+   * Method to delegate to validator set.
+   * @param delegator The user who is trying to delegate.
+   * @param memo Transaction memo.
+   * @param stdFee Fee options.
+   * @param signOptions Signing options.
+   * @param onFulfill Callback to handle tx fulfillment given raw response.
+   */
+  async sendDelegateToValidatorSetMsg(
+    delegator: string,
```
MaxMillington, line 2119: remove as a param and get directly from account store `this.base.address`

packages/stores/src/account/osmosis/index.ts
```diff
@@ -2040,6 +2039,130 @@ export class OsmosisAccountImpl {
...
+    coin: Coin, // Use the Coin object from "@keplr-wallet/unit"
+    memo: string = "",
+    stdFee: Partial<StdFee> = {},
```
MaxMillington, line 2122: probably wont need `stdFee` or `signOptions`

packages/stores/src/account/osmosis/index.ts
```diff
@@ -2040,6 +2039,130 @@ export class OsmosisAccountImpl {
...
+   * Method to delegate to validator set.
+   * @param delegator The user who is trying to delegate.
+   * @param memo Transaction memo.
+   * @param stdFee Fee options.
+   * @param signOptions Signing options.
+   * @param onFulfill Callback to handle tx fulfillment given raw response.
+   */
+  async sendDelegateToValidatorSetMsg(
+    delegator: string,
+    coin: Coin, // Use the Coin object from "@keplr-wallet/unit"
```
MaxMillington, line 2120: perhaps define type manually above and use here instead of using keplr wallet type

packages/stores/src/account/amino-converters.ts
```diff
@@ -24,6 +24,18 @@ const osmosisAminoConverters: Record<
...
+      "/osmosis.valsetpref.v1beta1.MsgDelegateToValidatorSet"
+    ],
+    aminoType: "osmosis/MsgDelegateToValidatorSet",
+  },
+  "/osmosis.valsetpref.v1beta1.MsgUndelegateFromValidatorSet": {
+    ...originalOsmosisAminoConverters[
+      "/osmosis.valsetpref.v1beta1.MsgUndelegateFromValidatorSet"
+    ],
+    aminoType: "osmosis/MsgUndelegateFromValidatorSet",
+  },
```
jonator, line 38: Let's update the osmosis commit has to Sishir's changes so we don't have to maintain these exceptions

packages/stores/src/account/osmosis/index.ts
```diff
@@ -21,10 +21,14 @@ import { DeliverTxResponse } from "../types";
 import { findNewClPositionId } from "./tx-response";
 import { DEFAULT_SLIPPAGE, osmosisMsgOpts } from "./types";
 
+interface CoinType {
+  currency: Currency;
+  amount: string;
+  denom: Currency;
+}
```
jonator, line 28: Fe things here:

* I'd prefer this bespoke type be inlined, as it makes the parameter type of the function easier to see immediately with the IDE.
* You repeat the Currency twice, if the denom should be a string, that would also repeat the denom info found in the first Currency member

packages/stores/src/account/osmosis/index.ts
```diff
@@ -2040,6 +2044,110 @@ export class OsmosisAccountImpl {
...
+              if (
+                bal.currency.coinMinimalDenom === coin.currency.coinMinimalDenom
+              ) {
+                bal.waitFreshResponse();
+              }
+            });
+
+          queries.osmosis?.queryAccountsPositions
+            .get(this.address)
+            .waitFreshResponse();
```
jonator, line 2094: Do we need to refresh positions?

We should be intentional when copying and pasting code.

packages/stores/src/account/osmosis/index.ts
```diff
@@ -2040,6 +2044,110 @@ export class OsmosisAccountImpl {
...
+        const msg = this.msgOpts.undelegateFromValidatorSet.messageComposer({
+          delegator: this.address,
+          coin: {
+            denom: coin.denom.coinMinimalDenom,
+            amount: coin.amount,
+          },
+        });
+
+        return [msg];
+      },
```
jonator, line 2071: For both functions here, the 3rd param is a list of msg objects

I think it'd be more readable if we remove the async function and directly return the message composer return value wrapped in a array literal

packages/web/components/cards/stake-dashboard.tsx
```diff
@@ -57,12 +57,12 @@ export const StakeDashboard: React.FC<{
         <div className="flex w-full flex-row justify-between py-10">
           <StakeBalances
             title={t("stake.stakeBalanceTitle")}
-            dollarAmount={`${dollarBalance}`}
+            dollarAmount={`${fiatBalance}`}
```
jonator, line 60: Nit: `fiatBalance.toString()` may be preferrable, as done below

packages/web/components/cards/stake-info-card.tsx
```diff
@@ -54,13 +54,17 @@ export const StakeInfoCard: FunctionComponent<{
 
   const handleInputChange = useCallback(
     (e: React.ChangeEvent<HTMLInputElement>) => {
+      let value = e.target.value;
+      if (value.startsWith(".")) {
+        value = "0" + value;
+      }
```
jonator, line 60: I'm really not OK with us not using an AmountConfig to manage the user input.

Doing otherwise is error prone.

jonator (changes requested):
Skim review, did not get a chance to run it yet

packages/web/utils/get-amount-primitive.ts
jonator, line 1: I believe our coin pretty and amount config should remove the need for such a util

packages/stores/src/account/amino-converters.ts
MaxMillington: Have rerun telescope twice and the route in the message is still wrong without the amino-converter

packages/stores/src/account/osmosis/index.ts
MaxMillington: I've refactored to remove `currency` but the reason I extracted into an interface is because this is used multiple times.  This keeps it DRYer

packages/stores/src/account/osmosis/index.ts
MaxMillington: I'm not following. Please paste code to demonstrate what you mean.

packages/web/utils/get-amount-primitive.ts
MaxMillington: I was not able to figure out how to do it with coin pretty. There is a correlate function in the amount config. I put this here to have our own

packages/stores/src/account/osmosis/index.ts
jonator: Replace the whole line range with:

```
[
  this.msgOpts.undelegateFromValidatorSet.messageComposer({
    delegator: this.address,
    coin: {
      denom: coin.denom.coinMinimalDenom,
      amount: coin.amount,
    },
  })
]
```

Less cluttered

packages/stores/src/account/osmosis/index.ts
jonator: DRY is nice but not a hard and fast rule. I prefer readability, proximity, explicitness, etc. over DRYness in situations like this

packages/stores/src/account/amino-converters.ts
jonator: https://github.com/osmosis-labs/osmosis/blob/1e3149f05a08c5af8d0df9b0392bb554dfc9bc78/proto/osmosis/valset-pref/v1beta1/tx.proto#L117

The Osmosis commit hasn't been updated to include Sishir's changes, which explains the same results

packages/web/components/cards/stake-info-card.tsx
jonator: Once the previews work I'll check out the amount UX

jonator (changes requested):
Some bugs around amount input:

Exception when clicking MAX button:
Screenshot 2023-08-04 at 9 25 30 AM

Exception when clicking HALF button:
Screenshot 2023-08-04 at 9 26 36 AM


I should not be capable of adding an amount with more decimals than the OSMO currency (6):
Screenshot 2023-08-04 at 9 26 17 AM

Got this error when delegating full amount:
Screenshot 2023-08-04 at 9 28 00 AM

I suggest we set up a dev sync to discuss the usefulness of the amount config for this use case. Also UI configs in general. Thoughts?

packages/web/utils/get-amount-primitive.ts
MaxMillington:
```
toCoin() doesn't seem to do it.

return new CoinPretty(osmo, 34).toCoin() ===> {denom: 'uosmo', amount: '34'};
```

MaxMillington:
@jonator Happy to have a discussion about using the keplr amount config vs implementing this ourself, but this isn't the place to do it. This PR is already very large, and meant just to implement the stake/unstake calls and UI, and we are taking an iterative process to build this feature behind a flag. We can easily add an amount config in a separate PR. We shouldn't block further development in the meantime

jonator (commented):
Can't comment on the line as it got missed, but in validator-squad.tsx it appears the modal is wrapped twice by observer 

https://github.com/osmosis-labs/osmosis-frontend/blob/cdf318e342e94783d92375c236e66248dafb7099/packages/web/modals/validator-squad.tsx#L34

jonator (commented):
Not sure if it's related to this PR, or from @mattupham , but it appears there's an issue with the frame of the virtualization here.

Virtualization may be overkill with even ~150 validators, but that's my opinion:

(This was as I was scrolling, and I paused to take this)
Screenshot 2023-08-07 at 11 11 45 AM

jonator (approved):
Mostly thoughts and nit comments.

packages/web/components/cards/unbonding-card.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import { useTranslation } from "react-multi-lang";
+
+import { Icon } from "~/components/assets";
+import { IconLink } from "~/components/cards/icon-link";
+import { OsmoverseCard } from "~/components/cards/osmoverse-card";
+import { useWindowSize } from "~/hooks";
+
+export const UnbondingCard: React.FunctionComponent = () => {
```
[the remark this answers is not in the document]
MaxMillington: Let's discuss this in weekly dev sync.

packages/web/pages/stake/index.tsx
```diff
@@ -2,47 +2,92 @@ import { Staking as StakingType } from "@keplr-wallet/stores";
...
+
+  const primitiveAmount = amountConfig.getAmountPrimitive();
+
+  const coin = useMemo(() => {
+    return { currency: osmo, amount: primitiveAmount.amount, denom: osmo };
+  }, [osmo, primitiveAmount]);
+
+  const stakeCall = useCallback(() => {
+    if (account?.address && account?.osmosis && coin?.amount) {
+      account.osmosis.sendDelegateToValidatorSetMsg(coin);
```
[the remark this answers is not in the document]
MaxMillington: My plan is to discuss error handling with syed to see what if anything he wants to introduce design-wise and address there. Right now, it bubbles up to our existing error boundary and the only indication to the user is the error that we get back from keplr extension.

MaxMillington:
merging to unblock. Any style changes can be addressed in a followup

PR #1934, opened 2023-07-29
Fix sf apr for stable pools

jonator:

Assume even weight for stable pools.

### ClickUp Task

### DIFF
--- a/packages/stores/src/queries/superfluid-pools/osmo-equivalent.ts
+++ b/packages/stores/src/queries/superfluid-pools/osmo-equivalent.ts
@@ -86,15 +86,19 @@ export class ObservableQuerySuperfluidOsmoEquivalent {
         const ratio = stakeAsset.weight.quo(pool.weightedPoolInfo.totalWeight);
         return ratio.toDec().mul(new Dec(1).sub(minimumRiskFactor));
       } else if (pool.stableSwapInfo) {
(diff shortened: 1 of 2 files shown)


packages/stores/src/queries/superfluid-pools/osmo-equivalent.ts
```diff
@@ -86,17 +86,13 @@ export class ObservableQuerySuperfluidOsmoEquivalent {
...
-        const otherScalingFactors = pool.stableSwapInfo.assets
-          .filter(({ denom }) => denom !== osmoCurrency.coinMinimalDenom)
-          .map(({ scalingFactor }) => new Dec(scalingFactor))
-          .reduce((acc, cur) => acc.add(cur), new Dec(0));
-        if (!stakeAsset) return new Dec(0);
+        if (!hasStakeAsset) return new Dec(0);
 
-        const ratio = new Dec(stakeAsset.scalingFactor).quo(
-          otherScalingFactors
+        const ratio = new Dec(1).quo(
```
asalzmann, line 94: Can you explain what this is doing? I'm having a hard time parsing this line.

I expect that for stableswap pools, the % of the pool that is eligible for staking is % OSMO in pool * 1-minimumRiskFactor.

I'd expect % OSMO in pool to be a function of the asset ratios.

PR #1935, opened 2023-07-31
Validator "Next Step" Modals

mattupham:

- we have a set of "next steps" if the user is new to staking, as well as if they have previous staking data
- add logic to show a different options + copy if a user has previous staking data or not

### ClickUp Task

## Brief Changelog

- add a new "bullish" style for buttons
- add another modal to stake page
- generate content based on the size of the user validators list
- add logic for routing to next steps

## Testing and Verifying

new user experience modal
prev stake data modal

### DIFF
--- a/packages/web/modals/validator-next-step.tsx
+++ b/packages/web/modals/validator-next-step.tsx
@@ -1,5 +1,4 @@
 import { Staking } from "@keplr-wallet/stores";
-import { observer } from "mobx-react-lite";
 import Link from "next/link";
 import { FunctionComponent } from "react";
 import { useTranslation } from "react-multi-lang";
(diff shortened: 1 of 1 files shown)


jonator (approved):

packages/web/modals/validator-next-step.tsx
```diff
@@ -0,0 +1,92 @@
...
+interface ExtendedModalBaseProps extends ModalBaseProps {
+  usersValidatorsMap: Map<string, Staking.Delegation>;
+  setShowValidatorModal: (val: boolean) => void;
+}
+
+export const ValidatorNextStepModal: FunctionComponent<ExtendedModalBaseProps> =
+  observer((props) => <ValidatorSquadContent {...props} />);
+
+const ValidatorSquadContent: FunctionComponent<ExtendedModalBaseProps> =
+  observer(
```
MaxMillington, line 19: Why is the `observer` needed?

MaxMillington (approved):
One question otherwise LGTM

packages/web/modals/validator-next-step.tsx
```diff
@@ -0,0 +1,92 @@
...
+    ({ onRequestClose, isOpen, usersValidatorsMap, setShowValidatorModal }) => {
+      // i18n
+      const t = useTranslation();
+
+      const isNewUser = usersValidatorsMap.size === 0;
```
MaxMillington, line 24: @mattupham we should touch base with @syed8rahman next week on this. Presumably, we don't want to show the new modal every time someone without validators comes to the page. We should probably store some key in indexedDB. But let's confirm with @syed8rahman first

packages/web/modals/validator-next-step.tsx
```diff
@@ -0,0 +1,92 @@
...
+              <p className="text-base font-thin">
+                {t("stake.validatorNextStep.existingUser.description")}
+              </p>
+              <div className="flex w-full gap-8">
+                <Button
+                  className="w-full"
+                  mode="primary-bullish"
+                  onClick={() => {
+                    onRequestClose();
+                    alert("make stake call now");
```
MaxMillington, line 71: The stake call is in my branch. Will try to get that in tomorrow hopefully

PR #1936, opened 2023-07-31
Web worker router (multi threading)

jonator:

You may notice that the swap page has become quite laggy since the launch of CL, especially when the page loads but also if you enter a large in amount with a CL and perhaps split route involved.

**Before** When the page loads, a lot of things happen (basically simultaneously):
1. Many JS bundles and assets load after the SSR'd HTML page
2.  Our root store loads, thus triggering various queries for pools, IBC trace denoms, user balances, wallets, etc. If this isn't the first time loading the page recently, much of this data is pulled from localstorage.
  * Notably: about 1000 pools load in two chunks (pagination), most are filtered based on TVL (currently filters to ~170 on main, ~270 on frontier)
3. The price store loads, triggers some individual pool queries, then starts calculating a few prices from pool weights and prices depending on the visible tokens
4. When the pools load, many reactions are triggered by MobX
  * ⭐️ In the trade token in UI config, many reactions are initially triggered by all pools loading, some individual pools loading, and a few prices loading from coingecko. This causes heavy load on the main thread as a lot of intense computation is triggered for _each_ of these reactions. If a user interacts with the UI, it may be blocked by searching for routes or split quotes, or React may be updating the DOM constantly.
  * Next.js is getting fancy with loading the token images, with each image being a separate React component.

**Now**, with web workers, the issue here on this bullet (⭐️ ) is greatly alleviated. As yes, many reactions are happening on page load, but instead of expensive calculations in the router being synchronously triggered, instead they are simply queued in the background thread (similar to a thread pool). The UI remains snappy, since each reaction is quickly addressed and the main thread is quickly available to react to the user's input.

I'm glad I figured this out, as I was worried about increasing performance issues looming with:
* More pools being added, with most of them being the more expensive CL pools. With CL released, we're looking to migrate as much liquidity as soon as possible.
* Was not a fan of server side routing if it could be avoided, given maintenance concerns there
* Most users don't have M1/2 Macs like us, so any performance issues we see are best case. I've noticed _substantial_ performance improvements on mobile. Before, my phone (iPhone 13Pro) would freeze quite often.
* This seemed to also fix the glitchiness with token images. I think they were not getting a chance to update/render properly due to the clog on main thread.

I'm hoping this solution allows us to scale for some time now (perhaps not up to 100k pools like Uniswap has). Even if it takes several seconds, it's always OK to show a loading indicator in the UI instead of the UI freezing and being unreactive during those seconds. I'm hoping this solution will scale to several thousand pools, and split routes with 5+ CL pools involved; all with the UI remaining completely responsive. Also, now we can more aggressively refresh the pool data in the background, as we know it won't block user input.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/860rcgraa)

## Brief Changelog

* Adds some loading UI to swap tool
* Set up Next.js webpack config to work with web workers
* Write web worker as a new implementation of the router interface, which is merely a proxy to the existing router
  * Con: had to write manual encoders/decoders to pass data between main thread and worker
* Pools package: add default tick data provider and pool amount provider, dependencies needed by CL pool instance (and can now be used by others)
* Minor: update route vis for CL pool to use spread factor in stead of swap fee
* Crank up router params to increase granularity of multihop and split search, given this newfound luxury

TODO
- [x] Encode/decode errors as they come from router. Some expected.

## Testing and Verifying

I've seen some issues with reliability in generating quotes, so marking as a draft as I look into reproducing those and fixing those.


packages/pools/src/concentrated/fetch-pool-amount-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+  /** Address => Response */
+  protected amountsCache = new Map<
+    string,
+    Awaited<ReturnType<AmountsDataProvider["getPoolAmounts"]>>
+  >();
+
+  constructor(
+    protected readonly baseNodeUrl: string,
+    protected readonly poolId: string,
+    protected readonly cacheDurationMs = 1000 * 30, // 30 seconds
```
MaxMillington, line 23: Can you define as a constant?

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+    protected readonly amountFetcher: (
+      bech32Address: string
+    ) => Promise<BankAmountResponse> = async (bech32Address) => {
+      if (fetch === undefined)
+        throw new Error("Fetch method must be available in the environment");
+
+      const res = await fetch(
```
MaxMillington, line 30: Elsewhere in our app, we are using axios. Is there a reason to use the native `fetch` here?

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
jonator: It is, as a default param

MaxMillington, line 23: Our cache gets cleared only every 30 seconds. At scale, this could be an issue. I wonder if it would be worthwhile to implement something more sophisticated here such as LRU cache. https://www.npmjs.com/package/lru-cache

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+  >();
+
+  constructor(
+    protected readonly baseNodeUrl: string,
+    protected readonly poolId: string,
+    protected readonly cacheDurationMs = 1000 * 30, // 30 seconds
+    protected readonly amountFetcher: (
+      bech32Address: string
+    ) => Promise<BankAmountResponse> = async (bech32Address) => {
+      if (fetch === undefined)
```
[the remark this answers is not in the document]
jonator: I built this with the intention of other teams using it. I think fetch API is ubiquitous at this point, but there could be some server side JS engine (perhaps an older version) that doesn't support it.

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
jonator: We could look into that if it's an issue in the future, IMO. It would be a premature optimization and would increase the bundle size of the worker.

packages/pools/src/concentrated/fetch-tick-data-provider.ts
```diff
@@ -0,0 +1,185 @@
...
+    protected readonly tickFetcher: (
+      poolId: string,
+      tokenInDenom: string,
+      boundTickIndex: string
+    ) => Promise<TickDepthsResponse> = async (
+      poolId,
+      tokenInDenom,
+      boundTickIndex
+    ) => {
+      if (fetch === undefined)
```
MaxMillington, line 44: same questions as above. Why use fetch instead of axios? And why would it be undefined?

packages/pools/src/concentrated/fetch-tick-data-provider.ts
```diff
@@ -0,0 +1,185 @@
...
+        throw new Error("Fetch method must be available in the environment");
+
+      const res = await fetch(
+        `${this.baseNodeUrl}osmosis/concentratedliquidity/v1beta1/liquidity_net_in_direction?pool_id=${poolId}&token_in=${tokenInDenom}&use_cur_tick=true&bound_tick=${boundTickIndex}`
```
MaxMillington, line 48: Consider using `new URL`, and `url.searchParams.append` here

packages/pools/src/concentrated/fetch-tick-data-provider.ts
```diff
@@ -0,0 +1,185 @@
...
+  protected _zeroForOneTicks: LiquidityDepth[] = [];
+  protected _oneForZeroTicks: LiquidityDepth[] = [];
+
+  protected _zeroForOneBoundIndex = new Int(0);
+  protected _oneForZeroBoundIndex = new Int(0);
+
+  constructor(
+    protected readonly baseNodeUrl: string,
+    protected readonly poolId: string,
+    protected readonly nextTicksRampMultiplier = new Int(9),
```
MaxMillington, line 33: consider moving these default values to a constant

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+      ({ denom }) => denom === pool.token0
+    );
+    const token1AmountRaw = balancesRaw.find(
+      ({ denom }) => denom === pool.token1
+    );
+    const amounts = {
+      token0Amount: new Int(token0AmountRaw?.amount ?? "0"),
+      token1Amount: new Int(token1AmountRaw?.amount ?? "0"),
+    };
+
```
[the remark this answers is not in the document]
jonator: Good catch, added

packages/pools/src/concentrated/fetch-tick-data-provider.ts
```diff
@@ -0,0 +1,185 @@
...
+        allTicks: prevTicks,
+        isMaxTicks: true,
+      };
+    }
+
+    const tokenInDenom = zeroForOne ? pool.token0 : pool.token1;
+
+    const setLatestBoundTickIndex = (index: Int) => {
+      if (zeroForOne) {
+        this._zeroForOneBoundIndex = index;
```
MaxMillington, line 117: `requestinDirectionWithInitialTickBound` performs a few mutations. I suggest refactoring to use immutable data structures and avoid mutating `_zeroForOneBoundIndex`, `_oneForZeroTicks`, `_zeroForOneBoundIndex`, and `_oneForZeroBoundIndex`

packages/pools/src/concentrated/fetch-tick-data-provider.ts
jonator: See constants rant above

packages/pools/src/concentrated/fetch-tick-data-provider.ts
jonator: Responded in former comment

packages/pools/src/concentrated/fetch-tick-data-provider.ts
jonator: Will have GPT4 do this

jonator: https://chat.openai.com/share/3790441c-57e1-49bf-9fae-085447fd3825

It was great at this

packages/web/utils/background-routes/coding/out-given-in-router.ts
```diff
@@ -0,0 +1,187 @@
...
+};
+
+export function encodeRouteWithInAmount(
+  route: RouteWithInAmount
+): EncodedRouteWithInAmount {
+  return {
+    ...route,
+    initialAmount: route.initialAmount.toString(),
+    pools: route.pools
+      .map(encodePool)
```
MaxMillington, line 72: You can do this more efficiently in a single pass like this:

```
route.pools.reduce((accumulated, currentPool) => {
      const encodedPool = encodePool(currentPool);
      if (encodedPool) {
        accumulated.push(encodedPool);
      }
      return accumulated;
    }, []

File: packages/pools/src/concentrated/fetch-tick-data-provider.ts (hunk 8 above)
Comment (jonator) [reply packages/pools/src/concentrated/fetch-tick-data-provider.ts:124]: How so? These params are effectively a cache, as the provider may be requested tick depths within a previously fetched tick bound. That's how our mobx query store version of the provider works, and how it's intended to be used.

File: packages/web/utils/background-routes/coding/out-given-in-router.ts [hunk 10]
```diff
@@ -0,0 +1,187 @@
...
+}
+
+export function decodeRouteWithInAmount(
+  route: EncodedRouteWithInAmount
+): RouteWithInAmount {
+  return {
+    ...route,
+    initialAmount: new Int(route.initialAmount),
+    pools: route.pools
+      .map(decodePool)
```
MaxMillington, line 84: Same as above

packages/web/utils/background-routes/coding/error.ts
```diff
@@ -0,0 +1,47 @@
...
+  if ("error" in resp) {
+    const error = decodeError(resp.error);
+    if (throwError) throw error;
+    else return error;
+  }
+}
+
+export function decodeError(error: EncodedError): Error {
+  switch (error) {
+    case "NotEnoughLiquidityError":
```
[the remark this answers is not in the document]
jonator: For these expected errors, we use the instance to identify the error instead of the message. I'd prefer to use the class instance as the symbol to identify the error instead of the string message, as the string message should be able to be changed.

packages/web/utils/background-routes/index.ts
```diff
@@ -0,0 +1,166 @@
...
+} from "./coding";
+import {
+  EncodedRequest,
+  EncodedResponse,
+  OptimizedRoutesWorker,
+} from "./worker";
+
+/** Router that delegates search problem to background thread, useful for unblocking the main thread for UI/DOM updates. */
+export class BackgroundRoutes implements TokenOutGivenInRouter {
+  protected static singletonWorker: OptimizedRoutesWorker | null = null;
```
MaxMillington, line 26: Why is this a static variable instead of an instance variable?

jonator: Hmm I think my intention was to always share a single router thread, since then we'd share the cache in the router.

Could be a large assumption though, thoughts?

packages/web/utils/background-routes/index.ts
```diff
@@ -0,0 +1,166 @@
...
+  encodeRouteByTokenInParameters,
+} from "./coding";
+import {
+  EncodedRequest,
+  EncodedResponse,
+  OptimizedRoutesWorker,
+} from "./worker";
+
+/** Router that delegates search problem to background thread, useful for unblocking the main thread for UI/DOM updates. */
+export class BackgroundRoutes implements TokenOutGivenInRouter {
```
MaxMillington, line 25: I think this should work fine, but I wonder if the use of serial numbers and collecting pending responses is overcomplicating things? We aren't making any server connections here right? Typically, web workers are just passing messages back and forth https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers

Wondering what the impetus was behind using this model? Am I missing something?

MaxMillington (commented):
Glad you are addressing this. Few comments/questions.

packages/web/utils/background-routes/index.ts
MaxMillington: Theoretically, if there were multiple instances of BackgroundRoutes, there could be an issue since they would all share the same Worker. Seems like an opportunity for a bug to me. I think it makes more sense to make it an instance variable.

packages/web/utils/background-routes/coding/out-given-in-router.ts
```diff
@@ -0,0 +1,187 @@
...
+    },
+    params[1],
+  ];
+}
+
+// they happen to be the same
+export type EncodedGetOptimizedRoutesByTokenInParameters =
+  EncodedRouteByTokenInParameters;
+
+export function encodeGetOptimizedRoutesByTokenInParameters(
```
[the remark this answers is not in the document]
MaxMillington: I personally think it's a little confusing since we just have one function that returns another that does the same thing. But it's not a huge deal.

packages/web/utils/background-routes/index.ts
jonator: I use that to maintain the Promise-based API in the router interface.

We use the serial number to identify the message as it's passed to the worker, then we use the same serial number in the future to identify the right response in the response pool and resolve the promise with that response. I don't believe there's any guarantee of order as response messages are received from worker. The use of the loop and timeout isn't ideal, but it seems to work fine for the most part. It is a bit fragile.

I found this from Google: [Comlink](https://github.com/GoogleChromeLabs/comlink). Intends to modernize the API around web workers as a Promise-based RPC API, but IMO what we have works fine as we could avoid pulling in that dependency (which hasn't been updated in a while).

packages/pools/src/concentrated/fetch-tick-data-provider.ts
MaxMillington: I don't have an opinion on the mobx query store, but as a general principle, side effects that cause mutation of properties should be avoided because the instance is changed unexpectedly when these properties are changed in various methods. The state of the instance might become inconsistent depending on the order of function calls

It's hard for me to know if that would be an issue here, but one way to ensure that it isn't is to use a more 'functional programming' approach to make sure that we avoid side effects.

You could create a state class above, and then create a new instance of that state provider with your function calls in order to track state.

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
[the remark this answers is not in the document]
MaxMillington: Gotcha. In that case, despite adding to the bundle size, it might be better to use `axios` since axios will work across browser and nodejs environments, making the package more consistent/reliable

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
MaxMillington: I'll defer to you here since I don't really have a good sense of the use case here. I think the point about premature optimization is stronger than bundle size. If we are getting a lot of unique requests and our cache hit rate is low, than downloading the lru-cache package would be a clear solution.

packages/pools/src/concentrated/fetch-tick-data-provider.ts
jonator: I do prefer full separation of state and code with functional programming. 

But in this case, OOP is needed. Generally, the lifecycle of an objects data is considered to be the lifecycle of the instance. Here, without the object having to know; instances are created with CL pools and released with CL pools.

So, when the pools refresh, the pool is dropped from memory, with it's instances of providers.

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
jonator: My point is I don't think these in particular are worth extracting elsewhere since they are certainly only used in that one place. I do think constants at the top of file or in a file have their place.

But, when developers rely on their convention/habit of looking at the top of the file, they should only find variables that are important and heavily relied on. Especially those found in dedicated files (which at that point I think are config variables, which generally should be minimized). Otherwise, we risk losing signal to noise.

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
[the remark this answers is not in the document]
jonator: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility

It's been supported in all browsers since 2015, almost 10 years.

packages/pools/src/concentrated/fetch-tick-data-provider.ts
```diff
@@ -0,0 +1,185 @@
...
+      allTicks,
+      isMaxTicks: false,
+    };
+  }
+
+  async fetchTicks(
+    tokenInDenom: string,
+    boundTick: Int
+  ): Promise<LiquidityDepth[]> {
+    const rawDepths = await this.tickFetcher(
```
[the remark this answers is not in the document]
jonator: I think it's fine to allow errors bubble up and be addressed at higher level, there's no specific handling I need to do here.

packages/web/utils/background-routes/index.ts
jonator: This is not obvious, so I'll add a readme to explain this.

packages/web/utils/background-routes/index.ts
jonator: Also touched on this in top level README.

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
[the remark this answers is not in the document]
MaxMillington: Your previous comment indicated that you were concerned about `fetch` not existing in some NodeJS/server side environments hence the `if (fetch === undefined)` check. It's difficult to know what percentage of devs are using Node 18+ (which includes `fetch`) but I would imagine a substantial portion are still using a version older than Node 18 (which would not have a `fetch` implementation). `axios` will ensure universal compatibility. 

If it only runs in browsers, I would not be concerned. The likelihood of someone running a browser that doesn't support `fetch` is very low.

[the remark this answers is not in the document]
jonator: That's the reason the fetch function is dependency injected, to handle for that case where it's not supported in the server side VM.

For now, I think it's good to assume the modern option now, as it will only become stronger going forward.

packages/web/components/swap-tool/index.tsx
```diff
@@ -50,9 +52,9 @@ import { PromoDrawer } from "./promo-drawer";
 import { SplitRoute } from "./split-route";
 
 export const SwapTool: FunctionComponent<{
-  // IMPORTANT: Pools should be memoized!!
-  pools: ObservableQueryPool[];
-
+  /* IMPORTANT: Pools should be memoized!! */
+  memoedPools: ObservableQueryPool[];
+  dataLoading?: boolean;
```
JoseRFelix, line 57: Perhaps we should rename this to `isDataLoading` to follow the boolean convention in this component

packages/web/components/swap-tool/index.tsx
```diff
@@ -528,14 +544,14 @@ export const SwapTool: FunctionComponent<{
...
-                        .toString()}{" "}
-                      {tradeTokenInConfig.sendCurrency.coinDenom}
+                      {formatPretty(
+                        queries.queryBalances
+                          .getQueryBech32Address(account?.address ?? "")
+                          .getBalanceFromCurrency(
+                            tradeTokenInConfig.sendCurrency
+                          ),
+                        { maxDecimals: 2 }
+                      )}
```
JoseRFelix, line 554: I personally liked the previous decimal precision better. It gave me a better idea of the balance available. What if we increase the `maxDecimals` to `4`?

JoseRFelix: How about this approach: if we have a number that is less than `1`, we display it with `6` decimals; but if the number is 1 or more, we show it with just `2` decimlas?

JoseRFelix:
I think it would look better if the whole section was behind a loader instead of the value: 
image

packages/pools/src/concentrated/fetch-pool-amount-provider.ts
```diff
@@ -0,0 +1,85 @@
+import { Int } from "@keplr-wallet/unit";
```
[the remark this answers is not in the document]
JoseRFelix: Maybe `DataProvider` suffix could be an alternative to just `Provider`. What do you think?

packages/web/hooks/use-some-previous.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect, useRef } from "react";
+
+/** Inspired by `usePrevious` from react-use, but includes a condition for using the previous value. */
+export function useSomePrevious<T>(value: T, condition: (prev: T) => boolean) {
```
JoseRFelix, line 4: What do you think if we rename this hook to `usePreviousWhen`?

packages/web/utils/background-routes/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      serialNumber,
+      ...request,
+    });
+
+    return new Promise(async (resolve, reject) => {
+      try {
+        const tId = setTimeout(() => {
+          resolve(TIMEOUT);
+        }, timeoutMs);
+        let maxIters = 1_000_000;
```
JoseRFelix, line 163:
```suggestion
        let maxIterations = 1_000_000;
```

packages/web/utils/background-routes/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  encodeRouteByTokenInParameters,
+} from "./coding";
+import {
+  EncodedRequest,
+  EncodedResponse,
+  OptimizedRoutesWorker,
+} from "./worker";
+
+/** Timeouts are expected with web workers, depending on the speed of the web worker's event loop as events (requests) are queued. */
+const TIMEOUT = Symbol("Timeout");
```
JoseRFelix, line 26:
```suggestion
const TIMEOUT_SYMBOL = Symbol("Timeout");
```

JoseRFelix (changes requested):
Some minor nits

JoseRFelix (approved):
LGTM! A breakthrough for performance 🚀

PR #1937, opened 2023-07-31
[WIP] Reposition and optimize token images

dogemos:

Token images are currently not standardized in positioning nor optimized. This PR repositions/edits token icons to follow icon keylines, and then was optimized by using ImgOptim at 80% quality.

### ClickUp Task

n/a

## Brief Changelog

- Reposition to standardize most token icons to fit within icon keylines
- Run ImageOptim at 80% to minimize file size
- Standardize all token sizes to 200x200px

## Testing and Verifying

- Verified by checking preview link

n/a


packages/web/public/tokens/8ball.svg
jonator, line 1: Looks like this one lost some content

jonator:
Thanks for doing this!

jonator:
FYI next.js optimizes images for us at build time, but it doesn't do anything to SVGs. So trimming SVGs of detail should improve performance.

JeremyParish69:
After reviewing, I'm seeing that this ruins most of the images. I recommend we do not proceed with this

regarding: -Reposition to standardize most token icons to fit within icon keylines
It looks like some non-circular token logos been shrunk to fit inside a circle and sometimes have had the extra background space filled in the same color as the original background was. This was not applied consistently at all.
-I'd like to ask: is this something we want to enforce? Many token logos either are an irregular shape, but all of it's opaque content fits within an inscribed circle. And then there are token logos that are irregularly shaped or are perfectly square that have opaque content beyond the inscribed circle; do we want those to be shrunk such that they would fit inside a circle? Do we want 0 transparency within the circle?

regarding: -Standardize all token sizes to 200x200px 
This was not applied consistently at all. However, if 200x200 is a good size, then I can start encouraging or personally re-create token logos to be exactly this size for the frontend going forward. 

It also inconsistently applies a padding on token logos which will shrink the token logos, so we'll have tokens appear to be different sizes because of the transparent padding on some of them. We should aim to have 0 additional padding.

JeremyParish69 (changes requested):
This should not proceed. We should address this in stages. Logos that aren't circular should be handled individually. This should not be applied to SVGs. Resizing to 200x200 can be handled separately. Then a final optimization of PNGs shouldn't cause much trouble.

jonator:
Stale

PR #1938, opened 2023-07-31
Merge Master to Stage

JoseRFelix:


packages/web/.eslintrc.json
```diff
@@ -18,6 +18,12 @@
...
     "simple-import-sort/imports": "error",
     "simple-import-sort/exports": "error",
-    "unused-imports/no-unused-imports": "error"
+    "unused-imports/no-unused-imports": "error",
+    "no-restricted-imports": [
+      "error",
+      {
+        "patterns": [".", "..", "./*", "../*"]
+      }
+    ]
```
jonator, line 27: I think this was removed on stage

packages/web/components/alert/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from "./info";
-export * from "./toast";
-export * from "./tx-event-toast";
-export * from "./types";
+export * from "~/components/alert/info";
+export * from "~/components/alert/toast";
+export * from "~/components/alert/tx-event-toast";
+export * from "~/components/alert/types";
```
jonator, line 4: These absolute imports everywhere were reverted on stage

jonator (changes requested):
These seem to mostly reintroduce absolute imports. Though I reverted that on stage as it's not preferable as far as readability. We should prefer how it is on stage.

jonator (changes requested):
@JoseRFelix can we try to restore the relative imports as they were in all files. I only saw changes to a few

jonator:
> @jonator Do we want to also revert them in the component files? I believe they are more readable than previous relative imports

We can leave those

JoseRFelix:
@jonator I believe I removed all absolute imports from the index files on this [commit](https://github.com/osmosis-labs/osmosis-frontend/pull/1938/commits/f581e49fba0a3d8b2e15fd144d5111d1371abfa4)

PR #1942, opened 2023-08-02
performance: add unsafe IBC currency registrar

jonator:

Assumes our IBC asset config is valid, and forgoes the IBC module trace denom query and proceeds to register the IBC currency as a valid currency on the Osmosis chain store.

This greatly improves performance on _first_ load in a given browser (before cache is populated).

Long overdue, as the number of assets and pools has increased quite a bit, and caused this to become a performance issue. Also, first time users are prone to lose interest if the initial load is slow. Big win.

Background: Keplr queries for IBC denom state in their registrar since users are capable of inputting IBC asset info manually and adding it to various chains. With that, it's necessary to query and validate that the IBC currency actually exists and has channels.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/863haenpe)

## Brief Changelog

* Removed Keplr's IBC currency registrar
* Created an unsafe IBC currency registrar, that trusts the config and assigns

## Testing and Verifying

Notice immense performance improvements on first time load (either by clearing cache, opening in incognito, or opening a preview link with a unique origin).


jonator:
@JeremyParish69 can you confirm if this breaks any assets? Especially multihop like PSTAKE?

jonator:
Have spotted issues with IBC transfer

packages/stores/src/currency-registrar/unsafe-ibc.ts
```diff
@@ -0,0 +1,150 @@
...
+  { portId: string; channelId: string }[] // IBC path (configured)
+];
+
+/** Will register IBC currencies to the Osmosis chain in the chain store, without querying for IBC trace denom from IBC module.
+ *  It assumes the given IBC asset config is valid.
+ *  Use for major performance boost if working with many IBC assets (as we are on Osmosis). */
+export class UnsafeIbcCurrencyRegistrar<C extends ChainInfo = ChainInfo> {
+  /** IBC hash (ibc/XXXXX) => [chainId, coinMinimalDenom] (uatom) */
+  // include chain ID, because nothing is stopping currencies from having the same config on multiple chains
+  // example: uluna on columbus (original Terra) and uluna on phoenix (new Terra)
```
JoseRFelix, line 23:
```suggestion
/** 
    IBC hash (ibc/XXXXX) => [chainId, coinMinimalDenom, IBC path]
  
    include chain ID, because nothing is stopping currencies from having the same config on multiple chains
    example: uluna on columbus (original Terra) and uluna on phoenix (new Terra)
  */
```

packages/stores/src/currency-registrar/unsafe-ibc.ts
```diff
@@ -0,0 +1,150 @@
...
+type OriginChainCurrencyInfo = [
+  string, // chain ID
+  string, // coinMinimalDenom
+  { portId: string; channelId: string }[] // IBC path (configured)
+];
+
+/** Will register IBC currencies to the Osmosis chain in the chain store, without querying for IBC trace denom from IBC module.
+ *  It assumes the given IBC asset config is valid.
+ *  Use for major performance boost if working with many IBC assets (as we are on Osmosis). */
+export class UnsafeIbcCurrencyRegistrar<C extends ChainInfo = ChainInfo> {
```
JoseRFelix, line 20: Do you think this PR could be a good opportunity to create some basic unit tests for this class in particular? Given the importance of the IBC registrar, I believe implementing these tests would be highly beneficial.

JoseRFelix (changes requested):

packages/stores/src/currency-registrar/unsafe-ibc.ts
jonator: I agree, but given the lack of a mock chain store and that Aris had no issues, I'd like to not block the release this week

JoseRFelix (approved):

packages/stores/src/currency-registrar/unsafe-ibc.ts
jonator: OK I added some tests, chain store was easier to create than I thought

JoseRFelix (approved):
🙌

packages/stores/src/currency-registrar/__tests__/unsafe-ibc.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+    // but do expect it to be on the origin chain
+    const cosmosChain = chainStoreMock.getChain("cosmoshub-4");
+    const shdOnCosmosCurrency = cosmosChain.findCurrency(shdOnCosmos);
+    expect(shdOnCosmosCurrency).toBeDefined();
+  });
+
+  test("correctly handles unknown IBC currency", () => {
+    // Call unsafeRegisterIbcCurrency with an IBC hash denom that isn't in osmosisIbcAssetsMock
+    // Expect the result to be an object representing an unknown currency
+  });
```
JoseRFelix, line 116: This test is empty

JoseRFelix (changes requested):

JoseRFelix (approved):

PR #1944, opened 2023-08-02
WIP Update react 2

MaxMillington:


package.json
```diff
@@ -47,20 +48,19 @@
...
     "lerna": "^4.0.0",
     "lint-staged": "^12.1.3",
     "pre-commit": "^1.2.2",
     "prettier": "^2.8.8",
     "ts-jest": "^27.1.2",
     "turbo": "^1.9.3",
-    "typescript": "4.5.4"
-  },
-  "dependencies": {}
+    "typescript": "latest"
```
jonator, line 64: Is this temporary? Could be a footgun in the future

packages/proto-codecs/package.json
```diff
@@ -17,7 +17,6 @@
     "transpiler"
   ],
   "license": "Apache-2.0",
-  "private": false,
```
jonator, line 20: Why was this removed? We use it when publishing packages.

packages/web/components/animation/dynamic-lottie-animation.tsx
```diff
@@ -1,15 +1,13 @@
...
 
 const Lottie = dynamic(() => import("lottie-react"), { ssr: false });
 
-/** Loads the lottie library component, and it's
- *  animation data dynamically (outside of the main bundle). */
 export const DynamicLottieAnimation: FunctionComponent<
   {
     /** JSON file source, likely a dynamic import: `import("./lottie.json")`. */
     futureLottieContents: Promise<any>;
-  } & Omit<ComponentProps<typeof Lottie>, "animationData">
```
jonator, line 12: There's no alternative `ComponentProps` type from React? Is there a problem getting types from React 18?

packages/web/hooks/ui-config/use-superfluid-pool.ts
```diff
@@ -17,7 +17,7 @@ export function useSuperfluidPool(): {
   const account = accountStore.getWallet(chainId);
 
   const delegateSharesToValidator = useCallback(
```
jonator, line 19: There seems to be a typescript issue with React 18. `useCallback` should be able to infer the types from the type of the callback. IMO this is important to reduce clutter across the app. Generally, we should minimize type definitions and annotations as much as possible, if it's not possible to infer the type.

packages/web/hooks/wallet-select.tsx
```diff
@@ -29,8 +30,8 @@ const [WalletSelectInnerProvider, useWalletSelect] = createContext<{
 
 export { useWalletSelect };
 
-export const WalletSelectProvider: FunctionComponent = observer(
-  ({ children }) => {
+export const WalletSelectProvider: FunctionComponent<{ children?: ReactNode }> =
```
jonator, line 33: I believe children are an array?

packages/web/modals/base.tsx
```diff
@@ -9,6 +15,9 @@ import IconButton from "../components/buttons/icon-button";
 
 setAppElement("body");
 
+// react-modal has had some issues fully supporting React 18, causing a type error in our project
+const ModalSafeForReact18 = ReactModal as ComponentType<ReactModal["props"]>;
```
jonator, line 19: Have we looked into a new modal provider?

jonator (changes requested):
Stopped reviewing... this is looking like a nightmare, yes 😅

I had none of these issues when I successfully upgraded to React 18 a year ago. The changes were only to package.json files.

My intuition is typescript is not working properly here with React. Do we need to update typescript? I don't remember whether I did...

PR #1950, opened 2023-08-03
Test Unsafe KSM

JeremyParish69:

To add KSM to Unsafe Registrar

## Brief Changelog

-adds KSM to ibc-assets, source-chain-config, prices, and token logo

## Testing and Verifying

This change has not been tested locally, because this is a test with the new unsafe registrar


jonator:
Screenshot 2023-08-03 at 2 35 00 PM

jonator (changes requested):
My current implementation expects the IBC denom in the chain infos currencies list (as with PSTAKE), instead of the origin denom ("4" with KSM)

We should pick one and stick with it. As I said in slack, I vote to use the IBC denom since that's the representation on the counterparty chain for Osmosis (vs origin chain in this multi hop)

We've already implemented the denom being the counterparty IBC form, as we have with PSTAKE.

jonator (approved):

JeremyParish69:
> My current implementation expects the IBC denom in the chain infos currencies list (as with PSTAKE), instead of the origin denom ("4" with KSM)
> 
> We should pick one and stick with it. As I said in slack, I vote to use the IBC denom since that's the representation on the counterparty chain for Osmosis (vs origin chain in this multi hop)
> 
> We've already implemented the denom being the counterparty IBC form, as we have with PSTAKE.

Ah, yes, thanks. I copy-pasted the wrong config by accident.

PR #1952, opened 2023-08-04
feat: Simulate Transaction to Estimate Gas Fee

JoseRFelix:
### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/864dw75ug)

### Testing plan

- [x] All transactions within the app should open wallet confirmation with a gas value set. Since Txs are being simulated some may fail because the user lacks the necessary funds like a real transaction. This is the expected behavior. 
- [x] Can do CW20 transactions (Wynd, etc.)

### DIFF
--- a/packages/stores/src/account/__tests_e2e__/swap-exact-out.spec.ts
+++ b/packages/stores/src/account/__tests_e2e__/swap-exact-out.spec.ts
@@ -73,8 +73,6 @@ describe("Test Osmosis Swap Exact Amount Out Tx", () => {
             "10",
             undefined,
             undefined,
-            undefined,
-            undefined,
             (tx) => {
               if (tx.code) reject(tx);
               else resolve(tx);
@@ -102,8 +100,6 @@ describe("Test Osmosis Swap Exact Amount Out Tx", () => {
               amount: "1",
             },
             "10",
-            undefined,
-            undefined,
             undefined,
             undefined,
             (tx) => {
@@ -159,8 +155,6 @@ describe("Test Osmosis Swap Exact Amount Out Tx", () => {
             .mul(new Dec(1))
             .truncate()
             .toString(),
-          undefined,
-          undefined,
           undefined,
           undefined,
           (tx) => {
@@ -260,8 +254,6 @@ describe("Test Osmosis Swap Exact Amount Out Tx", () => {
             .toString(),
           undefined,
           undefined,
-          undefined,
-          undefined,
           (tx) => {
             if (tx.code) reject(tx.rawLog);
             else resolve(tx);
@@ -348,8 +340,6 @@ describe("Test Osmosis Swap Exact Amount Out Tx", () => {
           ],
           tokenOut,
           estimated.tokenIn.toCoin().amount,
-          undefined,
-          undefined,
           undefined,
           undefined,
           (tx) => {
@@ -446,8 +436,6 @@ describe("Test Osmosis Swap Exact Amount Out Tx", () => {
             ],
             tokenOut,
             "10",
-            undefined,
-            undefined,
             undefined,
             undefined,
             (tx) => {
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -826,7 +826,7 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
         );
       }
 
-      const multiplier = 1.13;
+      const multiplier = 1.5;
       return {
         /**
          * The gas amount is multiplied by a specific factor to provide additional
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -821,7 +821,7 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
         );
       }
 
-      const multiplier = 1.06;
+      const multiplier = 1.13;
       return {
         /**
          * The gas amount is multiplied by a specific factor to provide additional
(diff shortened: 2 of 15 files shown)


packages/stores/src/account/__tests_e2e__/swap-exact-out.spec.ts
```diff
@@ -73,8 +73,6 @@ describe("Test Osmosis Swap Exact Amount Out Tx", () => {
             "10",
             undefined,
             undefined,
-            undefined,
-            undefined,
```
jonator, line 77: Big fan of removing some of these undefined params, too many

packages/stores/src/account/utils.ts
```diff
@@ -1,15 +1,28 @@
...
 import { OfflineDirectSigner, OfflineSigner } from "@cosmjs/proto-signing";
+import { StdFee } from "@cosmjs/stargate";
 import {
   ChainName,
   Endpoints,
   ExtendedHttpEndpoint,
   Logger,
 } from "@cosmos-kit/core";
 
+export interface TxFee extends StdFee {
```
jonator, line 11: Can this name be improved, or are you trying to remain generic?

If not, how about something like `EstimateableFee` or similar

packages/web/pages/api/simulate-transaction.ts
```diff
@@ -0,0 +1,95 @@
...
+      {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+          tx_bytes: body.tx_bytes,
+        }),
+      }
+    );
```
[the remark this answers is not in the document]
jonator: @JoseRFelix note this CodeQL error, with this request containing user-provided values.

Should we do some validation? Or should we expect the node to handle anything?

packages/stores/src/account/base.ts
```diff
@@ -749,4 +764,101 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
+            return { amount: amount.amount, denom: amount.denom };
+          }),
+        }),
+      }).finish(),
+      // Because of the validation of tx itself, the signature must exist.
+      // However, since they do not actually verify the signature, it is okay to use any value.
+      signatures: [new Uint8Array(64)],
+    }).finish();
+
+    const restEndpoint = getEndpointString(await wallet.getRestEndpoint(true));
```
jonator, line 810: One thing I noticed with Keplr's work that we may want to handle (depending on if we want to simulate IBC transactions on counterparty chains, likely yes), is they check the version of the Cosmos SDK being used for simulating the fee.

We could either do that check, or just accept a failed request to handle that case. (the latter is fine with me)

packages/stores/src/account/base.ts
```diff
@@ -749,4 +764,101 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
+
+    try {
+      const result = await axios.post<{
+        gas_info: {
+          gas_used: string;
+        };
+      }>(this.options?.simulateUrl ?? "/api/simulate-transaction", {
```
jonator, line 817: I like the addition of this param with default. By the book 👍

jonator (commented):
Just pointing out Tony's gas simulator, in case he does anything clever. Not that we'd use it, as it appears to be used to help users select the gas option, but worth looking for gotchas there:
https://github.com/chainapsis/keplr-wallet/blob/63406e987e7cc537b0d8b4d14c024f126be8616c/packages/hooks/src/tx/gas-simulator.ts#L102

arispech (approved):
LGTM

packages/stores/src/account/utils.ts
JoseRFelix: I'm trying to remain generic. This should be the fee type we use going forward

JoseRFelix:
> @JoseRFelix note this CodeQL error, with this request containing user-provided values.
>
> Should we do some validation? Or should we expect the node to handle anything?

@jonator It seems to be a false positive since we are already [doing validation on the route](https://github.com/osmosis-labs/osmosis-frontend/pull/1952/files#diff-8bd783ee67d078cca13f9b4342028d3e16a61ab54cd01819fc7fa17b29d3583aR37-R44). Only the endpoints defined in our chain info are authorized.

jonator (approved):

jonator:
Looks like we were overestimating split multihop CL swap gas by over 500k (22 DAI -> USDC)

Before:
Screenshot 2023-08-07 at 8 55 55 PM

After:
Screenshot 2023-08-07 at 8 58 29 PM

packages/stores/src/account/base.ts
JoseRFelix: Agreed. I'm actually using [Keplr's check](https://github.com/chainapsis/keplr-wallet/blob/93a13dcceb91bea88dd63934d2fa8d1a234ab38c/packages/hooks/src/tx/gas-simulator.ts#L410-L418) to determine if we need the [backup fee](https://github.com/osmosis-labs/osmosis-frontend/blob/d6dd1c45a9de6b13d632732a2d8bb88f68b1624c/packages/stores/src/account/base.ts#L876-L878)

JoseRFelix:
@jonator What do you think if we include this in tomorrow's deployment?

JoseRFelix:
@jonator I agree. I scoped this out in this [task](https://app.clickup.com/t/866argtnc)

PR #1957, opened 2023-08-04
Fixes to swap router

jonator:

On stage, the swap tool seemed to work fine for the most part. But I noticed an issue with the swap tool on the CL pool detail page, as well as excessive requests for balances and ticks (8+ for one quote for balances).

* Fix memory leak issue in `BackgroundRoutes`
* Key breakthrough: for existing requests, instead of returning 0s or empty lists for ticks, I maintain a map of promises for in flight requests and await on those. This prevented all sorts of annoying race conditions in the UI, and is a more reliable solution to async UI. Now, even without debounce in our trade UI config, we know (at least with this BG router), that the server will not get spammed unnecessarily.
* Dispose all UI config's reactions

Now, as user input and pools data is updated in the background, the BG router properly handles these async requests.

### ClickUp Task

## Brief Changelog

- [x] Fix: pool detail page for CL pool showing "Not enough liquidity" error on first appearance (works when input amount entered)

## Testing and Verifying

* Swap tool switches currencies, manages input, loads spot price accurately
* Works on mobile devices

### DIFF
--- a/packages/pools/src/concentrated/fetch-pool-amount-provider.ts
+++ b/packages/pools/src/concentrated/fetch-pool-amount-provider.ts
@@ -9,19 +9,31 @@ type BankAmountResponse = {
   }[];
 };
 
-/** Provider of bank balances from bank module by bech32Address. Expected by concentrated liquidity pool. */
+/** Provider of bank balances from bank module by bech32Address. Expected by concentrated liquidity pool.
+ *  The instance lifecycle is assumed to follow the pool instance.
+ */
 export class FetchPoolAmountDataProvider implements AmountsDataProvider {
   /** Address => Response */
-  protected amountsCache = new Map<
+  protected static amountsCache = new Map<
     string,
     Awaited<ReturnType<AmountsDataProvider["getPoolAmounts"]>>
   >();
 
-  protected activeCacheTimeouts = new Map<
+  protected static activeCacheTimeouts = new Map<
     string,
     ReturnType<typeof setTimeout>
   >();
 
+  /** Set of `bech32Address: string` if fetching for that address. */
+  protected static inFlightFetchPerAccount = new Set<string>();
+
+  /**
+   * Creates a new provider. The instance lifecycle is assumed to follow the pool instance.
+   * @param baseNodeUrl Base URL of node to fetch balances from. Only used in default `amountFetcher`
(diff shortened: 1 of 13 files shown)


packages/pools/src/concentrated/fetch-pool-amount-provider.ts
```diff
@@ -9,19 +9,36 @@ type BankAmountResponse = {
...
 
-  protected activeCacheTimeouts = new Map<
+  /** Address => Timeout ID */
+  protected static _activeCacheTimeouts = new Map<
     string,
     ReturnType<typeof setTimeout>
   >();
 
+  /** Map of `bech32Address: string` => Promise if fetching for that address. */
+  protected static _inFlightFetchPerAccount = new Map<
```
JoseRFelix, line 30: What do you think if we make the name more generic since pool addresses could also be used here? 

```suggestion
  protected static _inFlightFetchPerAddress = new Map<
```

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
```diff
@@ -463,138 +465,148 @@ export class ObservableAddConcentratedLiquidityConfig {
...
-        );
-        this.setMaxRange(
-          this.initialCustomPriceRange[1]
-            .mul(multiplicationQuoteOverBase)
-            .toString()
-        );
-      }
-    });
+    this._disposers.push(
+      autorun(() => {
```
JoseRFelix, line 469: I'm not saying we should do this in this PR, but can you think of any way where we can merge this process into one call? Something like `this.extendedClass.addDisposable(() => { ... })`. I would love to avoid instances where devs forget to add to a disposer unintentionally.

JoseRFelix (approved):

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
jonator: All the autoruns? Some autoruns don't need to be disposed since they live on the root store context.  Those that live in a component subtree need to be disposed if navigated away.

And if you meant that the autoruns/reactions need to be combined, that could introduce issues as we only want to react to certail things.

PR #1964, opened 2023-08-08
staking analytics

mattupham:

- add various logEvent calls for staking analytics
- add property logic for logEvent

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/860rcjd9y)

## Brief Changelog

- page viewed
- staking started
- staking completed
- unstaking started
- unstaking completed
- squad option clicked
- select squad and stake clicked
- build squad clicked

### One example - properties for staking started / unstaking started
Screenshot 2023-08-09 at 11 06 52 PM

## Testing and Verifying
- will need Rafael to verify these in amplitude

### DIFF
--- a/packages/web/pages/stake/index.tsx
+++ b/packages/web/pages/stake/index.tsx
@@ -1,5 +1,6 @@
 import { Staking as StakingType } from "@keplr-wallet/stores";
 import { CoinPretty, Dec } from "@keplr-wallet/unit";
+import { DeliverTxResponse } from "@osmosis-labs/stores";
 import { observer } from "mobx-react-lite";
 import React, { useCallback, useEffect, useMemo, useState } from "react";
 import { useTranslation } from "react-multi-lang";
@@ -16,8 +17,6 @@ import { useFeatureFlags } from "~/hooks/use-feature-flags";
 import { ValidatorNextStepModal } from "~/modals/validator-next-step";
 import { ValidatorSquadModal } from "~/modals/validator-squad";
 import { useStore } from "~/stores";
-
-import { DeliverTxResponse } from "../../../stores/src/account/types";
 
 const getAmountDefault = (fraction: number | undefined): AmountDefault => {
   if (fraction === 0.5) return "half";


packages/web/modals/validator-next-step.tsx
```diff
@@ -60,6 +63,10 @@ export const ValidatorNextStepModal: FunctionComponent<
               className="w-full"
               mode="primary-bullish"
               onClick={() => {
```
MaxMillington, line 65: Maybe we should extract this to a these onClick functions to a handleClick using useCallback. It's better perf since it won't trigger as many child rerenders

packages/web/modals/validator-next-step.tsx
```diff
@@ -60,6 +63,10 @@ export const ValidatorNextStepModal: FunctionComponent<
...
+                logEvent([
+                  EventName.Stake.squadOptionClicked,
+                  { option: "keep" },
+                ]);
                 onRequestClose();
                 alert("make stake call now");
```
MaxMillington, line 71: Can do in a sep PR but the stake call is merged so you should be able to put it in here

packages/web/modals/validator-squad.tsx
```diff
@@ -313,6 +316,22 @@ export const ValidatorSquadModal: FunctionComponent<ValidatorSquadModalProps> =
         ? totalSize - (virtualRows?.[virtualRows.length - 1]?.end || 0)
         : 0;
 
+    const handleClick = useCallback(() => {
```
MaxMillington, line 319: like this 👍

packages/web/modals/validator-next-step.tsx
MaxMillington: Similar to what you do in validator-squad below

packages/web/pages/stake/index.tsx
```diff
@@ -71,33 +85,94 @@ export const Staking: React.FC = observer(() => {
...
+    const delegationsMap = new Map<string, StakingType.Delegation>();
+
+    userValidatorDelegations.forEach((delegation: StakingType.Delegation) => {
+      delegationsMap.set(delegation.delegation.validator_address, delegation);
+    });
+
+    return delegationsMap;
+  }, [userValidatorDelegations]);
+
+  const [amountDefault, setAmountDefault] = useState<AmountDefault>("input");
```
MaxMillington, line 104: nit: can you move this up with the other useState hooks?

MaxMillington:
Looks good, just one comment about inline functions

packages/web/pages/stake/index.tsx
```diff
@@ -71,33 +85,94 @@ export const Staking: React.FC = observer(() => {
...
+
+    logEvent([EventName.Stake.stakingStarted, properties]);
+
     if (account?.address && account?.osmosis && coin?.amount) {
-      account.osmosis.sendDelegateToValidatorSetMsg(coin);
+      account.osmosis.sendDelegateToValidatorSetMsg(
+        coin,
+        "",
+        (tx: any) =>
+          Boolean(tx?.code) &&
```
jonator, line 139: Nit but will add more type safety:

The base `signAndBroadcast` used in `sendDelegateToValidatorSetMsg` returns a `DeliverTxResponse` type for the `tx` passed to onFullfill.

In our osmosis account functions (here `sendDelegateToValidatorSetMsg`), we could increase type safety by using that type instead of washing to any; like `(tx: any) => ...`. Both in the osmosis account message and here.

Future devs may also benefit from using the type to discover the structure of the tx, like the fact that it has an error code.

jonator (approved):
LGTM, with a typescript type comment

packages/web/pages/stake/index.tsx
```diff
@@ -8,20 +8,34 @@ import { AlertBanner } from "~/components/alert-banner";
...
 import { ValidatorSquadModal } from "~/modals/validator-squad";
 import { useStore } from "~/stores";
 
+type AmountDefault = "half" | "max" | "input";
+
+interface Properties {
+  amountDefault: AmountDefault;
+  amount: string;
+  amountUSD: string | undefined;
+}
```
jonator, line 25: FYI we have a properties type defined in v2 for properties sent to amplitude. IMO we should isolate amplitude types there as much as possible, to avoid having amplitude types be confused with actual business-logic data types.

packages/web/pages/stake/index.tsx
```diff
@@ -71,33 +85,94 @@ export const Staking: React.FC = observer(() => {
...
+  const [amountDefault, setAmountDefault] = useState<AmountDefault>("input");
+
+  const getProperties = useCallback(() => {
+    const amountUSD = priceStore
+      .calculatePrice(
+        new CoinPretty(
+          osmo,
+          new Dec(amountConfig.amount).mul(
+            DecUtils.getTenExponentNInPrecisionRange(osmo.coinDecimals)
+          )
```
jonator, line 113: I believe CoinPretty accepts the amount without decimals. So we could do `amountConfig.toCoinPrimitive().amount` as the second arg.

packages/web/pages/stake/index.tsx
```diff
@@ -71,33 +85,94 @@ export const Staking: React.FC = observer(() => {
...
+    const amount = amountConfig.amount;
+
+    const properties: Properties = {
+      amountDefault,
+      amount,
+      amountUSD,
+    };
+
+    return properties;
+  }, [amountConfig.amount, amountDefault, osmo, priceStore]);
```
jonator, line 127: A few things I notice:
* It appears the stake event requires a lot of property data, which is fine, but can cause issues depending on how our code is organized.
* The fact that we have to query so much data at this top level, perhaps is a sign we're logging in the wrong layer? The wrong component? What if we log the stake event closer to where the data is being displayed? Querying at this top level will cause a lot of cascading renders as data loads, which is not a huge issue.

packages/web/pages/stake/index.tsx
jonator: I think it can be removed. Amount config, where the half and max are implemented, has a `fraction` member. Instead of storing state separately in a useState, we can use `fraction === 0.5` for half and `fraction ==== 1` for max.

packages/web/pages/stake/index.tsx
jonator: Please test this and confirm it works.

jonator: Also, similar to the other query data, we shouldn't need a callback here. We can just get the USD value of the current amount in the top level. The price store is observable, so if price data is refreshed, this component would rerender and the USD amount would update.

packages/web/pages/stake/index.tsx
MaxMillington: Oh wait no, I'm guessing the data team asked to track whether the amount was input via the 'half' button, the 'max' button, or manually 'input', in which case we need to track that

packages/web/pages/stake/index.tsx
```diff
@@ -71,33 +87,111 @@ export const Staking: React.FC = observer(() => {
...
+  const amount = amountConfig.amount || "0";
+  const amountUSD = priceStore
+    .calculatePrice(
+      new CoinPretty(
+        osmo,
+        new Dec(amountConfig.amount || 0).mul(
+          DecUtils.getTenExponentNInPrecisionRange(osmo.coinDecimals)
+        )
+      )
+    )
```
jonator, line 116:
```suggestion
    .calculatePrice(
      new CoinPretty(
        osmo,
        amountConfig.getAmountPrimitive().amount
      )
    )
```

What you have works, nice! The decimals need to be removed from the amount before passing to CoinPretty. What I was trying to say (my bad for using wrong function name) was that the amount config has the decimal conversion within it.

jonator (approved):

packages/web/pages/stake/index.tsx
jonator: ( I realized I selected too many lines so the suggestion is not directly usable )

jonator (approved):
LGTM

PR #1970, opened 2023-08-08
feat: Transpile Modules in Next Dev Server and Add Lint to CI

JoseRFelix:

Transpile local Osmosis modules using the Next.js configuration. Makes it easy to have local libraries and keep a slick, manageable dev experience. This will solve the issue with build collisions that required devs to clear the build folder on occasion. 

This PR also adds a Lint action that will improve the repo consistency, especially while merging asset PRs. 

### ClickUp Task

[Add Lint CI Action](https://app.clickup.com/t/866arqaqq)
[Transpile Next.js local Osmosis modules](https://app.clickup.com/t/866arqat3)

### DIFF
--- a/.github/workflows/lint-and-check-format.yml
+++ b/.github/workflows/lint-and-check-format.yml
@@ -1,0 +1,34 @@
+name: Lint and Check Format Workflow
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [18.x]
+        package: ["web", "stores", "pools", "math"]
+
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: "**/node_modules"
+          key: ${{ runner.OS }}-${{ matrix.node-version }}-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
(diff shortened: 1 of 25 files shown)


jonator:
@JoseRFelix will try to look at this soon

jonator (approved):
Since we're touching our modules, can we also use this as an opportunity to update turbo repo in root package.json?
Screenshot 2023-08-10 at 10 10 38 AM

jonator (commented):
Is the expected behavior to save a console log in the stores package, and the dev server would refresh automatically? I'm not able to get that to work

JoseRFelix:
> Is the expected behavior to save a console log in the stores package, and the dev server would refresh automatically? I'm not able to get that to work

The server is updating and refreshing, but the client is not. I can see the changes through the console logs on the server, but these updates are not being reflected on the client. I think the class instances are not being updated maybe because of Mobx.

jonator:
> > Is the expected behavior to save a console log in the stores package, and the dev server would refresh automatically? I'm not able to get that to work
> 
> The server is updating and refreshing, but the client is not. I can see the changes through the console logs on the server, but these updates are not being reflected on the client. I think the class instances are not being updated maybe because of Mobx.

Yeah that makes sense. The root store gets initialized once on load as a react context, and even if the source file gets updated in the browser, the root store data would not change.

JoseRFelix:
> Yeah that makes sense. The root store gets initialized once on load as a react context, and even if the source file gets updated in the browser, the root store data would not change.

The issue is specific to Next.js v12. After extensive debugging, I found that upgrading to v13 will resolve the problem, as it automatically remounts the store whenever there's an update. I was able to reproduce this with their [example](https://github.com/vercel/next.js/tree/canary/examples/with-mobx) by downgrading it to v12

PR #1979, opened 2023-08-09
Revert "Revert (#1971)"

MaxMillington:
reverting the revert. DO NOT MERGE until auto deployments are turned off


jonator (approved):
This looks correct, it looks like it's reenabling our release on master

PR #1981, opened 2023-08-09
feat: Support wallets without Suggest Chain

JoseRFelix:

Implement a solution to manage wallets that do not support the "suggest chain" feature. Users interacting with unsupported assets will encounter limitations within the user interface.

## Brief description

- Add XDefi to test a wallet that does not support suggest chain
- Add a framework to handle wallet-specific errors

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/866arv8ag)

## Testing and Verifying

- [x] User can connect to XDefi wallet 
- [x] User can connect to other wallets
- [x] User can do a transaction with XDefi 
- [x] Non-supported chains will be grayed out on the assets page. Test with XDefi and Keplr Mobile
- [x] If the user connects through a modal on an unsupported asset, he should see an error message and can't make deposits.


arispech (approved):
Working as expected

packages/stores/src/account/base.ts
```diff
@@ -309,6 +331,63 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
+    if (!wallet || wallet.walletStatus !== WalletStatus.Connected) {
+      return undefined;
+    }
+
+    const id = `${wallet.walletName}_${chainId}`;
+
+    let promiseObservable = this._walletToSupportChainPromise.get(id);
+
+    if (!promiseObservable) {
+      promiseObservable = fromPromise<boolean>(wallet.supportsChain(chainId));
```
jonator, line 355: FYI from looking close at docs, you can pass a previous promise based observable as second param to help with smooth transitions between pending states

https://github.com/mobxjs/mobx-utils#frompromise

packages/stores/src/account/base.ts
```diff
@@ -309,6 +331,63 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
+      this._walletToSupportChainPromise.set(id, promiseObservable);
```
jonator, line 356: Is there any scenario where we want to delete a promise? Or is it for the lifetime of the app?

jonator (changes requested):
Got this error on local dev when:

1. Clicking connect wallet
2. Selecting Leap
3. Then selecting Keplr mobile

This was on first load. Also, it seems directly going to the option avoids this.


Screenshot 2023-08-13 at 3 52 31 PM

JoseRFelix:
> Got this error on local dev when:
> 
> 1. Clicking connect wallet
> 2. Selecting Leap
> 3. Then selecting Keplr mobile
> 
> This was on first load. Also, it seems directly going to the option avoids this.
> 
> Screenshot 2023-08-13 at 3 52 31 PM

I couldn't reproduce the issue. It might be due to a conflict in the build folders. Could you try to delete the build folders and attempt to reproduce it again?

packages/stores/src/account/base.ts
JoseRFelix: Thank you. I'm unsure how to apply it here, especially since the promises are designed to execute just once. Perhaps you've noticed something I missed. Should we include the second parameter?

packages/stores/src/account/base.ts
JoseRFelix: It should be for the lifetime of the app. The wallet's available chains shouldn't change unless it has the "suggest chain" feature, which wouldn't need this method

jonator (approved):

PR #1985, opened 2023-08-10
Fall back to pre v17 queries

jonator:

Falls back to pre v17 queries in the CFMM <> CL link query store if pre v17.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/863hazkth)

## Brief Changelog

* I still use the new single links query store, but now check the node version and fall back to the old individual query stores as needed

## Testing and Verifying

Connect dev env to edgenet, which is on v17 dev env:

Screenshot 2023-08-10 at 5 17 58 PM

Change env to point to mainnet:
* You should be able to see now on stage that the old link queries are called. Remember, the 400s are expected since that's what's returned when a link doesn't exist for a give pool ID:
Screenshot 2023-08-10 at 5 02 18 PM

### DIFF
--- a/packages/stores/src/queries/pools/cfmm-cl-link.ts
+++ b/packages/stores/src/queries/pools/cfmm-cl-link.ts
@@ -1,12 +1,27 @@
 import { KVStore } from "@keplr-wallet/common";
-import { ChainGetter, ObservableChainQuery } from "@keplr-wallet/stores";
-import { makeObservable } from "mobx";
+import {
+  ChainGetter,
+  ObservableChainQuery,
+  ObservableChainQueryMap,
+} from "@keplr-wallet/stores";
+import { computed, makeObservable } from "mobx";
 import { computedFn } from "mobx-utils";
 
+import { ObservableQueryNodeInfo } from "../tendermint/node-info";
 import { MigrationRecords } from "./types";
 
 export class ObservableQueryCfmmConcentratedPoolLinks extends ObservableChainQuery<MigrationRecords> {
-  constructor(kvStore: KVStore, chainId: string, chainGetter: ChainGetter) {
+  // BACKWARDS COMPATIBLE QUERIES FOR PRE V17:
+  // TODO: can be safely removed after migration to v17, est Aug 22 '23
+  protected backCompatQueryCfmmToCL: ObservableQueryCfmmToConcentratedLiquidityPoolLinks;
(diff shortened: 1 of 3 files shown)


packages/stores/src/queries/pools/cfmm-cl-link.ts
```diff
@@ -1,28 +1,73 @@
...
+    if (this.queryNodeInfo.isDevelopmentEnv) return true;
+
+    return (this.queryNodeInfo.nodeVersion ?? 0) >= 17;
+  }
+
   /** Link to concentrated liquidity pool from the given balancer/cfmm pool ID.
    *  Returns `undefined` if not loaded, `false`, if nonexistent, or the CL pool ID if it exists. */
   readonly getLinkedConcentratedPoolId = computedFn(
     (cfmmPooLId): string | false | undefined => {
       if (!this.response) return;
```
JoseRFelix, line 64: What do you think about adding a guard to check if queryNodeInfo has a response? I believe this could prevent unexpected calls

```suggestion
      if (!this.response || this.canFetch()) return;
```

packages/stores/src/queries/pools/cfmm-cl-link.ts
```diff
@@ -38,6 +83,10 @@ export class ObservableQueryCfmmConcentratedPoolLinks extends ObservableChainQue
     (clPoolId): string | false | undefined => {
       if (!this.response) return;
```
JoseRFelix, line 84: Same as above

czarcas7ic (approved):

packages/stores/src/queries/pools/cfmm-cl-link.ts
jonator: Are you noticing unexpected calls? My thought was, the canFetch would start to return true once the node query lands. Then the relevant link query can become observed and start.

jonator: Oh I see. Looking at the edgenet preview, it appears the v17 query is not running at all.

JoseRFelix: On non-v17, I notice some extra calls occurring, but I believe this is unrelated to the comment I made. Here's the reference image: 
image.

In my comment, I want to prevent an additional call from being made if the node version query takes too long to resolve or fails to resolve altogether.

JoseRFelix (approved):

PR #1987, opened 2023-08-11
Fix StablecoinTypes

kamal-sutra:
To fix 'StablecoinTypes' display error in various Stablecoin pools, as well as updating and adding the types of Stablecoins.

Check the images below for before/after changes (tested in localhost:3000)

image
image
image
image
image
image

### DIFF
--- a/packages/stores/src/chain.ts
+++ b/packages/stores/src/chain.ts
@@ -13,7 +13,11 @@ export interface ChainInfoWithExplorer extends ChainInfo {
   /** Add optional stable coin peg info to currencies. */
   currencies: Array<
     AppCurrency & {
-      pegMechanism?: "collateralized" | "algorithmic" | "hybrid";
+      pegMechanism?:
+        | "fiat-backed"
+        | "algorithmic"
+        | "crypto-backed"
+        | "hybrid-pegged";
     }
   >;
 }
--- a/packages/web/localizations/en.json
+++ b/packages/web/localizations/en.json
@@ -13,7 +13,7 @@
     "scalingFactorControllerWarning_third": "to maintain the peg. The current scaling factor is",
     "scalingFactorInformation": "This factor allows the curve to adjust to predictable price changes, such as the value accruing to a liquid staking token.",
     "singleAsset": "Single asset",
-    "stablecoinWarning": "You are adding liquidity to {denom}, a {mechanism}-backed stablecoin.",
+    "stablecoinWarning": "You are adding liquidity to {denom}, a {mechanism} stablecoin.",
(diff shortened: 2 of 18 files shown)


packages/stores/src/chain.ts
```diff
@@ -13,7 +13,11 @@ export interface ChainInfoWithExplorer extends ChainInfo {
   /** Add optional stable coin peg info to currencies. */
   currencies: Array<
     AppCurrency & {
-      pegMechanism?: "collateralized" | "algorithmic" | "hybrid";
+      pegMechanism?:
+        | "fiat-backed"
+        | "algorithmic"
+        | "crypto-backed"
+        | "hybrid-pegged";
```
jonator, line 20: It's nice we get more granularity in the various mechanisms.

Would you mind defining each of these types please? Perhaps in a comment as well?

packages/web/localizations/en.json
```diff
@@ -13,7 +13,7 @@
     "scalingFactorControllerWarning_third": "to maintain the peg. The current scaling factor is",
     "scalingFactorInformation": "This factor allows the curve to adjust to predictable price changes, such as the value accruing to a liquid staking token.",
     "singleAsset": "Single asset",
-    "stablecoinWarning": "You are adding liquidity to {denom}, a {mechanism}-backed stablecoin.",
+    "stablecoinWarning": "You are adding liquidity to {denom}, a {mechanism} stablecoin.",
```
jonator, line 16: Thoughts on the fact that the types are not localized? We could have a different key for each type so that they get translated.

Also, we'll need to localize the remaining files. Please take a look at the readme in the root of the repo for how this is done.

packages/stores/src/chain.ts
kamal-sutra: Fiat-backed: Backed by a reserve of traditional fiat currencies, such as the USD/Euro/GBP etc, providing steadiness with a 1:1 peg and reduced volatility.

Algorithmic: Leverages algorithms and contracts to adapt supply based on demand, striving to stabilize value amidst market changes.

Crypto-backed: These are collateralized by other cryptocurrencies, ensuring stability through locked assets while enabling issuance of stablecoins.

Hybrid-pegged: Mixes collateral assets such as fiat and cryptocurrencies while also utilizing algorithmic mechanisms to maintain its stability.

packages/web/localizations/en.json
kamal-sutra: different keys for each type are localized now with most of the translations. just trying to find any help with cn, hk, tw languages.

jonator:
stale

PR #1989, opened 2023-08-11
Revert "feat: Transpile Modules in Next Dev Server and Add Lint to CI…

JoseRFelix:

This reverts commit 81fd5631e2bb84ba7766bbf3920daa2b22896079. 

There are some issues with Mobx states on the deposit/withdraw modal. I don't know if there are more affected sections throughout the app. Tackling this now would not be worth it since it would require a whole app QA. Instead, it would be more efficient to wait for the release of React 18 and Next 13, so that the QA can be conducted for both updates simultaneously.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/866arqat3)


jonator (approved):

jonator:
@JoseRFelix can we track this finding as a subtask in the upgrade react /next task?

PR #1990, opened 2023-08-14
show IBC denom instead of UNKNOWN

jonator:

From discussion with Jeremy and Sunny, we all agreed it would be more useful to display a slice of the IBC denom instead of "UNKNOWN" for the denom. This would at least make the UI usable when differentiating tokens.

Screenshot 2023-08-14 at 10 41 04 AM


MaxMillington (approved):

MaxMillington:
Just need to update the test case

MaxMillington (approved):

PR #1995, opened 2023-08-14
Cl Fixes

jonator:

Fixes related to CL
### ClickUp Task

## Brief Changelog

* Only allow collect rewards for positions in pool in pool detail page

### DIFF
--- a/packages/web/components/pool-detail/concentrated.tsx
+++ b/packages/web/components/pool-detail/concentrated.tsx
@@ -86,18 +86,80 @@ export const ConcentratedLiquidityPool: FunctionComponent<{ poolId: string }> =
       ? pool.concentratedLiquidityPoolInfo.currentPrice
       : undefined;
 
-    const userPositions = osmosisQueries.queryAccountsPositions.get(
-      account?.address ?? ""
-    ).positions;
+    const userPositionsInPool = osmosisQueries.queryAccountsPositions
+      .get(account?.address ?? "")
+      .positionsInPool(poolId);
 
-    const userPositionsInPool = userPositions.filter(
-      (position) => position.poolId === poolId
-    );
     const userHasPositionInPool = userPositionsInPool.length > 0;
 
-    const rewardedPositions = userPositions.filter(
+    const rewardedPositions = userPositionsInPool.filter(
       (position) => position.hasClaimableRewards
     );
+
+    const onClickCollectAllRewards = () => {
+      const fiat = priceStore.getFiatCurrency(priceStore.defaultVsCurrency);
+
+      if (!account) throw new Error("No account");
+
+      const rewardAmountUSD = rewardedPositions.reduce(
+        (acc, { totalClaimableRewards }) => {
+          const totalValue =
+            totalClaimableRewards.length > 0 && fiat
+              ? totalClaimableRewards.reduce(
+                  (sum, asset) =>
+                    sum.add(
+                      priceStore.calculatePrice(asset) ??
+                        new PricePretty(fiat, 0)
+                    ),
+                  new PricePretty(fiat, 0)
+                )
+              : undefined;
+          acc += Number(totalValue?.toDec().toString() ?? 0);
+          return acc;
+        },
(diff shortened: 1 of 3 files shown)


packages/web/components/pool-detail/concentrated.tsx
```diff
@@ -86,19 +86,81 @@ export const ConcentratedLiquidityPool: FunctionComponent<{ poolId: string }> =
...
-      (position) => position.poolId === poolId
-    );
     const userHasPositionInPool = userPositionsInPool.length > 0;
 
-    const rewardedPositions = userPositions.filter(
+    const rewardedPositions = userPositionsInPool.filter(
       (position) => position.hasClaimableRewards
     );
 
+    const onClickCollectAllRewards = () => {
```
MaxMillington, line 99: Is all of this function unique to CL? Could any of it be reused in staking? If so, can we extract into a function that is reusable in staking?

packages/web/components/pool-detail/concentrated.tsx
```diff
@@ -86,19 +86,81 @@ export const ConcentratedLiquidityPool: FunctionComponent<{ poolId: string }> =
...
+      const fiat = priceStore.getFiatCurrency(priceStore.defaultVsCurrency);
+
+      if (!account) throw new Error("No account");
+
+      const rewardAmountUSD = rewardedPositions.reduce(
```
MaxMillington, line 104: Two quick thoughts on this algorithm, 

1. We have nested `reduce` functions which causes a quadratic time complexity. Is it possible to restructure `rewardedPositions` so that we can accomplish this with a linear time complexity? I'm not sure what the structure of `rewardedPositions` is, otherwise I would suggest something. Let me know if you want to pair in case I'm not being clear
2. If we can't avoid the nested loop, maybe we should consider pulling out the inner `reduce` to be it's own named function just so this is easier to read. Something like this:
```
const calculateTotalValue = (rewards, fiat) => {
    if (!Array.isArray(rewards) || !rewards.length || !fiat) return 0;

    return rewards.reduce((sum, asset) => {
        const price = priceStore.calculatePrice(asset);
        return sum + (price ? Number(price.toDec().toString()) : 0);
    }, 0);
};

const rewardAmountUSD = rewardedPositions.reduce((acc, { totalClaimableRewards }) => {
    return acc + calculateTotalValue(totalClaimableRewards, fiat);
}, 0);
```

MaxMillington: I recognize that this perf issue already existed before, but I'm just noticing it now I guess 🙃

MaxMillington (commented):
Two questions

packages/web/components/pool-detail/concentrated.tsx
jonator: It's unique to CL- use of the word "rewards" here refers to the spread and incentive rewards a user's position accrues for providing liquidity in a CL pool. Vs inflationary rewards received for locking OSMO as staked OSMO.

Let me know if you have any other questions 👍

packages/web/components/pool-detail/concentrated.tsx
jonator: In response:

`[positions: [coins]]`

1. I think we have a 2d array, or tree with height of 2, that needs to be traversed here (see above): an array of positions > then for each position an array of coin rewards. 90%+ of positions will have only 2 coins as rewards; the coins that the pool is comprised of (as users swap those coins, tiny amounts are left in positions that are used for the swap as the reward). Then, if the pool's being externally incentivized, that will perhaps add one additional coin in the remaining 10% of cases if that coin is different from the coins in the pool.  With that knowledge, I wasn't really thinking about performance as most users will have a few positions; my guess is it'd only get noticeable if you had 100s of positions (we'd have performance concerns elsewhere in that case 😅)
2. I agree, we can make this code easier to scan with some extraction 👍

MaxMillington (approved):

JoseRFelix (approved):

PR #2003, opened 2023-08-14
Fix: map y reserve error to `NotEnoughLiquidity` error

jonator:

With stableswap pools, our math layer would throw a prototype error with `cannot input more than y reserve` as the message. This essentially meant that there was not enough liquidity in the pool for the given amount (confirmed with @AlpinYukseloglu ). 

Problem: that Error would bubble up and would not be handled by our routing and swap UI config layer. It was basically considered an unexpected error. Now, in our pools layer, we catch that specific prototype error and cast it into a `NotEnoughLiquidityError` so it can properly be handled in our router. 

The error would arise _repeatedly_ as our router was testing various splits into a route containing a stableswap pool.

UPDATE: the fix was we were letting some 0 liq stable pools (and CL pools) into the router. The fix was to remove those low liq pools from going to the router, but I still found a bug in the route splitter algorithm that wouldn't handle this case under specific circumstances.

Recorded here within our "Router improvements" parent task: https://app.clickup.com/t/86a0fe27e

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0e8mxt)

## Brief Changelog

* In stable pool implementation, the `cannot input more than y reserve` prototype error is rethrown as a `NotEnoughLiquidityError`

## Testing and Verifying

Try trading OSMO>STRD. The router finds a split containing a stable pool and tests swaps through the stable pool. Now, the console should be free of `cannot input more than y reserve` errors, and the swap tool should provide a valid quote now that it knows that route with the stable pool does not contain enough liquidity.

### DIFF
--- a/packages/math/src/pool/stable.ts
+++ b/packages/math/src/pool/stable.ts
@@ -1,4 +1,4 @@
-import { Coin, Dec, DecUtils, Int } from "@keplr-wallet/unit";
+import { Coin, Dec, DecUtils } from "@keplr-wallet/unit";
 
 import { BigDec } from "../big-dec";
 import { checkMultiplicativeErrorTolerance } from "../rounding";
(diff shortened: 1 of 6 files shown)


jonator:
Wait I thought I fixed it but it's not working now..

PR #2011, opened 2023-08-16
Update Stride Token Logo

JeremyParish69:

To update the Stride Token Logo
This was already changed like a month ago, but since the old version was cached under the same image name, the updated image never appeared. Thus, I have to change the image to force the deployment to not used the old cached version.

## Brief Changelog

- changes the name of the stride token logo image in source-chain-config.ts and of the actual image

## Testing and Verifying


jonator:
Any cleanup of images we can do?

JeremyParish69:
@jonator I don't think so? Depends what you mean--compressing, removing unused, and/or optimizing appearance? Even so, I proposed that be it's own task. This is the only case to my knowlede at this time.

PR #2013, opened 2023-08-17
Update Certik Endpoints

kamal-sutra:
Updating certik endpoints due to Cors restriction with previous one


JeremyParish69:
The issue of CORS blocking has been resolved for CTK Shentu and there's no longer any need to update the endpoints.

PR #2014, opened 2023-08-17
(unblock v17) fix: ignore unsupported pool types

jonator:

If we encounter an unexpected pool type, like CW pools, we simply exclude it from our query pools list.

### ClickUp Task


czarcas7ic (approved):
awesome, ty!

PR #2015, opened 2023-08-17
feat: Redirect User to Wallet Compatible Sites

JoseRFelix:

Currently, stake and vote links in the sidebar redirect users to external sites that lack Leap and Cosmostation wallet support. For a good UX experience, external sites should support currently connected wallets if possible. 

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/866agbjqx)

## Testing and Verifying

- [x] Redirect users to wallet-compatible sites when they click Stake or Vote. Note: xDefi will send users to Keplr given there is no site that is compatible with it.


arispech (approved):
Working as expected

PR #2025, opened 2023-08-20
Update Rebus Endpoints

CryptoAssassin1:
### ClickUp Task


CryptoAssassin1:
@JeremyParish69 i tested the withdraw locally its working fine, on Deposit we need to increase the gas fee i think 
photo_2023-08-20_12-44-29

JeremyParish69:
> @JeremyParish69 i tested the withdraw locally its working fine, on Deposit we need to increase the gas fee i think photo_2023-08-20_12-44-29
 @CryptoAssassin1 I thought this was a case where it doesn't deposit because it's an EVM asset. I think the gas error is new and unrelated.

CryptoAssassin1:
got it.

PR #2029, opened 2023-08-21
Asset Integrations: Add kava.USDT

JeremyParish69:

To add USDT from Kava

## Brief Changelog

-adds kava to ibc-assets, and source-chain-config
-defined usdt price as pool 1081 spot price (NOT via the 'tether' coingecko id)

## Testing and Verifying

Validation (to Kava IBC) successful

## Pending
Decision on whether to link out to their Kava EVM transfer UI, which is geoblocked for some users


packages/web/config/generate-chain-infos/source-chain-infos.ts
```diff
@@ -2613,6 +2613,15 @@ const mainnetChainInfos: SimplifiedChainInfo[] = [
         //coinGeckoId: "pool:usdx",
         coinImageUrl: "/tokens/usdx.svg",
       },
+      {
+        coinDenom: "USDT",
+        coinMinimalDenom: "erc20/tether/usdt",
+        coinDecimals: 6,
+        //coinGeckoId: "tether",
```
MaxMillington, line 2620: @JeremyParish69 why do some of our assets have `//coinGeckoId` commented out? I notice some of them have different coinGeckoIds?

packages/web/config/ibc-assets.ts
```diff
@@ -156,19 +156,15 @@ export const IBCAssetInfos: (IBCAsset & {
...
+        destChannelId: "channel-1",
+        coinMinimalDenom: "erc20/tether/usdt",
         isVerified: true,
-        originBridgeInfo: {
-          bridge: "axelar" as const,
-          wallets: ["metamask" as const, "walletconnect" as const],
-          method: "deposit-address" as const,
-          sourceChainTokens: [AxelarSourceChainConfigs.usdt.ethereum],
-        },
+        //depositUrlOverride: "https://app.kava.io/transfer",
```
MaxMillington, line 164: same question. Why is this commented out?

packages/web/config/generate-chain-infos/source-chain-infos.ts
JeremyParish69: CoinGeckoId is used for assigning a price to the asset. We used to use mostly ids that were actually from CoinGecko, but now we define custom ids (any id starting with 'pool:' is a custom id) in the price.ts file that looks at on-chain pool balances to derive a price.
If needed, it would be easier to switch to actual CoinGeckoIds if they are just commented out.
Also, for when we can separate chain suggestion from our displayed price, we'll want to use actual CoinGecko Ids (e.g., 'tether') for the Chain Suggestion to wallets, and we'll want to use our custom pricing for our frontend.

packages/web/config/ibc-assets.ts
JeremyParish69: Normal ibc transfers just go to IBC chains, and that's what we're doing for now (because it's commented out), but that only goes to Kava IBC, not Kava EVM (used by CEXes). The commented out URL is Kava's transfer page to turn USDT on Kava IBC to USDT on Kava EVM, but it's geoblocked. So we'll soon want override these Deposit and Withdraw links to redirect to some other page that can do that transfer (maybe this one once geoblock is removed).

MaxMillington (approved):

PR #2039, opened 2023-08-23
save repsonse to keep validators in local storage. Wire up stake call…

MaxMillington:

<!-- > Add a description of the overall background and high level changes that this PR introduces

This PR will save the response to keep validators in local storage so that users don't see it every time they press stake.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/8678m12z5)

## Testing and Verifying

### DIFF
--- a/packages/web/pages/stake/index.tsx
+++ b/packages/web/pages/stake/index.tsx
@@ -59,7 +59,7 @@ export const Staking: React.FC = observer(() => {
       setLoading(false);
     }
 
-    // checkFeatureFlag();
+    checkFeatureFlag();
 
     setLoading(false);
   }, [flags.staking]);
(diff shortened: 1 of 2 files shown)


packages/web/pages/stake/index.tsx
```diff
@@ -206,10 +208,14 @@ export const Staking: React.FC = observer(() => {
       return;
     }
 
-    // TODO add showValidatorNextStepModal here
+    const selectedKeepValidators = localStorage.getItem("keepValidators");
```
MaxMillington, line 211: Decided to use localhost because our use case is simple enough not to require indexedDb. in the future, we will look into storing this on chain with the `address` as a key.

jonator: https://github.com/osmosis-labs/osmosis-frontend/blob/581b32f4fe78e0fe1abf7ea70f86e3feea27f1fd/packages/web/hooks/window/use-localstorage-state.ts#L7

Heads up we do have this `useLocalstorageState` hook that may simplify calling code

PR #2042, opened 2023-08-23
fix: Re-Render Issue With Deposit/Withdraw Modal

JoseRFelix:


jonator (approved):
https://app.clickup.com/t/86a0jj2km

Task to track lint rule we should add to catch this

MaxMillington (approved):
@JoseRFelix do these changes mean that we will now get a type error if default value is undefined? Are there any other modals that could be susceptible to similar issue?

PR #2044, opened 2023-08-23
Hotfix: hide pools in blacklist

jonator:
### ClickUp Task

### DIFF
--- a/packages/web/config/feature-flag.ts
+++ b/packages/web/config/feature-flag.ts
@@ -44,7 +44,15 @@ export const Announcement:
 // Fiat ramps
 export const BUY_OSMO_TRANSAK = true;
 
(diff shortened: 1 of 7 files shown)


packages/web/config/feature-flag.ts
```diff
@@ -44,7 +44,7 @@ export const Announcement:
 // Fiat ramps
 export const BUY_OSMO_TRANSAK = true;
 
-export const HiddenPoolIds: string[] = [];
+export const BlacklistedPoolIds: string[] = ["3", "5", "7", "9", "15"];
```
JoseRFelix, line 47: Nit: I think it would be a good idea to add a comment to identify what these pools are

JoseRFelix (approved):

PR #2047, opened 2023-08-24
[Notifi] bug fix: breadcrumb always shows constantly/ Mobile view crash

eric-notifi:
### Reason 
This [commit](https://github.com/osmosis-labs/osmosis-frontend/commit/1e92b3d7e3b0bd9a7ce567cd9e0ca4246a3d4fe9) moved the `NotifiContextProvider` out of _app scope. it cause the issue because `NavBar` consume the context

### Solution
Conditionally wrap `NavBar` with provider so that the context is available in `NavBar`

### NOTE
I see that this [commit](https://github.com/osmosis-labs/osmosis-frontend/commit/1e92b3d7e3b0bd9a7ce567cd9e0ca4246a3d4fe9) is to fix "build flakiness". Not sure it it relates to this change. If there is any impact, please kindly point it out.


jonator (approved):

JoseRFelix:
@eric-notifi Can you please verify my changes in stages 🙏  I wrapped the context only for the modal where I believe it's required. I prefer to avoid wrapping the navbar to prevent unnecessary renders.

eric-notifi:
@JoseRFelix ... I understand we better avoid unneeded render as possible. However, in this case, FE will need to consume the `NotifiContext` in `NavBar` because we will need to let user know if there is any `unread notification` (the Breadcrumb feature) even through the notifi popover is not rendered.
Accordingly, I would say the better way is conditionally wrap the `NavBar`.

packages/web/components/layouts/main.tsx
```diff
@@ -44,11 +47,22 @@ export const MainLayout: FunctionComponent<{
           <NavbarOsmoPrice />
         </div>
       </article>
-      <NavBar
-        className="ml-sidebar md:ml-0"
-        title={selectedMenuItem?.label ?? ""}
-        menus={menus}
-      />
+      {/** NavBar needs to be wrapped in NotifiContextProvider because NavBar consumes its context */}
```
JoseRFelix, line 50: I think we can remove this comment

JoseRFelix:
@eric-notifi I just left one comment. After resolving that, we can merge. Now that we'll display the unread count on the notification, it does make sense to wrap the navigation because of the mobile button.

packages/web/components/layouts/main.tsx
eric-notifi: Cool.. ty! The comment is removed.

JoseRFelix (approved):

PR #2070, opened 2023-08-29
Feat: Convert to stake modal

jonator:

Similar to the upgrades flow, this feature provides the user with a modal accessible from the pools page that allows them to select share pools that are earning a lower APR then the staking APR. From that list, the user can select a share pool to migrate via a custom ConvertToStake message that handles the migration of all forms of gamm shares (available, bonded, unbonding, SF bonding & SF unbonding).

Screenshot 2023-08-28 at 8 30 59 PM

NOTE: blocked on working with the chain devs on issues serializing lockIds with a value of 0, as we had with the migrate position message. 

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/862jx2q5t)

## Brief Changelog

* Adds a `UserConvertToStakeConfig` that encapsulates all the logic around querying for stake opportunities, selecting the desired pool, as well as sending the respective stake opportunity via transaction to the chain
* Adds a modal that depends on the above ConvertToStake UI config. If the user has not selected a validator set preference, the modal will jump to the validator selection modal before sending the transaction.

## Testing and Verifying

- [x] Works with available GAMM shares
- [x] Works with bonded and unbonding superfluid shares
- [x] Works with bonded and unbonding normal GAMM shares
- [x] FIX: refreshing data after converting; pools still showing
- [x] Change only using validator select if _no_ delegations were ever made

Successful un bond and stake for locked shares initiated from frontend:
https://www.mintscan.io/osmosis/transactions/5AE07FA3B61943145FAB7160C8F378539A9DA8789D7FA43654A37970140659E1

The message and valsetprefs functionality is already present on mainnet and other dev chains.


packages/stores/src/account/osmosis/index.ts
```diff
@@ -1605,7 +1606,223 @@ export class OsmosisAccountImpl {
...
+    );
+
+    const stakeCurrency = this.chainGetter.getChain(this.chainId).stakeCurrency;
+
+    const involvedQueryPools: ObservableQueryPool[] = [];
+
+    /** Locally relevant function for calculating the minimum amount of staking token to stake by converting
+     *  GAMM shares and swapping the GAMM shares in the pool to the staking token.
+     *  Provided to msg for slippage and price impact protection of user. */
+    const calcMinAmtToStake = async (shares: CoinPrimitive): Promise<Int> => {
```
MaxMillington, line 1643: This whole function is pretty long. What about extracting calcMinAmtToStake and then just referenced in the sendUnbondAndConvertToStake ?

packages/stores/src/account/osmosis/index.ts
```diff
@@ -1605,7 +1606,223 @@ export class OsmosisAccountImpl {
...
+          return new Promise((resolve, reject) =>
+            querySharePool.pool
+              .getTokenOutByTokenIn(token, stakeCurrency.coinMinimalDenom)
+              .then((quote) => resolve(quote.amount))
+              .catch(reject)
+          );
+        }
+      });
+
+      return await Promise.all(swapPromises).then((swaps) =>
```
MaxMillington, line 1718: 👍

packages/stores/src/account/osmosis/index.ts
```diff
@@ -1605,7 +1606,223 @@ export class OsmosisAccountImpl {
...
+            coin.denom.includes("/pool/")
+          );
+
+          // this lock contains some unexpected asset
+          if (!poolAssetCoin)
+            throw new Error(
+              `Lock ID ${asset.lockId} contains some unexpected asset`
+            );
+
+          if (poolAssetCoin.denom.includes("gamm/pool/")) {
```
MaxMillington, line 1741: We should probably declare 'magic strings' as constants.

packages/stores/src/account/osmosis/index.ts
```diff
@@ -1605,7 +1606,223 @@ export class OsmosisAccountImpl {
...
+
+    return this.base.signAndBroadcast(
+      this.chainId,
+      "convertAndStake",
+      msgs,
+      memo,
+      undefined,
+      undefined,
+      (tx: DeliverTxResponse) => {
+        if (tx.code == null || tx.code === 0) {
```
MaxMillington, line 1792: tx.code can be null or 0?

packages/stores/src/queries/concentrated-liquidity/account-positions.ts
```diff
@@ -58,17 +60,62 @@ export class ObservableQueryAccountPositions extends ObservableChainQuery<{
...
     return this.positions.filter((position) => position.poolId === poolId);
   });
+
+  /** Aggregates the coins in the given positions array. */
+  protected readonly getPositionsAssets = computedFn(
+    (positions: ObservableQueryLiquidityPositionById[]) => {
+      return Array.from(
+        positions
+          .reduce((balances, position) => {
+            const addToMap = (coin: CoinPretty) => {
```
MaxMillington, line 93: I think you might be able to simplify to avoid `if` `else`

```
const addToMap = (coin: CoinPretty) => {
  balances.set(
    coin.currency.coinMinimalDenom,
    (balances.get(coin.currency.coinMinimalDenom) || new CoinPretty(0)).add(coin)
  );
};

```

packages/stores/src/queries/valset-pref/users-validator-prefs.ts
```diff
@@ -0,0 +1,65 @@
...
+import { computed, makeObservable } from "mobx";
+
+type QueryUserValidatorPreferencesResponse = {
+  preferences: {
+    val_oper_address: string;
+    weight: string;
+  }[];
+};
+
+export class ObservableQueryUserValidatorPreferences extends ObservableChainQuery<QueryUserValidatorPreferencesResponse> {
```
MaxMillington, line 16: What's the difference between validator preferences and user delegations? In `stake/index` we use `cosmosQueries.queryDelegations.getQueryBech32Address` to get the validators. Not sure if those are the same or not

packages/stores/src/ui-config/convert-to-stake.ts
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /** For each user owned pool, this provides a breakdown of shares
+   *  that are of lesser returns than the staking inflationary returns.
+   *
+   *  Further, it provides a callback for converting those assets to stake via the
+   *  given account store. This callback expects a validator address if the user
+   *  has not yet selected a validator set preference. */
+  @computed
+  get suggestedConvertibleAssetsPerPool(): SuggestedConvertToStakeAssets[] {
```
MaxMillington, line 51: This function is rather long, can any of the code be extracted into helper functions to make it easier to read?

packages/web/components/funnels/convert-to-stake/available-conversions.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <span className="subtitle1 text-bullish-100">
+            {t("convertToStake.stakedOsmo")}
+          </span>
+          <h6 className="mr-4 text-bullish-300">
+            {`${convertToStakeConfig.stakeApr.maxDecimals(1)}% ` +
+              t("convertToStake.APR")}
+          </h6>
+        </div>
+      </div>
+      <div className="lex-col flex h-60 flex-col gap-2 overflow-y-scroll">
```
MaxMillington, line 43: do you mean `flex-col`

MaxMillington (commented):
Reminder to run the translation function

packages/stores/src/queries/valset-pref/users-validator-prefs.ts
jonator: I meant to follow up on this today. I think you guys are supposed to integrate with valsetprefs query in staking project, given that you use the valsetpref staking message to stake. We should open this topic in the frontend channel.

packages/stores/src/ui-config/convert-to-stake.ts
jonator: What if I annotate chunks of code with high-level comments that don't repeat the code? Then, I can achieve the same effect without introducing new free-standing components to be reconciled.

packages/stores/src/account/osmosis/index.ts
jonator: Where should the constants live?

packages/stores/src/account/osmosis/index.ts
jonator: Rationale for placement here:
* Given that it's a closure, it contains some vars from the immediate scope
* I have high confidence that the logic is only relevant and useful within this function, given how bespoke it is

Let me know what you think given the above points.

packages/stores/src/account/osmosis/index.ts
MaxMillington: Yea, I think that makes sense since it would capture undefined.

packages/stores/src/queries/valset-pref/users-validator-prefs.ts
MaxMillington: Agree. It seems to work correctly as coded now, but if we are missing something we def need outside eyes to look at it

packages/stores/src/ui-config/convert-to-stake.ts
MaxMillington: I'll leave it up to you. I think that annotations help, but in general whenever I see 100+ lines in a function, my eyes start to glaze over 😆

sunnya97:
1. it currently sends users to the choose superfluid validator modal, even if they already have some OSMO staked (and so should be able to just use the valset prefs msg)
2. The modal currently says "select superfluid validator", but this isn't for superfluid

CleanShot 2023-09-01 at 15 55 44@2x

jonator:
@sunnya97 
1. It currently should directly use valsetpref it one is present
2. I'll fix this
3. This is left over from prev design. Should I remove it for all use cases?

sunnya97:
1. You should still be able to use the valset prefs stake msg, as long as they have any delegations (even if not via a set valset pref)
2. 👍 
3. Yeah I think so.  It looks weird.

JoseRFelix (approved):

JoseRFelix (changes requested):
I misclicked on my previous review 😅

packages/stores/src/queries/lockup/account-locked.ts
```diff
@@ -407,6 +407,14 @@ export class ObservableQueryAccountLockedInner extends ObservableChainQuery<Acco
       });
     }
   );
+
+  readonly getPeriodLockById = computedFn((lockId) => {
+    if (!this.response) {
+      return undefined;
```
JoseRFelix, line 413: Have you observed any problems with this? Previous state issues occurred because observables had an initial `undefined` value. I'm not sure if this would also be related with `computedFn`

packages/stores/src/ui-config/convert-to-stake.ts
```diff
@@ -0,0 +1,264 @@
...
+      // only include if better opportunity
+      if (currentApr.toDec().lte(this.stakeApr.toDec().quo(new Dec(100)))) {
+        conversions.push({
+          poolId: sharePoolId,
+          totalValue,
+          userPoolAssets,
+          currentApr,
+        });
+      }
+    });
```
JoseRFelix, line 112: hat do you think about using the reduce function? I believe using it would simplify understanding the loop

```tsx
const conversions: SuggestedConvertToStakeAssets[] = ownedSharePoolIds.reduce(
  (acc, sharePoolId) => {
    const { sharePoolDetail, superfluidPoolDetail, poolBonding } =
      this.derivedDataStore.getForPool(sharePoolId);

    if (
      !sharePoolDetail.querySharePool?.hasPoolAsset(
        stakeCurrency.coinMinimalDenom
      )
    ) {
      return acc;
    }

    if (
      !(
        sharePoolDetail.userStats?.bondedValue?.toDec().isPositive() ||
        sharePoolDetail.userUnlockingAssets.length > 0
      )
    ) {
      return acc;
    }

    const totalValue =
      sharePoolDetail.userStats?.totalShareValue ??
      new PricePretty(this.fiatCurrency, 0);
    const userPoolAssets = sharePoolDetail.userPoolAssets.map(
      ({ asset }) => asset
    );

    const currentApr =
      poolBonding.highestBondDuration?.aggregateApr ??
      superfluidPoolDetail.superfluidApr.add(sharePoolDetail.swapFeeApr);

    if (currentApr.toDec().lte(this.stakeApr.toDec().quo(new Dec(100)))) {
      acc.push({
        poolId: sharePoolId,
        totalValue,
        userPoolAssets,
        currentApr,
      });
    }

    return acc;
  },
  []
);

```

packages/stores/src/ui-config/convert-to-stake.ts
```diff
@@ -0,0 +1,264 @@
...
+    // TODO: returning false since we can't serialize this case in the message
+    // this case prevents the message from containing an empty validator to defer to valset of delegations
+    return false;
+
+    // return (
+    //   this.cosmosQueries.queryDelegations.getQueryBech32Address(
+    //     this.accountAddress
+    //   ).delegations.length > 0
+    // );
+  }
```
JoseRFelix, line 145: Is it safe to merge this to prod without these functional boolean flags?

JoseRFelix (commented):
It's looking very good; great job! It's refreshing to see how tidy the components are now, with most of the logic moved to the config. This change also facilitates reusing these stores on other platforms, such as mobile. 🚀

packages/stores/src/queries/lockup/account-locked.ts
jonator: I think it's fine with computeds. It will still register with the response to listen for changes for a given `lockId`.

packages/stores/src/ui-config/convert-to-stake.ts
jonator: Yeah, as these flags are used to ensure the user selects a validator every time.

Then since a validator is selected, we won't have to use a null value for the validator address in the message, therefore avoiding the nullable value signature error

packages/stores/src/queries/pool-share/index.ts
```diff
@@ -244,11 +244,11 @@ export class ObservableQueryPoolShare {
 
   fetch(bech32Address: string) {
     return Promise.all([
-      this.queryPools.fetch(),
```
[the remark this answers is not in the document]
jonator: Fetch will only fetch if stale, and is not async. `waitFreshResponse` will fetch regardless if it is stale, basically guaranteeing new data. Fetch should be rarely used, since the query store will automatically keep data up to date.

packages/stores/src/account/osmosis/index.ts
```diff
@@ -1605,7 +1611,232 @@ export class OsmosisAccountImpl {
...
+      "convertAndStake",
+      msgs,
+      memo,
+      undefined,
+      undefined,
+      (tx: DeliverTxResponse) => {
+        if (!Boolean(tx.code)) {
+          involvedQueryPools.forEach((queryPool) => {
+            // refresh pool that was exited
+            queryPool.waitFreshResponse();
```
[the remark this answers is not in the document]
jonator: No need to await here, as we don't immediately depend on any of the data within this scope.

MaxMillington (approved):

JoseRFelix (approved):

PR #2076, opened 2023-08-29
refactor: de-duplicate summing coins

ValarDragon:

De-duplicates some logic for summing coin pretty's. I plan to then make a PR speeding these up later.

## Testing and Verifying

This change has not been tested locally, because I expect CI & unit tests to catch any difference here.

- Does this pull request introduce a new feature or user-facing behavior changes? no
- How is the feature or change documented? N/A

### DIFF
--- a/packages/web/utils/total-value.ts
+++ b/packages/web/utils/total-value.ts
@@ -1,0 +1,36 @@
+import { FiatCurrency } from "@keplr-wallet/types";
+import { CoinPretty, PricePretty } from "@keplr-wallet/unit";
+import { PoolFallbackPriceStore } from "@osmosis-labs/stores";
+
+// TODO: Change this out of the price pretty overhead, in a later iteration.
+export function coinsToFiatValue(
+  priceStore: PoolFallbackPriceStore,
+  fiat: FiatCurrency,
+  coins: CoinPretty[]
+): number {
+  return Number(
+    coinsToFiatPricePretty(priceStore, fiat, coins).toDec().toString()
+  );
+}
+
+// TODO: Do we actually want this anywhere?
+export function coinsToFiatPricePretty(
+  priceStore: PoolFallbackPriceStore,
+  fiat: FiatCurrency,
+  coins: CoinPretty[]
+): PricePretty {
+  return coins.reduce(
+    (sum, asset) =>
+      sum.add(priceStore.calculatePrice(asset) ?? new PricePretty(fiat, 0)),
+    new PricePretty(fiat, 0)
+  );
+}
+
+export function objCoinsToFiatPricePretty(
+  priceStore: PoolFallbackPriceStore,
+  fiat: FiatCurrency,
+  coins: { asset: CoinPretty }[]
+): PricePretty {
+  let coinPrettyArr: CoinPretty[] = coins.map((obj) => obj.asset);
+  return coinsToFiatPricePretty(priceStore, fiat, coinPrettyArr);
+}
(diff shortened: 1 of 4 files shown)


packages/web/utils/total-value.ts
```diff
@@ -0,0 +1,36 @@
...
+  coins: CoinPretty[]
+): number {
+  return Number(
+    coinsToFiatPricePretty(priceStore, fiat, coins).toDec().toString()
+  );
+}
+
+// TODO: Do we actually want this anywhere?
+export function coinsToFiatPricePretty(
+  priceStore: PoolFallbackPriceStore,
```
jonator, line 18: IMO we'd get more type flexibility if we used `IPriceStore` instead of the concrete `PoolFallbackPriceStore` that we have.

Stepping back, I think we'd make good use of a similar function on a new `IPriceStore` that we could add to our impl at `PoolFallbackPriceStore` to take a list of coins and return a single price. Rather than have these free floating util functions. I think it may be more discoverable and of higher locality. Thoughts?

Stepping all the way back and longer term, and I've brought this up with tony and Josh. I think we should extract the keplr-wallet architecture into a cosmos-eco maintained set of packages where the price store mentioned above could include this functionality for the greater benefit of the Cosmos frontend ecosystem. I think the keplr framework we're using has offered tons of value and has been a great interchain foundation to build from but it should be decoupled from Keplr team's initiatives and maintained and branded separately. cc @pyramation cc @sunnya97

ValarDragon: * Will change the concrete price store struct to the interface
* I don't quite understand our stores abstractions, so I don't have an opinion. I generally prefer just getting everything into one function first that is obviously state compatible, and then in a second pass moving how we like longer term. Happy to go with whatever, until I get a better sense of how we use stores.


Pretty sure theres universal dependence to get off of keplr's type system for this stuff :)

e.g. In the relevant code, we should move to just floats. 
What I was hoping to do here was:
* In a later PR, get tests for the unit functions, and then actually change the internal business accounting to floats, and return a final price pretty.
* Have a discussion later on for a single function for going from PricePretty (or whatever "price type" we use) to display, so we don't do ad-hoc ToDec().Stirng()'s, and manually set precisions around the codebase.

jonator: We should definitely get on a call to discuss this stuff if you consider it high priority, as I think there's a lot of background on the state of the frontend

jonator:
@ValarDragon 

I remembered we also have this in the user assets store:

https://github.com/osmosis-labs/osmosis-frontend/blob/14d06648db4835109198e3784407d411796f8e42/packages/web/stores/assets/index.ts#L244

ValarDragon:
That'd be a functionality change for these (and also is a bit odd that its not already handled in the price store)

Can that be targetted for dedupping in a followup?

jonator:
> That'd be a functionality change for these (and also is a bit odd that its not already handled in the price store)

Yeah I agree this would be good functionality to add to the price store. To conclude changes to prices store:
* Calculate value of one _or more_ tokens
* Will handle special case for gamm shares, given assumptions we can make about the PoolPriceStore's use for just the Osmosis chain

Also, @JeremyParish69 has requested in the past that we somehow remove the need to store routing config for finding pools to calculate prices from. Thoughts on maintaining an instance of `OptimizedRoutes` here and using that to find prices through pools?

JeremyParish69:
It's moreso that I wanted to move the config out of this repo. There's a script that finds the largest liquidity route from USDC.axl and records that in assetlists repo.
e.g., https://github.com/osmosis-labs/assetlists/blob/f4ad686edc8608a2b8598c5fdf1b1d59b6bf47f5/osmosis-1/osmosis-1.assetlist.json#L10644

ValarDragon:
I feel like this is a lot of comments for a simple refactor PR, can we do this discussion in an issue?

Would like to just get more mobility in moving things around for simple refactors.

I think theres a significant amount of changes we should be doing to the price store :)

jonator:
@ValarDragon could you put your ideas here? I'm about to fix a display issue with the price of CL tokens and I can do some refactoring there. So far, I'll:
* Allow calculate price to take an arbitrary number of CoinPretty parameters (hoping this allows a back compat API)
* Improve the calculate price function to handle gamm shares which would be easy given the pools query store is supplied

jonator:
@ValarDragon LMK what you think https://github.com/osmosis-labs/osmosis-frontend/pull/2101

PR #2079, opened 2023-08-30
feat: Re-Enable Notifications for Leap

JoseRFelix:


jonator:
Thought: what about using a feature of "sign-direct" instead of "notifications" since that was the root cause of the issue?

JoseRFelix:
@jonator That's a great idea! I'll add it to this PR. Btw, let's not merge this one yet. I'm having some issues. Let the Notifi team know

JoseRFelix:
@jonator This is ready after Notifi fix in #2096. Btw, reverted the feature flag rename because I believe `notifications` is more descriptive than `sign-arbitrary`. Some wallets support sign-arbitrary, but have issues when enabling notifications.

jonator (approved):

PR #2083, opened 2023-08-30
feat: Add New Apps Counter

JoseRFelix:
image

## Testing
 
- [x] The "New Apps" count should show 6 today (30 Aug 2023). However, it will reset to zero tomorrow because the counter only counts apps as new for a duration of 31 days.
- [x] When the user navigates to `/apps`, the counter should hide and not show up after reload


packages/stores/src/queries-external/github/get-file.ts
```diff
@@ -0,0 +1,68 @@
...
+    protected readonly commitHash?: string,
+    protected readonly baseUrl = GITHUB_RAW_DEFAULT_BASEURL
+  ) {
+    super(
+      kvStore,
+      baseUrl,
+      `/${repo}/${commitHash ? commitHash : "main"}/${filePath}`
+    );
+
+    makeObservable(this);
```
MaxMillington, line 24: why would this function need to be observable?

packages/stores/src/queries-external/github/get-file.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export class ObservableQueryFileInner<
+  T
+> extends ObservableQueryExternalBase<T> {
+  constructor(
+    kvStore: KVStore,
+    protected readonly repo: string,
+    protected readonly filePath: string,
+    protected readonly commitHash?: string,
+    protected readonly baseUrl = GITHUB_RAW_DEFAULT_BASEURL
```
MaxMillington, line 16: 👍

packages/stores/src/queries-external/github/get-file.ts
```diff
@@ -0,0 +1,68 @@
+import { KVStore } from "@keplr-wallet/common";
+import { HasMapStore } from "@keplr-wallet/stores";
+import { makeObservable } from "mobx";
+
+import { GITHUB_RAW_DEFAULT_BASEURL } from "..";
+import { ObservableQueryExternalBase } from "../base";
+
+export class ObservableQueryFileInner<
+  T
+> extends ObservableQueryExternalBase<T> {
```
MaxMillington, line 10: Does this function need to be in the store package? As I understand it, part of the purpose of putting some queries in a separate stores package is because we will someday ideally have docs so that other devs can use our open source queries. Does this one fit that model? Or am I misunderstanding something? Perhaps @jonator can shed some light as well.

packages/web/hooks/window/use-localstorage-state.ts
```diff
@@ -26,16 +27,20 @@ export function useLocalStorageState<T>(
     }
   }, [key, setStoredValue]);
 
-  const setValue = (value: T) => {
-    try {
-      if (typeof window !== "undefined" && key) {
-        window.localStorage.setItem(key, JSON.stringify(value));
+  const setValue = useCallback(
```
MaxMillington, line 30: 👍

packages/web/hooks/window/use-localstorage-state.ts
```diff
@@ -26,16 +27,20 @@ export function useLocalStorageState<T>(
...
+        if (typeof window !== "undefined" && key) {
+          window.localStorage.setItem(key, JSON.stringify(value));
+        }
+
+        setStoredValue(value);
+      } catch (error) {
+        console.error(error);
       }
+    },
+    [key]
```
MaxMillington, line 42: is this the only dependency?

packages/web/pages/_app.tsx
```diff
@@ -185,6 +191,39 @@ const MainLayoutWrapper: FunctionComponent<{ children: ReactNode }> = observer(
...
+export const AppsBadge: FunctionComponent<{ appsLink: string }> = observer(
+  ({ appsLink }) => {
+    const { newApps, allApps } = useNewApps();
+    const router = useRouter();
+    const [viewedAppTitles, setViewedAppTitles] = useLocalStorageState<
+      string[]
+    >("viewed-apps", []);
+
+    useEffect(() => {
+      if (router.pathname === appsLink && allApps.length > 0) {
```
MaxMillington, line 204: can you put some comments here? I'm not exactly sure why you are doing it this way.

packages/web/pages/apps/index.tsx
```diff
@@ -75,16 +73,14 @@ export const AppStore: React.FC<AppStoreProps> = ({ apps }) => {
...
-      );
+      const appDataResults = searchResults.map((result) => result.item);
 
       setSearchValue(value);
       setFuzzySearchResults(appDataResults);
     }
   };
 
-  const iterableData = searchValue ? fuzzySearchResults : nonFeaturedApps;
+  const appsToDisplay = searchValue ? fuzzySearchResults : nonFeaturedApps;
```
MaxMillington, line 83: 👍

MaxMillington (commented):
Couple of questions. Looks good

packages/stores/src/queries-external/github/get-file.ts
JoseRFelix: Good catch. The `ObservableQueryExternalBase` is already calling the `makeObservable` internally. I'll remove this

packages/stores/src/queries-external/github/get-file.ts
JoseRFelix: I agree that this seems to be tightly coupled to our app. I'll move this to the `web` package.

packages/web/hooks/window/use-localstorage-state.ts
JoseRFelix: Yes, this was also confirmed by Eslint React rule

packages/web/pages/_app.tsx
JoseRFelix: I primarily did it to colocate the logic making it more straightforward. It also has the following benefits: 
1. It ensures that the viewed app titles are updated as soon as the user navigates to the apps page. Otherwise, we would need to rely on context, event emitters, or mobx to get the updated state within AppsBadge.
2. It reduces the complexity by not managing state in two files
3. It improves the performance of the apps page by avoiding re-renders whenever the viewed app titles change.

arispech (approved):
LGTM

PR #2084, opened 2023-08-31
feat: Hide Notifications Based on Feature Flag on Mobile Phones

JoseRFelix:
### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0nk2ut)

## Brief Changelog

- Add a feature flag to show/hide notifications on mobile phones

## Testing and Verifying

- [ ] Display notifications icon according to `mobile-notifications` feature flag state

### DIFF
--- a/packages/web/components/navbar/index.tsx
+++ b/packages/web/components/navbar/index.tsx
@@ -25,6 +25,7 @@ import {
   useAmplitudeAnalytics,
   useDisclosure,
   useLocalStorageState,
+  useWindowSize,
 } from "~/hooks";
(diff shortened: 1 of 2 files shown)


MaxMillington (approved):

jonator (commented):
What if we do the same thing we did with concentrated liquidity on mobile?

In our `useFeatureFlags` hook we could:
* If window size is mobile, use LDs mobileNotifications flag
* If window size is desktop, use LDs notifications flag

But to our app it's just concerned with a single `.notifications` flag

jonator (approved):

PR #2089, opened 2023-08-31
Unbonding UI

MaxMillington:

UI and logic for unbonding UI

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/8678hgn0w)

## Brief Changelog

Adds UI and logic for unbonding validators 

## Testing and Verifying

Screenshot 2023-08-31 at 3 50 29 PM


packages/web/components/cards/stake-info-card.tsx
```diff
@@ -118,7 +118,7 @@ export const StakeInfoCard: FunctionComponent<{
               outAmountValue ? "opacity-100" : "opacity-0"
             )}
           >
-            {`≈ $ ${outAmountValue || "0"}`}
+            {`≈ ${outAmountValue || "0"}`}
```
MaxMillington, line 121: noticed this UI bug

packages/web/pages/stake/index.tsx
```diff
@@ -268,6 +274,41 @@ export const Staking: React.FC = observer(() => {
 
   const showStakeLearnMore = !isWalletConnected || isNewUser;
 
+  const { unbondingBalances } = unbondingDelegationsQuery;
+  const unbondingInProcess = unbondingBalances.length > 0;
+
+  function groupByCompletionTime(
```
MaxMillington, line 280: As mentioned in my slack message, I think this is the only way to get a single stake object

packages/web/components/stake/unbonding-in-progress.tsx
```diff
@@ -0,0 +1,78 @@
...
+export const UnbondingInProgress: React.FC<{
+  unbondings: {
+    completionTime: string;
+    balance: CoinPretty;
+  }[];
+}> = ({ unbondings }) => {
+  const t = useTranslation();
+  const { priceStore } = useStore();
+
+  function formatUnbondings(
```
jonator, line 16: Is this function definition within the render body needed? It's somewhat hard to follow the logic, given the invocation in the render body below.

packages/web/pages/stake/index.tsx
```diff
@@ -268,6 +274,41 @@ export const Staking: React.FC = observer(() => {
...
+    array: Array<{
+      validatorAddress: string;
+      entries: { completionTime: string; balance: CoinPretty }[];
+    }>
+  ): { completionTime: string; balance: CoinPretty }[] {
+    const groupedObjects: Record<string, CoinPretty> = {};
```
jonator, line 286: Personally, I'd prefer a `Map` over using an object, with `Object.*` utils.

Also, perhaps it's possible to use flatMap or reduce over this "array" object? Instead of nested for loops. From experience, the only use case for `for` loops is when you want to save some iterations via `continue`. Otherwise functional programming is preferred.

Also, what is the content of completion time. Is it an ISO string? IMO proper Date or dayjs duration objects offer more type safety then string representations. These object ideally should be created at the query store later and consumed by the view.

MaxMillington: I can get rid of the nested loop. Good catch. 

Map and Object aren't going to have any functional difference here, since I am just storing strings and values.

`completionTime` is just whatever string is returned by the chain. I don't want to change the type since all I'm doing is a groupBy.

packages/web/components/stake/unbonding-in-progress.tsx
MaxMillington: given that this is just formatting for this view, it didn't seem like it was worth it to extract this function. Since this view is only rendered once, any performance difference is negligible.

jonator (approved):

PR #2091, opened 2023-08-31
fix: all CL pool APR showing 0%

JeremyParish69:

To allow CL pools to not only show 0% APR

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0nxe7u)

## Brief Changelog

-adds no-lock/e/ (CL) pool shares to the incentives query that feeds APR numbers

## Testing and Verifying

### DIFF
--- a/packages/web/pages/api/active-gauges.ts
+++ b/packages/web/pages/api/active-gauges.ts
@@ -29,7 +29,8 @@ export default async function activeGauges(
     data: data.filter(
       (gauge) =>
         !gauge.is_perpetual &&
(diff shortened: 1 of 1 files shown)


jonator:
@JeremyParish69 I don't think we have the logic to support this. It works differently than share pools (balancer, stable).

We're still waiting on the data team API to provide APRs for positions and Cl pools.

JeremyParish69:
Thanks, yes, that seems correc based on what I've seen (and this PR not fixing it)

PR #2115, opened 2023-09-07
[Token Info] Add Token Actions and Socials Layout

JoseRFelix:
Desktop

image

Mobile

image


packages/web/pages/assets/[denom].tsx
```diff
@@ -18,7 +21,85 @@ const AssetInfoPage: NextPage = () => {
...
+              id="coingecko"
+            />
+          }
+        />
+      </div>
+    </nav>
+  );
+};
+
+const SocialButton: FunctionComponent<{
```
Comment (jonator) [inline packages/web/pages/assets/[denom].tsx:84]: Thoughts on moving this to components/buttons

Comment (JoseRFelix) [reply packages/web/pages/assets/[denom].tsx:84]: I agree. This is a good opportunity to add a `LinkIconButton` component

jonator (approved):

PR #2125, opened 2023-09-10
mattupham/table performance + sorting

mattupham:

- previously the table was slow when rendering / checking boxes
- virtualization also had issues too where part of the table would not be visible on scroll 
- sorting was not working as expected due to performance challenges

### ClickUp Task

https://app.clickup.com/t/8678k6rvk
https://app.clickup.com/t/8678k6rxe

## Brief Changelog

- pull out all formatted data into functions with useCallback
- remove virtualization -> the table is surprisingly much more performant without virtualization
- clean up table rendering, add react table selectors - row.getIsSelected() - row.getToggleSelectedHandler()
- memoize table, reduce overall renders
- note, there's a still slight flash on the re-render with our custom components, but that's much better than the slowness we were experiencing previously

## Testing and Verifying

- after, snappier, sort working

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/8744bfa9-d43c-495d-a8c8-f96b7570ce4a

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/80ce2d35-e007-4705-a8f1-1592432acfd0

- before, slow, laggy, virtualization was a bit bugged

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/e2ef0b16-d9ea-4b69-b9e5-9938e62219fa

- rows selected, validator length, eth
Screenshot 2023-09-10 at 13 20 04

### DIFF
--- a/packages/web/modals/validator-squad.tsx
+++ b/packages/web/modals/validator-squad.tsx
@@ -68,6 +68,8 @@ export const ValidatorSquadModal: FunctionComponent<ValidatorSquadModalProps> =
 
     const defaultUserValidatorsSet = new Set(usersValidatorsMap.keys());
 
+    // TODO clean up the default validators
+    // @ts-ignore
     const [selectedValidators, setSelectedValidators] = useState(
       defaultUserValidatorsSet
     );


MaxMillington (approved):
Code looks fine to me, but I'm troubled that we are seeing flashing during re-renders. 

It's been 3-4 years since I used react-table, but react-table used to be the easy/reliable table library (not the most performant, but fine for a few hundred rows). The fact that we run into these issues with a fairly simple table isn't super inspiring. 

(i agree we don't need virtualization here)

@mattupham did you research `react-virtualized`? 

A few thoughts on what could be causing the flash

1) Try memoizing the individual custom rows? 
2) Make sure we are using `key` in our maps and make sure they stay the same across renders?
3) Did you try using React Dev Tools to inspect re-renders?

Feels like this isn't a react problem though. Feels like a react-table problem

packages/web/modals/validator-squad.tsx
```diff
@@ -98,22 +93,111 @@ export const ValidatorSquadModal: FunctionComponent<ValidatorSquadModalProps> =
...
-          })),
-      [validators, queryValidators, getVotingPower, getMyStake]
+          .map((validator) => {
+            const votingPower = getVotingPower(validator);
+            const myStake = getMyStake(validator);
+
+            const formattedVotingPower = getFormattedVotingPower(votingPower);
+            const formattedMyStake = getFormattedMyStake(myStake);
+
+            const commissions = getCommissions(validator);
```
jonator, line 162: Are these functions necessary? You'll see this concept in the book I recommended, but IMO if the sole purpose of a function is to label what is happening while at the same time offering little functionality (here, just creating a pretty object, or calling another function), I'd rather have those operations be inline, then annotate it with a high level comment.

Then, I achieve the same effect without having a new developer jump back and forth, or reconcile how the function is being used throughout the codebase (i.e. how many times is it called, is it exported, what does it do, does it modify global vars, etc)

A general point inspired by this minor and inconsequential instance (this overhead is more pronounced across multiple files). Of course this will work fine, but when done often can be quite cumbersome to new devs.

packages/web/modals/validator-squad.tsx
```diff
@@ -62,19 +61,15 @@ export const ValidatorSquadModal: FunctionComponent<ValidatorSquadModalProps> =
...
 
     const { logEvent } = useAmplitudeAnalytics();
 
     const defaultUserValidatorsSet = new Set(usersValidatorsMap.keys());
 
+    // TODO clean up the default validators
+    // @ts-ignore
     const [selectedValidators, setSelectedValidators] = useState(
       defaultUserValidatorsSet
     );
```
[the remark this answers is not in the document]
jonator: Any ideas how we can store state in the table rows, so we don't have to render the whole table? One issue with this that comes to mind is then collecting that state when the user clicks next on the table, once their set is selected.

packages/web/modals/validator-squad.tsx
jonator: OK, this is fine in isolation, just a general philosophical point 👍

jonator (commented):
I'm not seeing any issues with perf in the preview page. The validator images aren't loading, though.

Also, thoughts on making the whole row clickable and not just the checkbox? Also, a little hover animation is nice. See the convert to stake modal to see what I mean.

jonator (approved):
A follow up would be good for the table UI polishing, otherwise looks good.

PR #2129, opened 2023-09-11
fix: Swap Memory Leak

JoseRFelix:
### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0rkhne)

## Brief Changelog

- Remove unnecessary pool fetch on `use-token-swap-query-params.ts`

## Testing and Verifying

- [ ] Swap router shouldn't bloat the heap

### DIFF
--- a/packages/stores/src/ui-config/trade-token-in-config.ts
+++ b/packages/stores/src/ui-config/trade-token-in-config.ts
@@ -88,9 +88,8 @@ export class ObservableTradeTokenInConfig extends AmountConfig {
 
   @override
   get sendCurrency(): AppCurrency {
+    // Return the initial send currency when pools are not fetched yet.
     if (this.sendableCurrencies.length === 0) {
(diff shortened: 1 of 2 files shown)


packages/stores/src/account/types.ts
```diff
@@ -61,4 +62,15 @@ export type RegistryWallet = Wallet & {
    * the notifications button.
    */
   features: Array<"notifications">;
+
+  signOptions?: SignOptions;
 };
+
+export type AccountStoreWallet<Injects extends Record<string, any>[] = []> =
```
jonator: Moving some stuff around here?

jonator (approved):
Great work

packages/stores/src/account/types.ts
JoseRFelix: We can disregard this. These were changes from this PR to `master`

PR #2140, opened 2023-09-13
add call to hasvalsetPref

MaxMillington:

Adding a call to check if the user hasValsetPref, which is diff than existing delegations

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0twmkd)

## Brief Changelog

- adds query for hasValsetPref to index page

## Testing and Verifying

### DIFF
--- a/packages/web/pages/stake/index.tsx
+++ b/packages/web/pages/stake/index.tsx
@@ -44,10 +44,17 @@ export const Staking: React.FC = observer(() => {
   const account = accountStore.getWallet(osmosisChainId);
   const address = account?.address ?? "";
   const queries = queriesStore.get(osmosisChainId);
+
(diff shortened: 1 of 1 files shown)


packages/web/pages/stake/index.tsx
```diff
@@ -204,7 +211,7 @@ export const Staking: React.FC = observer(() => {
     squadSize,
   ]);
 
-  const isNewUser = usersValidatorsMap.size === 0;
+  const isNewUser = !userHasValPrefs && usersValidatorsMap.size === 0;
```
MaxMillington, line 214: including both here, as there are cases when `userHasValPref` is false but the user has delegated validators

PR #2155, opened 2023-09-18
Add server pools query & transmuter pool

jonator:
### Server pools query
This PR moves us in the direction of the serverful architecture; sending less queries and processing less data on the client. This was necessary in this PR since transmuter pools contained nearly no useful routing data; the tokens within the pool. I added 2 api routes pool/[id] and /pools that can be used to fetch processed pools on the server. Further, we used to have fallback logic on the frontend where if the imperator stream query was not available, the node pools query would be used. Now, that logic is on the server; if the stream query fails, the node pools query is used to fetch pools. Also, for some pools, like CL and CosmWasm, the balance of tokens in that pool are also fetched and added to the response model.

### Transmuter pool
This PR adds the transmuter pool as a new pool type; it allows 1:1 zero slippage and zero fee swaps of two correlated assets. The server pools query also queries it's balances and appends them to the response model. These are then used in the routable pool on the client to allow swaps to be simulated through it. Further, additional metrics, such as volume, can be added later as that data becomes available in the pools stream query. Client side config is used to remark which CosmWasm pool is a transmuter pool vs some other custom pool via code ID. If a CosmWasm pool does not contain that code ID it is ignored. In the future, I'm looking to somehow query for wasm files that can be used to simulate arbitrary pool implementations on the client (or in the future on the server). That work is not included, only transmuter pools are supported. Also, a future PR is needed to include the Transmuter pools as preferred pools in the router.

### Edge function caching
With @JoseRFelix 's recommendation, I use `cachified` package, along with the `lru-cache` package to cache pools queries on each edge function for no more than 30 seconds. Each edge function distributed across zones maintains their own cache to handle traffic.

### @osmosis-labs/web Queries folder
I also added a queries folder that somewhat matches the structure of the queries folders in our stores package, except just for fetch calls for use in our edge functions. For node queries, it's nice to match the module architecture we have on the Osmosis node. Also, there are folders for external queries (like to imperator stream query) as well as a `complex` folder where queries with additional processing can be organized.

### Tests
Since I removed the query pools store for the node, our e2e tests will no longer work. We may need to add it back for tests to work, despite cosmwasm pools not being supported.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0prr7e)

## Brief Changelog

* Replaced node pools and pool query as well as imperator filtered pools query with a single pools query mentioned above
* Added Transmuter pool to pools package
* Updated existing pool models to include balances and metrics (metrics are not used now, but can be used in the future to prevent additional querying and thus increase performance)

## Testing and Verifying

- [ ] Pools page loads with all pools
- [ ] Refreshing reveals caching is working
- [ ] Individual pool pages working
- [ ] Can swap through transmuter pools. E.g. Atom <> Noble USDC


JoseRFelix:
@jonator Limit is set to 200 on the endpoint, but it's responding with 1170 pools

image
image

jonator:
@JoseRFelix fixed

JoseRFelix:
@jonator I've observed that the new `pools` query is quite slow, especially during a cold start.

image

Additionally, every time I access the token selection drawer, the content vanishes and then reappears. Could this be tied to how we're caching the pools query?

packages/web/queries/complex/pools.ts
```diff
@@ -0,0 +1,325 @@
...
+
+        // fall back to pools query on node
+        return await getPoolsFromNode();
+      }
+    },
+    ttl: 30 * 1000, // 30 seconds
+  });
+}
+
+export async function queryPoolRawFromFilteredPool(
```
JoseRFelix, line 113: We can drop the `async` keyword from the function because there's no `await` used inside. Even with `Promise.all`, the operations will be processed in sequence, not concurrently, due to the absence of `await`. If our goal is true parallel processing, we'd have to leverage the [worker_threads](https://nodejs.org/api/worker_threads.html) module in Node.js. Node.js is inherently single-threaded, processing one task at a time.

packages/web/queries/complex/pools.ts
```diff
@@ -0,0 +1,325 @@
...
+            order_key: "liquidity",
+          },
+          { offset: 0, limit: Number(numPools.num_pools) }
+        );
+        const queryPoolRawPromises = filteredPoolsResponse.pools.map(
+          queryPoolRawFromFilteredPool
+        );
+        const queryPoolRawResults = await Promise.all(queryPoolRawPromises);
+
+        const cosmwasmPools = await getCosmwasmPools();
```
JoseRFelix, line 89: Since we don't have any dependencies coming from `queryFilteredPools`, we can place `getCosmwasmPools` and `queryFilteredPools` in a promise.all to execute them concurrently

packages/web/queries/complex/pools.ts
```diff
@@ -0,0 +1,325 @@
...
+  const sharePoolBase = {
+    "@type": `/${filteredPool.type}`,
+    id: filteredPool.pool_id.toString(),
+    pool_params: {
+      exit_fee: new Dec(filteredPool.exit_fees.toString())
+        .mul(DecUtils.getTenExponentN(-2))
+        .toString(),
+      swap_fee: new Dec(filteredPool.swap_fees.toString())
+        .mul(DecUtils.getTenExponentN(-2))
+        .toString(),
```
JoseRFelix, line 182: Do we need to use @keplr/units for these functions? Btw, how much does Keplr bloat the bundle size of the functions?

packages/web/queries/complex/pools.ts
```diff
@@ -0,0 +1,325 @@
...
+
+      return cosmwasmPools.map((pool, index) => {
+        return {
+          ...pool,
+          tokens: poolBalances[index].balances.map((balance) => ({
+            denom: balance.denom,
+            amount: balance.amount,
+          })),
+        };
+      }) as CosmwasmPoolRaw[];
```
JoseRFelix, line 267: Could we split this logic? For the paginated pools query, can we utilize the result from the Imperator query? If I recall correctly, that result also had the cosmwasm pools.

packages/web/queries/complex/pools.ts
```diff
@@ -0,0 +1,325 @@
...
+    cache: cosmwasmPoolsCache,
+    async getFreshValue() {
+      const { pools } = await queryPools();
+      const cosmwasmPools = pools.filter(
+        (pool) => pool["@type"] === "/osmosis.cosmwasmpool.v1beta1.CosmWasmPool"
+      ) as CosmwasmPoolRaw[];
+
+      const poolBalancesPromises = cosmwasmPools.map(
+        async (pool) => await queryBalances(pool.contract_address)
+      );
```
JoseRFelix, line 256: Do you believe it's possible to cache the balances?

jonator: I cache them within this cachified call, along with querying pools

jonator:
@JoseRFelix can we try to review this PR, it may contain some of the pool query fixes you mention here. It seems more reliable

https://github.com/osmosis-labs/osmosis-frontend/pull/2157

jonator:
> @jonator I've observed that the new `pools` query is quite slow, especially during a cold start.
> 
> image
> 
> Additionally, every time I access the token selection drawer, the content vanishes and then reappears. Could this be tied to how we're caching the pools query?

Hopefully in prod only the first user will experience this?

packages/web/queries/complex/pools.ts
JoseRFelix: Yes, because it uses `await`, we can execute this concurrently

packages/web/queries/complex/pools.ts
jonator: I can remove those. But how high prio are concerns for bigger bundles on server? I imagine it's low impact since it's not going over wire

jonator:
Yeah initial load in previews is very slow

packages/pools/src/cosmwasm/transmuter.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+  getSpotPriceOutOverInWithoutSwapFee(
+    tokenInDenom: string,
+    tokenOutDenom: string
+  ): Dec {
+    validateDenoms(this, tokenInDenom, tokenOutDenom);
+    return new Dec(1);
+  }
+
+  // Interface: RoutablePool
```
JoseRFelix, line 70: Is this comment needed?

packages/pools/src/cosmwasm/transmuter.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  getLimitAmountByTokenIn(denom: string): Int {
+    const amount = this.raw.tokens.find(
+      ({ denom: tokenDenom }) => tokenDenom === denom
+    )?.amount;
+    if (!amount) throw new Error("Invalid denom");
+    return new Int(amount);
+  }
+
+  // transmute() https://github.com/osmosis-labs/transmuter/blob/c14cfa7a6adff5336a94c3caee7f94544757cd92/contracts/transmuter/src/transmuter_pool/transmute.rs#L8
```
JoseRFelix, line 80: Same as above

JoseRFelix (commented):
Left some final comments. After these, we can merge

JoseRFelix (approved):
LGTM, great work! 🚀

CryptoAssassin1:
Working as expected.

PR #2157, opened 2023-09-19
Add transmuter pool to router

jonator:

Make sure transmuter pool is included as a routable pool.
Screenshot 2023-09-18 at 10 22 14 PM

Screenshot 2023-09-18 at 10 58 50 PM

### DIFF
--- a/packages/pools/src/concentrated/pool.ts
+++ b/packages/pools/src/concentrated/pool.ts
@@ -391,7 +391,7 @@ export class ConcentratedLiquidityPool implements BasePool, RoutablePool {
     };
   }
 
-  async getLimitAmountByTokenIn(denom: string): Promise<Int> {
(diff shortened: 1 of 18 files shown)


JoseRFelix:
@jonator I've observed that the new `pools` query is quite slow, especially during a cold start.

image

Additionally, every time I access the token selection drawer, the content vanishes and then reappears. Could this be tied to how we're caching the pools query?

JoseRFelix:
@jonator The Transmuter pools aren't visible on the pools page. I assume they'll be added in a future PR?

jonator:
> @jonator The Transmuter pools aren't visible on the pools page. I assume they'll be added in a future PR?

Yeah they're not, we can add them if it's asked for. The detail page works though, and works for all types of pool if there's not a pool page for that pool type. It uses the base interface

PR #2171, opened 2023-09-20
Line for disabling routes through blocked pools

ValarDragon:
### ClickUp Task


packages/pools/src/router/routes.ts
```diff
@@ -92,6 +92,7 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
     maxSplit = 2,
     maxSplitIterations = 10,
   }: OptimizedRoutesParams) {
+    const blocked_pool_ids = ["1078", "1079"];
```
sunnya97, line 95:
```suggestion
    const blocked_pool_ids = [];
```

PR #2176, opened 2023-09-21
Replace libsodium with a no-op functionality

ValarDragon:

This performance boost is copied from Keplr!

Copied from my code comment:
```
    // Replace libsodium with a no-op API. It is only imported from within cosmJS to support
    // argon2i and ed25519, both functionalities which in the context of Cosmos would only get used within
    // an extension wallet. Libsodium is ~190kb gzipped, 500kb parsed, so this meaningfully reduces client load.
    // (And it gets bundled twice)
    //
    // It should never be getting used. This is copied from what Keplr does:
    // https://github.com/chainapsis/keplr-wallet/blob/master/package.json#L103-L104
```

Thank you @JoseRFelix for the help with understanding how we use webpack vs turbopack, and @MaxMillington for discussions leading to this!

### ClickUp Task

N/A

## Brief Changelog

Lower bundle size by marking 

## Testing and Verifying

I have tested that yarn analyze no longer shows the copies of libsodium in the client bundle, saving 190kb gzipped, 500kb parsed from the main bundle, and the same savings in a second bundle.

Have not tested that all FE functionality generally works. (My understanding is QA does this?)

- Does this pull request introduce a new feature or user-facing behavior changes? no
- How is the feature or change documented?  Code comment

### DIFF
--- a/packages/web/next.config.js
+++ b/packages/web/next.config.js
@@ -47,6 +47,13 @@ const config = {
           "noop",
           "index.js"
         ),
+        "libsodium-sumo": path.resolve(__dirname, "etc", "noop", "index.js"),
+        "libsodium-wrappers-sumo": path.resolve(
+          __dirname,
+          "etc",
+          "noop",
+          "index.js"
+        ),
       },
     };
 
--- a/packages/web/next.config.js
+++ b/packages/web/next.config.js
@@ -1,3 +1,5 @@
+const path = require("path");
+
 /** @type {import('next').NextConfig} */
 const config = {
   reactStrictMode: true,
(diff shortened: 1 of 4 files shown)


packages/web/next.config.js
```diff
@@ -27,6 +29,27 @@ const config = {
...
+      ...config.resolve, // This spreads existing resolve configuration (if any)
+      alias: {
+        ...config.resolve.alias, // This spreads any existing alias configurations
+        libsodium: path.resolve(__dirname, "etc", "noop", "index.js"),
+        "libsodium-wrappers": path.resolve(
+          __dirname,
+          "etc",
+          "noop",
+          "index.js"
+        ),
```
webmaster128, line 49: For more recent versions of CosmJS you need to patch out `libsodium-wrappers-sumo`. So ideally you just do the same thing for all four:

```
libsodium
libsodium-wrappers
libsodium-sumo
libsodium-wrappers-sumo
```

ValarDragon: Oh awesome, thank you!

MaxMillington (approved):
👍

JoseRFelix:
Thank you @ValarDragon! 🚀  I've sent this to the QA team. Let's hold off on merging until they've finished reviewing it.

CryptoAssassin1:
Working fine.

JoseRFelix (approved):

PR #2181, opened 2023-09-22
fix: Trade Pair Pools Page

JoseRFelix:

This PR fixes the infinite loop when opening trade pair in the individual pools page


sunnya97:
utack

sunnya97 (approved):

PR #2183, opened 2023-09-22
Replace bip39 with bip39-light

ValarDragon:

```
        // bip39 is only used in the context of extension wallet, so we can replace it.
        // replacing it with a no-op breaks build, so we can at least replace it with a lighter weight version for now.
        // ideally this becomes replaced with an API-compatible no-op.
```

Saves 70kb gzipped, 200kb parsed size.

## Testing and Verifying

We should just sanity check the FE works. I don't know of any reason why bip39 would get loaded. Its worth sanity checking that other languages don't break when you select them.

### DIFF
--- a/packages/web/next.config.js
+++ b/packages/web/next.config.js
@@ -54,6 +54,10 @@ const config = {
           "noop",
           "index.js"
         ),
+        // bip39 is only used in the context of the extension wallet, so we can replace it.
+        // replacing it with a no-op breaks build, so we can at least replace it with a lighter weight version for now.
+        // ideally this becomes replaced with an API-compatible no-op.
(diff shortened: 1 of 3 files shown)


ValarDragon:
cc @CryptoAssassin1 for testing!

CryptoAssassin1:
:white_check_mark: All the languages
:white_check_mark: All the wallets
:white_check_mark: Enable notifications
:white_check_mark: Deposit/withdraw
:white_check_mark: Collect rewards
:white_check_mark: Create position
:white_check_mark: Swap
:white_check_mark: All the pages are loading fine

ValarDragon:
Thank you!!

JoseRFelix (approved):

PR #2184, opened 2023-09-22
fix: transmuter router issues

p0mvn:

Transmuter pools provide 1:1 no slippage swaps. As a result, these pools should be preferred in the router.

If router detects a direct candidate route where there is a single pool AND token in matches AND token out matches, use that pool only without the split.

### ClickUp Task


packages/pools/src/router/routes.ts
```diff
@@ -150,6 +150,50 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
 
     let routes = this.getCandidateRoutes(tokenIn.denom, tokenOutDenom);
 
+    for (let i = 0; i < routes.length; i++) {
```
MaxMillington, line 153: `routes.forEach((route) => { })`

packages/pools/src/router/routes.ts
```diff
@@ -150,6 +150,50 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
...
+        }
+
+        if (tokenOutDenom != denomA && tokenOutDenom != denomB) {
+          continue;
+        }
+
+        // Transmuter pools expect a 1:1 swap with no slippage.
+        // If transmuter pool is encountered that has token in and token out
+        // return it as a single pool swap.
+        const cwPoolIDs = ["1175", "1176", "1211", "1212"];
```
MaxMillington, line 177: Can we declare these pools as constants at the top of the page? What does `cw` stand for?

packages/pools/src/router/routes.ts
MaxMillington: actually it looks like maybe you want a `map` here? `map` returns a new array

packages/pools/src/router/routes.ts
```diff
@@ -282,6 +326,8 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
         const outDenom = route.tokenOutDenoms[i];
 
         let poolSwapFee = pool.swapFee;
+        console.log("pool", pool.id.toString());
```
MaxMillington: remove logs

packages/pools/src/router/routes.ts
```diff
@@ -371,6 +417,8 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
           .sub(new Dec(1))
       : new Dec(0);
 
+    console.log("totalSwapFee", totalSwapFee.toString());
```
MaxMillington: remove logs

packages/pools/src/router/routes.ts
p0mvn: I want to return from the caller function if I successfully find the desired route.

If I use for each, I would be returning from the callback, right?

packages/pools/src/router/routes.ts
p0mvn: Addressed

packages/pools/src/router/routes.ts
p0mvn: Done. Cw stands for cosmwasm. Changed to transmuter for clarity.

Transmuter is a special case of a cosmwasm pool

packages/pools/src/router/routes.ts
```diff
@@ -150,6 +155,55 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
...
+
+        const directOutAmount = (
+          await this.calculateTokenOutByTokenIn([
+            { ...curRoute, initialAmount: tokenIn.amount },
+          ])
+        ).amount;
+
+        return [{ ...curRoute, initialAmount: directOutAmount }];
+      }
+    }
```
JoseRFelix, line 205: Is there any reason why we don't check this directly on `this.getCandidateRoutes`?

p0mvn: No reason. I will try now and see if it's a better locaton

p0mvn: `getCandidateRoutes` is not  async.

I think it is better to keep this as is. The hotfix searches for transmuter pool in routes, if it finds one, it does an async calculation to determine the result and returns that as a substitute to the original split route right away.

packages/pools/src/router/routes.ts
p0mvn: I was unable to get this working because I'm looking to return from `getOptimizedRoutesByTokenIn` (the caller) if certain criteria are matches.

With `map` or `forEach`, I would be returning from the callback.

Additionally, with `map`, it would make it return an array of a different type - `[]RouteWithInAmount`. routes is `[]Route` type. This change would make further handling messier. I think it's best to simply return in a for loop.

Please let me know if that makes sense

packages/pools/src/router/routes.ts
p0mvn: I like this. Thank you.

Applied:
[4345984](https://github.com/osmosis-labs/osmosis-frontend/pull/2184/commits/4345984e0f8dd8599db0c129c601875396d81cad)

JoseRFelix (approved):

PR #2185, opened 2023-09-22
fix: Individual Pools Query

JoseRFelix:

Individual get pool by id is failing because the API route is trying to get the id from search parameters instead of the pathname.

### DIFF
--- a/packages/web/pages/api/pool/[id].ts
+++ b/packages/web/pages/api/pool/[id].ts
@@ -6,8 +6,8 @@ type Response = {
 
 export default async function pools(req: Request) {
   const url = new URL(req.url);
-  const poolIdParam = url.searchParams.has("id")
-    ? url.searchParams.get("id")
+  const poolIdParam = url.pathname.split("/").slice(-1)
+    ? url.pathname.split("/").slice(-1)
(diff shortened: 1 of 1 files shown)


MaxMillington (approved):

packages/web/pages/api/pool/[id].ts
```diff
@@ -6,8 +6,8 @@ type Response = {
 
 export default async function pools(req: Request) {
   const url = new URL(req.url);
-  const poolIdParam = url.searchParams.has("id")
-    ? url.searchParams.get("id")
+  const poolIdParam = url.pathname.split("/").slice(-1)
+    ? url.pathname.split("/").slice(-1)
     : undefined;
```
Comment (p0mvn) [inline packages/web/pages/api/pool/[id].ts:11]: q: why does it have to be a conditional and not initialized directly?

p0mvn (approved):
utACK

packages/web/pages/api/pool/[id].ts
Comment (JoseRFelix) [reply packages/web/pages/api/pool/[id].ts]: Indeed, that piece of code was redundant. I have uploaded an updated version with enhanced validation behavior.

PR #2191, opened 2023-09-25
feat: Add Market Cap Column to New Assets Table

JoseRFelix:

Add Market Cap column to new assets table

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0xuq45)

## Testing and Verifying

- [ ] Should display market cap value in its column 
- [ ] Should display "-" when market cap is not available

image


jonator (approved):
Heads up the token page PR may also contain a market cap query. We should reconcile them

PR #2193, opened 2023-09-26
Bring Keplr Store & Hooks into Osmosis

ValarDragon:
Huge thank you to @delivan @Thunnini for work going into this

- Copy hooks from node modules into this repo
- Copy store from node modules into this repo
- Change tsconfig to ignore a bunch of things (TODO: Move this to per-package)
  - We are getting mystery ts-config issues. We should aim to get these keplr-packages brought into local packages entirely with time, to mitigate these issues.
- Change our imports for this

Subsequent PR's should be deleting significant amounts of this code & hopefully dropping some of the bundle size.


ValarDragon:
Main bundle size before this PR:
* Stat: 20.09MB
* Parsed: 8.84MB
* Gzip: 1.95MB

After this PR:
* Stat: 19.95MB
* Parsed: 8.82MB
* Gzip: 1.95MB

So this shows that its correctly not getting imported again

ValarDragon:
@CryptoAssassin1 can you please help with testing this branch?

It has a fairly wide surface area touched unfortunately. If theres a way to look out for caching issues, that would be great

I tried a swap and it worked, which makes me think things _should_ work?

CryptoAssassin1:
@ValarDragon sure im sharing it with QA team when we are done will comment here.

CryptoAssassin1:
:white_check_mark: Swaps
:white_check_mark: Deposit/withdraw
:white_check_mark: Swaps inside pools Balancer/CL pools
:white_check_mark: Collect Rewards
:white_check_mark: Create position
:white_check_mark: Remove position

@ValarDragon and all other features with different wallets worked well.

ValarDragon:
Awesome thank you both!

PR #2203, opened 2023-09-27
Remove keplr-wallet/crypto and therefore crypto-JS import

ValarDragon:
This PR removes the import keplr-wallet/crypto, and by proxy crypto-js.

I anticipate this saving 50kb of parsed size, and 10kb off of the gzip. I cannot validate this right now, due to seemingly unrelated build issues.


jonator (approved):

ValarDragon:
This PR succeeds at removing our direct import of @keplr-wallet/crypto, but its unfortunately still transitively brought in. Hopefully that gets unwound with more of the forking.

PR #2210, opened 2023-09-28
[Wallet Addition] Add Station Wallet

JoseRFelix:


sunnya97:
What's needed here to get this over the line?

CryptoAssassin1 (approved):
Working fine.

alecande11:
> When Max Deposit amount is selected, gas is not correctly calculated and the tx fails after consuming gas fees.
@kamal-sutra 
I've been debugging on this issue a little bit, and I'm pretty sure the issue is that we are calculating the tx fee in different ways.

## Example
I have 2.353451 STRD on my wallet, when I click on the "max" button you set a lower send amount (2.353031 STRD) to account for the fee that you have calculated (0.000420 STRD).
Screenshot 2024-01-11 at 11 38 40
So far everything is fine.

Later when I click on the "Deposit STRD" button you send this tx request to our extension.  
Screenshot 2024-01-11 at 11 43 36
As you can see you set `107024` as the gas amount.

Our extension, before signing calculates the tx fee (our gas price on stride is 0.005ustrd):
```
fee = 107024 * 0.005ustrd ≈ 535ustrd ≈ 0.000535 STRD
```
which ends up being higher than the fee you previously calculated, resulting in a failed tx for insufficient balance.

### Solution
Those imo are some of the possible causes for the issue:
- we are using different gas price sources (but that's unlikely since this is happening on pretty much all chains)
- the gas amount you used to calculate the max amount, is different from the gas amount you set on the tx request to our extension (maybe the one you set on tx request is including a gas adjustment?)

So I don't think this something we can fix on our extension, but rather on your max amount calculation or on the way you calculate gas for the tx request, but if during debugging you find out it's somehow something on our side, just lmk and I'll happy to assist!

CryptoAssassin1:
@kamal-sutra its same on the main app too when deposit Assets need to leave something for the gas fee.

kamal-sutra:
> > When Max Deposit amount is selected, gas is not correctly calculated and the tx fails after consuming gas fees.
> 
> @kamal-sutra I've been debugging on this issue a little bit, and I'm pretty sure the issue is that we are calculating the tx fee in different ways.
> 
> ## Example
> I have 2.353451 STRD on my wallet, when I click on the "max" button you set a lower send amount (2.353031 STRD) to account for the fee that you have calculated (0.000420 STRD). Screenshot 2024-01-11 at 11 38 40 So far everything is fine.
> 
> Later when I click on the "Deposit STRD" button you send this tx request to our extension. Screenshot 2024-01-11 at 11 43 36 As you can see you set `107024` as the gas amount.
> 
> Our extension, before signing calculates the tx fee (our gas price on stride is 0.005ustrd):
> 
> ```
> fee = 107024 * 0.005ustrd ≈ 535ustrd ≈ 0.000535 STRD
> ```
> 
> which ends up being higher than the fee you previously calculated, resulting in a failed tx for insufficient balance.
> 
> ### Solution
> Those imo are some of the possible causes for the issue:
> 
> * we are using different gas price sources (but that's unlikely since this is happening on pretty much all chains)
> * the gas amount you used to calculate the max amount, is different from the gas amount you set on the tx request to our extension (maybe the one you set on tx request is including a gas adjustment?)
> 
> So I don't think this something we can fix on our extension, but rather on your max amount calculation or on the way you calculate gas for the tx request, but if during debugging you find out it's somehow something on our side, just lmk and I'll happy to assist!

Thank you for clarification. IBC transfers working fine now

kamal-sutra:
@alecande11 small bug- USDC (noble) deposit is taking usdc.axl as fee and failing the tx. Please check.

image
image
image

kamal-sutra:
All other swaps, ibc transfers, CL/Classic pool txs are working well. Above bug is not a blocker (should be easy fix).
Good to merge.

CryptoAssassin1:
@kamal-sutra tx is not failing can you please check again

PR #2212, opened 2023-09-28
CL APR: pools table

jonator:

Adds APR to pools table. It is the APR for a full range position. Notice the APR for pools with stablecoins is low since so much liquidity is concentrated at $1.

Screenshot 2023-09-28 at 11 02 09 AM

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0yfc5a)


JoseRFelix:
@jonator Do you think it would make sense that for stables we set the range closer to 1 so it's more accurate?

jonator:
> @jonator Do you think it would make sense that for stables we set the range closer to 1 so it's more accurate?

What range do you propose? IMO it's nice that it illustrates the potentially low APR even if you concentrate since there's no impermanent loss

JoseRFelix (approved):
Nvm @jonator, having two range criteria would make it confusing. Maybe we could add this in a future version as needed

PR #2217, opened 2023-09-29
Add try catch to response

jonator:

If the APR query fails, it may return an unexpected response. I wrapped the entire APR reference in a try catch to handle all unexpected responses.

### ClickUp Task


arispech (approved):
The page does not crash anymore when moving the ticks

PR #2229, opened 2023-10-03
Router: Increase necessary TVL of pools

jonator:

Seems the limit for routable pools' TVL was lowered, causing low liq pools to be chosen. This increases the limit and therefore increases the liquidity of chosen pools for routing.

### ClickUp Task

## Testing and Verifying

USDC.axl > ATOM route should now be fixed and should be lower price impact.


CryptoAssassin1:
@jonator pool page and swap page is not loading.
Screenshot from 2023-10-03 21-56-24

PR #2232, opened 2023-10-04
Upgrading Next from 12 to 13

DavideSegullo:

The goal is to upgrade next to the latest release in order to enable the new features and also take advantage of the new benefits for DX (as they have improved the dev server as well).

In addition, by updating next, we also have the new version of React, which allows us to access the new features as well as being able to use the new library releases.

## Brief Changelog

- Upgrade next from v12 to v13
- Upgrade React from v16 to v18
- Re-rendering fixes

## Testing and Verifying

This change has been tested locally by rebuilding the website and verified content and links are expected.

Definitely more testing needs to be done, to make sure there are no bugs around, especially related to re-rendering.


JoseRFelix:
@DavideSegullo If this is still WIP, let's change this PR to a draft.

DavideSegullo:
@JoseRFelix  I did other tests and it seems to works fine, can you review it? Thank you! Maybe we can also try to deploy a test version, to also check that vercel ci/cd ecc works fine

MaxMillington:
Will need to have a test deploy for QA. We tried to do this a couple of months ago but found many bugs caused by the upgrade. We abandoned it because we got busy, but we definitely need this upgrade.

@DavideSegullo

DavideSegullo:
> Will need to have a test deploy for QA. We tried to do this a couple of months ago but found many bugs caused by the upgrade. We abandoned it because we got busy, but we definitely need this upgrade.
> 
> @DavideSegullo 

I can help with that, we can try to deploy and if there is any errors, I'll try to fix it!

packages/web/pages/api/pools.ts
```diff
@@ -37,6 +37,6 @@ export default async function pools(req: Request) {
 }
 
 export const config = {
-  runtime: "experimental-edge",
+  runtime: "edge",
```
MaxMillington, line 40: what is this change?

packages/web/components/complex/paginated-table.tsx
```diff
@@ -79,7 +79,10 @@ export const PaginatedTable = ({
           return (
             <Link
               key={row.original.queryPool.id}
+              prefetch={false}
```
MaxMillington, line 82: why is prefetch false? Wouldn't it be better to allow next to prefetch in the background? Or was the pre-fetch causing perf issues?

packages/web/components/cards/pool-card.tsx
```diff
@@ -32,51 +32,52 @@ export const PoolCard: FunctionComponent<
...
-              <div className="subtitle1 flex items-center gap-1 text-white-mid">
-                {t("pools.poolId", { id: poolId })}
-                {isSupercharged && (
-                  <Icon id="lightning-small" height={16} width={16} />
-                )}
-              </div>
+    <Link
+      href={`/pool/${poolId}`}
+      passHref
+      prefetch={false}
```
MaxMillington, line 38: Same question. Why is prefetch false?

MaxMillington:
> > Will need to have a test deploy for QA. We tried to do this a couple of months ago but found many bugs caused by the upgrade. We abandoned it because we got busy, but we definitely need this upgrade.
> > @DavideSegullo
> 
> I can help with that, we can try to deploy and if there is any errors, I'll try to fix it!

Thanks!

packages/web/pages/api/pools.ts
DavideSegullo: They changed it, it's was an experimental feature, but now it's stable. In fact, I inserted it because a warning message appeared suggesting replacing it

packages/web/components/complex/paginated-table.tsx
DavideSegullo: No performance regression in particular, I entered it while I was fixing a re-rendering bug on the pools page, we can reset it to true. Even considering when pools appear with pagination, that's an acceptable amount of prefetches

packages/web/components/cards/pool-card.tsx
DavideSegullo: No performance regression in particular, I entered it while I was fixing a re-rendering bug on the pools page, we can reset it to true. Even considering when pools appear with pagination, that's an acceptable amount of prefetches

DavideSegullo: Same reason as above

packages/web/components/complex/paginated-table.tsx
DavideSegullo: I can change this, if you agree

DavideSegullo:
> > > Will need to have a test deploy for QA. We tried to do this a couple of months ago but found many bugs caused by the upgrade. We abandoned it because we got busy, but we definitely need this upgrade.
> > > @DavideSegullo
> > 
> > 
> > I can help with that, we can try to deploy and if there is any errors, I'll try to fix it!
> 
> Thanks! FYI I'll authorize the vercel deploy to preview

Your deploy is working, did you find any issues?

I tried also the build, i think I'm missing some parameters (Sentry, maybe something for the swap tool, actually it doesn't work in dev mod on "stage" branch too),pages etc. are fine you can check here: https://osmosis-frontend-fix-web.vercel.app

JoseRFelix:
@DavideSegullo I did some quick QA and found an issue. When trying to change the language, the change is delayed and the user has to refresh the page to see the change.

DavideSegullo:
> @DavideSegullo I did some quick QA and found an issue. When trying to change the language, the change is delayed and the user has to refresh the page.

Thank you, I'll check that

DavideSegullo:
> > @DavideSegullo I did some quick QA and found an issue. When trying to change the language, the change is delayed and the user has to refresh the page.
> 
> Thank you, I'll check that

The solution was to replace "react-multi-lang" with a homemade, react 18-compatible solution; the library used some old workarounds that are no longer usable.

I also took the opportunity to improve it, you will also find an autocomplete of keys when using the "t" function.

This modification allows you to continue working as you did before, but I would suggest migrating to a more full-bodied library that also supports i18n for next (so you can use sub-paths).

This way we also have one less dependency, and we should also recover some KB, the hook weighs much less.

packages/web/components/complex/paginated-table.tsx
MaxMillington: I think it makes sense to do `prefetch={true}`

MaxMillington:
@DavideSegullo does this mean we can remove `react-multi-lang` from the `package.json`?

packages/web/hooks/language/context.tsx
```diff
@@ -0,0 +1,73 @@
...
+  defaultLanguage?: string;
+  defaultTranslations?: Translations;
+}
+
+const DEFAULT_LANGUAGE = "en";
+
+let globalTranslations: Translations = { en };
+let globalLanguage: string = DEFAULT_LANGUAGE;
+
+export const t: MultiLanguageT = (path, args) =>
```
MaxMillington, line 29: why is the `t` function declared twice?

DavideSegullo:
> @DavideSegullo does this mean we can remove `react-multi-lang` from the `package.json`?
> 
> We should probably use something like `react-intl` I would think? Wouldn't it make sense to just do that instead of custom rolling the `useMultiLanguage` hook?

Exactly, I didn't use a new library, because we can integrate that in another branch, so we can also integrate sub-path using one of the library suggest by next: https://nextjs.org/docs/pages/building-your-application/routing/internationalization.

In addition, sub-paths will allow us to index the site in multiple languages and perhaps eliminate the need to have a key in local storage (We would only manage it by uri and not with the state manager)

Ye I removed 'react-multi-lang' here: https://github.com/osmosis-labs/osmosis-frontend/pull/2232/commits/cc8467462340a47bef7f9315bdd8595f7845dc04

packages/web/hooks/language/context.tsx
DavideSegullo: Because there are some parts of the application, which access "t" without being React components, for example, the function that creates a "toast" does not have direct access to the react context

JoseRFelix:
@DavideSegullo Was there a specific rationale for renaming `useTranslation` to `useMultiLanguage`? I felt the original name conveyed its purpose well. Perhaps we could consider rolling it back minimizing changes to this PR?

DavideSegullo:
> @DavideSegullo Was there a specific rationale for renaming `useTranslation` to `useMultiLanguage`? I felt the original name conveyed its purpose well. Perhaps we could consider rolling it back minimizing changes to this PR?

Yes, because "useTranslation" used to return only the "t" function, and there were several separate methods to access the information, now "useMultiLanguage" returns all the state with a single hook, so it is no longer strictly related to "t".

Do you want to bring back the modification of the new hook? And then migrate to a new language system with another pr?

I had thought of doing a separate one, but if we merge without this change there are bugs on the multilingual, if you prefer I can create another branch from this one to fix.

How would you prefer to proceed?

packages/web/components/complex/paginated-table.tsx
DavideSegullo: I'll add it

packages/web/components/cards/pool-card.tsx
DavideSegullo: I enabled prefetch here

packages/web/components/complex/paginated-table.tsx
DavideSegullo: I found out, that "prefetch" is also disabled on "stage" branch: https://github.com/osmosis-labs/osmosis-frontend/blob/stage/packages/web/components/cards/pool-card.tsx#L35

JoseRFelix:
Although `useMultiLanguage` offers more than just the `t` method, I still believe `useTranslation` is a more descriptive name. My thought process is that `useTranslation` suggests it encompasses all aspects required to translate a string, including its state. On the other hand, `useMultiLanguage` gives me the impression that it only includes the available languages or the current selected language.

DavideSegullo:
> Although `useMultiLanguage` offers more than just the `t` method, I still believe `useTranslation` is a more descriptive name. My thought process is that `useTranslation` suggests it encompasses all aspects required to translate a string, including its state. On the other hand, `useMultiLanguage` gives me the impression that it only includes the available languages or the current selected language.

Ye, It seems legit, I'll rename it to make it more descriptive, maybe "useMultiLanguage" can be a bit confusing!

DavideSegullo:
Closed in favor of https://github.com/osmosis-labs/osmosis-frontend/pull/2237

PR #2242, opened 2023-10-12
Jonator/router fixes

jonator:

Will attempt routes and splits and if they fail it will try other routes. If all fail it's reported as there being no routes.

Also adds logger that can be used to debug router quicker.

### ClickUp Task


p0mvn (approved):
Changes make sense to me

PR #2246, opened 2023-10-13
fix: proper fix for no routes for this trade bug

p0mvn:

Properly fixes the "no routes for this trade bug" (screenshot attached)

The problem stemmed from the fact that the router selected 4 4-hop routes where, for some of them, the number of token out denoms mismatched the number of pools, falling validation as a result.

In https://github.com/osmosis-labs/osmosis-frontend/pull/2242, we made a hotfix to increase the number of routes considered from 4 to 6. This allowed router to pick up routes with fewer hops and make routing work.

However, that fix is inappropriate and only decreases the frequency of the issue occurring.

It was observed that routes with fewer hops never had this mismatch. As a result, a guess was made that we somehow modify the shared inputs in the recursive function. In JavaScript arrays are pass-by-reference. As a result, if we make an invalid mutation, it would be applied to all recursive calls.

It looks like this is indeed what was happening. By making deep copies of shared array inputs. The manual test started passing against ATOM / USDC axl pool when the max routes is 4

image

### ClickUp Task

### DIFF
--- a/packages/pools/src/router/routes.ts
+++ b/packages/pools/src/router/routes.ts
@@ -522,7 +522,13 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
       }
 
       for (let i = 0; i < pools.length; i++) {
-        if (poolsUsed[i]) {
+        // Create a deep copy of the current route, token outs
+        // and pools used.
+        const currentPoolRoute = currentRoute.slice();
+        const currentPoolTokenOuts = currentTokenOuts.slice();
+        const currentPoolsUsed = poolsUsed.slice();
+
+        if (currentPoolsUsed[i]) {
           continue; // skip pool
         }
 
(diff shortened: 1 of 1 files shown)


packages/pools/src/router/routes.ts
```diff
@@ -522,7 +522,13 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
       }
 
       for (let i = 0; i < pools.length; i++) {
-        if (poolsUsed[i]) {
+        // Create a deep copy of the current route, token outs
+        // and pools used.
+        const currentPoolRoute = currentRoute.slice();
+        const currentPoolTokenOuts = currentTokenOuts.slice();
+        const currentPoolsUsed = poolsUsed.slice();
```
p0mvn, line 529: I believe that we incorrectly applied mutations to shared array inputs that are passed by reference. As a result, decided to create deep copies for all of them

jonator:
I think this is unrelated. But the way we shift transmuter pools to the front of the greedy candidate route search causes it to be unoptimally picked first through so many pools as we see here:

Screenshot 2023-10-13 at 7 36 07 AM

A quick fix would be setting max hops to 3? Thoughts?

jonator (approved):

packages/pools/src/router/routes.ts
ValarDragon: Ahh nice job!

PR #2253, opened 2023-10-13
Router fix: Add current sqrt price from tick data provider

jonator:
### ClickUp Task

### DIFF
--- a/packages/pools/src/concentrated/fetch-tick-data-provider.ts
+++ b/packages/pools/src/concentrated/fetch-tick-data-provider.ts
@@ -1,5 +1,6 @@
 import { Dec, Int } from "@keplr-wallet/unit";
 import {
+  BigDec,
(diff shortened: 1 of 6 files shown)


packages/pools/src/concentrated/fetch-tick-data-provider.ts
```diff
@@ -256,9 +260,10 @@ export class FetchTickDataProvider implements TickDataProvider {
 
 function serializeRequest(
   response: TickDepthsResponse
-): SerializedTickDepthsResponse {
+): DeserializedTickDepthsResponse {
```
czarcas7ic, line 263: Is this just a naming convention thing? I would think the serializeRequest has a serializeResponse

czarcas7ic (approved):
High level ACK

JoseRFelix (approved):
🚀

PR #2259, opened 2023-10-16
CL Add Liq: use pool tick spacing instead of 1000

jonator:

Uses pool tick spacing param instead of 1000 for rounding the tick when a user is creating a position. This should allow users to make tighter positions with no issue.

I was told by @czarcas7ic that we no longer need to round to 1000th nearest tick, which was a concern at launch.

### ClickUp Task

### DIFF
--- a/packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
+++ b/packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
@@ -205,8 +205,7 @@ export class ObservableAddConcentratedLiquidityConfig {
 
   /** Used to ensure ticks are cleanly divisible by. */
   protected get tickDivisor() {
-    // TODO: use tickspacing from pool going forward
-    return new Int(1000);
+    return new Int(this.pool?.tickSpacing ?? 100);
   }
 
   @computed


packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
```diff
@@ -205,8 +205,7 @@ export class ObservableAddConcentratedLiquidityConfig {
 
   /** Used to ensure ticks are cleanly divisible by. */
   protected get tickDivisor() {
-    // TODO: use tickspacing from pool going forward
-    return new Int(1000);
+    return new Int(this.pool?.tickSpacing ?? 100);
```
p0mvn, line 208: Why can't we always use the pool's tick spacing? What is the purpose of harcoding 100? Is it because we might not have the pool data available in some edge cases yet?

p0mvn (approved):

packages/stores/src/ui-config/manage-liquidity/add-concentrated-liquidity.ts
jonator: Yeah the pool data needs to load

kamal-sutra:
Honestly a huge QOL improvement for FE users. Thanks both.

PR #2260, opened 2023-10-16
disable transmuter

jonator:
### ClickUp Task


JoseRFelix (approved):

jonator:
@ValarDragon do we still want this? I think you requested this IIRC

ValarDragon:
I think no longer, lets close

PR #2264, opened 2023-10-16
fix: failing to skip direct quotes with low liq

p0mvn:

Currently, we fail to skip direct quotes with low liquidity due to `Promise.all` never resolving. This stems from not being able to catch errors for individual quotes inside the map.

The solution is to add `await` inside the map logic so that we can handle any error from individual quotes and skip them silently.

Tested locally.


jonator (approved):

jonator (approved):

ValarDragon:
Great job!

PR #2252, opened 2023-10-13
Cosmos Kit Upgrade type fix

NoahSaso:
## What is the purpose of the change

This fixes some bugs introduced by breaking changes in the latest Cosmos Kit upgrade.

## Brief Changelog

- Type error fix
- Change wallet image tag to `img` from Next's `Image` to support externally linked wallet images
- Fix Cosmos Kit breaking change fee setting

## Testing and Verifying

This change has been tested locally by rebuilding the website.

PR #2272, opened 2023-10-18
Update cosmos kit packages

NoahSaso:

This updates Cosmos Kit packages to make a handful of UX improvements:
- Fix a fee bug introduced in a recent version of Cosmos Kit. The workaround is removed in this PR since the fix was implemented.
- Fix Keplr Mobile deep linking support.
- Add a simple placeholer image for the iframe wallet when the outer wallet is disconnected.
- Added recursive architecture to the iframes so that nested apps can use their own iframes internally.

## Testing and Verifying

This change has been tested locally. Test it by trying to submit a transaction and watching it succeed without having to manually set a fee in Keplr.


NoahSaso:
@JoseRFelix this undoes a fix from the past PR (#2252) by updating Cosmos Kit which just patched it.

JoseRFelix (approved):
Thank you, LGTM!

NoahSaso:
snuck another cosmos kit version update in that adds a placeholder image for the iframe wallet (UX improvement)

packages/stores/package.json
```diff
@@ -53,7 +53,7 @@
     "@cosmjs/math": "0.29.0",
     "@cosmjs/proto-signing": "0.29.0",
     "@cosmjs/stargate": "0.29.0",
-    "@cosmos-kit/core": "2.7.0",
```
JoseRFelix, line 56: This version should match `web/package.json`

JoseRFelix (commented):
Thank you for this PR! Left a small comment

packages/stores/package.json
NoahSaso: oops great catch

NoahSaso:
> Thank you for this PR! Left a small comment

nice thank you! fixed

PR #2277, opened 2023-10-20
Update Pool liquidity-24h Volume display order

kamal-sutra:
On balancer pools, 24hr Trading Volume is shown first and Pool Liquidity next. It's vice-versa on CL pools.

This PR makes the display consistent among all pools with Pool Liquidity first and 24h Trading Volume next.

**Before:**
image
image

**After changes (testing on localhost):**
image


jonator (approved):

sunnya97:
Good catch @kamal-sutra!

I actually think it should be the other way around.  Volume first, then liquidity.  Will create a PR to fix

kamal-sutra:
@sunnya97 Agree. Also if you can show Assets' composition inside CL pools- the amounts of each asset (as in balancer pools)

PR #2289, opened 2023-10-23
Mattupham/validator table highlight row

mattupham:
## What is the purpose of the change

- add row highlight on hover and select
- fix issue with checkbox losing height / width on render
- move table into validator squad component, weird props issues when it was in a separate component

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0td4pp)

## Testing and Verifying

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/9772deb8-0897-46d7-8a3a-a48180a37b6a

PR #2291, opened 2023-10-24
Mattupham/validator search updates 2

mattupham:
- we need to merge in #2289 before this PR

- addresses searching bugs as seen in the clickup tasks below, once filtered, Row IDs were not matching the previously selected row data
- currently our search implementation with useFilteredData does not work well with the core data model in the React Table package, we haven't run into issues before with useFilteredData because other tables throughout our app don't make use of selection (checkboxes)
- React Table assumes all of the data passed in isn't filtered, and the underlying functions in React Table will manage this for us
- refactor to use the React Table Filters example, make use of `@tanstack/match-sorter-utils`
- example [here](https://tanstack.com/table/v8/docs/examples/react/filters)

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a0yd2hg)
[ClickUp Task URL](https://app.clickup.com/t/86a0yd2n2)

## Brief Changelog

- add fuzzyFilter function
- add globalFilter
- add getFilteredRowModel
- update debounced search input to use these functions above

## Testing and Verifying

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/ef320b06-e287-4219-82d4-5395d7d42c93

### DIFF
--- a/packages/web/pages/stake/index.tsx
+++ b/packages/web/pages/stake/index.tsx
@@ -333,7 +333,6 @@ export const Staking: React.FC = observer(() => {
 
   const disableMainStakeCardButton =
     Boolean(isWalletConnected) && Number(amountConfig.amount) <= 0;
-  ///////
 
   return (
     <main className="m-auto flex max-w-container flex-col gap-5 bg-osmoverse-900 p-8 md:p-3">


packages/web/modals/validator-squad.tsx
```diff
@@ -157,241 +208,222 @@ export const ValidatorSquadModal: FunctionComponent<ValidatorSquadModalProps> =
...
+                  const formattedWebsite = props.row.original.formattedWebsite;
+                  const website = props.row.original.website;
+
+                  const operatorAddress = props.row.original.operatorAddress;
+
+                  const imageUrl =
+                    queryValidators.getValidatorThumbnail(operatorAddress);
+
+                  return (
+                    <div className="flex max-w-[15.625rem] items-center gap-3 sm:w-[300px]">
```
MaxMillington, line 298: why the mix of rem and px? And do we need decimal places with our rem?

MaxMillington: I agree that REM is better

MaxMillington (approved):

PR #2296, opened 2023-10-24
WIP: No longer non-zero native osmosis assets at top

ValarDragon:
WIP PR, goal is to no longer show 0 balance osmosis assets at the top, but instead have them filtering by TVL.

Stage of the PR is not ready for review, just sanity checking my understanding of the logic here.


sunnya97:
@ValarDragon shouldn't this just push them to a different spot down the list, instead of hiding them altogether

ValarDragon:
Yes, this PR was not ready for review

PR #2301, opened 2023-10-25
Make assets table no longer sort 0-balance native assets to top

ValarDragon:
I know assets v2 is coming out, but I think this is a nice win in the interrim. (It currently makes the assets page low utility for me as a user). Currently putting 0 balance osmosis native assets to the top prevents a lot of utility for me.

This PR:
- Reduces some code duplication in constructing the TableCell object
- Changes FiatValueRaw from a string to a Decimal
- Removes `isCw20` from native asset (seems like a leftover bug that it was there)
- Removes sorting 0 balance native assets to the top. Right now it puts them at the bottom which isn't ideal, but I think significantly better than now. Should get unified with however else we determine default asset sorting.


ValarDragon:
This takes my logged in view from 
image
to what I really want:
image

sunnya97 (approved):

PR #2306, opened 2023-10-25
Fix swap tool init & router pool filter

jonator:

Removes the disposer function so the spot quote and user input will continue to request quotes. However, since the disposer is not being used it may cause a memory leak if the app is used for a very long time.

### ClickUp Task

### DIFF
--- a/packages/pools/src/router/__tests__/routes.spec.ts
+++ b/packages/pools/src/router/__tests__/routes.spec.ts
@@ -375,8 +375,8 @@ describe("OptimizedRoutes", () => {
         route.pools.map((pool) => pool.id)
       );
 
-      expect(normRoute1PoolIds.includes("8")).toBeTruthy(); // includes preferred pool, shortest route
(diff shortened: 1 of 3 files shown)


packages/pools/src/router/__tests__/routes.spec.ts
```diff
@@ -375,8 +375,8 @@ describe("OptimizedRoutes", () => {
         route.pools.map((pool) => pool.id)
       );
 
-      expect(normRoute1PoolIds.includes("8")).toBeTruthy(); // includes preferred pool, shortest route
-      expect(prefRoute1PoolIds.includes("8")).toBeTruthy(); // NOT preferred pool, but high liq route
+      expect(normRoute1PoolIds.includes("8")).toBeFalsy(); // includes preferred pool, shortest route
+      expect(prefRoute1PoolIds.includes("8")).toBeFalsy(); // NOT preferred pool, but high liq route
```
JoseRFelix, line 379: Is this expected? If so, let's update the comment

JoseRFelix (approved):
Left a small nit

PR #2311, opened 2023-10-26
bug: Fix IBCX Price

JeremyParish69:

To fix the price of IBCX token, which on most pages wasn't working.

## Brief Changelog

- update IBCX pool price to use 1042 instead of 1022

## Testing and Verifying

## Discussion
It suddenly started working as soon as I added the CoinGeckoId property to the testnet IBCX.


packages/web/config/generate-chain-infos/source-chain-infos.ts
```diff
@@ -50,6 +50,7 @@ export const testnetChainInfos: SimplifiedChainInfo[] = [
         coinMinimalDenom:
           "factory/osmo13t90mkyvdnmn9wm8hfen6jk3hnlt8uqx8savlvjd5xghy5z6ye2qymy6cy/uibcx",
         coinDecimals: 6,
+        coinGeckoId: "usd-coin",
```
ValarDragon, line 53: confused by this line, but it matches Ion for now

ValarDragon (approved):

PR #2316, opened 2023-10-26
[Multi Bridge] Support Nomic With Bridge Transfer V2

JoseRFelix:
## Description

This is a temporary workaround until we create a Bridge Provider for Nomic, and support custom UI for Bridge Providers

### Testing

- [ ] Display nomic bridge even when multi bridge v2 is enabled


packages/web/config/generate-chain-infos/source-chain-infos.ts
```diff
@@ -4410,6 +4378,38 @@ export const mainnetChainInfos: SimplifiedChainInfo[] = [
     features: ["ibc-transfer", "ibc-go"],
     explorerUrlToTx: "https://vota-explorer.dorafactory.org/dora/tx/{txHash}",
   },
+  {
+    rpc: "", // TODO
+    rest: "", // TODO
```
jonator, line 4383: Should we follow up on this? @JeremyParish69

jonator (approved):

packages/web/config/generate-chain-infos/source-chain-infos.ts
JoseRFelix: Since Nomic's mainnet is not live, we don't have these values, but they should be available by the time #2317 is merged

PR #2317, opened 2023-10-27
Nomic integration tweaks

cwlittle:
This pull request fixes minor issues in the initial integration of the Nomic Bitcoin Bridge.

## Brief Changelog
- Fix a display issue with miner fees
- Update the relayer endpoint
- Display minimum deposits

## Testing and Verifying
These changes have been locally tested 

N/A

### DIFF
--- a/packages/web/integrations/nomic/transfer.tsx
+++ b/packages/web/integrations/nomic/transfer.tsx
@@ -102,7 +102,9 @@ const NomicTransfer: FunctionComponent<
 
     useEffect(() => {
       if (!osmosisAccount || !osmosisAccount.address) return;
-      const relayers = IS_TESTNET ? ["http://10.16.57.124:8999"] : [];
+      const relayers = IS_TESTNET
+        ? ["https://testnet-relayer.nomic.io:8443"]
+        : [];
       console.log(balanceOnOsmosis.destChannelId);
       generateDepositAddress({
         relayers,
--- a/packages/web/integrations/nomic/transfer.tsx
+++ b/packages/web/integrations/nomic/transfer.tsx
@@ -33,13 +33,6 @@ export const displayBtc = (num: number): string => {
   return resStr.replace(/\.?0+$/, "") + " BTC";
 };
 
-const calculateMinDeposit = (
-  minerFeeSats: number | undefined,
-  bridgeFee: number | undefined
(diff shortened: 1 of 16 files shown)


packages/web/integrations/nomic/transfer.tsx
```diff
@@ -33,6 +33,13 @@ export const displayBtc = (num: number): string => {
   return resStr.replace(/\.?0+$/, "") + " BTC";
 };
 
+const calculateMinDeposit = (
+  minerFeeSats: number | undefined,
+  bridgeFee: number | undefined
+): number => {
+  return 1000 / (1 - (bridgeFee as number)) + (minerFeeSats as number) * 1e8;
+};
```
JoseRFelix, line 41: We need to manage undefined values effectively. Perhaps we can default both bridgeFee and minerFeeStas to 0 or simply exit early?

JoseRFelix (changes requested):
Left a single comment

packages/web/integrations/nomic/transfer.tsx
cwlittle: I didn't want an improper value to stick in the case of race condition. Defaulting to 0 will always yield a result that is less than the actual minimum deposit. In any case, since `minerFee` and `bridgeFee` are pulled from the same response as `reachedCapacityLimit`, we can be sure that the user won't be able to proceed to the deposit page without the client first setting those values. However, I've updated this to make that contract more explicit.

packages/web/integrations/nomic/transfer.tsx
```diff
@@ -121,6 +117,9 @@ const NomicTransfer: FunctionComponent<
         if (res.code === 0) {
           setBridgeFee(res.bridgeFeeRate);
           setMinerFee(res.minerFeeRate);
+          setMinimumDeposit(
+            1000 / (1 - res.bridgeFeeRate) + res.minerFeeRate * 1e8
+          );
           setDepositAddress(res.bitcoinAddress);
           setQrBlob(res.qrCodeData);
           setReachedCapacityLimit(false);
```
JoseRFelix, line 125: What are your thoughts on creating a single-state object for all these values?

JoseRFelix (approved):

PR #2330, opened 2023-10-30
Nomic finalization

cwlittle:
Finalize changes for Nomic integration


packages/web/integrations/nomic/transfer.tsx
```diff
@@ -387,19 +387,13 @@ const NomicTransfer: FunctionComponent<
...
-                    <span>{t("assets.nomic.bridgeFee")}</span>
-                    <SkeletonLoader
-                      isLoaded={Boolean(bridgeInfo?.bridgeFeeRate)}
-                    >
-                      <span>{`${(
-                        (bridgeInfo?.bridgeFeeRate as number) * 100
-                      ).toFixed(2)}%`}</span>
+                    <span>{t("assets.nomic.bitcoinMinerFee")}</span>
+                    <SkeletonLoader isLoaded={Boolean(bridgeInfo)}>
+                      <span>{`~0.00002 BTC`}</span>
```
JoseRFelix, line 396: Is this expected?

cwlittle: Yes. The bridge fee only applies to deposits.

JoseRFelix: Could this change in the future? If so, how can we keep this updated?

cwlittle: We have some relayer updates that are coming soon that will provide this value. We'll open a PR to get that integrated.

JoseRFelix (approved):

PR #2332, opened 2023-10-31
Prevent trying to withdraw nBTC with a zero balance

mappum:
Fixes an issue where the incorrect denom was selected when the user tried to withdraw nBTC without having an nBTC balance. h/t @kamal-sutra

### DIFF
--- a/packages/web/integrations/nomic/transfer.tsx
+++ b/packages/web/integrations/nomic/transfer.tsx
@@ -181,18 +181,6 @@ const NomicTransfer: FunctionComponent<
         return;
       }
 
-      if (withdrawAmountConfig.sendCurrency.coinDenom !== "usat") {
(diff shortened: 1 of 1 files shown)


packages/web/integrations/nomic/transfer.tsx
```diff
@@ -181,6 +181,18 @@ const NomicTransfer: FunctionComponent<
         return;
       }
 
+      if (withdrawAmountConfig.sendCurrency.coinDenom !== "usat") {
```
JoseRFelix, line 184: What are your thoughts on disabling the submit button instead of waiting for the user to click withdraw

kamal-sutra:
Working fine now
image

JoseRFelix (approved):

PR #2354, opened 2023-11-01
CL mobile fixes

jonator:
### ClickUp Task

### DIFF
--- a/packages/web/components/pool-detail/base-pool.tsx
+++ b/packages/web/components/pool-detail/base-pool.tsx
@@ -87,7 +87,7 @@ export const BasePoolDetails: FunctionComponent<{
                     {t("pool.swapFee")}
                   </span>
                   <h4 className="text-osmoverse-100">
-                    {new RatePretty(pool.swapFee).maxDecimals(0).toString()}
+                    {new RatePretty(pool.swapFee).maxDecimals(2).toString()}
                   </h4>
                 </div>
               </div>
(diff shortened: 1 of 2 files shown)


arispech (approved):
LGTM

jonator:
QA approves

ValarDragon (dismissed):

packages/web/components/pool-detail/base-pool.tsx
```diff
@@ -87,7 +87,7 @@ export const BasePoolDetails: FunctionComponent<{
                     {t("pool.swapFee")}
                   </span>
                   <h4 className="text-osmoverse-100">
-                    {new RatePretty(pool.swapFee).maxDecimals(0).toString()}
+                    {new RatePretty(pool.swapFee).maxDecimals(2).toString()}
```
ValarDragon, line 90: Why did this change?

ValarDragon (approved):
CSS changes LGTM. Confused by the swap fee display change

packages/web/components/pool-detail/base-pool.tsx
ValarDragon: Ah this would never work, since its always 1 to 2 decimal points.

ValarDragon: Is there somewhere that makes this a mobile only component, or did this not work on desktop either?

jonator: We have `isMobile` that we could use

PR #2362, opened 2023-11-02
Lower routable pools 1k

jonator:
### ClickUp Task

### DIFF
--- a/packages/web/hooks/data/use-routable-pools.ts
+++ b/packages/web/hooks/data/use-routable-pools.ts
@@ -93,7 +93,7 @@ export function useRoutablePools(
           return pool
             .computeTotalValueLocked(priceStore)
             .toDec()
-            .gte(new Dec(10_000));
+            .gte(new Dec(1_000));
         })
         .sort((a, b) => {
           // sort by TVL to find routes amongst most valuable pools first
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Does this apply regardless of pool type, and regardless of whether the Unverified assets setting is toggled?

jonator:
> Does this apply regardless of pool type, and regardless of whether the Unverified assets setting is toggled?

Yes

JeremyParish69 (changes requested):
The preview does not seem to load pools below 10k (e.g., DYS, RAC, SUI, CRE--all missing) , and I can't select MEDAS, which has a couple pools above 10k.

MaxMillington (approved):

sunnya97 (approved):
Confirmed that the assets that @JeremyParish69 mentioned are now visible.

JeremyParish69 (approved):
LGTM. After turning OFF the Unverified setting, refreshing, and searching the swap widget, I can find assets that can only route through pools barely over $1000 (e.g., DYS). I couple of the assets I mentioned earlier no longer have >1000 liq in any pool, so not all are searchable, but that's to be expected.

PR #2366, opened 2023-11-03
Publish Stage

jonator:
### ClickUp Task


MaxMillington (approved):

sunnya97:
oops squashed by accident 😬

Is there a way to default the button for stage -> master to not be squash?

MaxMillington:
> oops squashed by accident 😬
> 
> Is there a way to default the button for stage -> master to not be squash?

@sunnya97 I don't think so, at least we weren't able to figure it out. We either have to remove squashing for everything or not. 

So long as we don't need to revert, we should be fine

PR #2367, opened 2023-11-03
Readable errors & keep modal open after failed tx & fix Inlang/localizations

jonator:

* Add human readable errors for common chain errors:
  * Slippage issue when swapping
Screenshot 2023-11-03 at 2 00 03 PM

  * Invalid amounts when adding CL liquidity (frontend should prevent this)
Screenshot 2023-11-03 at 6 57 53 PM

* Keep CL modals open when tx fails (or user rejects tx)

### ClickUp Task


packages/web/components/alert/prettify.ts
```diff
@@ -39,11 +46,32 @@ export function prettifyTxError(
...
+          return ["errors.sequenceNumber", { sequence }];
         }
       }
     }
 
+    // Failed swap due to slippage
+    const matchFailedSwapSlippage = message.match(regexFailedSwapSlippage);
+    if (matchFailedSwapSlippage) {
+      if (matchFailedSwapSlippage.length >= 2) {
+        const denom = matchFailedSwapSlippage[1];
```
MaxMillington, line 58: how do we know it will always be the [1] index?

packages/web/components/alert/toast.tsx
```diff
@@ -75,7 +79,11 @@ const ErrorToast: FunctionComponent<Alert> = ({ message, caption }) => (
     </div>
     <div className="text-white-high">
       <h6 className="mb-2 text-lg md:text-base">{t(message)}</h6>
-      {caption && <p className="text-sm md:text-xs">{t(caption)}</p>}
+      {caption && (
+        <p className="text-sm md:text-xs">
+          {typeof caption === "string" ? t(caption) : t(...caption)}
```
MaxMillington, line 84: how does this work? What data structure is `caption`?

jonator: It's either a string or an array containing params to `t()`. It will be a string in the case it's an unexpected error string from the chain, in which case it will just be displayed. Otherwise if it's an array, the error string from the chain matched some regex and was converted into a translation string, that may contain a string localization parameter (like the name of a token, or amount of a token).

packages/web/components/alert/prettify.ts
jonator: I think it's reasonable to expect it will be 1 since the chain errors are highly deterministic, and the regex is specific enough to assume it will only match that error case coming from chain

MaxMillington (approved):

PR #2369, opened 2023-11-06
Revert master

jonator:
The reverts the revert we did on master to restore the LD functionality. We can use this PR to add the changes to stage and also fix issues with LD.


packages/web/pages/_app.tsx
```diff
@@ -311,7 +323,11 @@ const ldAnonymousContext = {
   anonymous: true,
 };
 
-export default withLDProvider({
+const myID = undefined;
```
MaxMillington, line 326:
```suggestion
const myID = process.env.NEXT_PUBLIC_LAUNCH_DARKLY_CLIENT_SIDE_ID;
```

MaxMillington (approved):

PR #2372, opened 2023-11-06
small style adjustments

MaxMillington:
Screenshot 2023-11-06 at 4 01 53 PM

### DIFF
--- a/packages/web/components/main-menu.tsx
+++ b/packages/web/components/main-menu.tsx
@@ -64,25 +64,29 @@ const MorePopover: FunctionComponent<{
 }> = ({ item, secondaryMenus }) => {
   return (
     <Popover className="relative flex">
-      <Popover.Button className="h-full w-full px-4 py-3 focus:outline-none">
-        <MenuItemContent menu={item} />
-      </Popover.Button>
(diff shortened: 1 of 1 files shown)


packages/web/components/main-menu.tsx
```diff
@@ -174,11 +178,11 @@ const MenuItemContent: React.FC<{
               {badge}
             </div>
             <div
-              className={`transition-visibility transition-opacity duration-300 ease-in-out ${
+              className={`transition-visibility mt-0 transition-opacity duration-300 ease-in-out ${
```
jonator, line 181: There's an `mt-1` tag within this same string. IMO using `classNames` utility function instead of string interpolation helps make this a bit easier to avoid.

jonator (commented):
It seems the vertical alignment from the items-center is lost here. Is it possible to add that back when not open? More specifically: in assets button notice the top of the A is aligned with the top of the pie icon.

Screenshot 2023-11-06 at 6 20 44 PM

jonator (changes requested):
~~It seems we lost the vertical alignment on the icon and the text here. We should see if we can maintain the vertical alignment when the item is not open.~~

Accidentally reviewed twice

jonator (approved):

PR #2382, opened 2023-11-08
Edge function router

jonator:

@p0mvn requests we add our router to an edge function so that he can compare our routes against his during automated testing.

Example route:
`/api/router/token-out-given-in?tokenInDenom=uion&tokenInAmount=100&tokenOutDenom=uosmo`

IBC denoms will need to be URL encoded

### DIFF
--- a/packages/pools/src/router/routes.ts
+++ b/packages/pools/src/router/routes.ts
@@ -442,7 +442,7 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
    *  @param tokenOutDenom The output token denom.
    *  @param opts Options for the search.
    */
-  protected getCandidateRoutes(
+  getCandidateRoutes(
     tokenInDenom: string,
     tokenOutDenom: string,
     pools = this._sortedPools
--- a/packages/pools/src/router/routes.ts
+++ b/packages/pools/src/router/routes.ts
@@ -31,10 +31,6 @@ export type OptimizedRoutesParams = {
   /** **Ordered** IDs of pools to be prioritized in route selection.
    *  The first pools in the array will be selected first. */
   preferredPoolIds?: string[];
-  /** IDs of pools receiving OSMO incentives. */
-  incentivizedPoolIds: string[];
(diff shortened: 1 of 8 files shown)


packages/web/pages/api/router/token-out-given-in.ts
```diff
@@ -0,0 +1,147 @@
...
+    initialAmount: string;
+    pools: {
+      id: string;
+    }[];
+    tokenOutDenoms: string[];
+    tokenInDenom: string;
+  }[];
+};
+
+export default async function routeTokenOutGivenIn(req: Request) {
```
p0mvn, line 33: Do you mind also exposing the candidate route selection logic?

That is, for the given two denoms, return the candidate routes

packages/pools/src/router/routes.ts
```diff
@@ -31,10 +31,6 @@ export type OptimizedRoutesParams = {
   /** **Ordered** IDs of pools to be prioritized in route selection.
    *  The first pools in the array will be selected first. */
   preferredPoolIds?: string[];
-  /** IDs of pools receiving OSMO incentives. */
-  incentivizedPoolIds: string[];
-  /** Min/base denom of stake currency of chain (OSMO) */
-  stakeCurrencyMinDenom: string;
```
p0mvn, line 37: Is this a drive-by change?

p0mvn (approved):
LGTM

packages/web/pages/api/router/token-out-given-in.ts
jonator: You want candidate routes returned in API?

packages/pools/src/router/routes.ts
jonator: I forgot to remove it when I removed the OSMO 2 hop discount, as that's all it was used for. So yes you could say it's drive by.

packages/web/pages/api/router/token-out-given-in.ts
p0mvn: Yeah

PR #2385, opened 2023-11-09
Fix route sorting & fix stage build

jonator:

Route sorting would be lost when filtering candidate routes for unique pools.

### ClickUp Task


packages/pools/src/router/routes.ts
```diff
@@ -188,7 +188,7 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
         return includedRoutes;
       } else {
         route.pools.forEach(({ id }) => uniquePoolIds.add(id));
-        includedRoutes.unshift(route);
```
ValarDragon, line 191: LGTM to in isolation, though not seeing the corresponding sort line

jonator: Before they're filtered, they're sorted on line 182

packages/pools/src/router/__tests__/routes.spec.ts
```diff
@@ -265,104 +265,6 @@ describe("OptimizedRoutes", () => {
       ]);
       expect(scaledOut.amount.toString()).toEqual(normalOut.amount.toString());
     });
-
-    it("uses preferred pool IDs - lifts preferred direct pool into used route", async () => {
```
jonator, line 269: No longer desired behavior

packages/pools/src/router/routes.ts
ValarDragon: Hrmm, thats sorting by path length, not sorting by net price to user though?

MaxMillington (approved):

packages/pools/src/router/routes.ts
jonator: Yeah it's just sorting by length. Price comparison will happen later

This PR contains a critical bug fix, we can follow up with improvements

PR #2407, opened 2023-11-13
add superflow

sunnya97:

Adds superflow QA tool.  Should only impact stage.osmosis.zone

Just followed instructions here

CleanShot 2023-11-13 at 13 31 08@2x

### DIFF
--- a/packages/web/pages/_app.tsx
+++ b/packages/web/pages/_app.tsx
@@ -1,6 +1,8 @@
 import "../styles/globals.css"; // eslint-disable-line no-restricted-imports
 import "react-toastify/dist/ReactToastify.css"; // some styles overridden in globals.css
 
+// import superflow
+import { initSuperflow } from "@usesuperflow/client";
 import axios from "axios";
 import dayjs from "dayjs";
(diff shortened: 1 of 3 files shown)


packages/web/pages/_app.tsx
```diff
@@ -63,6 +65,7 @@ const DEFAULT_LANGUAGE = "en";
 
 function MyApp({ Component, pageProps }: AppProps) {
   useAmplitudeAnalytics({ init: true });
+  initSuperflow("GbkZQ3DHV4rsGongQlYg", { projectId: "2059891376305922" });
```
jonator, line 68: NIT: IDK how big superflow is, but we could look into dynamic imports for only if we're on stage to reduce bundle size

jonator (approved):

packages/web/pages/_app.tsx
jonator: @MaxMillington https://www.npmjs.com/package/@usesuperflow/client

It's not too big

sunnya97:
@jonator why is `Jest Tests Workflow / test (18.x, web)` isn't working?

PR #2408, opened 2023-11-13
chore: mergify backport to master configuration

p0mvn:

This is a convenient configuration for we need to backport individual PRs from stage to master without publishing all updates.

In such a case, the label "A:backport/master" can be applied to PR. Upon application, mergify opens a new PR against `master`.

This is useful for emergencies when there are a lot of updates on stage but we want to merge an individual fix

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,15 @@
+defaults:
+  actions:
+    backport:
+      assignees:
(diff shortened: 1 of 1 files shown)


p0mvn:
@Mergifyio refresh

sunnya97 (approved):

PR #2423, opened 2023-11-14
Display transmuter pools in pools table

sunnya97:

This PR makes it so that we show transmuter pools in the pools table.  WHen clicked, it links out to the celatone page for the pool.


sunnya97:
What's the easiest way to make the external link open in a new tab while keeping the internal link in the same tab?

jonator (approved):

PR #2426, opened 2023-11-14
prototype: add smart account page and signature verification

PaddyMc:

Prototype for spendlimits


MaxMillington (commented):
TLDR 🤣

packages/stores/src/account/base.ts
```diff
@@ -682,95 +693,185 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
-      value: signedTxBody,
-    };
-
-    const signedTxBodyBytes = wallet?.signingStargateOptions?.registry?.encode(
-      signedTxBodyEncodeObject
+    const sessionKey = localStorage.getItem("SessionKey");
+    if (sessionKey == null) {
+      throw new Error("session key not found or broken");
+    }
+    const privateKey = new PrivKeySecp256k1(fromBase64(sessionKey));
```
JoseRFelix, line 744: Will this be able to support `PrivKeySecp256r1`?

sunnya97 (changes requested):
asdf

sunnya97:
- [ ] dsfdf

packages/proto-codecs/scripts/codegen.ts
```diff
@@ -27,6 +27,7 @@ const includedPackages: AvailablePackages[] = [
   "osmosis.concentratedliquidity.v1beta1",
```
sunnya97, line 27: This is broken

packages/stores/src/account/base.ts
PaddyMc: It should if we edit the code to generate an R1 signature, not sure if it's needed atm

JoseRFelix (approved):

PR #2438, opened 2023-11-15
feat: Use Commit Hash to Generate Asset Lists

JoseRFelix:

Use commit hash to generate asset lists to avoid waiting for GitHub's CDN to refresh its cache.

## Testing

- [ ] Assets and chains should load like production


sunnya97:
This can be done via url_param?

JoseRFelix:
> This can be done via url_param?

This can't be done until we render the whole app through the server

jonator (approved):

sunnya97:
We should probably have a test asset to check this with?

kamal-sutra:
@JoseRFelix @JeremyParish69 Found only these differences. Rest all good. Please check/update.
image

PR #2442, opened 2023-11-15
feat: Display Unlisted Assets Based on Query Key

JoseRFelix:

To view unlisted assets for testing, append the following query parameter to the Osmosis URL:

```
?show_unlisted_assets=true
```

They'll be enabled for the tab's session. If you'd like to disable it, either open a new tab without the query parameter or restart the browser.

## Testing and Verifying

- [ ] Unlisted assets are enabled for a session when `?show_unlisted_assets=true` is appended to the URL
- [ ] Unlisted assets should not be visible by default


jonator (approved):

JeremyParish69 (changes requested):
As of the lasted successful build, I am still able to see tokens that should be unlisted (sqOSMO and sqATOM). Upon appending ?show_unlisted_assets to the assets page, I still cannot see QWOYN, which has been added to assetlist.

Also, when appending ?show_unlisted_assets to the URL, as soon as I click to a different page in the nav bar, the ?show_unlisted_assets is removed from the URL. It will be confusing for validators to have to manually add that every time they switch to a new page. Having ?show_unlisted_assets persist is desired.

JoseRFelix:
> As of the lasted successful build, I am still able to see tokens that should be unlisted (sqOSMO and sqATOM). Upon appending ?show_unlisted_assets to the assets page, I still cannot see QWOYN, which has been added to assetlist.
> 
> Also, when appending ?show_unlisted_assets to the URL, as soon as I click to a different page in the nav bar, the ?show_unlisted_assets is removed from the URL. It will be confusing for validators to have to manually add that every time they switch to a new page. Having ?show_unlisted_assets persist is desired.

My bad I forgot to add the =true to the instructions. Please try with this param `?show_unlisted_assets=true`. It should look like this https://osmosis-frontend-git-jose-use-feature-flag-to-ffce6e-osmo-labs.vercel.app/?show_unlisted_assets=true. 

Btw, it's fine if it goes away. Unlisted assets will be visible for that session

JeremyParish69 (approved):
works well!

PR #2450, opened 2023-11-16
Mattupham/telescope upgrade unstaking v20 fix 2

mattupham:

- in order to avoid a chain upgrade, we had to do a workaround for direct sign on unstaking
- I generated the new types with telescope, this broke other parts of the app that relied on BigInt since telescope was using LegacyDev
- I upgraded telescope to the newest version, but this broke the build
- I communicated this to cosmology and they were able to release an updated telescope version that builds for us related thread [here](https://osmosis-network.slack.com/archives/C046FSSBZNJ/p1700092676093459?thread_ts=1699636503.212459&cid=C046FSSBZNJ) -  (1.0.17)
- I upgraded telescope, ran telescope, and updated the unstaking fixes in this preview deploy above

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1edwxf)

## Brief Changelog

- update telescope
- run yarn generate
- update types
- update unstaking call
- update disabled staking button based on current amount config

## Testing and Verifying

in QA now

### DIFF
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -41,7 +41,6 @@ import {
   cosmosProtoRegistry,
   cosmwasmProtoRegistry,
   ibcProtoRegistry,
-  osmosis,
   osmosisProtoRegistry,
 } from "@osmosis-labs/proto-codecs";
 import type { AssetList, Chain } from "@osmosis-labs/types";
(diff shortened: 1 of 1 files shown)


MaxMillington (approved):
Looks like there is some feedback in the vercel comments, but I'm not sure it's related to this PR.

jonator (approved):

PR #2458, opened 2023-11-17
feat: Open Create Position When Query Param is Present

JoseRFelix:

Open the CL create position modal with the `open_create_position` query parameter. URLs will look like the following: `https://app.osmosis.zone/pool/1267?open_create_position=true`.

## Testing and Verifying

- [ ] Should open create position modal when passing `?open_create_position=true` to the URL


kamal-sutra (approved):
url remains same when clicking Create position. Works as intended.
image

CryptoAssassin1:
worked fine.
Screenshot from 2023-11-18 00-52-12

CryptoAssassin1 (approved):

PR #2466, opened 2023-11-23
Add neurton.wstETH Price

JeremyParish69:

Defines a price for wstETH using pool 1292

## Brief Changelog

- Define a price for wstETH using pool 1292 as pool:wsteth
(wstETH.axl's is titled 'pool:wsteth-wei', so there is no name conflict)

## Testing and Verifying

## Pending
Validation


JeremyParish69:
No longer applicable since prices are now fetched elsewhere

PR #2483, opened 2023-11-30
feat: Earn Page

DavideSegullo:

Give users clear paths to not just acquire, but also earn more on assets of interest.

### ClickUp Task

[ClickUp Task](https://app.clickup.com/37420681/v/dc/13nzm9-23853/13nzm9-49233)


packages/web/components/earn/allocation/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+    <div className="flex flex-col gap-7 1.5xl:flex-1">
+      <div className="flex items-center gap-3.5">
+        <h5 className="text-lg font-semibold leading-normal text-osmoverse-100 1.5xl:hidden">
+          {t("earnPage.allocation")}
+        </h5>
+        <p className="text-sm font-semibold text-wosmongton-300">5 tokens</p>
+      </div>
+      <div className="relative flex flex-col justify-between gap-12">
+        <div className="flex gap-4">
+          <button className="text-sm font-semibold text-osmoverse-100">
```
MaxMillington, line 93: I think we have existing `<Button />` component. Are we able to use?

packages/web/localizations/pl.json
```diff
@@ -828,22 +828,41 @@
...
+      "all": "Wszystkie tokeny"
+    },
+    "rewardTypes": {
+      "all": "Wszystko",
+      "single": "Pojedynczy",
+      "multi": "Wielo"
+    }
   }
-}
+}
```
jonator, line 868: I think this will fail the linter, can be fixed with `yarn lint:fix`

packages/web/components/radio-with-options/radio-with-options.tsx
```diff
@@ -58,7 +62,8 @@ export const RadioWithOptions = ({
           }
           value={value}
         >
-          {label}
+          {/**the following condition checks if the label passed is from an i18n translation path */}
+          {label.includes(".") ? t(label) : label}
```
jonator, line 66: I will mention it can be possible to not have a "." in a translation path.

packages/web/localizations/pl.json
fabryscript: Solved in [702f7df](https://github.com/osmosis-labs/osmosis-frontend/pull/2483/commits/702f7dfbf4126e4e8915da58aa7c422af346f12c)

packages/web/components/earn/allocation/index.tsx
fabryscript: These buttons are going to be replaced by `Tab` instances, as I already did in the Position / Allocation layout on smaller devices

Screenshot 2023-12-14 alle 11 01 42

packages/web/components/radio-with-options/radio-with-options.tsx
fabryscript: Got it! Do you think it will be better if I pass a flag as a prop that determines the presence of a label that represents an i18n path? (e.g. `labelIsI18nPath: boolean`)

DavideSegullo: We should remove `useTranslation` from radio-with-options, if not, the component is not generic

DavideSegullo: We removed useTranslation inside the component to fix this problem and also improve the component

MaxMillington (approved):
Looks like it just needs a rebase

PR #2489, opened 2023-11-30
DO NOT MERGE disclaimer translation

MaxMillington:
Adding translations for the disclaimer. 

I'm waiting on confirmation from legal on whether or not we should translate these disclaimers, but if they give us the green light this should be good to go


MaxMillington:
closing as not a priority and haven't heard from legal

PR #2493, opened 2023-12-01
One Click Trading

JoseRFelix:

This PR introduces 1-Click Trading, which aims to streamline the trading process by allowing users to execute trades with a single click. It will initially be hidden behind a feature flag because it is not yet live on-chain.

## Brief Changelog

- Adds 1-Click Trading behind a Feature Flag

## Testing and Verifying

- [ ] App should work like production while 1CT flag is off


packages/web/pages/_app.tsx
```diff
@@ -216,6 +216,12 @@ const MainLayoutWrapper: FunctionComponent<{ children: ReactNode }> = observer(
           icon: <Icon id="pool" className="h-5 w-5" />,
           selectionTest: /\/pools/,
         },
+        {
+          label: "Smart Account",
```
jonator, line 220: Reminder to add localization

packages/web/pages/smart-accounts/index.tsx
```diff
@@ -0,0 +1,424 @@
...
+import { OsmoverseCard } from "~/components/cards/osmoverse-card";
+import { CheckBox } from "~/components/control";
+import Spinner from "~/components/spinner";
+import { EventName } from "~/config";
+import { useAmplitudeAnalytics } from "~/hooks";
+import { Authenticator } from "~/server/queries/osmosis/authenticators";
+import { useStore } from "~/stores";
+import { api, RouterOutputs } from "~/utils/trpc";
+
+const SmartAccounts: NextPage = observer(function () {
```
jonator, line 19: Perhaps, similar to `useSwap` we can encapsulate the query/mutation/tx functionality into a `useAuthenticators` (or some better named) hook. In the future this would make it easier to reuse this in the mobile app.

Perhaps you intend to do this as this is a draft and a POC page it seems.

packages/web/pages/_app.tsx
JoseRFelix: Thank you for the feedback! This route will only be used for testing. We'll most likely remove it on release

jonator (commented):
Som thoughts

packages/web/pages/smart-accounts/index.tsx
```diff
@@ -0,0 +1,424 @@
...
+import type { NextPage } from "next";
+import { FunctionComponent } from "react";
+
+import { Button } from "~/components/buttons";
+import { OsmoverseCard } from "~/components/cards/osmoverse-card";
+import { CheckBox } from "~/components/control";
+import Spinner from "~/components/spinner";
+import { EventName } from "~/config";
+import { useAmplitudeAnalytics } from "~/hooks";
+import { Authenticator } from "~/server/queries/osmosis/authenticators";
```
jonator, line 15: Would be good to import this as a type. Otherwise, I believe it would import the server code into FE IIRC.

`import type { Authenticator } from "~/server/queries/osmosis/authenticators";`

packages/web/pages/smart-accounts/index.tsx
```diff
@@ -0,0 +1,424 @@
...
+    api.edge.authenticators.getCosmosAccount.useQuery(
+      {
+        address: osmosisAddress,
+      },
+      {
+        enabled: osmosisAddress.length > 0,
+      }
+    );
+
+  const removeAuthenticator = useMutation(async ({ id }: { id: string }) => {
```
jonator, line 49: As a follow up from Slack discussion, would be good to extract this mutation into a separate file that can be reused for sending txs. Then we can reliable update all relevant queries when this mutation is triggered.

jonator (commented):
More thoughts

packages/web/pages/smart-accounts/index.tsx
JoseRFelix: Yeah, this is more of a POC, but the real impl. I will create a separate hook for the mutations. I don't think we should create separate hooks for trpc queries though since we lose their devx benefits.

packages/web/pages/_app.tsx
MaxMillington, line 220: should be conditionally added based on feature flag

JoseRFelix:
@jonator @MaxMillington The smart account page is now hidden. I plan to add the feature flag in the next PRs for layout updates. For now, this PR will be a temporary solution because we can't merge it to the stage yet, due to the ongoing development of the Smart Account chain feature. Any updates to the proto in the main branch won't be reflected in the smart accounts branch, and I want to ensure we don't delay other updates

JoseRFelix:
@jonator This is not a POC. Despite having Feature Flags, we cannot merge this PR yet because any updates to the proto in the main branch won't be reflected in the chain smart accounts branch, and I want to ensure we don't delay other updates.

kamal-sutra (approved):

CryptoAssassin1 (approved):

jonator (approved):

PR #2494, opened 2023-12-02
feat(web): fixed broken links

omniwired:
It fixes 2 broken links:

- https://docs.osmosis.zone/overview/getting-started/#bonded-liquidity-gauges
- https://docs.osmosis.zone/overview/getting-started/#swap-fees

Both of them at the creation of this s PR are broken in production.

I've fixed the links but it seems the docs changed so the fix is not truly 1:1.

Feel free to point me into the right direction if changes are needed.


jonator (approved):

jonator:
@omniwired can you please update your branch from stage? We try to avoid pushing directly to master. Thanks for the contribution!

PR #2495, opened 2023-12-04
fix: transmuter is not being picked up in router

p0mvn:

Transmuter is currently not being picked up in the router:
https://app.osmosis.zone/?from=USDC.axl&to=USDC

The bug stems from the fact that we change the order of pools with the `unshift` call while validating preferred pools. As a result, all the preferred pools end up getting reversed in terms of the order. Eventually, due to the max split parameter, tranmsuter ends up being filtered out.

The solution is to maintain two arrays:
1. Preferred pool array that maintains the original order where transmuter is most preferred
2. Non-preferred pools

Then, merge the two arrays into 1 while preserving the order.

## Testing and Verifying

Tested locally.

### DIFF
--- a/packages/pools/src/router/routes.ts
+++ b/packages/pools/src/router/routes.ts
@@ -194,19 +194,27 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
     }, [] as Route[]);
 
     // prioritize (pick) routes by preference
+
+    // Maintain order of preferred routes.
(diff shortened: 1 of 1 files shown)


kamal-sutra (approved):
works fine.
image

CryptoAssassin1 (approved):
Worked fine.

jonator (approved):

PR #2503, opened 2023-12-05
feat: astroport pcl

gustavomauricio:

Add Astroport PCL pool interface and quote logic so those can be considered on the router

### DIFF
--- a/packages/pools/src/types.ts
+++ b/packages/pools/src/types.ts
@@ -3,7 +3,7 @@ import {
   ConcentratedLiquidityPoolRaw,
   TickDataProvider,
 } from "./concentrated";
-import { AstroportPclPool, CosmwasmPoolRaw, TransmuterPool } from "./cosmwasm";
+import { CosmwasmPoolRaw, TransmuterPool } from "./cosmwasm";
 import { StablePool, StablePoolRaw } from "./stable";
 import { WeightedPool, WeightedPoolRaw } from "./weighted";
 
@@ -49,14 +49,8 @@ export function makeStaticPoolFromRaw(
     );
   }
   if (rawPool["@type"] === COSMWASM_POOL_TYPE) {
-    rawPool = rawPool as CosmwasmPoolRaw;
-
-    // differentiate cosmoswasm pools by code id - 5005 would be for testnet
-    if (rawPool.code_id === "5005") {
-      return new AstroportPclPool(rawPool);
-    }
-
-    return new TransmuterPool(rawPool);
+    // currently only support transmuter pools
+    return new TransmuterPool(rawPool as CosmwasmPoolRaw);
   }
 
   // Query pool should not be created without a supported pool
--- a/packages/pools/src/types.ts
+++ b/packages/pools/src/types.ts
@@ -3,7 +3,7 @@ import {
   ConcentratedLiquidityPoolRaw,
   TickDataProvider,
 } from "./concentrated";
-import { CosmwasmPoolRaw, TransmuterPool } from "./cosmwasm";
+import { AstroportPclPool, CosmwasmPoolRaw, TransmuterPool } from "./cosmwasm";
 import { StablePool, StablePoolRaw } from "./stable";
 import { WeightedPool, WeightedPoolRaw } from "./weighted";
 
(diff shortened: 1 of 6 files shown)


packages/pools/src/interface.ts
```diff
@@ -2,7 +2,12 @@ import { Dec, Int } from "@keplr-wallet/unit";
 
 /** Interface for base pool data and basic operations on that data. */
 export interface BasePool {
-  get type(): "weighted" | "stable" | "concentrated" | "transmuter";
+  get type():
+    | "weighted"
+    | "stable"
+    | "concentrated"
+    | "transmuter"
+    | "astroport-pcl";
```
jonator, line 10: It appears your branch is quite out of date. I updated this to be it's own type, can you please merge stage into your branch? Thanks.

p0mvn (commented):
The high-level implementation looks good to me.

Is there anything we need to update with regard to pool type priority?

For example, we currently have transmuter as preferred. Is there anything similar we need to do for this pool type

packages/pools/src/types.ts
```diff
@@ -48,8 +49,14 @@ export function makeStaticPoolFromRaw(
     );
   }
   if (rawPool["@type"] === COSMWASM_POOL_TYPE) {
-    // currently only support transmuter pools
-    return new TransmuterPool(rawPool as CosmwasmPoolRaw);
+    rawPool = rawPool as CosmwasmPoolRaw;
+
+    // differentiate cosmoswasm pools by code id - 5005 would be for testnet
+    if (rawPool.code_id === "5005") {
```
sunnya97, line 55: This should be done in the web/config like transmuter code ids:

https://github.com/osmosis-labs/osmosis-frontend/blob/stage/packages/web/config/feature-flag.ts#L50

sunnya97 (changes requested):

jonator (changes requested):
https://github.com/osmosis-labs/osmosis-frontend/pull/2535

Can you please target the branch in this PR and take a look at it's changes for how to add your own code ID checks to then use your own cosmwasm pool implementation.

packages/pools/src/utils.ts
```diff
@@ -0,0 +1,27 @@
+// should be defined somewhere "project" wide
+const REST_ENDPOINT = "https://lcd.osmotest5.osmosis.zone";
```
jonator, line 2: https://github.com/osmosis-labs/osmosis-frontend/blob/b80b0235a6c3eafe3e9e0a8bf3436e7a553638c2/packages/web/server/queries/cosmos/bank/balances.ts#L16

Please reference this for getting the base URL to query from. Also, you may need to pass in the base URL from the transmuter pool since we only have access to chain configs from web package.

packages/pools/src/types.ts
jonator: See latest comment with link to PR containing the branch to target instead of stage.

packages/web/server/queries/complex/route-token-out-given-in.ts
```diff
@@ -100,7 +102,13 @@ export async function getRouter(
           }
 
           if (pool["@type"] === COSMWASM_POOL_TYPE) {
-            return new TransmuterPool(pool as CosmwasmPoolRaw);
+            pool = pool as CosmwasmPoolRaw;
+
+            if (AstroportPoolCodeIds.includes(pool.code_id)) {
+              return new AstroportPclPool(pool);
```
jonator, line 108: Let's pass in the base node query URL here instead of hardcoding in pools package

packages/pools/src/utils.ts
gustavomauricio: Wasn't really sure how you guys wanted me to handle that so I left that comment in there.

Updated now as you suggested, thanks

packages/web/config/feature-flag.ts
```diff
@@ -51,6 +51,9 @@ export const BlacklistedPoolIds: string[] = ["895"];
  *  NOTE: this is only used in the pools query for pools page and pool detail page. */
 export const TransmuterPoolCodeIds = IS_TESTNET ? ["3084"] : ["148"];
 
+/** Cosmwasm Code Ids confirmed to be transmuter pools in current env. */
+export const AstroportPoolCodeIds = IS_TESTNET ? ["5005"] : [""];
```
jonator, line 55: We'll need another PR once you deploy your contract to mainnet, of course.

jonator (approved):

PR #2504, opened 2023-12-05
One Click Trading Queries

JoseRFelix:
### ClickUp Task


packages/web/server/api/edge-routers/authenticators.ts
```diff
@@ -0,0 +1,46 @@
...
+import { queryCosmosAccount } from "~/server/queries/cosmos/auth";
+
+interface NestedAuthenticator {
+  authenticator_type: Authenticator["type"];
+  data: Authenticator["data"];
+}
+
+export const authenticatorsRouter = createTRPCRouter({
+  getAuthenticators: publicProcedure
+    .input(z.object({ address: z.string() }))
```
jonator, line 17: If this must be the user's OSMO address, would be nice to have param name be more specific and have it be `userOsmoAddress` as we have with assets procedures

packages/web/server/queries/authenticators/index.ts
jonator, line 1: Could this be placed in `osmosis` folder since it's an Osmosis query?

packages/stores/src/account/base.ts
```diff
@@ -664,9 +669,11 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
       chainId: chainId,
     };
 
-    const localStorageIsOneClickEnabled =
-      localStorage.getItem("isOneClickTradingEnabled") === "true";
-    if (localStorageIsOneClickEnabled) {
+    const isOneClickTradingEnabled = await this.isOneCLickTradingEnabled();
+
+    console.log(isOneClickTradingEnabled);
```
jonator: Committed log

JoseRFelix: Thank you for the review!

PR #2516, opened 2023-12-08
Fix: increase default slippage for creating position

jonator:
### ClickUp Task


jonator:
According to @czarcas7ic , the ideal fix is a single message that handles the superfluid staking position situation to directly use the amount instead of having to calculate slippage values. Coming in a future release.

PR #2517, opened 2023-12-08
Cleanup test stuff

jonator:
* Add spec to test files
* Move test utils to tests folder


packages/web/components/navbar-osmosis-update/__tests__/navbar-osmosis-update.spec.tsx
```diff
@@ -6,8 +6,11 @@ import { resetLDMocks } from "jest-launchdarkly-mock";
 import { rest } from "msw";
 
 import NavbarOsmosisUpdate from "~/components/navbar-osmosis-update";
-import { server } from "~/tests/msw";
-import { mockFeatureFlags, renderWithProviders } from "~/utils/test-utils";
+import {
+  mockFeatureFlags,
+  renderWithProviders,
+} from "~/utils/__tests__/test-utils.spec";
```
JoseRFelix, line 12: What are your thoughts on placing this in `~/tests/test-utils` instead? I'm not particularly a fan of adding spec to the utils file since they don't have a test inside

MaxMillington: My two cents is that I generally like to keep tests in the same directory as the files they are testing as it's easier to navigate as opposed to an entirely separate directory solely for tests. But I'm not going to go to the mattresses over it

MaxMillington (approved):

PR #2521, opened 2023-12-09
Move to multiple client queries

jonator:

Use multiple separate (unbatched) queries to the tRPC router function for each type of supported router. The UI will then reconcile the responses and show the fastest quote first, replacing with slower quotes as they are received all while the user input is unchanged. This functionality is supported mostly by React Queries `useQueries` hook with minimal business logic for finding the response with the highest out amount.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1mq9ae)

## Brief Changelog

* Add ability to disable batching for certain tRPC queries
* Use the `useQueries` hook to make tRPC calls per supported router (maintaining FF support)

## Testing and Verifying

* Swaps work still


packages/web/hooks/use-swap.ts
```diff
@@ -531,3 +505,95 @@ function useSwapAsset(
     isLoading: isLoading && !existingAsset,
   };
 }
+
+/** Iterates over available and identical routers and sends input to each one individually.
+ *  Results are reduced to best result by out amount.
+ *  Also returns the number of routers that have fetched and errored. */
+function useQueryBestQuote(
```
JoseRFelix, line 512: What are your thoughts on renaming this function to `useRouterBestQuote`?

packages/web/hooks/use-swap.ts
```diff
@@ -531,3 +505,95 @@ function useSwapAsset(
...
+  enabled: boolean,
+  routerKeys = ["legacy", "sidecar", "tfm"] as RouterKey[]
+) {
+  const featureFlags = useFeatureFlags();
+  const availableRouterKeys: RouterKey[] = useMemo(
+    () =>
+      !featureFlags._isInitialized
+        ? []
+        : routerKeys.filter(
+            (key) => featureFlags.sidecarRouter || key !== "sidecar"
```
JoseRFelix, line 523: I believe using a ternary can make it easier to read. 

```suggestion
            (key) => featureFlags.sidecarRouter ?  true : key !== "sidecar"
```

packages/web/hooks/use-swap.ts
```diff
@@ -531,3 +505,95 @@ function useSwapAsset(
...
+          if (cur && best.amount.toDec().lt(cur.amount.toDec())) return cur;
+          return best;
+        }, undefined)
+    );
+  }, [routerResults]);
+
+  const numSucceeded = routerResults.filter(
+    ({ isSuccess }) => isSuccess
+  ).length;
+  const oneIsSuccessful = Boolean(numSucceeded);
```
JoseRFelix, line 578:
```suggestion
  const isOneSuccessful = Boolean(numSucceeded);
```

packages/web/integrations/tfm/router.ts
```diff
@@ -59,18 +59,22 @@ export class TfmRemoteRouter implements TokenOutGivenInRouter {
         priceImpactTokenOut,
       };
     } catch (e) {
+      // perhaps this is from the apiClient, i.e. a timeout
```
JoseRFelix, line 62: We should add some unit tests for these routers. We can add this on another PR. I don't want to block this significant performance improvement.

JoseRFelix (commented):
Looks good! Left some minor comments

packages/web/hooks/use-swap.ts
jonator: I like this, but I'll include "query" since it's a wrapper around `useQueries`

packages/web/integrations/tfm/router.ts
jonator: Like mocking the response and validating the output? Sounds good. These were originally simple query clients but there is some logic added.

packages/web/hooks/use-swap.ts
```diff
@@ -531,3 +519,96 @@ function useSwapAsset(
...
+        },
+        {
+          enabled: enabled && featureFlags._isInitialized,
+
+          // quotes should not be considered fresh for long, otherwise
+          // the gas simulation will fail due to slippage and the user would see errors
+          staleTime: 5_000,
+          cacheTime: 5_000,
+          // don't retry quote, just display the issue to user
+          retry: false,
```
JoseRFelix, line 558: Thoughts on enabling retry only when the error is an unexpected (instanceof Error)?

jonator: Is it possible to dynamically change that value from prior results?

JoseRFelix: Yes! retry can be a [function](https://github.com/osmosis-labs/osmosis-frontend/blob/245c244b6d29dbfa9197edb9c27110154452cdaf/packages/web/modals/bridge-transfer-v2.tsx#L349-L356) that will receive the amount of failures and the last error.

ValarDragon:
I couldn't tell from skim of the pr, is there some sort of "router request id", that prevents old responses from overwriting display on newer requests?

E.g. i send requests for a $20 swap and then a $200 swap

A slow $20 request comes in after all $200 swap requests complete. Do I get a $20 quote result?

jonator:
@ValarDragon all querying is now handled by a lib called "React Query" (tRPC wraps it). It's the industry standard for querying in react apps. It should only return responses from queries that were given the _current_ input. So if you type in another value it should no longer return responses (or incoming or current) for the prior input value. I'll see if I can replicate this.

ValarDragon:
Thanks for the reply, was just asking if it was handled by the lib! I hadn't seen an issue, just didn't see what was handling this race condition from the code.

Glad the underlying library is handling it!

jonator:
Yes @ValarDragon another benefit of moving to tRPC is we can start using react query, which is a highly maintained/developed and featureful (with useful defaults) querying lib that we can start using. I think it will help us write much less code, like in this case.

kamal-sutra (commented):
Swaps loading fast now and Txs working well

JoseRFelix (approved):
LGTM!

PR #2522, opened 2023-12-10
fix: Select Recommended Assets

JoseRFelix:

The initial attempt to select a recommended asset like USDT was broken because the asset was not available in the assets array during the first fetch. To resolve this, we now also include a search for the asset within the recommended assets array.

## Testing and Verifying

- [ ] Can Select Tokens not available on first fetch like USDT and TIA


jonator (approved):

jonator:
The recommended assets are not loading initially. At least on mobile

jonator (approved):

PR #2525, opened 2023-12-11
Update nois.svg

kaisbaccour:
Change the Nois token from an ugly square to a circular coin.


jonator:
@JeremyParish69 arent images pulled from asset list now? I think we should delete all the tokens in the tokens folder now since they should be generated from the asset list at build time

JeremyParish69:
> @JeremyParish69 arent images pulled from asset list now? I think we should delete all the tokens in the tokens folder now since they should be generated from the asset list at build time

That's correct. This change has been made at the Chain Registry, which is where we source from now.

PR #2532, opened 2023-12-11
Mattupham/update primitive amount

mattupham:

primitive amount was using just a single amount config - update to use the dynamic amount config (stake / unstake)

### ClickUp Task


packages/web/pages/stake.tsx
```diff
@@ -214,6 +219,8 @@ export const Staking: React.FC = observer(() => {
       },
     ]);
 
+    console.log("coin: ", coin);
```
czarcas7ic, line 222: Not sure if you want to keep the log or not

czarcas7ic (approved):
high level ack

PR #2534, opened 2023-12-11
Force swap in pool in pool detail page

jonator:

Restores the prior functionality of the trade tokens modal on the pool detail page of forcing a swap through the pool the user is viewing instead of finding the optimal route through all pools.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1mq7j8)

## Brief Changelog

* Add `forcePoolId` to router interface
* Implement above in sidecar and legacy routers, for TFM, it returns 0 as out amount since their API does not support this
* Add above as parameter to query in pool detail page swap modal
* Hide route visualization if in modal

## Testing and Verifying

* All pool detail pages should force a swap through that page's pool
* The route visualization should be hidden on pool detail page trade token modals

### DIFF
--- a/packages/pools/src/router/routes.ts
+++ b/packages/pools/src/router/routes.ts
@@ -131,8 +131,23 @@ export class OptimizedRoutes implements TokenOutGivenInRouter {
 
   async routeByTokenIn(
     tokenIn: Token,
-    tokenOutDenom: string
+    tokenOutDenom: string,
+    forcePoolId?: string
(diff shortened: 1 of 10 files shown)


arispech (approved):
Working as expected

JoseRFelix (approved):

CryptoAssassin1:
Worked fine

kamal-sutra (commented):
Works fine

PR #2535, opened 2023-12-11
Astroport PCL pool

jonator:
This PR improves transmuter handling and serves as a reference for Astroport adding their custom cosmwasm pool.

For now, Astroport's pool will only be included in router but not in pools query.

Their code ID will need to be added once they deploy their contract to mainnet.


gnujoow (approved):

jonator:
Closing for now, and can reopen, but I think by January we'd like sidecar to support routing through cosmwasm pools via wasm execution. cc @p0mvn . In near term we'd like to deprecate our off chain TypeScript routing/simulation logic in favor of the sidecar implementation.

PR #2550, opened 2023-12-13
feat: Only Track Transfer When ETH TX is Successful and Do not Select Whole Input on Click

JoseRFelix:
### ClickUp Task

[Only Track Transfer When ETH TX is Successful](https://app.clickup.com/t/86a1np0h7)
[Do not Select Whole Input on Click](https://app.clickup.com/t/86a1nwern)

## Testing and Verifying

- [ ] If Tx fails, does not add the tx to track on the history table
- [ ] Clicking on input does not select the content
- [ ] Can bridge assets

### DIFF
--- a/packages/web/components/input/input-box.tsx
+++ b/packages/web/components/input/input-box.tsx
@@ -127,7 +127,6 @@ export const InputBox: FunctionComponent<Props> = ({
               onFocus && onFocus(e);
             }}
             onInput={(e: any) => onInput(e.target.value)}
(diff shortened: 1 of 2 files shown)


jonator (approved):

arispech (approved):
:white_check_mark: If Tx fails, does not add the tx to track on the history table
 :white_check_mark: Clicking on input does not select the content
  :white_check_mark:Can bridge assets

PR #2553, opened 2023-12-14
Add assets info tRPC procedure

jonator:

Adds the assets info tRPC function in prep for assets page rebuild. Also, I refactored the sort and search utils by moving them into the utils folder. I refactored the assets functions to map over a given list of assets and add additional info such as user data and market cap info. This makes them more composable and functional. The market cap info coming from data services is cached in memory for 5 minutes. 

### ClickUp Task

## Brief Changelog

* Add map asset market cap info function and queries
* Move sort and search into utils folder and add tests
* Remove sorting from getAssets function (not used)
* Update comments and improve type composition

## Testing and Verifying

* Swap tool still works as expected
* All tests continue to pass (including existing getAssets tests)


DavideSegullo (commented):
It seems good! So now we make a new request for each row? but only when they're visible (That's what I see from network tab)? At least to get some details like coingecko coin ecc., right?

jonator:
> It seems good! So now we make a new request for each row? but only when they're visible (That's what I see from network tab)? At least to get some details like coingecko coin ecc., right?

No we make requests in pages depending on the infinite query params (cursor, limit). React query will handle this state.

DavideSegullo (approved):

packages/web/utils/__tests__/compare.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { Dec, Int } from "@keplr-wallet/unit";
+
+import {
+  compareCommon,
+  compareDec,
+  compareDefinedMember,
+  CompareResult,
+} from "../compare";
+
+describe("compareCommon", () => {
```
JoseRFelix, line 10: Please take a look at these [tests](https://github.com/osmosis-labs/osmosis-frontend/blob/stage/packages/web/utils/__tests__/number.ts#L11) that leverage jest in case. It makes it much easier and readable to assert the same function with multiple values.

JoseRFelix (approved):

PR #2555, opened 2023-12-14
feat: Add asset base copy to clipboard 

DavideSegullo:

this PR adds a "copy to clipboard" button on the token detail page, in the section where the description of the displayed asset is located. The purpose of the button is to give the opportunity to view the raw denom of a token (useful for example for developers or those who want to do in-depth research)

### ClickUp Task


packages/web/components/token-details/token-details.tsx
```diff
@@ -118,11 +120,7 @@ const TokenDetails = ({
     }
   }, [coinGeckoId]);
 
-  const name = useMemo(() => {
-    if (details) {
-      return details.name;
-    }
-
+  const currency = useMemo(() => {
```
jonator, line 123: I know this is unrelated to the PR, but we have a handy `getAsset` tRPC function that can take any denom (base IBC denom or ticker symbol) and return currency info. This may be more scalable than importing the chainlist into the page.

I'd appreciate it if you could open a PR and use that in place of importing the chainlist anywhere the chainlist is used. LMK if you have any other questions. You can see an example of `getAsset` tRPC query in the `useSwap` hook.

packages/web/tailwind.config.js
```diff
@@ -220,6 +220,7 @@ module.exports = {
         earnpositions: "minmax(0, 1fr), 1px, minmax(0, 284px)",
       },
       width: {
+        25: "6.25rem",
```
jonator, line 223: In general, IMO we should avoid updating the global tailwind config to keep it lean and organized. I think it's OK to hardcode values if absolutely needed. Won't die on this hill though. Thoughts?

packages/web/components/token-details/token-details.tsx
DavideSegullo: I'll open a new PR for that, thank you!

packages/web/tailwind.config.js
DavideSegullo: Hardcode values can be fine but we should avoid them, having everything in configuration allows us to make changes without having to check all the files. I think we will generally solve this problem when we have a stable design system configuration (and consequently tailwind as well).

PR #2556, opened 2023-12-15
Add tRPC panel (swagger for tRPC)

jonator:

[TPRC Panel GitHub](https://github.com/iway1/trpc-panel)

Adds a local dev tool that can be visited by going to /api/panel in localhost. It allows devs to run tRPC functions from a user friendly UI, similar to swagger. There's no build step, it generates the UI on the fly from our tRPC app router.

Unfortunately, for some reason it's unable to generate input UI for zod objects that have been composed together, such as our use of `GetInfiniteAssetsInputSchema`. But I think it's worth decomposing those for each procedure so we can use this tool.

Screenshot 2023-12-14 at 7 07 46 PM

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -116,12 +116,12 @@
     "sha.js": "^2.4.11",
     "sharp": "^0.30.4",
     "superjson": "^2.2.0",
+    "trpc-panel": "^1.3.4",
     "utility-types": "^3.10.0",
     "web3-utils": "^1.7.4",
     "zod": "^3.22.4"
   },
   "devDependencies": {
-    "tsx": "^4.6.2",
     "@simbathesailor/use-what-changed": "^2.0.0",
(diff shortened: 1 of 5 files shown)


packages/web/package.json
```diff
@@ -116,12 +116,12 @@
     "sha.js": "^2.4.11",
     "sharp": "^0.30.4",
     "superjson": "^2.2.0",
+    "trpc-panel": "^1.3.4",
```
JoseRFelix, line 119: Should this be a dev dependency?

jonator:
I'm not sure this is going to be useful. Not only does it not handle composed input zod types, but it doesn't handle optional objects with required members. It thinks the child required member is required even though the containing object is optional and blocks input. We can keep an eye on the project and add it if they make improvements to the zod input object parsing.

Issue #2518, opened 2023-12-08
Automatically deduct fees when swapping with MAX [100 USDC reward!] 


MaxMillington:
🐛  Bug Bounty: Fix Needed! 🛠️ 
Reward: :trophy: 100 USDC 🏆 

We are excited to announce a bug bounty for third-party developers! If you've got the skills and know-how, here's your chance to earn some 💸  and contribute to our project.

Issue Description 📄
Automatically deduct fees when swapping MAX

Expected Behavior 💡
As a user, fees should be deducted when I'm swapping assets and select the 'MAX' option

Current Behavior ❌
Currently, fees are not deducted

Requirements for the Bounty 🏁

 Submit a pull request with a verified fix to the issue.
 The fix should not introduce any new bugs or vulnerabilities.
 Code should follow our project's style and contribution guidelines.
 Your solution will be reviewed by our team, and the bounty will be awarded upon approval.
How to Claim Your Bounty 💰

 Fork our repository.
 Create a new branch for your fix.
 Submit a pull request to our main branch.
 Include detailed information about your fix and how it addresses the issue.

 Once reviewed and merged, the 100 USDC will be transferred to your wallet.
We look forward to your contributions and thank you for helping improve our project!

**Please Note**: All bounty claimants must undergo a KYC (Know Your Customer) process with our finance team to receive the reward. This is a mandatory step to ensure compliance with financial regulations.

BalthazarB:
attempting this

HarjobandeepSingh:
@MaxMillington in local project swap function is not working then how i can test and debug it 
![image](https://github.com/osmosis-labs/osmosis-frontend/assets/81474463/7aa345d0-5457-4cca-85a8-36cc9966adc1)

MaxMillington:
@HarjobandeepSingh Did you run yarn, yarn build, and then yarn run dev? It should work on mainnet locally. Make sure to pull the latest changes.

HarjobandeepSingh:
> @HarjobandeepSingh Did you run yarn, yarn build, and then yarn run dev? It should work on mainnet locally. Make sure to pull the latest changes. 

Yes i did

HarjobandeepSingh:
> @HarjobandeepSingh Did you run yarn, yarn build, and then yarn run dev? It should work on mainnet locally. Make sure to pull the latest changes.

i tried again but its still not working @MaxMillington

HarjobandeepSingh:
[...]

PR #2558, opened 2023-12-15
Fix/ Automatically deduct fees when swapping with MAX ##2518

HarjobandeepSingh:

This pull request aims to fix the issue where fees are not automatically deducted when a user selects the 'MAX' option during asset swapping.

## Brief Changelog

- Modified the calculation of `osmoOutAmount` in `add-liquidity.ts` to always deduct the `OSMO_MEDIUM_TX_FEE` from the balance before comparing it with the out amount.

## Testing and Verifying

This change has been tested locally by running the asset swapping functionality and verifying that the fee is correctly deducted when the 'MAX' option is selected.

- This pull request introduces a change in the user-facing behavior. Previously, fees were not deducted when the 'MAX' option was selected during asset swapping. With this change, the fee will be automatically deducted.


HarjobandeepSingh:
fix/ #2518

MaxMillington:
@HarjobandeepSingh looks like the build is failing here due to a syntax error? Are you able to build this locally?

jonator (changes requested):
This appears to be making changes to the add liquidity state, and not swapping?

HarjobandeepSingh:
> This appears to be making changes to the add liquidity state, and not swapping?

@jonator check again plz

HarjobandeepSingh:
@MaxMillington sir plz check now

HarjobandeepSingh:
@jonator plz check

jonator:
@HarjobandeepSingh this is not close. Swap modal changes would likely be in `useSwap` hook. And would only apply if it's the staking currency

PR #2559, opened 2023-12-15
Add Skip bridge integration

thal0x:

Adds bridge integration using the Skip API.

### ClickUp Task

## Brief Changelog

- Add Skip bridge integration

## Testing and Verifying


packages/web/integrations/bridges/skip/queries.ts
```diff
@@ -0,0 +1,96 @@
...
+      include_evm_assets: "true",
+    });
+
+    if (chainID) {
+      urlParams.set("chain_id", chainID);
+    }
+
+    const response = await fetch(
+      `https://api.skip.money/v1/fungible/assets?${urlParams.toString()}`
+    );
```
JoseRFelix, line 23: I recommend using the [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) class to compose the URL since it includes the search params object internally and guarantees that the url is valid. 
```suggestion
    const url = new URL("/v1/fungible/assets", "https://api.skip.money")
    url.searchParams.append("include_evm_assets", "true")

    if (chainID) {
     url.searchParams.set("chain_id", chainID)
    }

    const response = await fetch(url.toString());
```

packages/web/integrations/bridges/skip/queries.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  async chains() {
+    const urlParams = new URLSearchParams({
+      include_evm: "true",
+    });
+
+    const response = await fetch(
+      `https://api.skip.money/v1/info/chains?${urlParams.toString()}`
+    );
```
JoseRFelix, line 39: Same as above

packages/web/integrations/bridges/skip/queries.ts
```diff
@@ -0,0 +1,96 @@
...
+      urlParams.set("chain_id", chainID);
+    }
+
+    const response = await fetch(
+      `https://api.skip.money/v1/fungible/assets?${urlParams.toString()}`
+    );
+
+    const data = (await response.json()) as {
+      chain_to_assets_map: Record<string, { assets: Asset[] }>;
+    };
```
JoseRFelix, line 27: Please take a look at our [apiClient](https://github.com/osmosis-labs/osmosis-frontend/blob/10bc2e754a7af647e5be3f5be67b8f3b73bcd5ef/packages/web/server/queries/cosmos/rpc-status.ts#L51) that will handle response types and fetch deserialization for us.

packages/web/integrations/bridges/skip/types.ts
```diff
@@ -0,0 +1,194 @@
+export type Asset = {
```
JoseRFelix, line 1: Can you rename this to `SkipAsset`? I'd like to avoid asset type collision between Osmosis and Skip

packages/web/integrations/bridges/skip/types.ts
```diff
@@ -0,0 +1,194 @@
...
+  name?: string;
+  logo_uri?: string;
+  decimals?: number;
+  token_contract?: string;
+  description?: string;
+  coingecko_id?: string;
+  recommended_symbol?: string;
+};
+
+export type Chain = {
```
JoseRFelix, line 20: Same here, but it would be called 
```suggestion
export type SkipChain = {
```

packages/web/integrations/bridges/skip/types.ts
```diff
@@ -0,0 +1,194 @@
...
+  chain_id: string;
+};
+
+export type SwapOperation = {
+  pool: string;
+  denom_in: string;
+  denom_out: string;
+};
+
+export type AxelarTransfer = {
```
JoseRFelix, line 114: Same as above
```suggestion
export type SkipAxelarTransfer = {
```

packages/web/integrations/bridges/skip/types.ts
```diff
@@ -0,0 +1,194 @@
...
+  chain_id: string;
+  pfm_enabled: boolean;
+  cosmos_sdk_version: string;
+  supports_memo: boolean;
+  logo_uri?: string;
+  bech32_prefix: string;
+  chain_type: string;
+};
+
+export type RouteRequestBase = {
```
JoseRFelix, line 31: What do you think if we prefix all the types in this file with `Skip`?

packages/web/pages/api/bridge-transfer/quotes/[bridge].ts
```diff
@@ -39,6 +39,8 @@ export default async function quoteByBridge(
     });
   }
 
+  console.log("quoteByBridge: ", req.url);
```
Comment (JoseRFelix) [inline packages/web/pages/api/bridge-transfer/quotes/[bridge].ts:42]: Are these console logs expected?

JoseRFelix (changes requested):
Thank you for your valuable contribution! 🚀 I've provided some comments, primarily aimed at enhancing code readability. Additionally, I will pass this PR on to the QA team for testing.

packages/web/pages/api/bridge-transfer/quotes/[bridge].ts
Comment (thal0x) [reply packages/web/pages/api/bridge-transfer/quotes/[bridge].ts]: No sorry, was trying to debug something. Removed now!

packages/web/integrations/bridges/skip/types.ts
thal0x: Done

thal0x:
@JoseRFelix all feedback addressed, let me know if you need anything else

packages/web/hooks/use-feature-flags.ts
```diff
@@ -46,5 +46,6 @@ export const useFeatureFlags = () => {
       : launchdarklyFlags.notifications,
     _isInitialized: isInitialized,
     _isClientIDPresent: !!process.env.NEXT_PUBLIC_LAUNCH_DARKLY_CLIENT_SIDE_ID,
+    multiBridgeProviders: true,
```
JoseRFelix, line 49: I believe this leaked from the dev environment.

JoseRFelix (commented):
This looks great! I've left a comment for your review. Once we receive feedback from the QA team, we can proceed with merging.

JoseRFelix:
@thal0x Could you merge the changes from the stage branch? It includes some transaction history improvements. There should be no breaking changes.

arispech:
I just successfully deposited and withdrew MATIC through Skip, however, both tx did not show up on the FE transfer history table

CryptoAssassin1:
There is $70.6 transfer fee using skip and on Axelar its just $4.31
Screenshot from 2023-12-20 01-18-33
Screenshot from 2023-12-20 01-18-45

CryptoAssassin1:
Something is wrong with the Skip fees changing every min and giving more than triple fees check the comparison with Axelar fees.
AVAX
Skip:$70.6
Axelar:$4.31
ETH
Skip:$119
Axelar:$82
USDC.axl
Skip:$209
Axelar:$60.1

Screenshot from 2023-12-20 01-33-55
Screenshot from 2023-12-20 01-31-59

bpiv400:
It may be that we're using GMP and Axelar is not. Looking into it

thal0x:
> Something is wrong with the Skip fees changing every min and giving more than triple fees check the comparison with Axelar fees. AVAX Skip:$70.6 Axelar:$4.31 ETH Skip:$119 Axelar:$82 USDC.axl Skip:$209 Axelar:$60.1
> 
> Screenshot from 2023-12-20 01-33-55 Screenshot from 2023-12-20 01-31-59

After doing some investigation, there must be some difference in the implementations. They might be using deposit addresses, while ours uses GMP. Regardless fees on Avalanche have been extremely high lately, so we think this is correct on our end.

CryptoAssassin1:
@thal0x i let USDC.axl window open and the transfer fee jumped from $115 to $158 in few mins and then back to $112 and yesterday it was $200+ and Axelar fee is just $60

thal0x:
@CryptoAssassin1 yes the fee will be variable depending on how congested the relevant chain is

bpiv400:
Hi folks, 

I just wanted to give some context on how Axelar calculates fees and why you're observing differences right now and why you observe variability over time. 

The axelar gas fee computation is opaque but it depends on several factors: 
* Source chain 
* Destination chain 
* The token being transferred
* Gas cost on the source chain (This has been very high and variable on Avalanche recently, which is why you see the Skip fee changing)
* How much of a buffer the caller provides to (We provide a 1.1x multiplier on estimated gas by default since we've noticed sometimes if you don't buffer it, the transaction will just fail because the gas cost on the source chain will change. This is also what Axelar recommends.)  
* Whether the caller is using GMP (i.e. `callContractWithToken`) or transfer (i.e. `sendToken`) 

It's likely that the primary difference you're observing between Axelar's reported fees and the reported fees from Skip is a result of this last factor: We're always using `callContractWithToken`, rather than `sendToken`). 

Sometimes `callContractWithToken` is more expensive, sometimes `sendToken` is more expensive. It's not clear why this is the case, again because the fee logic is variable. We use `callContractWithToken` in all cases right now because we've observed roughly that it seems cheaper the majority of the time, though Axelar may change fees at whim behind the scenes for either or both (and we've seen them do this a few times). (We also prefer `callContractWithToken` because it allows us to swap for destination gas or perform a swap to another token upon arrival, which can be useful for more complex integrations) 

The most ideal solution here is for the Skip API to check both `callContractWithToken` and `sendToken` (at least when `sendToken` is a feasible choice) in real time each time a user requests a route. This is planned on our roadmap for January, which should bring our estimates in line the vast majority of the time. But as noted above, due to gas cost variability, you'll still see the fee fluctuate over time.

bpiv400:
Bumping this -- anything needed to get this merged?

JoseRFelix (changes requested):
@bpiv400 @thal0x I think we're lacking a 'transfer-status-source' class to monitor transactions in the history table. Could you please check out this [implementation](https://github.com/thal0x/osmosis-frontend/blob/9aed072c8ba4712088268417f84f12d506f2d826/packages/web/integrations/bridges/squid/squid-transfer-status-source.ts#L15) for reference examples?

CryptoAssassin1:
Deposited Matic successfully and its still showing Pending.
Screenshot from 2024-01-04 00-32-58

thal0x:
@CryptoAssassin1 could you send me the transaction hash, this may be an API issue. We will investigate

JoseRFelix:
@thal0x https://axelarscan.io/gmp/0x733cf1ed530b4f600d617d2ac83c8a666f74d350a06d2460c8cf36b4d15b6673

thal0x:
@JoseRFelix @CryptoAssassin1 Ok looks like this was an API issue, we're looking into what caused this.

CryptoAssassin1:
@thal0x Thank you

JoseRFelix:
@thal0x Please notify us as soon as the bug is resolved.  We would like to promptly re-test it

thal0x:
@JoseRFelix the bug has been deployed so everything should be good.

Just a heads up, we have been encountering issues with the IBC congestion on Osmosis lately as I'm sure you're all aware

JoseRFelix:
@CryptoAssassin1 This is ready for a re-review

CryptoAssassin1 (approved):
Looking good.
Screenshot from 2024-01-09 18-16-06

JoseRFelix (approved):
Thank you @thal0x for the contribution!

PR #2560, opened 2023-12-15
Move APR breakdown behind FF

jonator:

https://app.launchdarkly.com/default/development/features/apr-breakdown/targeting

### ClickUp Task

### DIFF
--- a/packages/web/components/complex/all-pools-table.tsx
+++ b/packages/web/components/complex/all-pools-table.tsx
@@ -22,6 +22,7 @@ import {
   useRef,
   useState,
 } from "react";
+import { ReactNode } from "react";
 
 import { Icon } from "~/components/assets";
 import { PaginatedTable } from "~/components/complex/paginated-table";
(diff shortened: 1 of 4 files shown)


packages/web/components/complex/all-pools-table.tsx
```diff
@@ -463,14 +465,12 @@ export const AllPoolsTable: FunctionComponent<{
...
-                        apr={pool.apr}
-                      />
-                    ) : (
+                    ) : Boolean(pool.concentratedPoolDetail) &&
+                      flags.aprBreakdown ? (
+                      <ClAprBreakdownCell poolId={pool.queryPool.id} />
+                    ) : flags._isInitialized ? (
                       pool.apr.toString()
-                    )
+                    ) : null
```
JoseRFelix, line 473: I think this conditional is a little difficult to understand because it's too nested with ternaries. What are your thoughts on refactoring it like this: 

```tsx
let value: ReactNode | null;
if (isAPRTooHigh) {
  // Only display warning when APR is too high
  value = (
    <Tooltip
      className="w-5"
      content={t("highPoolInflationWarning")}
    >
      <p className="flex items-center gap-1.5">
        <Icon
          id="alert-triangle"
          className="h-4 w-4 text-osmoverse-400"
        />
        {pool.apr.toString()}
      </p>
    </Tooltip>
  );
} else if (Boolean(pool.concentratedPoolDetail) && flags.aprBreakdown) {
  value = <ClAprBreakdownCell poolId={pool.queryPool.id} />;
} else if (flags._isInitialized) {
  value = pool.apr.toString();
} else {
  value = null;
}

 <MetricLoaderCell
                  isLoading={
                    queriesOsmosis.queryIncentivizedPools.isAprFetching
                  }
                  value={value} />
```

JoseRFelix (approved):

PR #2571, opened 2023-12-16
New assets table (WIP)

jonator:

The new assets page is behind a FF: `new-assets-table`.

This PR adds the initial implementation, which is subject to change. But, it makes many of the tRPC function modifications and query additions needed to get the data for the table. Also, search, sort, and price chart duration selection has been added.

As I discuss with the design team, we will add collapsed/canonical assets, deposit withdraw, as well as mobile UI.

### ClickUp Task

## Brief Changelog

* Some refactoring of assets server functions to make asset-related functions more composable.
* Some refactoring/cleanup of utilities, such as those for sorting and searching from tRPC functions
* Moved entire assets page behind FF, moved old code
* New assets table. Uses react table with virtualization and useInfiniteQuery.


packages/web/components/table/asset-info.tsx
```diff
@@ -0,0 +1,669 @@
...
+    {
+      enabled: !isLoadingWallet,
+      getNextPageParam: (lastPage) => lastPage.nextCursor,
+      initialCursor: 0,
+    }
+  );
+  const assetsData = useMemo(() => {
+    const assets = assetPagesData?.pages.flatMap((page) => page?.items) ?? [];
+    if (selectedView === "myTokens")
+      return assets.filter((a) => a.amount?.toDec().isPositive());
```
JoseRFelix, line 116: Is there any reason why we are not doing this in the query itself?

packages/web/components/table/asset-info.tsx
```diff
@@ -0,0 +1,669 @@
...
+    enableFilters: false,
+    getCoreRowModel: getCoreRowModel(),
+    filterFns: {
+      /**
+       * these filters, even though they are not used in this table instance,
+       * are necessary to suppress errors derived by the "@tanstack/table-core"
+       * module declaration in the earn page.
+       *
+       * @fabryscript
+       */
```
JoseRFelix, line 227: We should look into why we need this. Perhaps we can upgrade @tanstack/table

packages/web/components/table/asset-info.tsx
```diff
@@ -0,0 +1,669 @@
...
+  row: {
+    original: { coinDenom, priceChange24h },
+  },
+  timeFrame,
+}) => {
+  const { data: recentPrices } =
+    api.edge.assets.getAssetHistoricalPrice.useQuery({
+      coinDenom,
+      timeFrame,
+    });
```
JoseRFelix, line 466: What are your thoughts on adding a stale time of 30 seconds to avoid fetching constantly while user scrolls

packages/web/components/table/asset-info.tsx
```diff
@@ -0,0 +1,669 @@
...
+      width={80}
+      height={50}
+      lineWidth={2}
+      data={recentPriceCloses}
+      color={
+        isBullish
+          ? theme.colors.bullish[400]
+          : isBearish
+          ? theme.colors.ammelia[400]
+          : theme.colors.wosmongton[200]
```
JoseRFelix, line 490: Thoughts on using an `if` statement?

```suggestion
let color: string;
if (isBullish) {
  color = theme.colors.bullish[400];
} else if (isBearish) {
  color = theme.colors.ammelia[400];
} else {
  color = theme.colors.wosmongton[200];
}
```

packages/web/components/table/asset-info.tsx
```diff
@@ -0,0 +1,669 @@
...
+
+  const [searchInput, _, setSearchInput, searchQueryInput] =
+    useSearchQueryInput();
+
+  // Pass search query in an effect to prevent rendering the entire table on every input change
+  // Only on debounced search query input
+  useEffect(
+    () => setSearchQuery(searchQueryInput),
+    [setSearchQuery, searchQueryInput]
+  );
```
JoseRFelix, line 630: If the `searchQueryInput` is debounced, why do we need this `useEffect`?

packages/web/components/table/asset-info.tsx
```diff
@@ -0,0 +1,669 @@
...
+}> = ({
+  selectedTimeFrame,
+  setSelectedTimeFrame,
+  setSearchQuery,
+  selectedView,
+  setSelectedView,
+}) => {
+  const { t } = useTranslation();
+
+  const [searchInput, _, setSearchInput, searchQueryInput] =
```
JoseRFelix, line 622: Thoughts on making this an object instead of an array? It's faster to get the properties through autocomplete than Typescript hover.

packages/web/hooks/input/use-search-query-input.ts
```diff
@@ -0,0 +1,32 @@
...
+  // get selectable currencies for trading, including user balances if wallect connected
+  const [searchInput, setSearchInput] = useState<string>("");
+
+  // generate debounced search from user inputs
+  const [debouncedSearchInput, setDebouncedSearchInput] =
+    useDebouncedState<string>("", debounceMs);
+  useEffect(() => {
+    const sanitizedSearchInput = searchInput.replace(/#/g, "").trim();
+    setDebouncedSearchInput(sanitizedSearchInput);
+  }, [setDebouncedSearchInput, searchInput]);
```
JoseRFelix, line 19: Is there any reason why we are using an effect instead of creating a callback function wrapping setDebouncedSearchInput?

packages/web/hooks/user-settings/use-user-favorite-asset-denoms.ts
```diff
@@ -0,0 +1,33 @@
+import { useCallback } from "react";
+
+import { useLocalStorageState } from "../window/use-localstorage-state";
+
+export function useUserFavoriteAssetDenoms() {
+  const [favoritesList, onSetFavoritesList] = useLocalStorageState(
```
JoseRFelix, line 6: Since `onSetFavoritesList` is not an event, but a set function I believe we should remove the `on`
```suggestion
  const [favoritesList, setFavoritesList] = useLocalStorageState(
```

packages/web/hooks/user-settings/use-user-favorite-asset-denoms.ts
```diff
@@ -0,0 +1,33 @@
...
+    "favoritesList",
+    ["OSMO", "ATOM", "TIA"]
+  );
+
+  const addFavoriteDenom = useCallback(
```
JoseRFelix, line 11: Since this is an event callback based on an action by the user, what are your thoughts on appending `on`?

```suggestion
  const onAddFavoriteDenom = useCallback(
```

packages/web/hooks/user-settings/use-user-favorite-asset-denoms.ts
```diff
@@ -0,0 +1,33 @@
...
+    (denom: string) => {
+      if (favoritesList.includes(denom)) return;
+      onSetFavoritesList([...favoritesList, denom]);
+    },
+    [favoritesList, onSetFavoritesList]
+  );
+
+  const removeFavoriteDenom = useCallback(
```
JoseRFelix, line 19: Same as above

packages/web/hooks/window/use-window-size.ts
```diff
@@ -10,6 +8,21 @@ export interface WindowSize {
...
+ */
+export const enum Breakpoint {
+  SM = 640,
+  MD = 768,
+  LG = 1024,
+  XLG = 1152,
+  XL = 1280,
+  XLHALF = 1408,
+  XXL = 1536,
+  XXXL = 1792,
```
JoseRFelix, line 23: Thoughts on making this lowercase?
```suggestion
  sm = 640,
  md = 768,
  lg = 1024,
  xlg = 1152,
  xl = 1280,
  xlhalf = 1408,
  xxl = 1536,
  xxxl = 1792,
```

packages/web/server/queries/complex/assets/price.ts
```diff
@@ -236,3 +241,86 @@ export async function calcAssetValue({
...
+  } else if (timeFrame === "1M") {
+    numRecentFrames = 30; // Last month of prices with bars as 1 day
+  }
+
+  return getAssetHistoricalPrice({
+    coinDenom,
+    timeFrameMinutes,
+    numRecentFrames,
+  });
+}
```
JoseRFelix, line 326: What are your thoughts on merging this function with `getAssetHistoricalPrice` and handle these timeframes in a conditional? Handling nested functions can be quite taxing to understand.

packages/web/utils/__tests__/compare.spec.ts
```diff
@@ -52,3 +57,41 @@ cases(
...
+    const b = { member: undefined };
+    expect(compareDefinedMember(a, b, "member")).toBe(-1);
+  });
+
+  it("should return 0 if a and b are nil values", () => {
+    const a = { member: null };
+    const b = { member: undefined };
+    expect(compareDefinedMember(a, b, "member")).toBe(0);
+  });
+});
```
JoseRFelix, line 97: Thoughts on using jest-in-case?

JoseRFelix (commented):
Nice job! This is looking very clean. Left some comments

packages/web/components/table/asset-info.tsx
jonator: I talked to Davide about it last week, it relates to a TypeScript module override when using the filter options. He said he will open a PR later to fix.

DavideSegullo: Yes, it's due to tanstack table, I'll fix that in a different PR, we need to override their types. I also opened a PR to them https://github.com/TanStack/table/pull/5215

DavideSegullo: here the PR: https://github.com/osmosis-labs/osmosis-frontend/pull/2613

sunnya97:
The font feels a bit off.  Like it's more bold than it should be.

sunnya97:
The market rank isn't supposed to be rank on Osmosis, but rather global crypto-wide market cap (just use coingecko's)

jonator:
> The market rank isn't supposed to be rank on Osmosis, but rather global crypto-wide market cap (just use coingecko's)

What about tokens not on CoinGecko? Fall back to imperator?

packages/web/components/table/asset-info.tsx
jonator: As the comment states, this prevents rerenders on the whole table as the user types each character, but rather only rerenders the table when the debounce expires.

packages/web/hooks/input/use-search-query-input.ts
jonator: Thanks for noticing this😅, seems I got carried away with some of the effects in this one

packages/web/components/table/asset-info.tsx
jonator: Merged stage into this branch, and removed these filters.

packages/web/utils/__tests__/compare.spec.ts
jonator: My only opposition to it is it breaks my Jest UI extension for each of the test cases, however it does make tests easier to add to and update, so I'll make the change.

packages/web/components/table/asset-info.tsx
JoseRFelix: I suggest that we eliminate the `useSearchQueryInput` here and instead, have the parent component utilize `useSearchQueryInput`

packages/web/hooks/use-swap.ts
```diff
@@ -316,20 +315,12 @@ export function useSwapAssets({
...
-
-  const inputSearch = useMemo(
-    () => (debouncedSearchInput ? { query: debouncedSearchInput } : undefined),
-    [debouncedSearchInput]
-  );
+  const {
+    searchInput: assetsQueryInput,
+    setSearchInput: setAssetsQueryInput,
+    queryInput,
+  } = useSearchQueryInput();
```
JoseRFelix, line 323: Thoughts on using `useDebouncedState` directly here instead of creating a separate hook for the search query input? My concern is that the separate hook might be overly complex for what we need to achieve: a denounced query state

packages/web/hooks/use-swap.ts
```diff
@@ -340,11 +331,11 @@ export function useSwapAssets({
   const { data: searchAssets, isLoading: isLoadingSearchAssets } =
     api.edge.assets.getAssets.useQuery(
```
JoseRFelix, line 332: I suggest combining this query with the useInfiniteQuery to create a single query. By doing so, we would have less code, a simpler implementation, and we could take advantage of the benefits offered by paginated search

packages/web/hooks/use-swap.ts
```diff
@@ -518,16 +509,16 @@ function useSwapAsset<TAsset extends Asset>(
   );
   const queryEnabled =
     Boolean(minDenomOrSymbol) && !isLoadingWallet && !existingAsset;
```
JoseRFelix, line 511: I believe using a Boolean for this purpose might lead to an unusual edge case. For instance, if a user searches for the string "0", the result would incorrectly be interpreted as false

```suggestion
    !isNil(minDenomOrSymbol) && typeof minDenomOrSymbol === "string" && !isLoadingWallet && !existingAsset;
```

packages/web/hooks/use-swap.ts
jonator: Can we merge this and I address this in a future small PR?

packages/web/components/table/asset-info.tsx
```diff
@@ -0,0 +1,653 @@
...
+
+              if (isFavorite) onRemoveFavorite();
+              else onSetFavorite();
+            }}
+            height={24}
+            width={24}
+          />
+        </div>
+      ) : (
+        <div style={{ height: 24, width: 24 }} />
```
JoseRFelix, line 388:
```suggestion
        <div className="h-[24px] w-[24px]" />
```

JoseRFelix (approved):
Left some nits, but other than that LGTM!

JoseRFelix (approved):
Left some nits, but other than that LGTM!

PR #2608, opened 2023-12-21
fix: Withdraw UI on Mobile

JoseRFelix:

For mobile users who cannot use Metamask, provide an option to manually enter their wallet address, enabling them to transfer assets directly to this address.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1g2bf8)

## Testing and Verifying

- [ ] Can transfer to manually input evm address
- [ ] Deposit button should be hidden in mobile
- [ ] Can still transfer on desktop

### DIFF
--- a/packages/web/modals/pre-transfer.tsx
+++ b/packages/web/modals/pre-transfer.tsx
@@ -10,8 +10,8 @@ import { TokenSelect } from "~/components/control";
 import { UNSTABLE_MSG } from "~/config";
 import { useTranslation } from "~/hooks";
 import { useWindowSize } from "~/hooks";
+import { useCoinFiatValue } from "~/hooks/queries/assets/use-coin-fiat-value";
 import { ModalBase, ModalBaseProps } from "~/modals";
-import { useStore } from "~/stores";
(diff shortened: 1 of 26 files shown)


packages/web/modals/pre-transfer.tsx
```diff
@@ -41,16 +42,18 @@ export const PreTransferModal: FunctionComponent<
   const { t } = useTranslation();
 
   const tokenValue = priceStore.calculatePrice(
```
jonator, line 44: Thoughts on using `useCoinFiatValue` instead of price store here? Would be nice, going forward, to remove all refs to price/query stores.

JoseRFelix: I was not aware this hook existed. I have added this on my latest commit. Thank you!

arispech (approved):
Can transfer to manually input evm address :white_check_mark:
 Deposit button should be hidden in mobile :white_check_mark:
 Can still transfer on desktop :white_check_mark:

MaxMillington (approved):

PR #2615, opened 2023-12-22
Publish Stage

jonator:
* Fix ko.json
* Small fix to swap tool half/max logic


DavideSegullo (commented):
Everything seems okay to me, but I just notice one thing in the new assets table, if I do a search using a full word like "bitsong" it doesn't find any results, whereas if I type in "bit" it finds it right away.
Same thing if I try to search as "BTSG", I don't know how many other assets this applies to

sunnya97:
Have a lot of QA stuff on new asset page, but probably okay to merge since its behind FF

sunnya97 (approved):

jonator:
@DavideSegullo @sunnya97 yep assets page is very much a WIP, but I made some minor fixes/cleanup to code elsewhere, even though the assets page is behind FF

PR #2616, opened 2023-12-22
fix: Assets That do not Have an IBC Trace

JoseRFelix:

Previously, we considered an empty `traces` array a native Osmosis token. Unfortunately, there will be assets that are native to Osmosis and also have a trace like `liquid-stake`. This PR finds the last ibc denom instead of assuming it's the last entry.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1raryg)

## Testing and Verifying

- [ ] Should be able to do IBC transfers
- [ ] Balances should be computed correctly
- [ ] All listed assets should be visible


JeremyParish69 (approved):
This is desired and will allow us to remove otherwise needless overrides.

jonator (approved):

CryptoAssassin1 (approved):

PR #2624, opened 2023-12-28
Add default feature flags for development mode

daniel-farina:

This pull request modifies the `useFeatureFlags` hook to only enable certain feature flags and set `_isInitialized` to `true` when running in a development environment.

### ClickUp Task

## Brief Changelog

- Modified the `useFeatureFlags` hook to conditionally spread `defaultFlags` and set `_isInitialized` based on the `NODE_ENV` environment variable.

## Testing and Verifying

This change has been tested locally by running the application in both development and production environments and verifying that the feature flags and `_isInitialized` are set as expected. It has not been tested on stage.

- Does this pull request introduce a new feature or user-facing behavior changes? No

### DIFF
--- a/packages/web/hooks/use-feature-flags.ts
+++ b/packages/web/hooks/use-feature-flags.ts
@@ -69,7 +69,7 @@ export const useFeatureFlags = () => {
     notifications: isMobile
       ? launchdarklyFlags.mobileNotifications
       : launchdarklyFlags.notifications,
-    _isInitialized: process.env.NODE_ENV === "development",
+    _isInitialized: process.env.NODE_ENV === "development" || isInitialized,
     _isClientIDPresent: !!process.env.NEXT_PUBLIC_LAUNCH_DARKLY_CLIENT_SIDE_ID,
   } as Record<ModifiedFlags, boolean>;
 };
--- a/packages/web/hooks/use-feature-flags.ts
+++ b/packages/web/hooks/use-feature-flags.ts
@@ -28,6 +28,29 @@ type ModifiedFlags =
   | "_isInitialized"
   | "_isClientIDPresent";
 
+// These default flags will be passed to the app if the app is running in development mode
+const defaultFlags: Record<ModifiedFlags, boolean> = {
+  concentratedLiquidity: true,
+  staking: false,
+  swapsAdBanner: true,
+  notifications: true,
+  convertToStake: true,
+  upgrades: true,
+  tokenInfo: true,
+  newAssetsTable: false,
+  sidebarOsmoChangeAndChart: true,
+  multiBridgeProviders: true,
+  unlistedAssets: false,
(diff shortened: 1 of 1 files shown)


daniel-farina:
I added `|| isInitialized` to keep the same logic as before

packages/web/hooks/use-feature-flags.ts
```diff
@@ -42,10 +65,11 @@ export const useFeatureFlags = () => {
 
   return {
     ...launchdarklyFlags,
+    ...(process.env.NODE_ENV === "development" ? defaultFlags : {}),
```
jonator, line 68: Thanks for this! This will help external contributors.

We have a dev env for LaunchDarkly that allows us to test FFs in local development. Perhaps we could also check to see if `NEXT_PUBLIC_LAUNCH_DARKLY_CLIENT_SIDE_ID` is env var is populated, and if it is, don't use default flags.

Also, we should probably put this spread operator after `notifications` below so that it get's overridden as well.

packages/web/hooks/use-feature-flags.ts
```diff
@@ -28,6 +28,29 @@ type ModifiedFlags =
...
+  multiBridgeProviders: true,
+  unlistedAssets: false,
+  earnPage: false,
+  sidecarRouter: true,
+  legacyRouter: true,
+  tfmRouter: true,
+  osmosisUpdatesPopUp: false,
+  aprBreakdown: true,
+  _isInitialized: false,
+  _isClientIDPresent: false,
```
jonator, line 51: These two can be removed

packages/web/hooks/use-feature-flags.ts
```diff
@@ -28,6 +28,29 @@ type ModifiedFlags =
   | "_isInitialized"
   | "_isClientIDPresent";
 
+// These default flags will be passed to the app if the app is running in development mode
```
jonator, line 31: Would also be nice to add a note that these defaults are for external contributors. Team members should prefer the `NEXT_PUBLIC_LAUNCH_DARKLY_CLIENT_SIDE_ID` with the dev ID set so that they can test with LD itself.

jonator (changes requested):

packages/web/hooks/use-feature-flags.ts
MaxMillington: yea i think we want if `!NEXT_PUBLIC_LAUNCH_DARKLY_CLIENT_SIDE_ID` && `process.env.NODE_ENV === 'development` the use defaultFlags. Otherwise LD wont have control over dev env

MaxMillington (changes requested):

MaxMillington:
Fixed [here](https://github.com/osmosis-labs/osmosis-frontend/pull/2749)

PR #2627, opened 2024-01-02
Use tRPC for sidebar OSMO asset price data

jonator:

* Use tRPC for getting OSMO currency, asset info (price change) and historical price data. This allows us to leverage fine-grained caching on the edge functions thereby reducing direct data queries.
* Remove/reduce dependency on current chain, price, and queries stores
* Remove adding buy OSMO clicked metadata to amplitude, as we currently aren't querying that [event](https://app.amplitude.com/data/osmosis-zone/Osmosis%20%F0%9F%A7%AA/events/main/latest/Sidebar%3A%20Buy%20OSMO%20clicked?view=All&tab=USED%2520BY&propertyValidityFilter=All%2520Properties). We can re-add metadata if we want in the future, at a time where our tRPC query infra is more mature.

### DIFF
--- a/packages/web/components/navbar-osmo-price.tsx
+++ b/packages/web/components/navbar-osmo-price.tsx
@@ -1,5 +1,4 @@
 import { WalletStatus } from "@cosmos-kit/core";
-import { CoinPretty, Dec, DecUtils } from "@keplr-wallet/unit";
 import classNames from "classnames";
 import { observer } from "mobx-react-lite";
 import Image from "next/image";
@@ -20,10 +19,10 @@ import {
 import { FiatOnrampSelectionModal } from "~/modals";
 import { useStore } from "~/stores";
 import { theme } from "~/tailwind.config";
-import { getLastDayChartData } from "~/utils/chart";
+import { api } from "~/utils/trpc";
 
 const NavbarOsmoPrice = observer(() => {
-  const { accountStore, priceStore, chainStore, assetsStore } = useStore();
+  const { accountStore, chainStore } = useStore();
   const { t } = useTranslation();
   const { logEvent } = useAmplitudeAnalytics();
   const flags = useFeatureFlags();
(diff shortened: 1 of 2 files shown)


packages/web/components/navbar-osmo-price.tsx
```diff
@@ -34,19 +33,15 @@ const NavbarOsmoPrice = observer(() => {
...
-      DecUtils.getTenExponentNInPrecisionRange(
-        chainStore.osmosis.stakeCurrency.coinDecimals
-      )
-    )
+  const { data: osmoCurrency } = api.edge.assets.getAsset.useQuery({
+    findMinDenomOrSymbol: "OSMO",
+  });
+  const { data: osmoPrice } = api.edge.assets.getAssetPrice.useQuery(
+    osmoCurrency!,
+    { enabled: Boolean(osmoCurrency) }
```
JoseRFelix, line 44: What do you think about creating a single query for all this data, except for the chart data, to minimize the number of requests?

jonator: My goal with these was to keep the queries single purpose and composable. Are you proposing we add prices to the getAsset(s) procedures?

They will likely be batched so networking overhead shouldn't be a concern.

MaxMillington: I'm inclined to agree with John here as I think single purpose queries are more scalable and easier to maintain.

MaxMillington (approved):

JoseRFelix (approved):

packages/web/components/navbar-osmo-price.tsx
jonator: Also, these particular queries are used individually elsewhere already. I think batching is added to tRPC as to prevent networking from being a concern when composing queries, allowing us to focus on organizing based on the types of data involved.

JoseRFelix: I agree, this applies to queries that aren't dependent on each other. However, in this specific case, we can't batch the query since we must first wait for `getAsset` to complete before we can begin fetching its price

PR #2631, opened 2024-01-02
Mattupham/staking lottie buttons

mattupham:

- add updated buttons with lottie animations

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1ue36x)

## Brief Changelog

- replace buttons
- add background clipped folder image
- add lottie animation
- make resizeable
- add mask styles with arbitrary values

## Testing and Verifying

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/9a2ac680-f5f8-42f6-890a-709b2e88baa2

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/7fac2c66-8c72-4ff6-8403-34f1aaf24ccc

Screenshot 2024-01-02 at 11 52 23 AM


packages/web/components/cards/rewards-card.tsx
```diff
@@ -1,55 +1,71 @@
...
-  tooltipContent,
   disabledTooltipContent,
   onClick,
-  image = null,
   disabled,
+  globalLottieFileKey,
+  position,
 }) => {
-  const ConditionalWrapper: React.FC<{ children: React.ReactNode }> = ({
+  const ConditionalTooltip: React.FC<{ children: React.ReactNode }> = ({
```
jonator, line 23: Nit: I think children come out of the box with the FC type, and are of a more specific type than ReactNode.

packages/web/public/right-grid.png
jonator, line 1: Thoughts on placing these images into one of the folders in the public folder?

packages/web/styles/globals.css
```diff
@@ -335,3 +335,15 @@
...
+  mask-image: url("./folder-left-tab.svg");
+  mask-size: contain;
+  mask-repeat: no-repeat;
+}
+
+.folder-right {
+  mask-image: url("./folder-right-tab.svg");
+  mask-size: contain;
+  mask-repeat: no-repeat;
+}
```
jonator, line 349: IMO we should avoid putting feature-specific styles in globals.css. (Even though there may be some there already). I think this is more fit for defining our design system instead. In isolation it's not a big deal, but over the long term it could be come quite messy. Thoughts on finding a way to co-locate these styles with the markup?

jonator (approved):

PR #2634, opened 2024-01-02
feat: Refactor Profile to TRPC

JoseRFelix:
### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1uagwj)

## Testing and Verifying

- [ ] Should display user balance
- [ ] Should display user balance and USD equivalent value when opening modal
- [ ] Balance should be invalidated after performing a transaction

### DIFF
--- a/packages/web/components/navbar/index.tsx
+++ b/packages/web/components/navbar/index.tsx
@@ -147,12 +147,12 @@ export const NavBar: FunctionComponent<
       }
     }, [onOpenFrontierMigration, onOpenSettings, query, userSettings]);
 
-    const account = accountStore.getWallet(chainId);
+    const wallet = accountStore.getWallet(chainId);
     const walletSupportsNotifications =
-      account?.walletInfo?.features?.includes("notifications");
+      wallet?.walletInfo?.features?.includes("notifications");
 
     const { data: icnsQuery, isLoading: isLoadingICNSQuery } = useICNSName({
-      address: account?.address ?? "",
+      address: wallet?.address ?? "",
     });
 
     // announcement banner
@@ -207,7 +207,7 @@ export const NavBar: FunctionComponent<
                     link: (e) => {
                       e.stopPropagation();
                       e.preventDefault();
-                      if (!account) return;
+                      if (!wallet) return;
                       onOpenNotifi();
                       closeMobileMainMenu();
                     },
@@ -376,7 +376,11 @@ export const NavBar: FunctionComponent<
             />
             <ClientOnly>
               <SkeletonLoader
-                isLoaded={!isWalletLoading && !isLoadingICNSQuery}
+                isLoaded={
+                  wallet?.isWalletConnected
+                    ? !isWalletLoading && !isLoadingICNSQuery
+                    : !isWalletLoading
+                }
               >
                 <WalletInfo
                   className="md:hidden"
@@ -448,7 +452,7 @@ const WalletInfo: FunctionComponent<
     );
 
   return (
-    <SkeletonLoader isLoaded={!isLoadingUserOsmoAsset}>
+    <SkeletonLoader isLoaded={walletConnected ? !isLoadingUserOsmoAsset : true}>
       <div className={className}>
         {!walletConnected ? (
           <Button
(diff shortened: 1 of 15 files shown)


packages/web/modals/profile.tsx
```diff
@@ -58,7 +60,6 @@ export const ProfileModal: FunctionComponent<
     accountStore,
     priceStore,
```
jonator, line 61: Would it be possible to remove need to use price store entirely?

We have a `getAssetPrice` tRPC procedure and a use `useCoinFiatValue` hook.

jonator (commented):
NIT

Would be nice as we do this migration to also remove as many references as possible to assets store, price store, in addition to queries store.

Longer term we can remove uses of chain store, particularly for getting Osmosis chain ID.

packages/web/modals/profile.tsx
JoseRFelix: Thank you for the feedback! I have removed the price store

packages/web/stores/index.tsx
```diff
@@ -1,11 +1,25 @@
...
 
 export const StoreProvider: FunctionComponent = ({ children }) => {
-  const [rootStore] = useState(() => new RootStore());
+  const apiUtils = api.useUtils();
+  const [rootStore] = useState(
+    () =>
+      new RootStore({
+        txEvents: {
+          onBroadcastFailed: () => {
+            apiUtils.edge.assets.getAsset.invalidate(); // Invalidate user balance
```
jonator, line 15: Are we expected to remember to add all relevant queries here? IMO this may be unsustainable and error prone if we want to use tRPC for almost everything. I'm imagining a future with many bespoke queries.

To reopen the convo from just before christmas; perhaps we can use some sort of event emitter pattern (doesn't have to use event emitters since you said they're error prone) or similar to notify views what category of node data was affected by the tx. Then, we could invalidate queries closer to where they are used, which would help keep related code co-located. For common/foundational queries though, we could invalidate them in one place as we do here to avoid having to repeatedly invalidate for every query.

jonator (approved):
Just some questions and curiosities.

kamal-sutra (approved):

packages/web/components/navbar/index.tsx
```diff
@@ -146,12 +147,13 @@ export const NavBar: FunctionComponent<
       }
     }, [onOpenFrontierMigration, onOpenSettings, query, userSettings]);
 
-    const account = accountStore.getWallet(chainId);
+    const wallet = accountStore.getWallet(chainId);
```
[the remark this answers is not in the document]
JoseRFelix: We changed it to match Cosmoskit's nomenclature. Additionally, this change clarifies things, as account is a broader term than wallet. Looking forward, account could encompass things like oauth or one click trading

packages/web/stores/index.tsx
JoseRFelix: It's unlikely that we'll need to monitor relevant queries for the global transaction events. This situation is unique because the user's balance needs to be invalidated, regardless of the transaction type.

As you mentioned, co-location seems to be a better strategy. My recommendation is to incorporate this within our upcoming useMutations. Relying on event-driven invalidation based on category is too prone to errors. It can lead to multiple invalidations across different components, potentially causing an n+1 refetching problem. This is because there's no mechanism to prevent repeated invalidations and they don't happen in a sequence.

packages/web/modals/profile.tsx
```diff
@@ -91,6 +91,16 @@ export const ProfileModal: FunctionComponent<
...
   const address = wallet?.address ?? "";
 
+  const { data: userOsmoAsset } = api.edge.assets.getAsset.useQuery(
+    {
+      findMinDenomOrSymbol: "OSMO",
+      userOsmoAddress: wallet?.address ?? "",
+    },
+    {
+      enabled: Boolean(wallet?.address) && typeof wallet?.address === "string",
+    }
```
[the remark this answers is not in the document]
jonator: Like `wallet.address as string`, instead of using `?` operators

CryptoAssassin1 (approved):

PR #2640, opened 2024-01-04
feat: Migrate Ad Banner

JoseRFelix:

Fetch ad banners from [Osmosis CMS](https://github.com/osmosis-labs/fe-content/blob/main/cms/swap-rotating-banner.json) instead of leveraging a hardcoded JSON to get swap rotating banner ads. This will avoid PRs on this repo to update CMS content. 

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1udkvj)

## Testing and Verifying

- [x] Ads banner should be visible


packages/web/pages/index.tsx
```diff
@@ -66,11 +48,36 @@ const Home = ({ ads }: InferGetStaticPropsType<typeof getStaticProps>) => {
...
-          <SwapTool ads={ads} />
+          {featureFlags.swapsAdBanner && <SwapAdsBanner />}
+          <SwapTool />
         </div>
       </div>
     </main>
   );
 };
 
+const SwapAdsBanner = () => {
```
MaxMillington, line 59: just to confirm, the tanstack query will cache this call for us? otherwise, so it won't just fetch the banners on every load?

Remember, fe-content is a normal repo, meaning we will get rate limited (1k requests per hour if i remember correctly)

@JoseRFelix

JoseRFelix: Yes, I will also add a local cache for 1 hour. Additionally, it's worth noting that this request is made entirely through the client and does not utilize our edge functions, so we won't face any rate-limiting issues

CryptoAssassin1 (approved):

jonator (approved):

PR #2646, opened 2024-01-05
Add pools query functions

jonator:

Adds starting point for pools table, and some basic pool query functions with caching. Also, I added a concept of "providers" for pools that allow us to use different data sources (node, sqs, indexer) for pools and easily swap them out. This also helps contain some of the complexity with querying data for pools, as in the case of the node pools query we need to also query balances for CL and cosmwasm pools to later calculate TVL. Also, these queries attempt to use our price functions for calculating TVL.

This will eventually be used for pools table.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1uagzk)

## Brief Changelog

* Added `getPools` function, with typesafe filtering and sorting
* Added node and sidecar pool providers
* Added tests for sidecar pool provider, as it will be the preferred provider
* Reorganize some of the assets code to make it more consistent with pools, and to split it up by type of data


DavideSegullo (approved):

packages/web/server/queries/complex/assets/index.ts
```diff
@@ -66,231 +58,73 @@ export async function getAssets({
...
-  asset,
-  userOsmoAddress,
-}: {
-  assetList?: AssetList[];
-  asset: TAsset;
-  userOsmoAddress?: string;
-}): Promise<TAsset & MaybeUserAssetInfo> {
-  if (!userOsmoAddress) return asset;
+/** Transform given asset list into an array of minimal asset types for user in frontend. */
+function reduceAssetList(
```
JoseRFelix, line 75: What do you think about renaming this function to `simplifyAssetListForDisplay`?

packages/web/server/queries/complex/pools/providers/__tests__/sidecar.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+  getAsset: jest.fn(),
+  calcAssetValue: jest.fn(),
+}));
+
+describe("getPoolsFromSidecar", () => {
+  server.use(
+    rest.get("https://sqs.osmosis.zone/pools/all", (_req, res, ctx) => {
+      return res(ctx.json(mockSidecarResponse));
+    })
+  );
```
JoseRFelix, line 29: This should be within the test or in a `beforeEach` hook because the msw handlers are reset after each test.

packages/web/server/queries/indexer/pool-aprs.ts
```diff
@@ -0,0 +1,18 @@
...
+  osmosis: number;
+  boost: number;
+  total_apr: number;
+};
+
+/** Queries numia for a breakdown of APRs per pool. */
+export function queryPoolAprs(): Promise<PoolApr[]> {
+  const url = new URL("/pools_apr", NUMIA_INDEXER_BASEURL);
+  return apiClient(url.toString());
+}
```
JoseRFelix, line 18: I believe we already have a query like this within the Numia namespace in queries, following Staking.

PR #2650, opened 2024-01-08
Feat: add token details stats section

DavideSegullo:
The goal of this PR is to add a new stats section on the token detail page, through which you can do the following things:

- View asset balance
- View staking balance (Only for OSMO)
- View a summary of the pools you participate in with the asset and also the balances, in the pools card 
- Deposit asset
- Withdraw asset
- Buy OSMO (Only for OSMO)

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1jb1r8)


sunnya97:
We shouldn't have a buy button on assets that we don't support buying.

packages/web/components/your-balance/your-balance.tsx
```diff
@@ -170,70 +179,101 @@ const ActionButton = ({
...
+  const {
+    isOpen: isFiatOnrampSelectionOpen,
+    onOpen: onOpenFiatOnrampSelection,
+    onClose: onCloseFiatOnrampSelection,
+  } = useDisclosure();
+  const account = accountStore.getWallet(chainStore.osmosis.chainId);
+  const fiat = priceStore.getFiatCurrency(priceStore.defaultVsCurrency);
+  const chainName = chainStore.getChainFromCurrency(denom)?.chainName;
+
+  const { data } = api.edge.assets.getAssetInfo.useQuery({
```
jonator, line 197: It would be better to use getAssetPrice, then pass DEFAULT_VS_CURRENCY to PricePretty. 

I'd like to remove additional references to priceStore to eventually remove it.

packages/web/localizations/en.json
```diff
@@ -830,7 +827,8 @@
     "tvl": "TVL",
     "backButton": "All tokens",
     "ariaBackButton": "All tokens back button",
-    "noData": "N.D."
+    "noData": "N.D.",
+    "buyToken": "Buy {coinDenom}"
```
jonator, line 831: Looks like these still need to be translated?

packages/web/server/api/edge-routers/assets-router.ts
```diff
@@ -105,7 +105,10 @@ export const assetsRouter = createTRPCRouter({
         asset: userAsset,
       });
 
-      return userMarketInfoAsset;
+      return {
+        ...userAsset,
+        ...userMarketInfoAsset,
+      };
```
jonator, line 111: This shold not be needed, as `getAssetMarketInfo` should include the user asset data. Was that not the case?

packages/web/localizations/en.json
DavideSegullo: Yes, we're still on draft, that's why a lot of things are still missing

packages/web/server/api/edge-routers/assets-router.ts
DavideSegullo: Nope, "amount" field was missing

packages/web/components/your-balance/your-balance.tsx
DavideSegullo: Oh yes, good idea! I'll refactor this one

packages/web/localizations/en.json
DavideSegullo: Fixed

packages/web/server/api/edge-routers/assets-router.ts
jonator: I think amount is missing if the user has no balance for that asset

packages/web/components/your-balance/your-balance.tsx
DavideSegullo: Fixed here: https://github.com/osmosis-labs/osmosis-frontend/pull/2650/commits/e24752256a6a799cba62e6f9510bf43f1a1ca888

packages/web/server/api/edge-routers/assets-router.ts
DavideSegullo: Yes I checked the balance, without the spread it always came back zero, instead with the use of the spread operator between the two objects I get the balance.
Yes it is better to return formatted object with zero value anyway, so we also bring back here the fact that we do not have to hold a reference to fiat currency

jonator: OK I'll need to fix this

jonator: @DavideSegullo fixed in stage

DavideSegullo: Perfect, I'll rebase to address It, thank you!

MaxMillington (approved):

CryptoAssassin1 (approved):

kamal-sutra (approved):

PR #2651, opened 2024-01-08
Remove $

MaxMillington:


packages/web/components/cards/unbonding-card.tsx
```diff
@@ -23,9 +23,9 @@ export const UnbondingCard: React.FunctionComponent = () => {
         <div className="flex flex-col pt-1 text-lg">
           {/* TODO figure out how to get estimated unbonding time */}
```
MaxMillington, line 24: @mattupham fast follow up after release. i had forgotten about this

PR #2659, opened 2024-01-09
Remove unlisted keyword filter of asset list

jonator:
When creating asset tRPC procedure, I was under the impression that the "osmosis-unlisted" keyword meant that it should be filtered. This caused confusion when unlisted assets were not appearing in the frontend.

I removed the filter, and will consider adding a flag later that can be used to treat that keyword as a filter of the assets page.

Also fixed direct swap hanging issue in legacy router.

### DIFF
--- a/packages/web/server/queries/complex/assets/__tests__/assets.spec.ts
+++ b/packages/web/server/queries/complex/assets/__tests__/assets.spec.ts
@@ -29,15 +29,6 @@ describe("getAssets", () => {
       expect(assets[0].coinDenom).toEqual("stLUNA");
     });
 
-    it("should not return unlisted assets", async () => {
-      const assets = await getAssets({
-        search: { query: "PYTH" },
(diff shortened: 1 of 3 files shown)


JoseRFelix:
@jonator It looks like tests are failing

JeremyParish69 (changes requested):
The Pool Detail swap modal now shows the unlisted token, but there are still console errors and there is no quoted price after inputting a swap in amount.

JeremyParish69 (changes requested):
The Pool Detail swap modal now shows the unlisted token, but there are still console errors and there is no quoted price after inputting a swap in amount.
In addition, the frontend ignores the unlisted flag, which is a step backwards. We still need to hide tokens that aren't ready.

jonator:
Closing in favor of: https://github.com/osmosis-labs/osmosis-frontend/pull/2696

Issue #2661, opened 2024-01-09
Pools number should show up in links preview [100 USDC Reward!]


MaxMillington:
🐛  Bug Bounty: Fix Needed! 🛠️ 
Reward: :trophy: 100 USDC 🏆 

We are excited to announce a bug bounty for third-party developers! If you've got the skills and know-how, here's your chance to earn some 💸  and contribute to our project.

Issue Description 📄
Pools number should show up in links preview 

Expected Behavior 💡
As a user, if I copy the link to a pools page and paste it on (e.g.) twitter, the pools number should show in the thumbnail preview

Current Behavior ❌
Currently, we do not have anything in the meta tags, so it just shows '#'
![image (1)](https://github.com/osmosis-labs/osmosis-frontend/assets/11064595/022b65ab-78c2-4d45-a046-1051fa1317ee)

Requirements for the Bounty 🏁

 Submit a pull request with a verified fix to the issue.
 The fix should not introduce any new bugs or vulnerabilities.
 Code should follow our project's style and contribution guidelines.
 Your solution will be reviewed by our team, and the bounty will be awarded upon approval.
How to Claim Your Bounty 💰

 Fork our repository.
 Create a new branch for your fix.
 Submit a pull request to our main branch.
 Include detailed information about your fix and how it addresses the issue.

 Once reviewed and merged, the 100 USDC will be transferred to your wallet.
We look forward to your contributions and thank you for helping improve our project!

**Please Note**: All bounty claimants must undergo a KYC (Know Your Customer) process with our finance team to receive the reward. This is a mandatory step to ensure compliance with financial regulations.

jonator:
This should be doable with SSR.

gnujoow:
> This should be doable with SSR.
i just wonder if osmosis app is going to convert to ssr.

MaxMillington:
@gnujoow we already use SSR in a fair number of spots. Feels free to check it out!

MaxMillington:
Fixed by @fabryscript

PR #2666, opened 2024-01-10
Fix issue with pool page title

fabryscript:

This pull request fixes the issue #2661

## Brief Changelog

- This makes the computation of the `poolId` parameter server-side in order to match the SEO timing which shows the page title in the browser

## Testing and Verifying

### DIFF
--- a/packages/web/pages/pool/[id].tsx
+++ b/packages/web/pages/pool/[id].tsx
@@ -113,11 +113,10 @@ const Pool: FunctionComponent<Props> = observer(
 
 export const getStaticPaths: GetStaticPaths = async () => {
   const { num_pools } = await queryNumPools();
-  let paths: { params: { id: string } }[] = [];
 
-  for (let i = 0; i <= Number(num_pools); i++) {
-    paths.push({ params: { id: String(i + 1) } });
-  }
+  const paths = Array.from({ length: Number(num_pools) + 1 }, (_, i) => ({
+    params: { id: String(i + 1) },
+  }));
 
   return { paths, fallback: "blocking" };
 };
--- a/packages/web/pages/pool/[id].tsx
+++ b/packages/web/pages/pool/[id].tsx
@@ -1,5 +1,5 @@
 import { observer } from "mobx-react-lite";
-import { GetServerSideProps, InferGetServerSidePropsType } from "next";
+import { GetStaticPaths, GetStaticProps, InferGetStaticPropsType } from "next";
 import { useRouter } from "next/router";
 import { NextSeo } from "next-seo";
 import { FunctionComponent, useEffect, useMemo, useState } from "react";
@@ -14,6 +14,7 @@ import { useTranslation, useWindowSize } from "~/hooks";
 import { useNavBar } from "~/hooks";
 import { useFeatureFlags } from "~/hooks/use-feature-flags";
 import { TradeTokens } from "~/modals";
+import { queryNumPools } from "~/server/queries/osmosis";
 import { useStore } from "~/stores";
 
 interface Props {
@@ -21,7 +22,7 @@ interface Props {
 }
 
 const Pool: FunctionComponent<Props> = observer(
-  ({ poolId }: InferGetServerSidePropsType<typeof getServerSideProps>) => {
+  ({ poolId }: InferGetStaticPropsType<typeof getStaticProps>) => {
     const router = useRouter();
     const { chainStore, queriesStore } = useStore();
     const { chainId } = chainStore.osmosis;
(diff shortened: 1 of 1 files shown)


CryptoAssassin1 (approved):
Fixed.
Screenshot from 2024-01-11 17-02-06

DavideSegullo (commented):
the only side effect of using `getServerSideProps` is that the loads are slow, just see it by clicking a few rows from the table, maybe there is a better approach?

fabryscript:
> the only side effect of using `getServerSideProps` is that the loads are slow, just see it by clicking a few rows from the table, maybe there is a better approach? 

Yes, I'm going to try a different approach based on `getStaticProps`

kamal-sutra (approved):

MaxMillington:
@fabryscript ping us again when you make the change

fabryscript:
Hi @MaxMillington, I've successfully made the change and now the page uses `getStaticProps` instead.

packages/web/pages/pool/[id].tsx
```diff
@@ -13,97 +14,117 @@ import { useTranslation, useWindowSize } from "~/hooks";
...
+  for (let i = 0; i <= Number(num_pools); i++) {
+    paths.push({ params: { id: String(i + 1) } });
+  }
+
+  return { paths, fallback: "blocking" };
+};
+
+export const getStaticProps: GetStaticProps = async ({ params }) => {
+  const id = params?.id as string;
+  return { props: { poolId: id ?? "-" } };
```
Comment (DavideSegullo) [inline packages/web/pages/pool/[id].tsx:127]: since the page already refers to a pool, perhaps it makes sense to rename the parameter to 'id', putting 'pool' before it is redundant

kamal-sutra (approved):

DavideSegullo (changes requested):

CryptoAssassin1:
Build failed @kamal-sutra @MaxMillington

MaxMillington:
@fabryscript the build is failing. are you building locally? looks like type errors

packages/web/pages/pool/[id].tsx
```diff
@@ -13,97 +14,117 @@ import { useTranslation, useWindowSize } from "~/hooks";
...
+  }
+);
+
+export const getStaticPaths: GetStaticPaths = async () => {
+  const { num_pools } = await queryNumPools();
+  let paths: { params: { id: string } }[] = [];
+
+  for (let i = 0; i <= Number(num_pools); i++) {
+    paths.push({ params: { id: String(i + 1) } });
+  }
```
Comment (jonator) [inline packages/web/pages/pool/[id].tsx:120]: Nit: I think a `map` (vs `for`) would be more readable here

packages/web/pages/pool/[id].tsx
```diff
@@ -13,97 +14,117 @@ import { useTranslation, useWindowSize } from "~/hooks";
...
+
+  return { paths, fallback: "blocking" };
+};
+
+export const getStaticProps: GetStaticProps = async ({ params }) => {
```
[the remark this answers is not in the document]
Comment (fabryscript) [reply packages/web/pages/pool/[id].tsx:125]: I don't think so, the reason being that the number of pools is a value which doesn't change so often, hence why I wouldn't consider using revalidation

fabryscript:
> @fabryscript the build is failing. are you building locally? looks like type errors

I've re-built it locally and everything seems working fine. Could you please tell me the details of the errors that Vercel is throwing?

packages/web/pages/pool/[id].tsx
Comment (fabryscript) [reply packages/web/pages/pool/[id].tsx]: Done in [8522a86](https://github.com/osmosis-labs/osmosis-frontend/pull/2666/commits/8522a8648512037f733cbf66efb3eec94283657e)

packages/web/pages/pool/[id].tsx
Comment (fabryscript) [reply packages/web/pages/pool/[id].tsx]: Done in 2bb4d66

MaxMillington:
@fabryscript 
Screenshot 2024-01-16 at 2 25 25 PM
Screenshot 2024-01-16 at 2 25 35 PM
Screenshot 2024-01-16 at 2 25 44 PM

fabryscript:
> @fabryscript 
> 
> Screenshot 2024-01-16 at 2 25 25 PM
> 
> Screenshot 2024-01-16 at 2 25 35 PM
> 
> Screenshot 2024-01-16 at 2 25 44 PM
> 
> 

As you can see these errors are not related to my code but to the code which is responsible for generating the `[denom].tsx` page directory. Could you please trigger a new build again and see if it works? There might be a problem with the Twitter API.

MaxMillington:
@fabryscript looks like it worked.

fabryscript:
> @fabryscript looks like it worked.

Good to know!

MaxMillington (approved):

DavideSegullo (approved):

MaxMillington:
@fabryscript please email finance@osmosis.team to claim the bounty!

fabryscript:
I have seen that there might be some problems with `getStaticProps` relative to the fact that the build fails for OOM reasons. I might work on a fix to this, maybe by switching back to full SSR with `getServerSideProps`.

MaxMillington:
> I have seen that there might be some problems with `getStaticProps` relative to the fact that the build fails for OOM reasons.

could there be a memory leak? Do we need to revert in the meantime?

jonator:
@MaxMillington https://github.com/osmosis-labs/osmosis-frontend/pull/2701

This PR reverts the changes as a bundle

PR #2679, opened 2024-01-11
close modal when clicking x

MaxMillington:


packages/web/components/stake/stake-onboarding.tsx
```diff
@@ -107,6 +107,7 @@ export const StakeOnboarding: React.FC<StakeOnboardingProps> = ({
       <StakeIntroModal
         isOpen={showStakeIntroModal}
         onRequestClose={handleClose}
+        onRequestSimpleClose={() => setShowStakeIntroModal(false)}
```
MaxMillington, line 110: need diff function to handle `X` vs clicking button

PR #2681, opened 2024-01-12
feat: Add CL Positions Query

JoseRFelix:

This PR primarily introduces functions to retrieve user positions and detailed position information. It involves adding queries for this purpose. However, the trpc router may still undergo modifications to align with UI requirements once these are integrated.

There is also a fix for superfluid APRs. In production, we were using the outdated `cosmos/bank/v1beta1/supply/${denom}` endpoint. The correct endpoint to use is `/cosmos/bank/v1beta1/supply/by_denom?denom=${denom}`

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1uaudw)

## Brief Changelog

- Add queries and functions to get user CL position data
- Move `BondStatus` to @osmosis-labs/types
- Fix CL superfluid APRs


packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
```diff
@@ -0,0 +1,297 @@
...
+import { UserOsmoAddressSchema } from "~/server/queries/complex/parameter-types";
+import { queryPositionPerformance } from "~/server/queries/imperator";
+import { queryAddressCLPositions } from "~/server/queries/osmosis/concentrated";
+
+export const concentratedLiquidityRouter = createTRPCRouter({
+  getUserPositions: publicProcedure
+    .input(
+      z
+        .object({
+          sortDirection: z.enum(["asc", "desc"]).default("desc").optional(),
```
jonator, line 19: I've been burned here where default doesn't actually get set by zod if an "optional" follows. Just a heads up (this may work)

packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
```diff
@@ -0,0 +1,297 @@
...
+  isUnbonding: boolean;
+}): PositionStatus {
+  const inRange = lowerPrice.lt(currentPrice) && upperPrice.gt(currentPrice);
+  const diff = new Dec(
+    Math.min(
+      Number(currentPrice.sub(lowerPrice).toString()),
+      Number(upperPrice.sub(currentPrice).toString())
+    )
+  );
+  const rangeDiff = upperPrice.sub(lowerPrice);
```
jonator, line 78: Could we put these types and fns into complex folder?

packages/web/server/queries/complex/assets/index.ts
```diff
@@ -36,8 +38,18 @@ export async function getAsset({
...
+    key: anyDenom,
+    getFreshValue: async () => {
+      const assets = await getAssets({
+        assetList,
+        findMinDenomOrSymbol: anyDenom,
+      });
+      return assets[0];
+    },
+    ttl: 10 * 1000, // 10 seconds
+  });
```
jonator, line 52: Is this just to save on the array access? I think `getAssets` is cached.

packages/web/server/queries/osmosis/concentrated/address-cl-positions.ts
jonator, line 1: IMO we should fully match path name of chain module: `concentrated` > `concentratedliquidity`

jonator (commented):
Initial pass

jonator:
I think there's more merge conflicts from the changes we made to throwing exceptions

jonator (commented):
Some things I noticed

packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
JoseRFelix: I've found that by adding `default`, we don't need to specify `optional`. The main change is that we have to use `z.input` instead of `z.infer` to determine the type. This makes sense because when used as input, a value can be optional, but when we're dealing with a type, we expect the value to always be defined.

packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
JoseRFelix: Should be on their respective folder on my latest commit

packages/web/server/queries/complex/assets/index.ts
JoseRFelix: You're right. Removed this cachified

packages/web/server/api/edge-routers/pools-router.ts
```diff
@@ -35,9 +35,9 @@ export const poolsRouter = createTRPCRouter({
     .input(
       GetInfinitePoolsSchema.and(
         z.object({
-          sort: createSortSchema(marketIncentivePoolsSortKeys)
-            .optional()
-            .default({ keyPath: "totalFiatValueLocked" }),
+          sort: createSortSchema(marketIncentivePoolsSortKeys).default({
+            keyPath: "totalFiatValueLocked",
+          }),
```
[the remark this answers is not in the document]
JoseRFelix: Following my earlier comment, I think this change won't cause any issues; instead, it will set the default value as intended. Previously, the value remained undefined, which negated the purpose of having a default.

packages/web/server/queries/complex/pools/index.ts
```diff
@@ -57,9 +61,9 @@ export async function getPool(poolId: string): Promise<Pool> {
  *  Params can be used to filter the results by a fuzzy search on the id, type, or coin denoms, as well as a specific id or type. */
 export async function getPools(
   params?: PoolFilter,
-  poolProvider: PoolProvider = getPoolsFromSidecar
+  poolsProvider: PoolsProvider = getPoolsFromSidecar
 ): Promise<Pool[]> {
-  let pools = await poolProvider();
+  let pools = await poolsProvider({ poolIds: params?.poolIds });
```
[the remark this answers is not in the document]
JoseRFelix: I agree it makes it simpler, however, it's slower than querying the exact pools. Fetching all pools takes ~1.3s, whereas querying with specific IDs completes in ~400ms. Do you think we'll have to add many providers because otherwise it will be a small price to pay for the performance boost?

packages/web/server/queries/osmosis/poolmanager/pools.ts
```diff
@@ -4,7 +4,7 @@ import { ChainList } from "~/config/generated/chain-list";
 
 export type WeightedPoolRawResponse = {
   "@type": "/osmosis.gamm.v1beta1.Pool";
-  id: string;
+  id: number;
```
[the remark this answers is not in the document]
JoseRFelix: Looks like a merge conflict. I fixed this on my latest commit

packages/web/server/queries/sidecar/pool.ts
```diff
@@ -0,0 +1,23 @@
...
+  sqs_model: {
+    total_value_locked_uosmo: string;
+    balances: {
+      denom: string;
+      amount: string;
+    }[];
+    pool_denoms: string[];
+    spread_factor: string;
+  };
+};
```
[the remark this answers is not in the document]
JoseRFelix: Yeah, thank you! I have brought your changes and we should be synced

packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
```diff
@@ -0,0 +1,110 @@
...
+    .input(
+      z.object({
+        positionId: z.string(),
+      })
+    )
+    .query(async ({ input: { positionId } }) => {
+      const [position, performance] = await Promise.all([
+        queryCLPosition({ id: positionId }),
+        queryPositionPerformance({ positionId }),
+      ]);
```
jonator, line 51: What is the benefit of wrapping in promise all? Does it send the requests simultaneously?

packages/web/server/queries/complex/assets/index.ts
```diff
@@ -139,6 +141,40 @@ function simplifyAssetListForDisplay(
   return assetListAssets.map(makeMinimalAsset);
 }
 
+export async function mapRawAssetsToCoinPretty({
```
jonator, line 144: A comment would be useful for when this should be used.

packages/web/server/queries/complex/assets/index.ts
```diff
@@ -139,6 +141,40 @@ function simplifyAssetListForDisplay(
...
+  rawAssets,
+  calculatePrice,
+}: {
+  rawAssets?: {
+    amount: string | number;
+    denom: string;
+  }[];
+  calculatePrice?: boolean;
+}): Promise<(CoinPretty & { fiatValue?: Dec })[]> {
```
jonator, line 153: At complex layer, I'd prefer PricePretty over Dec for fiat value.

Also, long term I'm thinking we should create our own unit package, and rename *Pretty to just *.

packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
```diff
@@ -0,0 +1,110 @@
...
+          rawClaimableSpreadRewards: position.position.claimable_spread_rewards,
+          calculatePrice: true,
+        }),
+        getTotalEarned({
+          totalIncentivesRewards: performance.total_incentives_rewards,
+          totalSpreadRewards: performance.total_spread_rewards,
+        }),
+      ]);
+
+      const totalPrincipalValue = principalAssets.reduce(
```
jonator, line 78: Perhaps we can create a util that accepts these complex types (anything that implements `toDec()`) and sums them?

Signature would be something like `function sum(values: number | string | { toDec() }): Dec`

packages/web/server/queries/complex/cosmos/distribution-params.ts
```diff
@@ -0,0 +1,22 @@
+import { RatePretty } from "@keplr-wallet/unit";
+import cachified, { CacheEntry } from "cachified";
+import { LRUCache } from "lru-cache";
+
+import { DEFAULT_LRU_OPTIONS } from "~/config/cache";
+import { queryDistributionParams } from "~/server/queries/cosmos/distribution";
+
+const distributionCache = new LRUCache<string, CacheEntry>(DEFAULT_LRU_OPTIONS);
+
+export function getDistributionParams({ chainId }: { chainId: string }) {
```
jonator, line 10: Since we're still early, I'm wondering if we somewhat replicate what Keplr did with the Chain ID > Node Queries map. 

We can have a default for Osmosis chain ID, but perhaps it's a function that can make it easy to make node queries to other chains in the chain registry.

We could do this later or not do it at all, as I don't think we have a strong need to query other chains. Perhaps the approach we can take is to just create additional query functions for querying arbitrary chains by ID as needed.

packages/web/server/queries/complex/osmosis/epochs.ts
```diff
@@ -0,0 +1,74 @@
...
+  });
+}
+
+export async function getEpoch({ identifier }: { identifier: string }) {
+  const epochs = await getEpochs();
+
+  const epoch = epochs.find((epoch) => epoch.identifier === identifier);
+  if (!epoch) {
+    console.info(`Epoch ${identifier} not found`);
+    return undefined;
```
jonator, line 66: I'd prefer to throw here, thoughts? I found not throwing on server procedures (in the case of debugging the price and asset functions with testnet asset list) makes debugging issues much more time consuming.

It shouldn't be an end user UX hit, as react query should just not return the data. Or we can display an error to the user.

packages/web/server/queries/complex/osmosis/osmosis-params.ts
```diff
@@ -0,0 +1,59 @@
...
+export async function getSuperfluidParams() {
+  return cachified({
+    cache: paramsCache,
+    key: "osmosis-superfluid-params",
+    ttl: 30 * 1000, // 30 seconds
+    getFreshValue: async () => {
+      const { params } = await querySuperfluidParams();
+
+      return {
+        minimumRiskFactor: new Dec(params.minimum_risk_factor),
```
jonator, line 55: I think complex folder is perfect for caching and returning more complex types 👍

packages/web/server/queries/complex/pools/apr.ts
jonator, line 1: Thoughts on placing these functions in the incentives file in this same folder? Those already return APRs. Also, they contain a superfluid APR value from numia as well.

packages/web/server/queries/complex/pools/osmo-equivalent.ts
```diff
@@ -0,0 +1,86 @@
...
+  StablePoolRawResponse,
+  WeightedPoolRawResponse,
+} from "~/server/queries/osmosis";
+
+export async function estimatePoolAPROsmoEquivalentMultiplier({
+  poolId,
+}: {
+  poolId: string;
+}) {
+  const pool = await getPool({ poolId });
```
jonator, line 18: Good use of new pool function, that relies on the providers.

packages/web/server/queries/complex/chain-inflation.ts
jonator, line 1: Should we offer this function? I think with taker fees and protorev the staking APR has become more complex, and likely should just be fetched from numia.

jonator (commented):
Got about half way and want to review more but I g2g. Here's a batch of comments

packages/web/server/queries/complex/concentrated-liquidity.ts
```diff
@@ -0,0 +1,526 @@
...
+            undelegatingPositionsPromise,
+            stakeCurrencyPromise,
+            initialRangeAprPromise,
+          ]);
+
+          const pool = pools.find((pool) => pool.id === position.pool_id);
+
+          if (!pool) {
+            console.error(`Pool (${position.pool_id}) not found`);
+            return undefined;
```
jonator, line 379: Would prefer a throw here instead of error log.

jonator, line 379: Would prefer to throw more aggressively

packages/web/server/queries/complex/concentrated-liquidity.ts
```diff
@@ -0,0 +1,526 @@
...
+          const userUnbondingPositionsPromise = getUserUnbondingPositions({
+            userOsmoAddress,
+          });
+          const delegatedPositionsPromise = getUserDelegatedClPositions({
+            userOsmoAddress,
+          });
+          const undelegatingPositionsPromise = getUserUndelegatingClPositions({
+            userOsmoAddress,
+          });
+          const stakeCurrencyPromise = getAsset({ anyDenom: "OSMO" });
```
jonator, line 350: Would it be more robust to get this from the osmosis chain config?

packages/web/server/queries/complex/concentrated-liquidity.ts
```diff
@@ -0,0 +1,526 @@
...
+        await queryUndelegatingClPositions({
+          bech32Address: userOsmoAddress,
+        });
+      return cl_pool_user_position_records;
+    },
+    ttl: 5 * 1000, // 5 seconds
+  });
+}
+
+export async function normalizePositions({
```
jonator, line 291: I'm not sure what you mean with the function title "normalizePositions". A comment or more descriptive function name would be appreviated

packages/web/server/queries/osmosis/concentratedliquidity/position-by-id.ts
```diff
@@ -0,0 +1,19 @@
+// "/osmosis/concentratedliquidity/v1beta1"/positions/${bech32Address}?pagination.limit=10000
```
jonator, line 1: Are these comments for others' reference?

jonator (commented):
Another review.

I'd like to get this through since I need some of the queries for the new assets page. But I think some tests with mock data would be very valuable. There's a lot of processing of queried data happening that I'd imagine would be tricky to debug if incorrect.

packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
JoseRFelix: It resolves promises concurrently. It's faster than if we use plain await

packages/web/server/queries/complex/cosmos/distribution-params.ts
JoseRFelix: I agree with adding this feature early. I've made a simple function called `createNodeQuery` to help us do this. It also simplifies fetching node data by removing the need to constantly re-enter the `URL`. Please take a look at my latest commit

packages/web/server/queries/complex/chain-inflation.ts
JoseRFelix: I have removed these methods on my latest commit. It is much better and cleaner to fetch them from Numia. Thank you for the idea!

packages/web/server/queries/osmosis/concentratedliquidity/position-by-id.ts
JoseRFelix: I overlooked this earlier. It has been removed in my latest commit.

JoseRFelix:
@jonator I'm missing some unit tests, but this is ready for another look

jonator (approved):

PR #2684, opened 2024-01-12
Testnet fix

jonator:
It seems the testnet asset list is not matching what's on testnet @JeremyParish69 

* don't filter unverified assets from pools on testnet


MaxMillington (approved):

JeremyParish69:
We don't use the 'unverified' keyword for any testnet assets, so I don't expect this to change anything...

PR #2685, opened 2024-01-12
Fix: chain suggestions

DavideSegullo:
Chain Suggestions compiled by our frontend are faulty and not showing balances within the wallet. (For example Passage config is broken on production, balance doesn't load inside Keplr).

This PR fixes the list of assets we pass to the `WalletManager` class to make sure we provide it with the native data of the chain it belongs to and not the IBC data.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1xqh15)

### DIFF
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -131,6 +131,20 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
 
   private _cache = new LRUCache<string, CacheEntry>({ max: 30 });
 
+  /**
+   * We make sure that the 'base' field always has as its value the native chain parameter
+   * and not values derived from the IBC connection with Osmosis
+   */
+  private get _assets() {
+    return this.assets.map((assetList) => ({
+      ...assetList,
+      assets: assetList.assets.map((asset) => ({
+        ...asset,
+        base: asset.traces ? getSourceDenomFromAssetList(asset) : asset.base,
+      })),
+    }));
+  }
+
   constructor(
     public readonly chains: (Chain & { features?: string[] })[],
     readonly osmosisChainId: string,
@@ -154,17 +168,6 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
     >
   ) {
     this._wallets = wallets;
-    /**
-     * We make sure that the 'base' field always has as its value the native chain parameter
-     * and not values derived from the IBC connection with Osmosis
-     */
-    this.assets = assets.map((assetList) => ({
-      ...assetList,
-      assets: assetList.assets.map((asset) => ({
-        ...asset,
-        base: getSourceDenomFromAssetList(asset),
-      })),
-    }));
     this._walletManager = this._createWalletManager(wallets);
     this.accountSetCreators = accountSetCreators;
 
@@ -174,7 +177,7 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
   private _createWalletManager(wallets: MainWalletBase[]) {
     this._walletManager = new WalletManager(
       this.chains,
-      this.assets as CosmologyAssetList[],
+      this._assets as CosmologyAssetList[],
       wallets,
       logger,
       true,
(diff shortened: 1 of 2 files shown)


packages/web/stores/root.ts
```diff
@@ -55,6 +55,29 @@ import {
...
+    };
+  });
+
+  return {
+    ...assetList,
+    assets,
+  };
+});
+
+console.log(RawAssetLists.find((el) => el.chain_id === "passage-2"));
```
MaxMillington, line 79: can we remove the log

MaxMillington (approved):
let's get jeremy to QA

packages/web/stores/root.ts
```diff
@@ -55,6 +55,29 @@ import {
...
+      ...asset,
+      base,
+    };
+  });
+
+  return {
+    ...assetList,
+    assets,
+  };
+});
```
JoseRFelix, line 77: I believe we shouldn't modify the assetlist at the root store level. Instead, how about we move this logic into the account store?

packages/web/stores/root.ts
DavideSegullo: Done

packages/web/stores/root.ts
```diff
@@ -55,6 +55,27 @@ import {
...
+    let base = asset.base;
+
+    if (ibcTrace) {
+      base = ibcTrace.counterparty.base_denom;
+    }
+
+    return {
+      ...asset,
+      base,
+    };
```
JoseRFelix, line 70: Btw, we can get the source denom using the [`getSourceDenomFromAssetList`](https://github.com/osmosis-labs/osmosis-frontend/blob/e6fb7743580039d926d1448df0070506f4f34c5b/packages/utils/src/asset-utils.ts#L18) function in @osmosis-labs/utils.

JoseRFelix (changes requested):

packages/web/stores/root.ts
DavideSegullo: Done

packages/web/stores/root.ts
DavideSegullo: You could do it, in general I wouldn't do it for a generic class that tends to be controlled by a 'parent,' but in this case it's to make sure the data is in the correct format, so I would say yes

DavideSegullo: Done

CryptoAssassin1:
Fixed.
Screenshot from 2024-01-13 01-35-42

CryptoAssassin1 (approved):

JoseRFelix:
@DavideSegullo Looks like tests are failing

DavideSegullo:
> @DavideSegullo Looks like tests are failing

Fixed

packages/stores/src/account/base.ts
```diff
@@ -126,6 +131,20 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
 
   private _cache = new LRUCache<string, CacheEntry>({ max: 30 });
 
+  /**
+   * We make sure that the 'base' field always has as its value the native chain parameter
+   * and not values derived from the IBC connection with Osmosis
+   */
+  private get _assets() {
```
JoseRFelix, line 138: What are your thoughts on renaming this to `walletManagerAssets` since we only need this for cosmos-kit?

DavideSegullo: Yes makes sense

DavideSegullo: Done

JoseRFelix (approved):

kamal-sutra (approved):

JeremyParish69:
FWIW: the Passage Chain Suggestion has now been added to Keplr's Chain Registry, so that would override whatever we suggested. But I was able to eventually find that Lambda chain is not in their registry, and I could see that the compiled Chain Suggestion correctly uses LAMB for its currency instead of Osmosis' ibc hash denom. So I think this fixes it!

PR #2696, opened 2024-01-16
fix: No Keyword Assets and Get Unlisted Assets from TRPC

JoseRFelix:
#### No Keyword assets

Assets that don't have a keyword should still be visible when the 'unverified assets' flag is activated because they are not unlisted, and also not verified.

#### Unlisted Assets
 Users who enable the 'unlisted asset preference' should be able to view these assets in both the swap tool and the asset table.

From a technical standpoint, we will introduce a new filter to the map trpc queries, specifically `getAssets` and `getAssetInfos`. However, we won't apply this filter to `getAsset` or `getAssetInfo`, as these more detailed queries are not directly visible to users unless they specifically choose the asset. Thus, it's acceptable to always retrieve unlisted assets for these queries. This approach ensures backward compatibility with other modules, like swap.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1ym43w)

## Brief Changelog

- Add `includeUnlistedAssets` to the map assets queries 

## Testing and Verifying

- [ ] Should see unlisted assets in the swap and assets table when enabled. To enable add the `show_unlisted_assets=true` query key to the URL.

### DIFF
--- a/packages/web/server/queries/complex/assets/index.ts
+++ b/packages/web/server/queries/complex/assets/index.ts
@@ -94,7 +94,7 @@ function simplifyAssetListForDisplay(
     .filter((asset) =>
       params.includeUnlisted
         ? true // Do not filter the unlisted assets
-        : asset.keywords && !asset.keywords.includes("osmosis-unlisted")
+        : !asset.keywords?.includes("osmosis-unlisted")
     );
 
   let assetListAssets = listedAssets.filter((asset) => {
--- a/packages/web/server/queries/complex/assets/index.ts
+++ b/packages/web/server/queries/complex/assets/index.ts
@@ -23,6 +23,7 @@ export type Asset = {
 export const AssetFilterSchema = z.object({
   search: SearchSchema.optional(),
   onlyVerified: z.boolean().default(false).optional(),
+  includeUnlisted: z.boolean().default(false).optional(),
 });
 /** Params for filtering assets. */
 export type AssetFilter = z.infer<typeof AssetFilterSchema>;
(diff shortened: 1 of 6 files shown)


kamal-sutra (approved):
Both Bad and Baddog showup on Swap tool and in pool 1356
image

image

jonator (changes requested):
See my broad take on slack:

https://osmosis-network.slack.com/archives/C03JT90769K/p1705399977402679?thread_ts=1705261704.588239&cid=C03JT90769K

CryptoAssassin1 (approved):

packages/web/server/queries/complex/assets/index.ts
```diff
@@ -33,11 +34,17 @@ const searchableAssetListAssetKeys = ["symbol", "base", "name", "display"];
...
 }: {
   assetList?: AssetList[];
   anyDenom: string;
-}): Promise<Asset | undefined> {
-  const assets = await getAssets({ assetList, findMinDenomOrSymbol: anyDenom });
+} & AssetFilter): Promise<Asset | undefined> {
+  const assets = await getAssets({
+    assetList,
+    findMinDenomOrSymbol: anyDenom,
+    includeUnlisted: true,
```
jonator, line 45: Do we want this hardcoded here?

JoseRFelix: I believe so, however, these params can be overriden if needed with the `...params`. For detailed queries, unlisted assets are included by default since these queries are like searches and are only used when a specific asset is selected.

jonator (approved):

jonator:
@JoseRFelix thoughts on having this flag be enabled by default on dev? I'm not sure if that's the solution, but I'm worried it will be tedious to always have to put this query param in place. Also, it this PR doesn't address the client/browser usage of the asset list which ignores the unlisted keyword (unlike tRPC getAssets function) cc @JeremyParish69

JoseRFelix:
> @JoseRFelix thoughts on having this flag be enabled by default on dev? I'm not sure if that's the solution, but I'm worried it will be tedious to always have to put this query param in place. 

I think we don't use unlisted assets much in development. Also, I prefer keeping our development server similar to production to simplify debugging. WDYT?

>Also, it this PR doesn't address the client/browser usage of the asset list which ignores the unlisted keyword (unlike tRPC getAssets function)

Previously, we managed unlisted assets in our assets store. This PR is just introducing support for them with trpc.

PR #2699, opened 2024-01-16
Squad "View or edit" button disabled - update

mattupham:

The "View or edit" button has a disabled state when you have insufficient funds for the value entered in the stake/unstake amount fields - basically the same behavior as the Stake and Unstake buttons.

This button should never be disabled if you are actively staking – i.e. you have at least one validator in your squad.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1vg6rg)


MaxMillington:
@mattupham do we still want this if we end up performing a redelegation on editing validator squad?

PR #2700, opened 2024-01-16
Fetch token pair historical data via tRPC

jonator:

To save on costs, mildly cache token pair historical data on our edge mid layer instead of always querying from clients. This also refactors some UI config store query logic, and adds some queries for pool data.

Screenshot 2024-01-17 at 2 59 07 PM
Screenshot 2024-01-17 at 2 59 37 PM

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1xguyj)

## Brief Changelog

* Add server side queries, with 3 minutes of caching per pool + tokens + time duration combo
* Move add CL config to web package, extracting historical price query store
* Replace historical price query store in ObservableHistoricalAndLiquidityData. This is used in the position detail views and charts on pool detail page
* Use getPool query for some pool data in above stores and on CL pool detail page

## Testing and Verifying

* All functionality should still work in CL add liquidity
* All data should be accurate on CL pool detail pages
* Increase CL liquidity should work
*


sunnya97:
This seems slower that what's live on app.osmosis.zone

DavideSegullo (approved):

jonator:
> This seems slower that what's live on app.osmosis.zone

I think it's because it adds an extra hop between browser and data services, through the edge network. But, it is cached for 3 mins so many users would only hop once (on the edge, which should be closer to their machine)

CryptoAssassin1 (approved):

PR #2702, opened 2024-01-17
Fix: disable pool page static generation

DavideSegullo:
Fix out of memory error related to static generation

### ClickUp Task


jonator (approved):
We prob go back to missing the pool ID in SSR HTML, but at least builds work now

PR #2706, opened 2024-01-18
add spread factors to pools table

sunnya97:

- Reduce the size of pool type icons
- Show spread factor for pools on the pools page

CleanShot 2024-02-03 at 02 03 24@2x


JoseRFelix (approved):
LGTM 🚀

PR #2713, opened 2024-01-19
Show cosmwasm pools in route vis

jonator:

Requested by @p0mvn .

Show cosmwasm pools in route vis, now that we're expected to route through astroport pools in addition to transmuter pools. Also, this updates the router procedure to use SQS pools query to get pool info instead of imperator. This should improve reliability.

Ignore the pool ID, but it will look like this:
Screenshot 2024-01-19 at 12 56 23 PM


sunnya97:
Can we make it so it recognizes Astroport PCL pools, and says "Astroport" with a little Astroport logo, instead of just custom?

jonator:
> Can we make it so it recognizes Astroport PCL pools, and says "Astroport" with a little Astroport logo, instead of just custom?

What are the code IDs for mainnet and testnet? @sunnya97

p0mvn:
@jonator just noticed the question and happened to have this info

for mainnet: https://github.com/osmosis-labs/sqs/blob/0ca01db2e227c9dd07018f69bafc6b0c109558ea/config.json#L23-L24

for testnet: https://github.com/osmosis-labs/sqs/blob/0ca01db2e227c9dd07018f69bafc6b0c109558ea/config-testnet.json#L23-L24

Note that we don't have a code ID for astroport on mainnet yet. Also, the higher transmuter code ID is the transmuter V2

Boss is tracking transmuter code IDs here: https://github.com/osmosis-labs/transmuter?tab=readme-ov-file#stored-codes

jonator:
@sunnya97 since Astroport is still not deployed on mainnet, thoughts on adding a follow up PR with an additional pool type for astroport later?

jonator:
@sunnya97 added astroport PCL

DavideSegullo (approved):

PR #2714, opened 2024-01-19
Mattupham/components demo page 1

mattupham:

- design needs an easy way to see our current components, in order to provide feedback

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a1ztxfk)

## Brief Changelog

- create a simple /components page, add control and button components to start
- redirect to / on prod (don't allow access)
- remove some unused components

## Testing and Verifying

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/00d889f1-b017-4126-941a-7f6910413d2d

### DIFF
--- a/packages/web/pages/components.tsx
+++ b/packages/web/pages/components.tsx
@@ -179,7 +179,6 @@ const Radios = () => {
 const Buttons = () => {
   return (
     <Card title="Buttons">
-      <Component title="Typography"></Component>
       <Component title="Regular">
         <Button onClick={() => console.log("clicked")}>Click</Button>
       </Component>


jonator (approved):
These PRs would also be a good opportunity to revisit and clean up naming, folder and file organization to make things easier to discover and more navigable.

One naming improvement I see could be to change the name of "stake tab" to be more generic, like "tabs" or something. Since future uses may not be specific to staking.

DavideSegullo (commented):
Looks good in general I only have two notes:

1. What do you think if we use composition pattern for components instead of using too many props? Like shadcn and other libraries do? example [here](https://ui.shadcn.com/docs/components/select)
2. Maybe we could think about supporting RSC by default, I think we can get it by simply specifying 'use client' on the client components (It could be out of scope, but can help us in the future if we choose to switch to the app router).

PR #2715, opened 2024-01-21
Prevent TVL calculation from crashing table

jonator:
### ClickUp Task


sunnya97 (approved):
utAck

PR #2720, opened 2024-01-23
Feat: add search filter inside pools page url query

DavideSegullo:
added search filter inside pools page query filters.

In the pools page, it is convenient to keep track in the query params also of the text string searched by the user, so that the page state can be restored when sharing the link.
It is also useful to be able to open the pools page in which I display for example only OSMO pools.

### ClickUp Task

### DIFF
--- a/packages/web/hooks/input/use-search-query-input.ts
+++ b/packages/web/hooks/input/use-search-query-input.ts
@@ -1,5 +1,6 @@
-import { useCallback, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo } from "react";
 
+import { useQueryParamState } from "~/hooks/window/use-query-param-state";
 import type { Search } from "~/utils/search";
 
 import { useDebouncedState } from "../use-debounced-state";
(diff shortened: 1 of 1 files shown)


kamal-sutra (approved):
Works as expected

CryptoAssassin1 (approved):

packages/web/hooks/input/use-search-query-input.ts
```diff
@@ -8,20 +9,30 @@ import { useDebouncedState } from "../use-debounced-state";
...
     },
-    [setDebouncedSearchInput]
+    [setDebouncedSearchInput, setSearchInput_]
   );
 
+  useEffect(() => {
+    if (searchInput && isReady) {
+      setSearchInput(searchInput);
+    }
+  }, [isReady]);
```
jonator, line 34: Is this effect needed? Is this initializing the state?

DavideSegullo: I would rather not, maybe some refactoring should be done, the problem is that a state is kept inside `AllPoolsTable` which is called `searchQuery` which is an object, but the method that sets it is injected inside `TableControls`, if I don't use the `useEffect` when there is the first load of the page, the filter state passed to trpc is the empty one, as the one `searchQuery` is undefined, because nobody called `setSearchQuery`.

I didn't make a complete change because I don't know what other cases you need to have as a `searchQuery` object. 
It may make sense to make the `search` parameter that we pass to trpc a useMemo (I see that we are not currently using the limit field)

DavideSegullo: If we can modify it that way, I can remove the effect, because the state would remain consistent

MaxMillington (approved):

PR #2721, opened 2024-01-23
Use SSR computation for poolId parameter

fabryscript:

This makes the computation of the `poolId` parameter present in the `[id].tsx` page server side in order to fix the issue of the id not displaying correctly in link previews

## Testing and Verifying

### DIFF
--- a/packages/web/pages/pool/[id].tsx
+++ b/packages/web/pages/pool/[id].tsx
@@ -1,4 +1,5 @@
 import { observer } from "mobx-react-lite";
+import { GetServerSideProps } from "next";
 import { useRouter } from "next/router";
 import { NextSeo } from "next-seo";
 import { FunctionComponent, useEffect, useMemo, useState } from "react";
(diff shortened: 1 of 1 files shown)


fabryscript:
As we saw previously in the PR #2666, the result of using `getStaticProps` & `getStaticPaths` is an OOM error which is probably caused by the eccessive size of the pages that needs to be iterated through in order to obtain all of the possible ids in `getStaticPaths`. For this reason, I do think that it is better to fix the issue by using SSR and accepting the tradeoff of the possibility that the initial loading time could be a little slower than usual

jonator:
From the preview, I haven't found the SSR to cause much of a noticeable perf problem.

jonator (approved):

DavideSegullo (approved):

fabryscript:
> @fabryscript what's the status of this one?

It seems to work fine

fabryscript:
@MaxMillington Are there any updates concerning to the bounty redeem? I have not yet recieved a response from the first mail they sent me

MaxMillington:
@fabryscript it takes around a month for them to process. If you have not received it after a month, let me know and I will take care of it

fabryscript:
> @fabryscript it takes around a month for them to process. If you have not received it after a month, let me know and I will take care of it

Perfect, thanks for the response

PR #2728, opened 2024-01-24
tRPC: User asset breakdown

jonator:

Screenshot 2024-01-24 at 5 01 21 PM

Adds the tRPC procedure needed for the new header in the assets page. This procedure iterates over all of a users bank balances (including GAMM shares), CL positions, locked GAMM shares, and staked OSMO and creates CoinPretty objects from the asset list and finds the total fiat value using our price function.

Some feedback would be appreciated on where these getUserAsset functions should live. Perhaps placing them in complex/assets/user would be too much long term. We could break them down into files by asset source such as bank, staking, CL, etc.

I want to add some unit tests with mock API requests to make sure the aggregation logic is sound once dev feedback is provided. But it's mostly just running simple queries and passing denoms by the asset list.

### ClickUp Task

## Brief Changelog

* Adds getUserAssets type functions in complex/assets/user file
* Adds a few missing queries
* Adds some utilities for aggregating and finding calculating the value of a list of coins

## Testing and Verifying

Use placeholder UI to ensure the balances appear accurate based on an account's assets. Any transactions should cause the balances to refresh.


packages/web/server/api/edge-routers/assets-router.ts
```diff
@@ -34,7 +35,7 @@ const GetInfiniteAssetsInputSchema = InfiniteQuerySchema.merge(
 ).merge(UserOsmoAddressSchema);
 
 export const assetsRouter = createTRPCRouter({
-  getAsset: publicProcedure
+  getUserAsset: publicProcedure
```
JoseRFelix, line 38: Given that the osmo address is optional, I think it would be better to use a more general name, such as getAssets. What do you think?

packages/web/utils/compare.ts
```diff
@@ -47,7 +47,8 @@ export function compareDec(a: Dec, b: Dec): CompareResult {
   return 0;
 }
 
-/** Prefers an object that has a given member defined and non-null. */
+/** Prefers an object that has a given member defined and non-null.
+ *  **DOES NOT** compare the value. */
 export function compareDefinedMember<T extends object>(
```
JoseRFelix, line 52: If this is only comparing definition, what if we rename this to `compareMemberDefinition`?

JoseRFelix (commented):
This looks very clean, great work 🔥

DavideSegullo (approved):

JoseRFelix (approved):
LGTM!

PR #2731, opened 2024-01-25
Fix: swap tool reset on page change

DavideSegullo:
In production, when I go to the detail page of a token and then change assets via the related assets section, the swap component does not update with the new default values to make the swap.
This PR fixes this misbehavior

### ClickUp Task

### DIFF
--- a/packages/web/hooks/use-swap.ts
+++ b/packages/web/hooks/use-swap.ts
@@ -451,6 +451,11 @@ function useToFromDenoms(
     initialToDenom
   );
 
+  useEffect(() => {
+    setToAssetState(initialToDenom);
+    setFromAssetState(initialFromDenom);
+  }, [initialFromDenom, initialToDenom]);
+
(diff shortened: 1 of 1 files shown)


jonator (commented):
Is it ok if the token drawers can now be used to select other tokens? IMO it's fine as long as the tokens are set initially to the token for the current token page

DavideSegullo:
> Is it ok if the token drawers can now be used to select other tokens? IMO it's fine as long as the tokens are set initially to the token for the current token page

Ye I think is fine, that's the right behaviour

jonator:
@DavideSegullo actually I take it back the drawers aren't available anymore, but this is fixed

jonator (approved):

kamal-sutra (approved):
Fixed. works fine now

CryptoAssassin1 (approved):

PR #2735, opened 2024-01-26
Pools table: allow simultaneous sort & search

jonator:
* Allow sorting while searching on pools table


CryptoAssassin1:
Search by pool number is buggy.

PR #2736, opened 2024-01-26
feat: timeout height (transaction TTL)

p0mvn:

Adding TX TTL for better feedback during spam and timely mempool eviction.

This PR introduces and additional field for transaction body and makes the necessary changes to serialize and sign it.

This will allow the chain to evict the transaction from mempool after reaching a certain height. This will be useful during times of spam or congestion where users will get some sort of feedback about transaction failure.

Today, due to a poor chain UX, the feedback is not always present.

The configuration is set to 10. At the current block time of 5 seconds, this gives user transactions approximately 50 seconds to succeed.

### Next steps

- We should set the timeout offset value via config: https://github.com/osmosis-labs/osmosis-frontend/issues/2740
- We should also update `cosmjs` and `cosmos-kit` deps as they are outdated, requiring us to reimplement some helpers in this PR: https://github.com/osmosis-labs/osmosis-frontend/issues/2738

### Test TX

https://www.mintscan.io/osmosis/tx/A336F54BDAB9EC2FAECE6D8F376C4BC2A19C565A04D444E4EAF7AAA4BE8B824B

A few more were tested from the address

### DIFF
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -3,10 +3,9 @@ import {
   AminoMsg,
   encodeSecp256k1Pubkey,
   OfflineAminoSigner,
-  StdFee,
 } from "@cosmjs/amino";
 import { fromBase64 } from "@cosmjs/encoding";
-import { Int53, Uint53 } from "@cosmjs/math";
+import { Int53 } from "@cosmjs/math";
 import {
   EncodeObject,
   encodePubkey,
@@ -85,6 +84,7 @@ import {
   getEndpointString,
   getWalletEndpoints,
   logger,
+  makeSignDocAmino,
   removeLastSlash,
   TxFee,
 } from "./utils";
@@ -1146,42 +1146,4 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
       },
     });
   }
-}
-
-/**
- * The document to be signed
- * Referenced from CosmjsL
- * https://github.com/cosmos/cosmjs/blob/287278004b9e6a682a1a0b1664ba54646f65a1a0/packages/amino/src/signdoc.ts#L21-L35
- *
- * We copied it over to work around dependency updates.
- */
-export interface StdSignDoc {
-  readonly chain_id: string;
-  readonly account_number: string;
-  readonly sequence: string;
-  readonly fee: StdFee;
-  readonly msgs: readonly AminoMsg[];
-  readonly memo: string;
-  readonly timeout_height?: string;
-}
-
-// Creates the document to be signed from given parameters.
-export function makeSignDocAmino(
-  msgs: readonly AminoMsg[],
-  fee: StdFee,
-  chainId: string,
-  memo: string | undefined,
-  accountNumber: number | string,
-  sequence: number | string,
-  timeout_height?: bigint
-): StdSignDoc {
-  return {
-    chain_id: chainId,
-    account_number: Uint53.fromString(accountNumber.toString()).toString(),
-    sequence: Uint53.fromString(sequence.toString()).toString(),
-    fee: fee,
-    msgs: msgs,
-    memo: memo || "",
-    ...(timeout_height && { timeout_height: timeout_height.toString() }),
-  };
 }
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -2,11 +2,11 @@ import type { AssetList as CosmologyAssetList } from "@chain-registry/types";
 import {
   AminoMsg,
   encodeSecp256k1Pubkey,
-  makeSignDoc as makeSignDocAmino,
   OfflineAminoSigner,
+  StdFee,
 } from "@cosmjs/amino";
 import { fromBase64 } from "@cosmjs/encoding";
-import { Int53 } from "@cosmjs/math";
+import { Int53, Uint53 } from "@cosmjs/math";
 import {
   EncodeObject,
   encodePubkey,
@@ -52,6 +52,7 @@ import {
   getSourceDenomFromAssetList,
   isNil,
 } from "@osmosis-labs/utils";
+// eslint-disable-next-line import/no-extraneous-dependencies
 import axios from "axios";
 import { Buffer } from "buffer/";
 import cachified, { CacheEntry } from "cachified";
@@ -63,6 +64,7 @@ import {
   TxBody,
   TxRaw,
 } from "cosmjs-types/cosmos/tx/v1beta1/tx";
+import Long from "long";
 import { LRUCache } from "lru-cache";
 import { action, makeObservable, observable, runInAction } from "mobx";
 import { fromPromise, IPromiseBasedObservable } from "mobx-utils";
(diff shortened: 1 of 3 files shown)


p0mvn:
I'm currently blocked on:
image


This points to the fact that on-chain signature verification does not match the signed data sent from client.

I suspect that the error stems from this [call](https://github.com/osmosis-labs/osmosis-frontend/blob/8b20ba00dfea04d758a1d956c60cacd9dd9ad41a/packages/stores/src/account/base.ts#L778) that operates on the outdated `StdSignDoc` type in the `cosmos-kit` dependency

I tried hardcoding `timeout_height` to zero. In that case, the TX would succeed. Currently, hardcoded it to non-zero value in the future for ease of debugging.

I also tried regenerating types but that did not seem to help

Guidance and suggestions are appreciated 🙏

packages/proto-codecs/scripts/get-proto.sh
```diff
@@ -6,7 +6,7 @@ GREEN='\033[0;32m' # Green color
 NC='\033[0m' # No Color
 
 PROTO_DIR="./chain-proto"
-OSMOSIS_COMMIT_HASH="b09f8724030c351e5cbc96b25a84ea0d114d3f34"
+OSMOSIS_COMMIT_HASH="3f92949da8c79208ad34cd7eda4a13275dabd617"
```
p0mvn, line 9: Note: this did not help :(

packages/stores/src/account/base.ts
```diff
@@ -759,6 +785,8 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
           signDoc
         ));
 
+    console.log("signed", signed);
```
p0mvn, line 788: Note: I think that the problem stems from the outdated call to dependencies on the few lines above

p0mvn:
I was able to execute a tx with non-zero timeout height:
https://www.mintscan.io/osmosis/tx/9E126F1D467E61D6EF22DAC52A16CA96E9518FE6CB9DF50C4DC6C9A4BED66502?height=13468122

with 
```
osmosisd tx poolmanager swap-exact-amount-in 1000000uosmo 2 --swap-route-pool-ids 2 --swap-route-denoms uion --from local --chain-id osmosis-1 --sign-mode amino-json --fees 2000uosmo --timeout-height 13567584
```

JoseRFelix:
@p0mvn Can you try upgrading cosmjs? Ignore the typescript errors for now. I just want to confirm something

p0mvn:
> @p0mvn Can you try upgrading cosmjs? Ignore the typescript errors for now. I just want to confirm something

@JoseRFelix attempted in separate PR here:
https://github.com/osmosis-labs/osmosis-frontend/pull/2737

packages/stores/src/account/base.ts
```diff
@@ -1116,3 +1150,41 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
+  return {
+    chain_id: chainId,
+    account_number: Uint53.fromString(accountNumber.toString()).toString(),
+    sequence: Uint53.fromString(sequence.toString()).toString(),
+    fee: fee,
+    msgs: msgs,
+    memo: memo || "",
+    ...(timeout_height && { timeout_height: timeout_height.toString() }),
+  };
+}
```
p0mvn, line 1190: These can be removed as part of: https://github.com/osmosis-labs/osmosis-frontend/issues/2738

packages/stores/src/account/base.ts
```diff
@@ -91,6 +93,14 @@ import { WalletConnectionInProgressError } from "./wallet-errors";
 
 export const GasMultiplier = 1.5;
 
+// The number of heights from current before transaction times out.
+// 10 heights * 5 second block time = 50 seconds before transaction
+// timeout and mempool eviction.
+const timeoutHeightOffset: bigint = BigInt(10);
```
p0mvn, line 98: Will move to Vercel config: https://github.com/osmosis-labs/osmosis-frontend/issues/2740

packages/stores/src/account/base.ts
```diff
@@ -892,6 +923,8 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
         true
       );
 
+      console.log("account", account.getSequence());
+
```
p0mvn, line 927:
```suggestion
```

kamal-sutra (approved):
All different txs are successful within 10 block heights of timeout

packages/stores/src/account/base.ts
```diff
@@ -1116,3 +1147,41 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
+ */
+export interface StdSignDoc {
+  readonly chain_id: string;
+  readonly account_number: string;
+  readonly sequence: string;
+  readonly fee: StdFee;
+  readonly msgs: readonly AminoMsg[];
+  readonly memo: string;
+  readonly timeout_height?: string;
+}
```
JoseRFelix, line 1166: Let's move this to the `./types` folder

packages/stores/src/account/base.ts
```diff
@@ -1116,3 +1147,41 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
+  return {
+    chain_id: chainId,
+    account_number: Uint53.fromString(accountNumber.toString()).toString(),
+    sequence: Uint53.fromString(sequence.toString()).toString(),
+    fee: fee,
+    msgs: msgs,
+    memo: memo || "",
+    ...(timeout_height && { timeout_height: timeout_height.toString() }),
+  };
+}
```
JoseRFelix, line 1187: Let's move this to the `./utils` folder

JoseRFelix (approved):
Thank you for cleaning this up! I left two nits, but this should be ready to go 🚀

CryptoAssassin1 (approved):
Worked fine

ValarDragon:
Lets increase this beyond 10 blocks imo. I recommend 50 blocks for now

p0mvn:
> Lets increase this beyond 10 blocks imo. I recommend 30 blocks for now
> 
> Its even worse imo if the tx would get in within 2 minutes, and we evict it. Remember that:
> 
> * Were trying to lower block times
> * Latest height queries can under relatively normal behavior be 1-2 blocks behind on heavy query serving nodes

Thinking from a user's perspective, would I even want to wait for my transaction for 2 minutes?

I think it depends. For something like increasing collateral when close to being liquidated - definitely.

However, for swaps, not really. For a smaller value swap, I would probably refresh way before 2 minutes and not care enough. For a larger swap value, I would prefer for it to timeout sooner (max 50 seconds or 10 blocks) so that I can get the updated information

Your suggestion is probably right - it is best to prefer the former case generally and add custom timeouts for swaps in the future. 

I'll change this to 30. This will be moved into Vercel env [here](https://github.com/osmosis-labs/osmosis-frontend/issues/2740) so we can discuss and adapt it more.

Please let me know what you think

packages/stores/src/account/base.ts
p0mvn: My personal preference is to avoid bloating utils files or packages and prevent them from growing out of control.

A general rule I aim to follow is to move something into `utils` if it is used in multiple sources. Otherwise, keep a helper function locally to where it is used.

However, happy to not over-index on preference and will apply the suggestion. Just wanted to share my thoughts.

p0mvn:
Applied all suggestions - merging

PR #2737, opened 2024-01-26
chore: try upgrade cosmjs for TTL

p0mvn:

Trying to upgrade cosmjs and cosmos-kit for TTL

### ClickUp Task


p0mvn:
I'm seeing this printed:
image

Note that `timeout_height` seems to be present there

p0mvn:
Please note that I'm able to let this succeed via CLI:
```
osmosisd tx poolmanager swap-exact-amount-in 1000000uosmo 2 --swap-route-pool-ids 2 --swap-route-denoms uion --from local --chain-id osmosis-1 --sign-mode amino-json --fees 2000uosmo --timeout-height 13567584 --output=json --generate-only | jq .
{
  "body": {
    "messages": [
      {
        "@type": "/osmosis.poolmanager.v1beta1.MsgSwapExactAmountIn",
        "sender": "osmo1q8709l2656zjtg567xnrxjr6j35a2pvwhxxms2",
        "routes": [
          {
            "pool_id": "2",
            "token_out_denom": "uion"
          }
        ],
        "token_in": {
          "denom": "uosmo",
          "amount": "1000000"
        },
        "token_out_min_amount": "2"
      }
    ],
    "memo": "",
    "timeout_height": "13567584",
    "extension_options": [],
    "non_critical_extension_options": []
  },
  "auth_info": {
    "signer_infos": [],
    "fee": {
      "amount": [
        {
          "denom": "uosmo",
          "amount": "2000"
        }
      ],
      "gas_limit": "200000",
      "payer": "",
      "granter": ""
    },
    "tip": null
  },
  "signatures": []
}
```

which matches body on the screenshot:
image


Therefore, I suspect that the issue must have something to do with the signing

p0mvn:
@JoseRFelix I connected FE deployment from this branch to my node, and it seems that what the timeout height received on the node is zero which is in conflict to what I see in `console.log` on the FE side.

Do you have any insight into why that could be? Is TX data being sent to node from a different code area from which I'm making the modification

image
- I confirmed that this is the TX I sent from FE

packages/stores/src/account/base.ts
```diff
@@ -770,13 +802,19 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
     };
 
+    console.log("signedTxBody", signedTxBody);
+
     const signedTxBodyEncodeObject = {
       typeUrl: "/cosmos.tx.v1beta1.TxBody",
       value: signedTxBody,
     };
 
+    console.log("signedTxBodyEncodeObject", signedTxBodyEncodeObject);
```
p0mvn, line 816: Here, I still see that the timeout height is set

packages/stores/src/account/base.ts
```diff
@@ -792,6 +830,14 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
       signMode
     );
 
+    console.log(
+      "decode bytes",
+      wallet?.signingStargateOptions?.registry?.decodeTxBody(
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        signedTxBodyBytes!
+      )
+    );
```
p0mvn, line 839: Here, it decodes to zero.

Therefore, something in the serialization logic is broken

PR #2742, opened 2024-01-27
Feat: improve pool search performance

DavideSegullo:
this pr fixes the performance issues introduced in the PR #2720 , the input had some delays. Also there are improvements in general, now the default values don't need to be in the uri (anyway if the user doesn't make changes they will always be those, so they are only set later, this avoids having to use the router even before it is needed and we don't have to wait until it is ready to make a request to trpc anymore)

### ClickUp Task


packages/web/components/complex/all-pools-table-v2.tsx
```diff
@@ -34,11 +38,67 @@ import { AprDisclaimerTooltip } from "../tooltip/apr-disclaimer";
...
+] as const;
+
+const marketIncentivePoolsSortKeys = [
+  "totalFiatValueLocked",
+  "feesSpent7dUsd",
+  "feesSpent24hUsd",
+  "volume7dUsd",
+  "volume24hUsd",
+  "aprBreakdown.total",
+] as const;
```
jonator, line 59: Would it be possible to import these from the server?

packages/web/components/complex/all-pools-table-v2.tsx
```diff
@@ -34,11 +38,67 @@ import { AprDisclaimerTooltip } from "../tooltip/apr-disclaimer";
...
+        parseAsStringLiteral<PoolTypeFilter>(poolTypes)
+      ).withDefault([...poolTypes]),
+      poolIncentivesFilter: parseAsArrayOf<PoolIncentiveFilter>(
+        parseAsStringLiteral<PoolIncentiveFilter>(incentiveTypes)
+      ).withDefault([...incentiveTypes]),
+    },
+    {
+      history: "push",
+    }
+  );
```
jonator, line 92: Would there be any unintended consequences of not using next router?

packages/web/components/complex/all-pools-table-v2.tsx
DavideSegullo: We can share them from a common folder, but we can't import them from the server folder because when you do this it also imports the trpc initializer and gives error

packages/web/components/complex/all-pools-table-v2.tsx
DavideSegullo: Internally uses it with some optimizations to avoid unnecessary re-rendering, the good part of the library is that it is also compatible with the router app, if we migrate in the future we don't have to rewrite all the utilities

packages/web/components/input/search-box.tsx
```diff
@@ -40,6 +48,11 @@ type SearchBoxProps = Omit<InputProps<string>, "currentValue"> &
   VariantProps<typeof searchBoxClasses> & {
     type?: string;
     currentValue?: string;
+    /**
+     * adds the ability to output the searchbox event using a debounce effect,
+     * which is useful to avoid making too many requests.
+     */
+    debounce?: number;
```
jonator, line 55: So you've built debounce feature directly into the component itself? Perhaps this should be used over all `useSearchQueryInput` uses. I'll keep this in mind going forward and with new assets table

packages/web/components/complex/all-pools-table-v2.tsx
```diff
@@ -90,23 +146,24 @@ export const AllPoolsTable: FunctionComponent<{
...
-            direction: sortDirection,
+            direction: sortParams.allPoolsSortDir,
           }
         : undefined,
       minLiquidityUsd: 1_000,
     },
     {
       getNextPageParam: (lastPage) => lastPage.nextCursor,
       initialCursor: 0,
-      enabled: isQueryParamsReady,
```
jonator, line 109: I used this to prevent queries from being sent before query param was loaded in future renders. It seems nuqs somehow loads query params on first render? As I don't see any queries with unloaded inputs, which is very nice.

jonator (approved):
Perhaps we should be using this everywhere instead of my custom `useQueryParamState`?

packages/web/components/input/search-box.tsx
DavideSegullo: Yes, because it allows us to avoid having a dedicated hook and allows us not to have a double state

CryptoAssassin1 (approved):
worked fine

packages/web/components/complex/all-pools-table-v2.tsx
DavideSegullo: Yes, because it actually works differently, it builds an object with default values, if we change the default values then it puts them in the url.
This avoids having to wait for the router to be ready, plus if you don't change the filters in the url there are no "redundant" strings

kamal-sutra (approved):

MaxMillington (approved):

PR #2743, opened 2024-01-28
Use tRPC for CL positions UI

jonator:

Use the tRPC procedure @JoseRFelix created (as a starting point) for fetching data needed for expandable user position cards.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a22kxqt)

## Brief Changelog

* Combines CL tRPC procedures into one
  * Removes some excessive querying in those procedures (some queries we're unnecessarily sent per user position vs per user)
* Removes some unnecessary backend code, and removes caching of user account queries
* Integrates procedures with UI
* Removes some refs to chain and price stores
* Query price and depth chart data only when position card is expanded. This was needed since we use tRPC for pair historical price data instead of MobX query stores, which used to only become observed and send the query when the card was expanded

## Testing and Verifying

CL position data should be the same.

* All behaviors, including superfluid staking and unbonding, should work as exists in prod
* All position tx modals and txs should work as expected


packages/utils/src/api-client.ts
```diff
@@ -59,7 +59,7 @@ export async function apiClient<T>(
         data = await response.json();
       } catch (e) {
         throw new ApiClientError({
-          message: UNEXPECTED_ERROR_MESSAGE,
+          message: `JSON parse error: ${e}`,
```
JoseRFelix, line 62: 👍

packages/web/server/queries/complex/assets/index.ts
```diff
@@ -86,26 +84,19 @@ export async function getAssets({
 }
 
 /**
- * This function maps raw assets to a CoinPretty. This is useful for
+ * This function coins to a CoinPretty if listed in asset list. This is useful for
```
JoseRFelix, line 87:
```suggestion
 * This function converts coins to a CoinPretty if listed in asset list. This is useful for
```

packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
```diff
@@ -24,80 +14,9 @@ export const concentratedLiquidityRouter = createTRPCRouter({
...
-          .sub(totalPrincipalValue.toDec())
-          .quo(totalPrincipalValue.toDec())
-      );
-
-      return {
-        principalAssets,
-        totalEarned,
-        roi,
-      };
-    }),
```
JoseRFelix, line 102: I believe we should leave position performance on a separate query. Maybe even separate the position details into more queries. It's nicer to see data load bit by bit rather than waiting for everything. Plus, if there's an error, it won't mess up the whole list like we've seen with Imperator

packages/web/utils/coin.ts
```diff
@@ -1,15 +1,20 @@
 import { CoinPretty } from "@keplr-wallet/unit";
 
 /** Combines coins of the same denom into the same CoinPretty object by sum. */
-export function aggregateCoinsByDenom(coins: CoinPretty[]): CoinPretty[] {
+export function aggregateCoinsByDenom<TCoin extends CoinPretty>(
+  coins: TCoin[]
```
JoseRFelix, line 5: I suggest we omit the `T` since we shouldn't mind if it's a generic or not
```suggestion
  coins: Coin[]
```

packages/web/server/queries/complex/assets/index.ts
```diff
@@ -86,26 +84,19 @@ export async function getAssets({
...
  *
- * @returns A promise that resolves to an array of CoinPretty objects. Each CoinPretty object represents an asset and has an optional 'fiatValue' property.
+ * @returns A promise that resolves to an array of CoinPretty objects. Each CoinPretty object represents an asset that is listed. Unlisted assets are filtered.
  */
-export async function mapAssetsToCoins({
-  rawAssets,
-  calculatePrice,
-}: {
-  rawAssets?: {
+export async function mapListedCoins(
```
JoseRFelix, line 94: What do you think if we rename this function to something more specific?
```suggestion
export async function mapRawCoinToPretty(
```

JoseRFelix (changes requested):
This is looking very clean 🔥 Left some comments

packages/web/server/api/edge-routers/concentrated-liquidity-router.ts
jonator: I think combining all details into a single query greatly simplifies code complexity, especially on the client. Also, ROI calculation in the header of the position card needs the metrics in the expanded card, so there's not a clean break of collapsed vs expanded data.

To address the potential error issues with imperator (which I've brought up to their team), what are your thoughts on making the imperator portions optional and returning undefined if there's issues?

JoseRFelix: Although it adds a bit more complexity, having a quicker initial load is worth it for users IMO. Splitting into more queries isn't too complicated, but we can handle that in a separate PR. For now, I agree that we shouldn't throw if Imperator fails, and just return `undefined` instead

JoseRFelix (approved):
LGTM

DavideSegullo (approved):

JoseRFelix:
@jonator I'm unable to open one of my positions which is superfluid. Notice how the console is triggering too many requests to keybase

https://github.com/osmosis-labs/osmosis-frontend/assets/21092519/0f026f03-959c-4ded-a117-c3eb8e1e857b

jonator:
> I'm unable to open one of my positions which is superfluid. Notice how the console is triggering too many requests to keybase

@JoseRFelix I think this is because of the validator select modal. I'll try to only render that when needed.

jonator:
@JoseRFelix that's unrelated to this PR. Not sure why it's affecting this functionality. I'm unable to replicate

JoseRFelix:
@jonator this error is related to this PR. Seeing this in local dev 
image

PR #2744, opened 2024-01-29
feat: feedback for tx ttl after timeout

p0mvn:

This PR extends #2736 with user feedback.

Now, once the transaction is broadcasted, we will repeatedly query the latest height until either transaction succeeds or timeout height is reached.

Made sure to return the same error independent from source (tx broadcast vs query)

## Error Display

image

This is the error that is returned after 10 heithgs (approx 50 seconds).

## Future

In the future, we can implement a countdown or a status bar.

## Testing

Tested by hardcoding the timeout offset to 1 height and repeating multiple tests like on the attached screenshot

This error can be triggered in broadcast mode by pausing in the "submit transaction Keplr view" for more than a minute.
image


packages/stores/src/tx/tracer.ts
```diff
@@ -273,14 +334,19 @@ export class TxTracer {
...
+            params,
+            resolver: resolve,
+            rejector: reject,
+          });
+
+          this.sendSubscribeTxRpc(id, params);
+          resolve({ id });
+        } catch (e) {
+          reject(e);
+        }
```
p0mvn, line 349: Note: this promise was not resolved properly. While I acknowledge that there are cleaner ways to address this by performing a bigger refactor to switch to async/await generally.

However, I opted in for a local change to limit the scope and risk around this critical but brittle part of the codebase.

packages/web/localizations/en.json
```diff
@@ -379,7 +379,8 @@
     "chartUnavailable": "Chart unavailable",
     "sequenceNumber": "You have too many concurrent transactions! Try resending after your prior tx lands on chain. (Sequence: {sequence})",
     "swapSlippage": "Swap for {coinDenom} failed due to slippage. Try increasing the slippage tolerance.",
-    "invalidAmounts": "Invalid amount(s)"
+    "invalidAmounts": "Invalid amount(s)",
+    "txTimedOutError": "Transaction timed out. Please retry."
```
p0mvn, line 383: Note: I considered also formatting height but I don't think it would be very useful for users in the current state.

If we do a count to timeout height in the future, it will become more useful

CryptoAssassin1 (approved):
Worked fine but getting tx failed message on ibc deposit/withdraw.
Screenshot from 2024-01-29 15-03-18

p0mvn:
Need to take a look at the IBC transfer issues reported in QA.

packages/web/stores/ibc-notifier.tsx
```diff
@@ -21,6 +21,8 @@ export const IbcNotifier: FunctionComponent = observer(() => {
     if (!historyHandlerAdded) {
       ibcTransferHistoryStore.addHistoryChangedHandler((history) => {
         if (chainStore.hasChain(history.destChainId)) {
+          console.log("history", history);
```
p0mvn, line 24: Note: This line is being triggered twice per IBC transfer. I'm failing to detect what is the retigger

JoseRFelix (approved):
LGTM!

Issue #2739, opened 2024-01-27
chore: set up VS code debugger for debugging the application with breakpoints e2e


p0mvn:
We need the ability to press a button and start an app with a debugger running for developer productivity.

Acceptance Criteria
- `launch.json` configuration
- I press a button and everything works
- Can stop on breakpoint

p0mvn:
I did some investigations. The closest I could get is being able to set breakpoints in the `web` package only with the following configuration:
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "node",
            "request": "launch",
            "name": "Next.js Node Debug",
            "runtimeExecutable": "${workspaceFolder}/node_modules/next/dist/bin/next",
            "cwd": "${workspaceFolder}/packages/web",
            "console": "integratedTerminal"
        }
    ]
}
```

Opened a discussion in the turbo repo: https://github.com/vercel/turbo/discussions/7135

Sources that I found useful:
- https://github.com/vercel/turbo/discussions/4860
- https://github.com/vercel/turbo/discussions/2639

MaxMillington:
Closing this as most JS devs use the chrome debugger tool to step through code

PR #2745, opened 2024-01-29
chore: web package debugger configuration for vs code

p0mvn:
Progress towards: https://github.com/osmosis-labs/osmosis-frontend/issues/2739

This PR adds a VS code debugger configuration.

It enables starting the app and setting breakpoints in the `web` package. I'm still investigating how to achieve setting breakpoints in other packages and tracking progress in #2739.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,13 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Next.js Node Debug",
+            "runtimeExecutable": "${workspaceFolder}/node_modules/next/dist/bin/next",
+            "cwd": "${workspaceFolder}/packages/web",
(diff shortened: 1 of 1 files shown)


jonator:
This is just for stepping through the backend?

jonator:
For client, it's possible to set breakpoints in the source maps in the source tab in chrome dev tools

MaxMillington (approved):
can I introduce you to `console.log` 😂

p0mvn:
> This is just for stepping through the backend?
> For client, it's possible to set breakpoints in the source maps in the source tab in chrome dev tools

This is still only relevant to the `web` package (client) but it is already helping me run and debug it directly through vs code, offering more convenience. I will continue investigating how to also enable other packages.

> can I introduce you to console.log 😂

I like being in the driver's seat 🤣

p0mvn:
Merging with one ACK since this is a config update

PR #2747, opened 2024-01-29
Router: silent failure for missing asset price config

jonator:

As with stDYDX, if an asset isn't configured for pricing yet we should not let that fail the router queries. Instead, we can simply omit the price data from the response.

### DIFF
--- a/packages/web/server/api/edge-routers/swap-router.ts
+++ b/packages/web/server/api/edge-routers/swap-router.ts
@@ -110,15 +110,15 @@ export const swapRouter = createTRPCRouter({
           ? await calcAssetValue({
               anyDenom: tokenInDenom,
               amount: quote.tokenInFeeAmount,
-            })
+            }).catch(() => null)
           : undefined;
         const tokenOutPrice = await getAssetPrice({
           asset: { coinMinimalDenom: tokenOutDenom },
-        });
+        }).catch(() => null);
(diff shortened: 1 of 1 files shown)


MaxMillington (commented):
at some point we are going to have to look into re-adding/fixing sentry. Feels scary to openly be adding a silent failure. We should probably make a note/comment or track this effort somewhere to go back and fix it in the future?

jonator:
> at some point we are going to have to look into re-adding/fixing sentry. Feels scary to openly be adding a silent failure. We should probably make a note/comment or track this effort somewhere to go back and fix it in the future

@MaxMillington yeah I still haven't reached a good conclusion as far as error management. I think there's many errors that are low prio that shouldn't completely fail the query, like this one. But it'd be good to still know about the errors somehow, as if it doesn't fail the query it won't be detected in vercel runtime logs.

I think sentry is meant to be the solution to this as there's likely a way we can log silent errors into issues on their platform, as well as the critical errors. Ideally, we can categorize them.

JeremyParish69 (approved):
I'm in agreement with the idea, but haven't found an example to test this on

jonator:
@MaxMillington 
Tracking: https://app.clickup.com/t/86a2bav4e

jonator:
Going to merge this. Once we get to adding rich logging to our tRPC backend we can log these silenced errors. But for now they are non-critical

PR #2748, opened 2024-01-29
Use sidecar pool provider

jonator:

See if sidecar can return pools missing from imperator pools query


sunnya97 (approved):

MaxMillington (approved):
nice

jonator:
@sunnya97 do you think we should actually use this? I think it's slightly less performant (+1-2s without any cache) than the imperator pools query since we have to calculate TVL ourselves from using our prices and the balances from the pools

CryptoAssassin1 (approved):

kamal-sutra (approved):

PR #2749, opened 2024-01-29
allow the app to work even if developer does not have LD key in env

MaxMillington:
This PR will allow third party devs to get all the feature flags working even if they do not have the LD client ID by providing default values to the existing flags.

### DIFF
--- a/packages/web/hooks/use-feature-flags.ts
+++ b/packages/web/hooks/use-feature-flags.ts
@@ -26,10 +26,33 @@ export type AvailableFlags =
   | "topAnnouncementBanner"
   | "tfmProTradingNavbarButton";
 
-type ModifiedFlags =
-  | Exclude<AvailableFlags, "mobileNotifications">
-  | "_isInitialized"
-  | "_isClientIDPresent";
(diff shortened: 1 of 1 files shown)


packages/web/hooks/use-feature-flags.ts
```diff
@@ -26,10 +27,33 @@ export type AvailableFlags =
...
+  unlistedAssets: false,
+  earnPage: false,
+  sidecarRouter: true,
+  legacyRouter: true,
+  tfmRouter: true,
+  osmosisUpdatesPopUp: false,
+  aprBreakdown: true,
+  newPoolsTable: true,
+  topAnnouncementBanner: true,
+  tfmProTradingNavbarButton: true,
```
jonator, line 55: How should we decide what is default true, should we not just have all be true?

DavideSegullo (approved):

jonator:
@MaxMillington looks like there's a build error

packages/web/hooks/use-feature-flags.ts
MaxMillington: 🤷. I'm ok with them all being true.

PR #2753, opened 2024-01-30
feat: configure timeout height via environment variable

### LINKED ISSUE #2740: chore: move timeout height offset to be configured from Vercel
In https://github.com/osmosis-labs/osmosis-frontend/pull/2736, we introduced a `timeoutHeightOffset` for computing the timeout height relative to latest for a transaction.

We should instead have this valuebe set via a Vercel environment variable

**Acceptance Criteria**
- timeout height is not hardcoded but set via Vercel environment variable
- Configured for all of stage, prod, testnet

p0mvn:

Closes #2740

Setting up tx timeout offset to be set via environment variable.

Already added the vercel env for all environments:
image

## Testing

- Tested swaps in preview

### DIFF
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -74,6 +74,7 @@ import { aminoConverters } from "./amino-converters";
 import {
   AccountStoreWallet,
   DeliverTxResponse,
+  NEXT_TX_TIMEOUT_HEIGHT_OFFSET,
   RegistryWallet,
   TxEvent,
   TxEvents,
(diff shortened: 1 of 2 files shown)


czarcas7ic:
@p0mvn shouldn't you also add this for signAmino method as well? Not just signDirect?

MaxMillington (approved):
Does everyone need to add this to their `env.local` as a result of this change?

p0mvn:
> Does everyone need to add this to their `env.local` as a result of this change?

No, if unset, it will fallback to the default value of 30

p0mvn:
Merging this with one ACK since the change is low-risk.

This will also help me trigger redeployment of SQS dev environment

PR #2757, opened 2024-01-30
feat: add memo field for frontend txs

czarcas7ic:

If the memo field is empty, adds "FE" to the memo field. This will help us detect which txs that fail most originate from the frontend, so we can tweak as necessary. 

I was hoping to make the memo "FE-{hash}", if anyone knows of a simple way to do this, lmk.

## Testing and Verifying

Here is this being tested against a swap and collect CL rewards:

Screenshot 2024-01-31 at 10 10 39 AM

Screenshot 2024-01-31 at 10 10 01 AM


sunnya97:
@czarcas7ic what is the `{hash}`?

czarcas7ic:
@sunnya97 git hash of the live frontend build

czarcas7ic:
Any issue with getting this merged?

jonator:
@czarcas7ic perhaps the current FE commit hash is available in a vercel env var? We can attempt to get it, and if not available we can omit it.

packages/stores/src/account/base.ts
```diff
@@ -733,6 +733,12 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
       throw new Error("Failed to retrieve account from signer");
     }
 
+    // If the memo is empty, set it to "FE" so we know it originated from the frontend for
+    // QA purposes.
+    if (memo === "") {
+      memo = "FE";
+    }
```
p0mvn, line 740: Could we append it to the bottom of the memo if the memo is present? 

At the same time, it might make sense to make it clear to users that this metadata instrumentation is done by our app to avoid confusion. E.g. append the following:
"The flag below is appended by the Osmosis frontend\nFE"

czarcas7ic: Agreed offline we want the memo field to be super lean to not increase any on chain storage.

czarcas7ic: Added memo append logic here https://github.com/osmosis-labs/osmosis-frontend/pull/2757/commits/3e191db4f56d111b074569c36e3961965a6345a2

p0mvn (approved):

PR #2760, opened 2024-01-31
fix: improve slippage error message

p0mvn:

The error message for slippage is confusing because we have a bug. It should be this:
https://github.com/osmosis-labs/osmosis-frontend/blob/5abb4d43936d145eca2c93135c6d56b7db70ee50/packages/web/localizations/en.json#L779

but users see this:
image

Fixed and tested by hardcoding an inappropriately high slippage:
image


packages/stores/src/tx/utils.ts
```diff
@@ -1,10 +1,5 @@
 export function isSlippageError(tx: any): boolean {
-  if (
-    tx &&
-    tx.code === 7 &&
-    tx.codespace === "gamm" &&
```
p0mvn, line 5: I removed these because slippage errors can come from modules other than gamm and have a different code.

Given current chain setup, I think basic string matching is the most appropriate here

jonator (approved):

MaxMillington (commented):
I think we have more languages than this?

MaxMillington:
> I think we have more languages than this?

e.g. we are missing ru.json and ja.json @p0mvn

p0mvn:
> > I think we have more languages than this?
> 
> e.g. we are missing ru.json and ja.json @p0mvn

I deleted all translations other than `en.json` and ran `yarn machine-translate` and `yarn lint:fix`.

I can tell that `ru.json` is already translated correctly

MaxMillington:
> > > I think we have more languages than this?
> > 
> > 
> > e.g. we are missing ru.json and ja.json @p0mvn
> 
> I deleted all translations other than `en.json` and ran `yarn machine-translate` and `yarn lint:fix`.
> 
> I can tell that `ru.json` is already translated correctly

👍

MaxMillington (approved):

PR #2777, opened 2024-02-05
Implement new Home page

fabryscript:
### ClickUp Task


fabryscript:
TODO: add translation labels

packages/web/components/chart/home-chart.tsx
```diff
@@ -0,0 +1,89 @@
+import { AreaData, MouseEventParams, Time } from "lightweight-charts";
```
jonator, line 1: Why not visx? We already import it.

packages/web/components/home/hooks.ts
```diff
@@ -0,0 +1,114 @@
+import { TokenHistoricalPrice } from "@osmosis-labs/stores/build/queries-external/token-historical-chart/types";
```
jonator, line 1: Let's use the type from the server file instead. Be sure to use `import type` to only import the TS type.

jonator (commented):
Initial scan

packages/web/components/chart/home-chart.tsx
DavideSegullo: We had started out using visx, but there are behaviors that it doesn't handle natively, and in the reference indicated they were using lightweight-chart.

I can think whether it is possible to do the same things with visx, but right now the biggest problem is that the handling of asset x is a bit buggy on visx

MaxMillington:
@fabryscript @DavideSegullo Can we close this PR? I think we are starting the new homepage designs from scratch

PR #2790, opened 2024-02-07
Fix: pool ratio percentage calc

DavideSegullo:

When opening new position with in an LP, the ratio of the two tokens normally shows the percentage. Now, since yesterday, it alsways shows 50 50 in both tokens but the value ratio changes. The % percentage value does not get updated. Please fix.

The token balance is showing 50%/50%, but when I put in the amounts, the ratio changes to 60%/40%. The transaction in Keplr shows the correct amounts.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a21c18d)


MaxMillington (approved):
👍

PR #2793, opened 2024-02-08
Hindi, Portuguese Translation corrections

kamal-sutra:
Few important terms like APR, TVL, Max are correctly translated.


MaxMillington (approved):
thank you

PR #2803, opened 2024-02-13
fix: pools timeout

JoseRFelix:

Following #2800, pools query sometimes took to long to load and the function was timing out. Most likely, the edge function was running out of memory because of individually cached pools. This PR reverts this change and increases the TTL. 

## Testing and Verifying

- [ ] Requests should not timeout after subsequent loads


jonator (approved):

kamal-sutra (approved):
Loads better

CryptoAssassin1 (approved):

PR #2820, opened 2024-02-19
Fix: cl pool strategy selection reset

DavideSegullo:
If you try to select a moderate or aggressive strategy, the card does not remain selected but the custom one is set by default.

https://app.osmosis.zone/pool/1325

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a27gv87)


CryptoAssassin1 (approved):
Fixed.

PR #2821, opened 2024-02-19
Integrate asset list v2

jonator:
[Asset list v2](https://raw.githubusercontent.com/osmosis-labs/assetlists/main/osmosis-1/generated/frontend/assetlist.json)

* Integrate asset list v2
* Maintain usage of prev chain list
* New asset list passes all unit tests:
  * Keplr compatible asset list tests
  * Assets store tests
  * Complex queries assets tests

[ClickUp Task URL](https://app.clickup.com/t/86a293kn6)


packages/types/src/asset-types.ts
```diff
@@ -3,218 +3,164 @@ import type {
...
-  relative_image_url: string;
-  coingecko_id?: string;
-  keywords?: string[];
-  origin_chain_name: string;
-  origin_chain_id: string;
-  price_info?: {
-    dest_coin_minimal_denom: string;
-    pool_id: string;
-  };
 }
```
JoseRFelix, line 96: What are your thoughts on making this type an intersection based on the chain type? It would improve type safety and autocompletion.  

```ts
interface CosmosCounterparty {
  chainId: string
  ...
}

interface EVMCounterparty {
  chainId: number
  ...
}

interface NonCosmosCounterparty {
  ...
}

export type Counterparty = CosmosCounterparty | EVMCounterparty | NonCosmosCounterparty
```

packages/web/config/asset-list/__tests__/asset-lists.spec.ts
```diff
@@ -795,11 +879,47 @@ describe("getKeplrCompatibleChain", () => {
             "coinGeckoId": undefined,
             "coinImageUrl": "/tokens/generated/dgl.png",
             "coinMinimalDenom": "factory/juno1u805lv20qc6jy7c3ttre7nct6uyl20pfky5r7e/DGL",
-            "contractAddress": "juno1u805lv20qc6jy7c3ttre7nct6uyl20pfky5r7e",
+            "contractAddress": undefined,
```
JoseRFelix, line 882: Is this expected? I see the contact address in the coinMinimalDenom.

packages/web/config/ibc-overrides.ts
```diff
@@ -33,7 +33,7 @@ type AdditionalData = Partial<
 const TestnetIBCAdditionalData: Partial<
```
JoseRFelix, line 33: Do we still need this file? All of these data should be on the new asset list

packages/web/stores/assets/assets-store.ts
```diff
@@ -157,49 +153,53 @@ export class ObservableAssets {
...
+        return !asset.preview;
       }) // Remove unlisted assets if preview assets is disabled
       .map((ibcAsset) => {
-        const chainInfo = this.chainStore.getChain(ibcAsset.origin_chain_id);
+        const chainInfo = this.chainStore.getChain(
+          ibcAsset.counterparty[0]?.chainId?.toString() ?? ""
+        );
 
-        const minimalDenom = getSourceDenomFromAssetList(ibcAsset);
+        const sourceDenom_ = ibcAsset.sourceDenom;
```
JoseRFelix, line 170: How about we rename `sourceDenom_` to `ibcAssetSourceDenom`? I believe it's more descriptive

packages/web/stores/assets/assets-store.ts
```diff
@@ -157,49 +153,53 @@ export class ObservableAssets {
...
+            `Unknown currency ${sourceDenom_} for ${
+              ibcAsset.counterparty![0].chainId
+            }`
           );
         }
 
-        const ibcTrace = getLastIbcTrace(ibcAsset.traces);
+        const ibcTransferMethod = ibcAsset.transferMethods.find(
+          ({ type }) => type === "ibc"
+        ) as IbcTransferMethod | undefined;
```
JoseRFelix, line 193: Perhaps we can create a utility for this? I say this because of the type. Having to constantly cast it as IbcTransferMethod could become a hassle.

JoseRFelix (commented):
Great work! 🚀  This looks much cleaner already. Left some minor comments

JoseRFelix (commented):
Great work! 🚀  This looks much cleaner already. Left some minor comments

packages/web/stores/assets/assets-store.ts
jonator: Typically I'd agree that repeated code is bad, but IMO this is more readable than a utility that saves a few LOC because we can see exactly what is going on. I noticed with the many utilities for assets that we had (some I removed) made asset-related code hard to read because I often found myself diving into the implementation to see what was going on.

packages/web/config/ibc-overrides.ts
jonator: I think it perhaps could be removed, but I'd like to get to removing that when we do the new bridge project that fully leverages the bridge providers. I'm hoping with that project we can delete a lot of this legacy code and config.

packages/web/config/asset-list/__tests__/asset-lists.spec.ts
jonator: See slack, but I don't believe it is needed

packages/web/config/ibc-overrides.ts
JoseRFelix: Sounds good. In the meantime, how about we move the deposit/withdraw override URLs since they change too frequently (need to open PRs) and are low risk to refactor?

jonator: I will adopt those with my work on new asset page. I'll include them in the table response.

JoseRFelix (approved):
LGTM!

kamal-sutra (approved):

CryptoAssassin1 (approved):

DavideSegullo (approved):

PR #2822, opened 2024-02-20
(1CT) Create One Click Trading Session

JoseRFelix:
### ClickUp Task

[1CT Logic](https://app.clickup.com/t/86a21b84u)
[1CT Profile Link](https://app.clickup.com/t/86a21b7z1)

## Brief Changelog

- Allow Creating 1CT session
- Remove Session from LocalStorage after disconnect
- Add 1CT Expiry and Activated toasts
- Add 1CT Intro Modal
- Display the 'welcome back screen' in the wallet modal after the user has started at least one 1CT session
- Display the remaining session time radial progress in the navbar

## Testing and Verifying

- [ ] Should allow user to create 1CT account
- [ ] Should allow to create a 1CT session with different parameters
- [ ] Should display a toast when the session expires
- [ ] Should display a toast when the session starts


packages/stores/src/account/osmosis/index.ts
```diff
@@ -2125,21 +2125,77 @@ export class OsmosisAccountImpl {
...
+            .balances.forEach((balance) => balance.waitFreshResponse());
+
+          queries.cosmos.queryDelegations
+            .getQueryBech32Address(this.address)
+            .waitFreshResponse();
+
+          queries.cosmos.queryRewards
+            .getQueryBech32Address(this.address)
+            .waitFreshResponse();
+        }
```
jonator, line 2177: I'm assuming we'll eventually remove these for when we finish tRPC migration?

packages/types/src/autheticator-types.ts
jonator, line 1: TBH I'm not a fan of this types package, similar to not being a fan of types.ts files. I think types are easier to understand when they're defined closer to where they are used. Thoughts?

packages/web/components/buttons/button.tsx
jonator, line 1: I think this button will be deprecated in favor of the new ui/button cc @mattupham

jonator (commented):
I think build is failing, but I added an initial review with high-level thoughts

packages/stores/src/account/osmosis/index.ts
JoseRFelix: That's right. In the future, I plan to eliminate these injected accounts and create message-composing functions. This way, we'll only need one function for broadcasting. This approach will simplify the process and offer the greatest flexibility for creating transactions, whether they have a single or multiple messages.

packages/types/src/autheticator-types.ts
JoseRFelix: I agree, however, we need a way to share the types between the `stores` and `web` packages. I think we can remove this package once we have the `server` package where we can co-locate these types with their utils.

packages/web/components/buttons/button.tsx
JoseRFelix: Good to know! I'll update this

DavideSegullo (commented):
I'm trying to initialized a session, but I get an error here:
getAccountPubKeyAndAuthenticators

Screenshot 2024-03-01 alle 11 05 13

JoseRFelix:
> I'm trying to initialized a session, but I get an error here: getAccountPubKeyAndAuthenticators
> 
> Screenshot 2024-03-01 alle 11 05 13

Previews will not work until we get an HTTPS link for the smart accounts node. I'll make sure to keep you posted once I have it. In the meantime, I would appreciate a code review on this PR

packages/web/components/alert/toast.tsx
```diff
@@ -1,11 +1,15 @@
...
-import { FunctionComponent } from "react";
+import React, { FunctionComponent } from "react";
 import { toast, ToastOptions } from "react-toastify";
 
 import { Alert, ToastType } from "~/components/alert";
 import { Icon } from "~/components/assets";
 import { t } from "~/hooks";
 
+export function dismissAllToasts() {
+  toast.dismiss();
```
DavideSegullo, line 10: Why do we need a different function for that? Do you assume you will need to add more shared logic here in the future?

packages/web/components/one-click-trading/screens/reset-period-screen.tsx
```diff
@@ -0,0 +1,80 @@
...
+  "day",
+  "week",
+  "month",
+  "year",
+] as OneClickTradingResetPeriods[];
+
+export function getResetPeriodTranslationKey(
+  id: OneClickTradingResetPeriods
+): string {
+  switch (id) {
```
DavideSegullo, line 19: If the key is always in the same format as the corresponding translation, couldn't we just check if "ResetPeriods" contains that key and then construct the string as "oneClickTrading.settings.resetPeriodScreen.periods.{key}"? Obviously if it doesn't at contains we return "oneClickTrading.settings.resetPeriodScreen.periods.unknown", so in the future if we add other keys we only need to update "ResetPeriods" and the translations and we won't have a long switch.

What do you think?

packages/web/components/one-click-trading/screens/session-period-screen.tsx
```diff
@@ -1,40 +1,45 @@
...
-    periodTranslationKey:
-      "oneClickTrading.settings.sessionPeriodScreen.periods.month",
-  },
-  {
-    id: "year",
-    periodTranslationKey:
-      "oneClickTrading.settings.sessionPeriodScreen.periods.year",
-  },
-] as const;
+export function getSessionPeriodTranslationKey(
```
DavideSegullo, line 9: How about we do the same thing proposed above with "ResetPeriods" here as well?

DavideSegullo (approved):

packages/web/components/one-click-trading/screens/reset-period-screen.tsx
JoseRFelix: This is a good idea! I'll add this in

PR #2823, opened 2024-02-20
fix: pools on testnet

p0mvn:

Fixes pools on testnet by removing broken filtering.

### DIFF
--- a/packages/web/server/queries/complex/pools/index.ts
+++ b/packages/web/server/queries/complex/pools/index.ts
@@ -63,13 +63,22 @@ export async function getPools(
 ): Promise<Pool[]> {
   let pools = await poolProvider({ poolIds: params?.poolIds });
 
-  if (params?.types || params?.minLiquidityUsd) {
-    pools = pools.filter(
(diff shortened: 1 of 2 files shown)


jonator (approved):
NIT improvement. There's a boolean IS_TESTNET var in env.ts that may be better to use over checking env var directly.

packages/web/server/queries/complex/pools/index.ts
```diff
@@ -63,13 +63,22 @@ export async function getPools(
...
+    pools = pools.filter(({ type }) =>
+      params?.types ? params.types.includes(type) : true
+    );
+  }
+
+  // Note: we do not want to filter the pools if we are in testnet because we do not have accurate pricing
+  // information.
+  if (
+    params?.minLiquidityUsd &&
+    process.env.NEXT_PUBLIC_IS_TESTNET !== "true"
```
MaxMillington, line 76: you can do `!process.env.NEXT_PUBLIC_IS_TESTNET` for future reference

p0mvn:
Updated to use `IS_TESTNET` variable

PR #2828, opened 2024-02-20
Remove pool 1423 from hardcoded boosts

mattupham:

- remove pool 1423 from hardcoded boost exclusion

### ClickUp Task


p0mvn (approved):
LGTM. Thanks

I can see the external boost in the preview

PR #2832, opened 2024-02-20
fix: Astroport PCl pool type not appearing in the v2 table

p0mvn:

Fixes Astroport PCL pool not appearing in the pools table


JoseRFelix (approved):

jonator (changes requested):
I missed this it seems

Use of pool types is confusing here as only transmuter cosmwasm pool is available in UI filter.

IMO this code needs to clarify that a bit more.

https://github.com/osmosis-labs/osmosis-frontend/pull/2832/files#diff-c937e719b3142806965b8e28631df9c71c2ceb92ed1d97917fceeb79c65ad904R43-R48

jonator (approved):

PR #2842, opened 2024-02-22
Optimize asset queries

jonator:

* Reduce excessive querying of market caps per asset
* Add batched querying for CoinGecko coins and prices
* Validates that a CoinGecko ID is active on CoinGecko before querying CoinGecko

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a293kn6)

## Brief Changelog

- Add [DataLoader](https://www.npmjs.com/package/dataloader#batch-function) package for batching queries by ID into a query by multiple IDs

## Testing and Verifying

Should notice improved performance for assets page v2. Also, the Vercel logs should show decreased requests.

### DIFF
--- a/packages/web/server/queries/complex/assets/price.ts
+++ b/packages/web/server/queries/complex/assets/price.ts
@@ -3,7 +3,6 @@ import { makeStaticPoolFromRaw, PoolRaw } from "@osmosis-labs/pools";
 import { Asset } from "@osmosis-labs/types";
 import { getAssetFromAssetList, isNil } from "@osmosis-labs/utils";
 import cachified, { CacheEntry } from "cachified";
-import DataLoader from "dataloader";
 import { LRUCache } from "lru-cache";
 
 import { DEFAULT_LRU_OPTIONS, LARGE_LRU_OPTIONS } from "~/config/cache";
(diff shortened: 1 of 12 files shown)


DavideSegullo (approved):

packages/web/server/queries/complex/assets/price.ts
```diff
@@ -40,22 +41,50 @@ async function getCoingeckoCoin({ denom }: { denom: string }) {
...
+    cache: pricesCache,
+    key: `prices-batch-loader-${currency}`,
+    getFreshValue: async () => {
+      return new DataLoader(
+        (ids: readonly string[]) => batchFetchCoingeckoPrices(ids, currency),
+        {
+          // workaround to work on Vercel Edge runtime
+          batchScheduleFn: (cb) => setTimeout(cb, 0),
+        }
+      );
```
JoseRFelix, line 77: I wonder if we should create an `EdgeDataLoader` so that we don't have to continuously assign the `batchScheduleFn`.

packages/web/server/queries/complex/assets/price.ts
```diff
@@ -14,6 +14,7 @@ import {
 } from "~/server/queries/coingecko";
```
JoseRFelix, line 14: What are your throughs on adding a data loader for calculating prices with pools on this PR?

jonator: You mean a data loader for fetching pools by ID?

jonator: I was looking into that and, with our use of pool providers, our direct use of this legacy pools query fn, and with the SQS prices API service nearing completion, I thought I'd wait for that then just remove the use of this query.

JoseRFelix (approved):
LGTM!

PR #2844, opened 2024-02-24
Add remote cache support, remove server imports

jonator:
Reduces client bundle size by removing import of server code for use in bridge providers' transfer status source objects. Adds remote cache object (implemented with Vercel KV). Found root bundle to be reduced from 2.14 to 2.11 mb.

It's also worth noting that I dove into our [Vercel KV store in prod](https://vercel.com/osmo-labs/osmosis-frontend/stores/kv/store_cGqHAHqmtP0OA8YD/cli) and tweaked the settings to auto prune keys when full, and added more read replicas: Washington DC, Dublin, and Singapore.

* Add remote cache and use in sidecar pool provider
  * Uses superjson to let us easily cache complex objects
* Refactor bridge providers to track transfer statuses directly in the respective transfer status source implementations
* Fix tests that are intended to run in node env (vs dom env)
* Loads env vars for use in jest tests

TODO
- [x] Fix test GH Action by ensuring env vars are set

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a2b5x5j)

## Testing and Verifying

* Transfer status from deposit/withdraws that use bridge providers should still work.
* Pools page should load, now from data in Vercel KV.

### DIFF
--- a/packages/web/server/queries/complex/pools/market.ts
+++ b/packages/web/server/queries/complex/pools/market.ts
@@ -23,7 +23,8 @@ export function getCachedPoolMarketMetricsMap(): Promise<
   return cachified({
     cache: metricPoolsCache,
     key: "pools-metrics-map",
-    ttl: 1000 * 60 * 5, // 5 minutes
+    ttl: 1000 * 10, // 10 seconds
+    staleWhileRevalidate: 1000 * 25, // 25 seconds, edge function run time limit
     getFreshValue: async () => {
       const map = new Map<string, PoolMarketMetrics>();
 
--- a/.github/workflows/jest-tests.yml
+++ b/.github/workflows/jest-tests.yml
@@ -1,7 +1,6 @@
 name: Jest Tests Workflow
 
 on: [push]
-env: ${{ secrets.TWITTER_KV_STORE_REST_API_TOKEN, secrets.TWITTER_KV_STORE_REST_API_URL }}
 
 jobs:
   test:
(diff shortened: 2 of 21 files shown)


JoseRFelix (approved):

packages/web/server/queries/complex/pools/market.ts
```diff
@@ -23,7 +23,8 @@ export function getCachedPoolMarketMetricsMap(): Promise<
   return cachified({
     cache: metricPoolsCache,
     key: "pools-metrics-map",
-    ttl: 1000 * 60 * 5, // 5 minutes
+    ttl: 1000 * 10, // 10 seconds
+    staleWhileRevalidate: 1000 * 25, // 25 seconds, edge function run time limit
```
p0mvn, line 27: Why are we lowering ttl to 10 seconds?

I think that it isn't really useful to update the volume data frequently. Infrequent update > 1 min makes sense to me. Reducing the TTL would increase the load on the data backend.

The `stalteWhileRevalidate` addition makes sense but I don't think that it would lower the load on the data backend

jonator: I think I'm going to use a remote cache and increase this value to 1 min, 1 hour respectively.

I have it 10s, 25s here because edge funcitons only live for 25s, meaning the in memory cache only lives for 25s at most. So, for 10s users will hit the cache, and the remaining 15s users will still hit cache and data will be updated in background.

p0mvn: The 1 min ttl makes sense. What is the reason for choosing such a high number for stale while revalidate? Wouldn't it be more than we need?

kamal-sutra (approved):

PR #2846, opened 2024-02-26
fix corners for pool cards

mattupham:

- pool card corners were cutting off

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a2b832w)

## Brief Changelog

apply border radius fix

## Testing and Verifying

Screenshot 2024-02-26 at 8 29 05 AM


jonator (changes requested):
It's hard to tell, but I think the corners are still not uniform, they bulge ever so slightly. Can you notice that too?

jonator (approved):

PR #2849, opened 2024-02-26
fix: hide volume column on error on pools page v2

p0mvn:

Currently, our pools page bricks if volume data is unavailable.

This PR fixes the issue by avoiding to format the volume column in such a case.

It calculates the number of pools that have volume formatted. If greater than half, volume column is displayed and the missing pools show "N/A"

If less than half, then we do not show the volume column

This is related to the `/fees/v1/pools` query

### ClickUp Task

## Testing and Verifying

Tested this by force throwing an error from the endpoint.

See:
image

I added the error mock here: https://github.com/osmosis-labs/osmosis-frontend/blob/a1660e2e3244adb0653f370a3ff6a9f7983d2d10/packages/web/server/queries/imperator/pools-fees.ts#L17-L20

### DIFF
--- a/packages/web/components/complex/all-pools-table-v2.tsx
+++ b/packages/web/components/complex/all-pools-table-v2.tsx
@@ -186,13 +186,15 @@ export const AllPoolsTable: FunctionComponent<{
 
   // If more than half of the pools have volume data, we should format the volume column.
   // Otherwise, we should not format the volume column.
-  let volumePresenceCount = 0;
-  poolsData.forEach((pool) => {
-    if (pool.volume24hUsd) {
-      volumePresenceCount++;
-    }
-  });
-  let shouldFormatVolume = volumePresenceCount > poolsData.length / 2;
+  const shouldDisplayVolumeColumn = useMemo(() => {
+    let volumePresenceCount = 0;
+    poolsData.forEach((pool) => {
+      if (pool.volume24hUsd) {
+        volumePresenceCount++;
+      }
+    });
+    return volumePresenceCount > poolsData.length / 2;
+  }, [poolsData]);
 
   // Define columns
   const columnHelper = createColumnHelper<Pool>();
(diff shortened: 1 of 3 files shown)


packages/web/components/complex/all-pools-table-v2.tsx
```diff
@@ -184,31 +184,51 @@ export const AllPoolsTable: FunctionComponent<{
...
 
+  // If more than half of the pools have volume data, we should format the volume column.
+  // Otherwise, we should not format the volume column.
+  let volumePresenceCount = 0;
+  poolsData.forEach((pool) => {
+    if (pool.volume24hUsd) {
+      volumePresenceCount++;
+    }
+  });
+  let shouldFormatVolume = volumePresenceCount > poolsData.length / 2;
```
JoseRFelix, line 195: This should be placed on a `useMemo`. Otherwise, it will call this expended calculation on every render

```suggestion
 const shouldFormatVolume = useMemo(() => {
    let volumePresenceCount = 0;
    poolsData.forEach((pool) => {
      if (pool.volume24hUsd) {
        volumePresenceCount++;
      }
    });
    return volumePresenceCount > poolsData.length / 2;
  }, [poolsData]);
```

JoseRFelix, line 195: Perhaps we should rename this to `shouldDisplayVolumeColumn`?

JoseRFelix (changes requested):
Thank you for the PR! Left some comments

p0mvn:
> Thank you for the PR! Left some comments

Thanks, addressed!

jonator (approved):

packages/web/components/complex/all-pools-table-v2.tsx
jonator: Also, you could return early if it reaches half of volume count.

Saving milliseconds.

But looks good

JoseRFelix (approved):
LGTM!

PR #2853, opened 2024-02-27
Use remote cache for pool volume data

jonator:

Help alleviate the instability of the pool volume query (pools/fees) by caching it's data in a shared cache. This prevents new edge functions from having to query the data and can instead rely on the cache, which has a generous stale time.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a2b5x5j)

## Brief Changelog

* Use remote cache for pools fees data (volume)
* Add generous staleWhileRevalidate setting to cache value (1 hour), allowing older data to be returned even if there's intermittent errors with the query service
* Add additional fault tolerance to RemoteCache object: falling back to in-memory cache if there's any kind of issue creating the remote cache client (Vercel KV)

## Testing and Verifying

* Pools volume data should appear
* CURLing the [remote cache](https://vercel.com/osmo-labs/osmosis-frontend/stores/kv/store_cGqHAHqmtP0OA8YD/settings) should show an entry for `pools-metrics-map`


packages/web/server/queries/complex/pools/market.ts
```diff
@@ -15,16 +14,16 @@ export type PoolMarketMetrics = Partial<{
...
 export function getCachedPoolMarketMetricsMap(): Promise<
   Map<string, PoolMarketMetrics>
 > {
   return cachified({
     cache: metricPoolsCache,
     key: "pools-metrics-map",
-    ttl: 1000 * 10, // 10 seconds
-    staleWhileRevalidate: 1000 * 25, // 25 seconds, edge function run time limit
+    ttl: 1000, // 1 min
+    staleWhileRevalidate: 1000 * 60, // 1 hour
```
p0mvn, line 26: The 1 min ttl makes sense. What is the reason for choosing such a high number for stale while revalidate? Wouldn't it be more than we need?

jonator: FYI I just fixed these values

What stale time are you thinking.

packages/web/utils/cache.ts
```diff
@@ -15,28 +15,40 @@ const isTestEnv = process.env.NODE_ENV === "test";
...
+      try {
+        this.kvStore = createClient({
+          url: process.env.TWITTER_KV_STORE_REST_API_URL!,
+          token: process.env.TWITTER_KV_STORE_REST_API_TOKEN!,
+        });
+      } catch (e) {
+        console.error(
+          "Failed to create RemoteCache client. Falling back to in-memory cache.."
+        );
+        this.fallbackCache = new LRUCache<string, CacheEntry>({ max: 50 });
```
JoseRFelix, line 39: Thoughts on using the `DEFAULT_LRU_OPTIONS` here?

jonator: I thought about that, but I wanted to avoid having web config be a dependency with this file.

Perhaps we can move that const to this file? Since this file manages cache utils?

packages/web/server/queries/complex/pools/market.ts
```diff
@@ -15,16 +14,16 @@ export type PoolMarketMetrics = Partial<{
...
 /** Get a cached Map with pool IDs mapped to market metrics for that pool. */
 export function getCachedPoolMarketMetricsMap(): Promise<
   Map<string, PoolMarketMetrics>
 > {
   return cachified({
     cache: metricPoolsCache,
     key: "pools-metrics-map",
-    ttl: 1000 * 10, // 10 seconds
-    staleWhileRevalidate: 1000 * 25, // 25 seconds, edge function run time limit
+    ttl: 1000, // 1 min
```
ValarDragon: Did the units of this change? (10000 -> became 1000, but 10s -> 1min)

ValarDragon (approved):

packages/web/server/queries/complex/pools/market.ts
p0mvn: I would expect 2 minutes at most to be useful and unsure why 1 hour is necessary.

From my understanding, after 1 minute we would update the cache in the background and `staleWhileRevalidate` represents how long the cache is still valid during the update.

The update workload will probably not take this long so I think we should lower it to be near of what we would expect

packages/web/server/queries/complex/pools/market.ts
jonator: I just fixed this with a later commit. LMK if this addresses you comment.

packages/web/server/queries/complex/pools/market.ts
ValarDragon: perhaps 30 min?

jonator: I'm down for 30 mins, as volumes quite a bit with market movement / hype

p0mvn: `staleWhileRevalidate`:
```
 * Amount of milliseconds that a value with exceeded ttl is still returned
     * while a fresh value is refreshed in the background
```

I think it makes sense to have this be at most 1-2 minutes. Do we expect the refresh workloads to run longer than that?

packages/web/utils/cache.ts
JoseRFelix: I agree. In the future, when we move `server` to a new package, we'll most likely need these variables.

packages/web/server/queries/complex/pools/market.ts
jonator: If we're more generous with the stale time, then offers more tolerance to the service being down for 1 min as you suggest

p0mvn: I see the goal you're trying to achieve with this now

p0mvn (approved):

JoseRFelix (approved):
LGTM!

packages/web/utils/cache.ts
```diff
@@ -15,28 +31,44 @@ const isTestEnv = process.env.NODE_ENV === "test";
...
+      console.warn(
+        "RemoteCache is not available in test environment. Falling back to in-memory cache.."
+      );
+      this.fallbackCache = new LRUCache<string, CacheEntry>(
+        DEFAULT_LRU_OPTIONS
+      );
     }
   }
 
   async get<T>(key: string) {
```
JoseRFelix, line 69: @jonator I believe we have to prefix keys based on the environment. IIRC, Vercel KV does not differ between envs so we could have some data conflicts

jonator: WDYM by environment? Aren't shared keys a key value prop with the remote cache?

JoseRFelix: I mean envs like production, testnet, and edgenet

jonator: Ah that's true. What are all of the environments?
* Mainnet
* Testnet
* Edgenet (any other chain env)

What should we use to derive these environment in your opinion?

JoseRFelix: I agree that other environments like testnet and edgenet can utilize in-memory caching. We can introduce a check for environment variables. If they're missing, we should default to the in-memory cache. Additionally, it would be wise to set up a KV for stage/previews so we can QA the changes

jonator: OK. I'll check for chain IDs set in `process.env.NEXT_PUBLIC_OSMOSIS_CHAIN_ID_OVERWRITE`, as those are not expected to change for mainnet and testnet. Should we check against hardcoded values or simply prefix with that variable value?

Then, if it's not set we could divert to in-memory cache.

You're saying we should have a separate KV store for stage vs prod?

jonator:
Converting to draft as we decide our strategy to segment cache for different Osmosis node envs (mainnet, testnet, edgenet), as well as deployment envs (prod, stage)

packages/web/utils/cache.ts
JoseRFelix: > You're saying we should have a separate KV store for stage vs prod?

Yes, we should distinguish between them by using different KV urls, particularly because they access the same underlying datasets, such as the asset list and chain list

jonator:
Investigating an issue with 400s when posting to remote cache

jonator:
I removed all uses of remote cache, as further discussion is needed around how we want to handle Vercels 1MB request limit

PR #2860, opened 2024-02-28
Fix initialization of CL add liq strategy

jonator:

Check for null, since 0 values would also return false.

### ClickUp Task


JoseRFelix (approved):

packages/web/hooks/ui-config/use-add-concentrated-liquidity-config.ts
```diff
@@ -821,8 +821,8 @@ export class ObservableAddConcentratedLiquidityConfig {
           this.pool &&
           this._baseDepositAmountIn.sendCurrency &&
           this._quoteDepositAmountIn.sendCurrency &&
-          this._minHistoricalPrice &&
-          this._maxHistoricalPrice &&
+          this._minHistoricalPrice !== null &&
+          this._maxHistoricalPrice !== null &&
```
p0mvn, line 825: Should we be checking for `undefined`?

p0mvn (approved):

PR #2866, opened 2024-02-29
Remove CoinGecko search from getAssetPrice

jonator:

If an asset is not in the asset list and doesn't have pool price ID or CoinGecko ID info, let's return an error instead of spamming CoinGecko to search for the CoinGecko ID. Searching CoinGecko through prices.osmosis.zone was causing excessive traffic.

This also slightly simplifies the logic and improves naming of the `getAssetPrice` function.

### DIFF
--- a/packages/web/server/queries/complex/assets/price.ts
+++ b/packages/web/server/queries/complex/assets/price.ts
@@ -8,7 +8,6 @@ import { LRUCache } from "lru-cache";
 import { AssetLists } from "~/config/generated/asset-lists";
 import {
   CoingeckoVsCurrencies,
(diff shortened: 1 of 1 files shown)


niccoloraspa:
Amazing thank you! 

Can I block any `/api/search` request to prices now?

MaxMillington (approved):

jonator:
@niccoloraspa yes

jonator:
I think for low liq tokens we were performing a lot of searches, but if they're low liq (don't have a pool to calculate prices with) they for sure won't yield a useful CoinGecko ID from the search

PR #2883, opened 2024-03-02
(1CT) Add Profile Section

JoseRFelix:
image
image

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a21b7zm)

## Brief Changelog

- Improve code quality by adding authenticator unit test and comments
- Make `createOneClickTradingSession` Mutation
- Allow creating/editing 1CT session from profile
- Handle deleting 1CT sessions

## Testing and Verifying

- [ ] Can create/edit a session from profile modal
- [ ] Can end session in profile


packages/web/server/api/edge-routers/one-click-trading-router.ts
```diff
@@ -40,6 +44,49 @@ export const oneClickTradingRouter = createTRPCRouter({
...
+            ...sub,
+            authenticatorId: authenticator.id,
+          }))
+        );
+
+      if (subAuthenticators.length === 0) {
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "Session not found: authenticators array is empty",
+        });
```
jonator, line 70: Do we need to throw TRPCError objects ourselves, or will tRPC convert normal errors into those objects?

packages/web/server/api/edge-routers/one-click-trading-router.ts
```diff
@@ -40,6 +44,49 @@ export const oneClickTradingRouter = createTRPCRouter({
   getNetworkFeeLimitStep: publicProcedure.query(async () =>
     getNetworkFeeLimitStep()
   ),
+  getSessionAuthenticator: publicProcedure
+    .input(z.object({ userOsmoAddress: z.string(), publicKey: z.string() }))
+    .query(async ({ input }) => {
```
jonator, line 49: Could the logic in this query be pulled into a `getSessionAuthenticator` funciton in complex folder? May make it more reusable, especially if we don't throw tRPC-specific errors.

packages/web/modals/profile.tsx
```diff
@@ -436,6 +497,182 @@ export const ProfileModal: FunctionComponent<
   );
 });
 
+const ProfileOneClickTradingSettings = ({
```
jonator, line 500: Could this component be placed in `one-click-trading` folder, to co-locate?

packages/web/hooks/one-click-trading/use-one-click-trading-info.ts
```diff
@@ -43,9 +82,31 @@ export const useOneClickTradingInfo = ({
     return () => clearTimeout(timeoutId);
   }, [isExpired, t, value?.info, onExpire]);
 
+  const getTimeRemaining = useCallback(() => {
```
jonator, line 85: Since this takes no params, why don't we use a `useMemo` to just return the remaining time?

packages/web/hooks/one-click-trading/use-one-click-trading-info.ts
```diff
@@ -43,9 +82,31 @@ export const useOneClickTradingInfo = ({
...
+    const oneClickTradingInfo = value?.info;
+    if (!oneClickTradingInfo) return 0;
+    const sessionEndDate = dayjs.unix(
+      unixNanoSecondsToSeconds(oneClickTradingInfo.sessionPeriod.end)
+    );
+
+    return sessionEndDate.unix() - dayjs().unix();
+  }, [value?.info]);
+
+  const getTotalSessionTime = useCallback(() => {
```
jonator, line 95: Same as above.

packages/web/hooks/one-click-trading/use-one-click-trading-info.ts
```diff
@@ -1,28 +1,67 @@
...
 
-export const useOneClickTradingInfo = ({
+/**
+ * This hook manages and provides information about the one-click trading session.
+ *
+ * This hook encapsulates the logic for fetching one-click trading session information,
+ * determining if the session is enabled and if it has expired. It also provides utility
+ * functions to calculate the remaining time for the session and the total session time.
+ *
+ * @param {Object} params - The parameters object.
```
jonator, line 17: NIT: would prefer to avoid specifying types in js-doc, would prefer just typescript.

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -1,54 +1,143 @@
...
+ * This hook is designed to fetch and maintain the default parameters for one-click trading,
+ * allowing for easy access and modification within components. If `oneClickTradingInfo` is provided, it uses
+ * this to set initial parameters. Otherwise, it fetches the default parameters from the application's API.
+ *
+ * The hook also provides a mechanism to reset the parameters to their initial values, which
+ * can be either the custom parameters passed at initialization or the fetched default parameters.
+ *
+ * This hook is primarily intended for the one click trading settings modal.
+ *
+ * @param {Object} params - The parameters object for initializing the hook.
```
jonator, line 47: Same as in -info file

jonator (approved):
Some surface-level code quality comments. But it looks like you know what you're doing.

packages/web/components/navbar/index.tsx
```diff
@@ -175,14 +174,16 @@ export const NavBar: FunctionComponent<
     const onClick1CT = () => {
       displayToast(
         {
-          message: t("oneClickTrading.toast.oneClickTradingActive"),
-          caption: "59 minutes remaining",
+          titleTranslationKey: t("oneClickTrading.toast.oneClickTradingActive"),
+          captionTranslationKey: "59 minutes remaining",
```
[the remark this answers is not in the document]
JoseRFelix: These are just all the toasts while I connect them. I'll remove these before going into production.

packages/web/server/api/edge-routers/one-click-trading-router.ts
JoseRFelix: Yes, since TRPC uses these errors to determine the query's status. Normal errors always default to a status of `500`

packages/web/hooks/one-click-trading/use-one-click-trading-info.ts
JoseRFelix: Since the info object contains a hardcoded session expiry time, it won't update automatically, returning inaccurate data. We also don't want to add an interval on this hook since it would trigger unnecessary re-renders. Therefore, we return a function so the components can handle the update logic themselves as we do on `WalletInfo`, setting a `setInterval`

PR #2901, opened 2024-03-06
Fetch prices from SQS

jonator:

Fetches prices in batches from SQS instead of calculating from pools. Should greatly increase performance. I also reorganize the code to fetching prices from multiple providers, similar to what we did with pools providers. To start, we'll have "pools" and "sidecar" price providers. This allows a one line change for where we get prices from.

### ClickUp Task

[ClickUp Task URL](https://app.clickup.com/t/86a2e6k30)

## Brief Changelog

* Create price providers
* Add SQS prices query
* Send batched requests to SQS, falling back to CoinGecko if no result for an asset

## Testing and Verifying

* Assets and prices load in swap tool and new assets page


jonator:
Marking as draft since SQS prod does not include prices query yet

p0mvn (approved):
🥇

PR #2923, opened 2024-03-08
feat: Separate Edge Function Load

JoseRFelix:
## Summary by CodeRabbit

- **New Features**
	- Implemented timeouts for pool data retrieval processes to enhance performance and reliability.
- **Refactor**
	- Updated cache configuration for more efficient data handling.
- **Bug Fixes**
	- Removed an outdated caching option to improve data freshness and accuracy.


ValarDragon:
@coderabbitai review

packages/web/pages/api/edge-trpc-assets/[trpc].ts
```diff
@@ -0,0 +1,32 @@
+import { fetchRequestHandler } from "@trpc/server/adapters/fetch";
+import { NextRequest } from "next/server";
+
+import { edgeRouter } from "~/server/api/edge-routers/edge-router";
+import { createEdgeTRPCContext } from "~/server/api/trpc";
+import { constructEdgeUrlPathname } from "~/utils/trpc-edge";
+
+// We're using the edge-runtime
+export const config = {
+  runtime: "edge",
```
Comment (jonator) [inline packages/web/pages/api/edge-trpc-assets/[trpc].ts:10]: Thoughts on serverless, since we determined it was a memory issue and not event loop queue issue? There are issues with edge, even without load

packages/web/server/api/edge-routers/pools-router.ts
```diff
@@ -108,59 +109,68 @@ export const poolsRouter = createTRPCRouter({
...
+                sortInput,
+                minLiquidityUsd,
+                types,
+                incentiveTypes,
+              }),
+              cursor,
+              limit,
+            }),
+          12_000,
+          "getMarketIncentivePools"
```
jonator, line 172: Why wrap pagination in timeout? Does this alleviate edge memory issues?

packages/web/server/queries/complex/pools/providers/indexer.ts
```diff
@@ -33,11 +33,14 @@ import { DEFAULT_VS_CURRENCY } from "../../assets/config";
 import { Pool } from "..";
 import { TransmuterPoolCodeIds } from "../env";
 
-const poolsCache = new LRUCache<string, CacheEntry>(DEFAULT_LRU_OPTIONS);
+const poolsCache = new LRUCache<string, CacheEntry>({ max: 1 });
```
jonator, line 36: Does this address the edge memory issue?

packages/web/server/api/edge-routers/pools-router.ts
JoseRFelix: Although it doesn't solve the issue directly, it allows the function to fail quickly, freeing up resources for other requests. Additionally, this enables the Front-End to retry the query with a different function in case of a timeout

packages/web/server/queries/complex/pools/providers/indexer.ts
JoseRFelix: Yes, it does. Furthermore, we only need one cache entry for the pools function because this cache won't be shared with other cached functions

packages/web/pages/api/edge-trpc-assets/[trpc].ts
Comment (JoseRFelix) [reply packages/web/pages/api/edge-trpc-assets/[trpc].ts:10]: I think this change should be made in a separate PR because it requires some considerations:

1. Our node functions are in just one region, so we need a plan to manage the extra load on the SQS server cc @p0mvn 
2. Based on benchmarks, this change adds ~1.5s to every query, on top of the initial ~10s needed for a cold start
3. We must rename our edge router, as keeping its current name would wrongly suggest that these routes run on the edge when they run on a node

packages/web/server/api/edge-routers/pools-router.ts
jonator: Is this common enough to consider adding the timeout inside of `maybeCachePaginatedItems`? Perhaps we can add an optional `timeoutName` param to make the timeout optional?

JoseRFelix: I believe it's a good idea to abstract this behavior. However, instead of increasing the complexity of `maybeCachePaginatedItems`, how about we create a `timeoutProcedure` within trpc so all routes can share a timeout? This would also give us visibility in the logs for all routes since timed-out functions return an error and status code.

packages/web/pages/api/edge-trpc-assets/[trpc].ts
Comment (jonator) [reply packages/web/pages/api/edge-trpc-assets/[trpc].ts:10]: 1. Is there no way to make node multi region
2. I think we're still having issues in Asia, I 
3. What if we remove `edge` entirely from our tRPC router namespaces? With how tRPC lets us run logic to determine a route's link, I think there's no reason to have the runtime be indicated by the query path. Rather, we can just organize routes by data type or feature

packages/web/server/api/edge-routers/pools-router.ts
jonator: Is that like a global tRPC timeout?

jonator (approved):

PR #2925, opened 2024-03-08
refactor: query SQS router directly

p0mvn:

Trying to query SQS directly and bypass tRPC for routing.


jonator:
@p0mvn could this be closed?

PR #2936, opened 2024-03-09
feat: get spot price from quote

p0mvn:

We don't need to fetch spot price. It can be retrieved directly from quote.

### ClickUp Task


jonator:
Done here:
https://github.com/osmosis-labs/osmosis-frontend/pull/2967

PR #2948, opened 2024-03-12
Get price from pools as fallback

jonator:
Fall back to calculating prices from pre-defined pool routes if not available from SQS. Will then fallback to CoinGecko if no price config (with pool ID) is found.

## Summary by CodeRabbit

- **Refactor**
	- Updated the method for fetching asset prices to improve reliability and accuracy.
	- Enhanced error handling for price retrieval.


packages/web/server/queries/complex/assets/price/providers/sidecar.ts
```diff
@@ -69,7 +69,7 @@ export function getPriceBatched(asset: Asset) {
         loader
           .load(asset.coinMinimalDenom)
           .then((price) => new Dec(price))
-          .catch(() => getPriceFromCoinGecko(asset))
+          .catch(() => getPriceFromPools(asset))
```
p0mvn, line 72: Is there a way we could add some metrics/events so that we know how many and which pool IDs are affected by fallback?

jonator: We could log warnings? Then Vercel logs would pick this up

p0mvn: What about amplitude? Would that be the right tool for this?

I'm worried that logs would only be polluting and unnecessary. Something like a metric/event would probably be cleaner here

jonator: I think a better tool would be Sentry (we recently revived our account for this). Thoughts? They have a Node.js package.

It has a good way of bucketing the same issues together. At least it did for the client side error handling. It would look at the call stack and message in exceptions and bundle them together by number of occurrences. So rather than a mile-long noisy log entries in Vercel, the same issue would get bundled together and tallied by # of occurences.

Many of them were duds or noise. We can easily flag an issue as a false alarm or non issue. This will silence it and allow other issues to surface and be noticed.

Thoughts? I think now that Kelsey fixed billing with Sentry I'll integrate it on client and server.
See:
https://docs.sentry.io/platforms/javascript/guides/nextjs/

jonator: Progress here: https://github.com/osmosis-labs/osmosis-frontend/pull/2954

p0mvn (approved):

PR #2959, opened 2024-03-13
Fix: custom slippage setting: check for NaN

jonator:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved number validation in slippage settings to ensure only valid numbers are accepted.
	- Enhanced boundary check to restrict manual slippage values within a specific range.


DavideSegullo (approved):

packages/stores/src/ui-config/slippage-config.ts
```diff
@@ -64,15 +64,16 @@ export class ObservableSlippageConfig {
 
   @action
   setManualSlippage(str: string) {
+    if (isNaN(Number(str))) return;
```
JoseRFelix, line 67: Thoughts on using `isNumeric` instead?

jonator: Where can that be imported from?

jonator: Given this is the stores package, I don't have access to utils in web package.

JoseRFelix: This function is a good candidate for `@osmosis-labs/utils`. This suggestion is a nit, so feel free to merge or move it to the utils folder

PR #2974, opened 2024-03-18
SQS: Get code ID and inOutSpotPrice from SQS quote response

jonator:
Gets code ID and in/out spot price returned from SQS.

* Code ID: if a cosmwasm pool in a route, we use the Code ID to get the specific Cosmwasm pool type from FE hardcoded values.
* Returns the in/out spot price from SQS, which used to be returned from querying the router by passing 1 in token in. This avoids the edge case where 1 token in is too much for the liquidity of the selected token pair. Example: swapping any BTC variant for a low liq token out.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new property to enhance pool result details.
	- Enhanced swap functionality with improved pool type determination based on new criteria.
- **Refactor**
	- Updated field names and processing logic for clearer financial terminology and data handling.
	- Adjusted internal logic for setting pool types to utilize new property efficiently.

### DIFF
--- a/packages/web/integrations/sidecar/router.ts
+++ b/packages/web/integrations/sidecar/router.ts
@@ -41,7 +41,7 @@ export class OsmosisSidecarRemoteRouter implements TokenOutGivenInRouter {
         route: routes,
         effective_fee,
         price_impact,
-        // in_out_spot_price,
+        in_base_out_quote_spot_price,
       } = await apiClient<SidecarQuoteResponse>(queryUrl.toString());
 
       const swapFee = new Dec(effective_fee);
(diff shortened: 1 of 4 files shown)


packages/web/integrations/sidecar/router.ts
```diff
@@ -53,13 +53,14 @@ export class OsmosisSidecarRemoteRouter implements TokenOutGivenInRouter {
         swapFee,
         priceImpactTokenOut: priceImpact,
         tokenInFeeAmount: tokenIn.amount.toDec().mul(swapFee).truncate(),
-        // inOutSpotPrice: new Dec(in_out_spot_price),
+        inOutSpotPrice: new Dec(in_base_out_quote_spot_price),
```
p0mvn, line 56: Consider following the same naming pattern with base and quote for clarity

p0mvn (approved):
utACK. Nice work!

PR #2992, opened 2024-03-25
Fix: update swap to match SQS return

jonator:
* Decimals we're recently removed from the `in_base_out_quote_spot_price` spot price. This updates our client to add the decimals ourselves on FE.
* This also adds a warning log if there's an issue getting an asset from the asset list in the swap tool

## Summary by CodeRabbit

- **New Features**
	- Enhanced swap tool calculation logic for more accurate spot price display.
- **Bug Fixes**
	- Updated swap functionality to correctly handle spot price calculations and error scenarios.
- **Chores**
	- Improved codebase by renaming variables for clarity and excluding irrelevant files.
- **Refactor**
	- Modified internal logic for handling spot prices in swaps, ensuring a more intuitive user experience.


JoseRFelix (approved):

packages/pools/src/router/types.ts
```diff
@@ -62,7 +62,7 @@ export type SplitTokenInQuote = Quote & {
   /** In amount after fees paid are subtracted. */
   tokenInFeeAmount?: Int;
   swapFee?: Dec;
-  inOutSpotPrice?: Dec;
+  inBaseOutQuoteSpotPrice?: Int;
```
p0mvn, line 65: Why is this an int now?

jonator: Since a base amount is now being returned, we've traditionally handled all base amounts as Ints (from SQS, Osmosis node, etc)

p0mvn: I'm not sure I follow. Could you elaborate? It's returned as a decimal from SQS:
image

p0mvn (changes requested):
looks like the int conversion is breaking

image

packages/web/hooks/use-swap.ts
```diff
@@ -107,24 +107,6 @@ export function useSwap({
...
-      tokenInAmount: DecUtils.getTenExponentN(
-        swapAssets.fromAsset?.coinDecimals ?? 0
-      )
-        .truncate()
-        .toString(),
-      tokenOutDenom: swapAssets.toAsset?.coinMinimalDenom ?? "",
-      forcePoolId: forceSwapInPoolId,
-    },
-    isToFromAssets && !Boolean(quote?.inOutSpotPrice)
-  );
```
p0mvn, line 126: This quote spot price was causing too many edge cases. Let's remove

packages/web/components/swap-tool/index.tsx
```diff
@@ -667,46 +673,42 @@ export const SwapTool: FunctionComponent<SwapToolProps> = observer(
...
-                            maxDecimals: Math.min(
-                              swapState.toAsset.coinDecimals,
-                              8
-                            ),
-                          }
-                        )
-                      : "0"
-                  }`}
-                </span>
+                {swapState.toAsset && isSwapAssetsSpotPriceAvailable && (
```
p0mvn, line 676: Made it so that the spot price is hidden if it is unavailable from quote or loading instead of being zero

packages/web/components/swap-tool/index.tsx
```diff
@@ -722,16 +724,21 @@ export const SwapTool: FunctionComponent<SwapToolProps> = observer(
...
-                    width={isMobile ? 14 : 18}
-                    className={classNames(
-                      "text-osmoverse-400 transition-all",
-                      showQuoteDetails ? "rotate-180" : "rotate-0",
-                      isQuoteDetailRelevant ? "opacity-100" : "opacity-0"
-                    )}
-                  />
+                  {
+                    // if unavailalble, hide the icon
+                    isSwapAssetsSpotPriceAvailable && (
```
p0mvn, line 729: When spot price is unavailable, this icon would be in an awkward place so made it hidden

p0mvn (approved):

PR #2993, opened 2024-03-25
refactor: cl position details clean up

p0mvn:

Light clean up. Tested locally.

### ClickUp Task

## Summary by CodeRabbit

- **Refactor**
	- Simplified the process of fetching user position details by removing unnecessary conditional logic and streamlining parameter requirements.

### DIFF
--- a/packages/server/src/queries/complex/concentrated-liquidity/index.ts
+++ b/packages/server/src/queries/complex/concentrated-liquidity/index.ts
@@ -230,20 +230,16 @@ export type UserPositionDetails = Awaited<
 /** Appends user and position details to a given set of positions.
  *  If positions are not provided, they will be fetched with the given user address. */
 export async function mapGetUserPositionDetails({
-  positions: initialPositions,
+  positions,
   userOsmoAddress,
   ...params
 }: {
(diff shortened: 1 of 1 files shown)


packages/server/src/queries/complex/concentrated-liquidity/index.ts
```diff
@@ -230,20 +230,16 @@ export type UserPositionDetails = Awaited<
 /** Appends user and position details to a given set of positions.
  *  If positions are not provided, they will be fetched with the given user address. */
 export async function mapGetUserPositionDetails({
```
jonator, line 232: I traced where this is being used. The caller simply fetches positions then passes them to this function. What if we query for positions in this function then call it `getUserPositionsDetails`?

I sense, that everywhere this function would be called, the caller would first have to fetch positions before calling this function. Moving that fetch here would slightly simplify things, but it could cause repeated querying of positions if other data is needed that is not returned by this function.

To get around that though, we could also return the original positions from that query in this function and that data would not have to be requeried in that case.

jonator:
closing as stale, can reopen

PR #2994, opened 2024-03-26
refactor: separate assets hook from swap hook

p0mvn:

Clean up the swap tool by separating the assets hook from the swap hook.

This leads to more readability and also avoiding React components rerendering due to quote updates when only assets are used within these components

### ClickUp Task

### DIFF
--- a/packages/web/components/control/token-select-with-drawer.tsx
+++ b/packages/web/components/control/token-select-with-drawer.tsx
@@ -9,13 +9,13 @@ import { TokenSelectDrawer } from "~/components/drawers/token-select-drawer";
 import { Disableable } from "~/components/types";
 import { EventName, SwapPage } from "~/config";
 import { useAmplitudeAnalytics, useWindowSize } from "~/hooks";
-import { SwapState } from "~/hooks/use-swap";
+import { SwapAssets } from "~/hooks/use-swap";
 
 /** Will display balances if provided `CoinPretty` objects. Assumes denoms are unique. */
 export const TokenSelectWithDrawer: FunctionComponent<
   {
     isFromSelect: boolean;
-    swapState: SwapState;
+    swapAssets: SwapAssets;
     dropdownOpen?: boolean;
     canSelectTokens?: boolean;
     page?: SwapPage;
@@ -25,7 +25,7 @@ export const TokenSelectWithDrawer: FunctionComponent<
 > = observer(
   ({
     isFromSelect,
-    swapState,
+    swapAssets,
     dropdownOpen,
     disabled,
     canSelectTokens = true,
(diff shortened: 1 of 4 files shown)


packages/web/components/control/token-select-with-drawer.tsx
```diff
@@ -9,13 +9,13 @@ import { TokenSelectDrawer } from "~/components/drawers/token-select-drawer";
...
 import { EventName, SwapPage } from "~/config";
 import { useAmplitudeAnalytics, useWindowSize } from "~/hooks";
-import { SwapState } from "~/hooks/use-swap";
+import { SwapAssets } from "~/hooks/use-swap";
 
 /** Will display balances if provided `CoinPretty` objects. Assumes denoms are unique. */
 export const TokenSelectWithDrawer: FunctionComponent<
   {
     isFromSelect: boolean;
-    swapState: SwapState;
```
p0mvn, line 18: For example, this component would rerender on quote update when it only should rerender if we change to and from assets

jonator (commented):
I think we can achieve the same perf optimizations if we pass in the array of selectable assets (as the `Currency` type from `@osmosis-labs/types`) as a prop to the token select drawer. That way, the drawer would only re-render when the list changes.

IMO it's cleaner if we have the asset selection be managed in the useSwap hook, so that future devs (mobile app?) just needs to use that hook to get swap functionality instead of also needing to set up the useSwapAssets first before using the useSwap hook.

Thoughts?

p0mvn:
> I think we can achieve the same perf optimizations if we pass in the array of selectable assets (as the `Currency` type from `@osmosis-labs/types`) as a prop to the token select drawer. That way, the drawer would only re-render when the list changes.
> 
> IMO it's cleaner if we have the asset selection be managed in the useSwap hook, so that future devs (mobile app?) just needs to use that hook to get swap functionality instead of also needing to set up the useSwapAssets first before using the useSwap hook.
> 
> Thoughts?

Separation of concern is a design principle that we should aim to adhere to.

Packaging several use cases into one function is not the right approach. It increases complexity.

If mobile needs to have assets, they can also call `useSwapAssets`.

With the original approach, there are several issues:
- complexity from doing several things in the swap hook
- increased number of input parameters to support these use cases
- excessive number of returns from having to return swap assets
- the previous points have led to excessive rerending because the swapState would be spread everywhere when only a few things are needed.

In my opinion, the above points make this change worthwhile

jonator:
I see, perhaps, its where the concerns are separated is where I disagree. To me, it's apparent that in order to swap, you will always need the to/from assets that are being swapped, so because of that I consider it within the same concern. So, the `useSwap` hook is justified in encapsulating that functionality and would therefore offer more utility to the caller.

p0mvn:
> I see, perhaps, its where the concerns are separated is where I disagree. To me, it's apparent that in order to swap, you will always need the to/from assets that are being swapped, so because of that I consider it within the same concern. So, the `useSwap` hook is justified in encapsulating that functionality and would therefore offer more utility to the caller.
> 
> But, you may be right as the to/from assets hook does make some assumptions around storing state in the query params, which likely won't be the case for the mobile app. So because of that, I think it's justified to keep it separate and accept your changes here. To clarify that, what if we rename `useSwapAssets` to `useQueryParamSwapAssets` instead?

I think it depends on which level of absraction we look at this.

In the context of he current system design, computing swap assets is separate from performing the swap.

PR #2999, opened 2024-03-26
Swap tool: calculate in/out spot price on client

jonator:
This PR removes the need to use a value returned from the router as the effective (execution) spot price of the trade. Instead, we calculate this on the client from the in token and quoted out token amounts.

## Summary by CodeRabbit

- **Refactor**
	- Simplified spot price calculation logic across the application.
- **Bug Fixes**
	- Improved error handling for spot price calculations in the swap functionality.
- **New Features**
	- Updated swap tool to display spot prices more accurately based on user input and quote data.


p0mvn (approved):
LGTM. Nice work.

Tested that the known bugs are fixed

PR #3009, opened 2024-03-29
chore: Add .direnv file for automatic Node.js version switching with fnm

Amosel:

Added [.direnv](https://direnv.net) config, to support that allows [fnm](https://github.com/Schniz/fnm) to automatically switch Node.js version.

### ClickUp Task

## Summary by CodeRabbit

- **Chores**
    - Enhanced Node.js version management to automatically align with the version specified in `package.json`.


.envrc
```diff
@@ -0,0 +1,6 @@
+use_node() {
+  local node_version=$(jq -r '.engines.node' package.json)
+  fnm use "$node_version"
```
jonator, line 3: What about nvm? Also, do these tools offer this functionality?

jonator: It's worth mentioning- for my purposes I use nvm and set the latest as a default alias. Perhaps devs can do something similar then we don't have to concern ourselves with maintaining this for all devs.

Amosel: You're right, this is narrowing the solution. Plus, version resolution is not trivial.
With `.nvmrc` we don't get unpredictable results looking forwards.

Amosel: I use `fnm env --use-on-cd` to get shell automatic update to version. Recommend that you explore fnm.

Amosel: You can use `fnm use --resolve-engines` and it also has a `fnm env --use-on-cd`

MaxMillington (approved):

DavideSegullo (approved):

PR #3026, opened 2024-04-02
[FE tRPC Test] Implement a happy path and variants for FE tRPC swap tool test

JoseRFelix:

This PR improves test coverage by adding E2E tests to our server package. This will ensure that the SQS returns quotes that are compatible with our FE. Tests have been implemented for both the ideal scenarios and several variant cases such as:

- Quote between USDT and USDC while also assessing that the token in amount is 5% or less of the token out amount.
- Quote for an [asset](https://github.com/osmosis-labs/sqs/actions/runs/8515867084/job/23324058331) SQS cannot compute a price. In this case, we opted for USK. 
- Quote between medium-volume tokens
- Quote between low-volume tokens
- Quote with transmuter in route. USDC.axl <> USDC
- Quote with Astroport pool in route. ASTRO <> OSMO
- Basic quote (OSMO <> ATOM) with TFM router.  

Moving forward, we can introduce additional tests as we aim to explore more edge cases involving specific quotes

## Brief Changelog

- Add E2E tests to `server` package
- Run these tests in CI. All PRs and the `stage` branch will execute these tests against both the production and staging SQS. The main (production) branch will exclusively run tests against the production SQS.

## Summary by CodeRabbit

- **Chores**
	- Updated GitHub Actions versions in CI/CD workflows for improved reliability and performance.
- **New Features**
	- Expanded linting and formatting checks to include additional packages for better code quality assurance.
	- Introduced end-to-end testing for server applications, enhancing test coverage and reliability.
- **Tests**
	- Added comprehensive end-to-end tests for the `swap-router` functionality, ensuring robustness and accuracy in token swapping operations.


packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
```diff
@@ -0,0 +1,223 @@
...
+import { swapRouter } from "../swap-router";
+
+const { createCallerFactory, router: createTRPCRouter } = initTRPC
+  .context<typeof createInnerTRPCContext>()
+  .create({
+    transformer: superjson,
+  });
+const createCaller = createCallerFactory(createTRPCRouter({ swapRouter }));
+const caller = createCaller({
+  assetLists: AssetLists,
```
jonator, line 18: Prob worth adding a comment explaining that we're smoke testing against prod data from SQS and the mock asset/chain lists are hardcoded snapshots of the prod generated lists. So, perhaps we can point out that if you're troubleshooting an issue, perhaps these snapshots need to be updated.

jonator (approved):
LGTM, nit

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
```diff
@@ -0,0 +1,223 @@
...
+  chainList: MockChains,
+});
+
+const percentageRegex = /^(<\s)?\d+\.\d+%$/;
+const atomMinimalDenom =
```
p0mvn, line 23: For my understanding, what does "minimal" stand for?

p0mvn (approved):
Great work!

Leaving some recommendations for future test cases:
- Quote between USDT and USDC. For such a quote, we will be able to make stricter assertions. For example, we can expect that token in amount is roughly equal to out amount with max relative error of say 5%
- We should trigger a case where SQS does not support a price so that a fallback is triggered. See [this](https://github.com/osmosis-labs/sqs/actions/runs/8515867084/job/23324058331) for unsupported prices.
- Quote between medium volume tokens
- Quote between low volume tokens
- Quote with transmuter in route. Maybe USDC and USDC.axl?
- Quote with Astroport pool in route. ASTRO and OSMO
- Basic quote (OSMO <> ATOM) with TFM router. This is so that we know that our fallback works if we ever need it. Note: some parts of it might be broken today

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
JoseRFelix: `minimalDenom` represents the base IBC hash denom in the Osmosis chain.

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
JoseRFelix: Thank you! This is a great suggestion. Also, because we are working on end-to-end tests, I'm considering fetching the asset and chain lists from our assetlist repo. I believe this approach will make these tests more resilient and help us avoid situations like this

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
p0mvn: This is outside of the scope of this PR. I know that the naming convention has been here before but would it make sense to standardize it with chain?

In SQS and chain, we are calling them as "ibcDenom"s. I'm noticing that the mobile is inheriting the FE naming convention.

I think that standardization across the stack would be helpful long term

JoseRFelix:
@p0mvn Thank you for the variant cases suggestion! In my latest commit, I've implemented all the recommended flows.

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
JoseRFelix: Thank you. Long term, it does make sense to ensure consistency across our entire stack. We, and maybe mobile too, chose this naming convention to maintain compatibility with Keplr's `CoinPretty` object, which follows the same definition and property name

A practical first step would be to create a dictionary document that outlines all terminology. This will help prevent similar situations from happening in the future. Following that, we'll need to devise a strategy for renaming these objects, a process that will likely require updates to our primitive objects.

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
```diff
@@ -0,0 +1,1160 @@
...
+    expect(split.initialAmount).toBeInstanceOf(Int);
+
+    expect(Array.isArray(split.pools)).toBeTruthy();
+    expect(split.pools.length).toBeGreaterThan(0);
+
+    expect(split.tokenInDenom).toBe(usdtMinimalDenom);
+
+    expect(Array.isArray(split.tokenOutDenoms)).toBeTruthy();
+    expect(split.tokenOutDenoms.length).toBeGreaterThan(0);
+  }
```
p0mvn, line 455: Are there any repeated checks that we could move into helpers for readability?

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
```diff
@@ -0,0 +1,1160 @@
...
+  expect(parseFloat(amount)).toBeGreaterThan(0);
+
+  // Swap fee
+  // expect(reply.swapFee).toBeInstanceOf(RatePretty);
+  // // Should match with the format of "0.1%"
+  // expect(reply.swapFee?.toString()).toMatch(percentageRegex);
+
+  // const swapFee = reply.swapFee!.toDec().toString();
+  // expect(isNumeric(swapFee)).toBeTruthy();
+  // expect(parseFloat(swapFee)).toBeGreaterThan(0);
```
p0mvn, line 839: Is this something that TFM doesn't return? If so, what is our plan around this? I propose making an issue, outlining all fields that are missing. Then, we will share it with them, requesting to add the missing fields to the API

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
```diff
@@ -0,0 +1,1160 @@
...
+  expect(isNumeric(reply.timeMs)).toBeTruthy();
+
+  // Token in fee amount fiat value
+  // expect(reply.tokenInFeeAmountFiatValue).toBeInstanceOf(PricePretty);
+
+  // const tokenInFeeAmountFiatValue = reply
+  //   .tokenInFeeAmountFiatValue!.toDec()
+  //   .toString();
+  // expect(isNumeric(tokenInFeeAmountFiatValue)).toBeTruthy();
+  // expect(parseFloat(tokenInFeeAmountFiatValue)).toBeGreaterThan(0);
```
p0mvn, line 892: ditto

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
```diff
@@ -0,0 +1,1160 @@
...
+
+  // amount fiat value
+  expect(reply.amountFiatValue).toBeInstanceOf(PricePretty);
+
+  const amountFiatValue = reply.amountFiatValue!.toDec().toString();
+  expect(isNumeric(amountFiatValue)).toBeTruthy();
+  expect(parseFloat(amountFiatValue)).toBeGreaterThan(0);
+});
+
+async function getSortedPoolsWithVolume() {
```
p0mvn, line 909: nit: consider adding a short spec outlining the details of the helper

p0mvn (commented):
Nice work! Overall, looks great 💪 

Left some minor suggestions and questions. Let me know what you think

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
JoseRFelix: Thank you for the review! My latest commit abstracts the assertion to a function and it's much cleaner

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
JoseRFelix: We are only missing the swap fee from their quote. I can reach out to them. Also, do you know the formula to compute the effective fee?

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
JoseRFelix: This value depends on the swap fee. It will be resolved if TFM adds it to their quotes.

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
JoseRFelix: Added it on my last commit

packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
p0mvn: Sounds great, yeah please let them know.

Here's the effective fee:
https://github.com/osmosis-labs/sqs/blob/664ad77accf906ea61cd88eff0294f1be6fca779/router/usecase/quote.go#L92

It's computed above

PR #3047, opened 2024-04-05
Fix: user balances query

DavideSegullo:

The goal of this PR is to fix all services dependent on the `queryBalances` service, such as the trpc `getUserAssetsBreakdown` route, because of paging limitations if a user has many balances at the moment he is not shown some of them.

To solve this problem I introduced a pageLimit of 1000 elements.

### Linear Task


CryptoAssassin1 (approved):
Fixed

PR #3056, opened 2024-04-09
docs: comment for batch loader cache 

p0mvn:

Lower sidecar batch loader prices cache to 1 minute.

Upon discussing this with @jonator , realized that this is a batch loader cache. Propose lowering it to 1 minute still as 10 minutes feels excessive.

### Linear Task


JoseRFelix (approved):

ValarDragon (approved):

jonator:
@p0mvn do you still think this change is worth committing, even though it doesn't affect data flows?

p0mvn:
> @p0mvn do you still think this change is worth committing, even though it doesn't affect data flows? I think a descriptive comment could have helped avoid this

I'm happy to not commit this change if there isn't a clear benefit to avoid causing unexpected instabilities. Let me add a comment and merge this

PR #3063, opened 2024-04-09
Feat: improve token info chart price fallback

DavideSegullo:

This PR improves fallback handling for token info page graphs, in particular even if a token explicitly wants to use coingecko, however we try to retrieve the data from numia, if we get no results then we use coingecko services.

image (2)

### Linear Task


JeremyParish69 (approved):
Looks good for TNKR, which is a Preview asset in CG. USDC also looks fine. stIBCX has no CG page, and also looks good.

sunnya97 (approved):

PR #3070, opened 2024-04-09
refactor/test: swap tool token in/out & fee fiat pricing

p0mvn:

This PR changes the algorithm for computing swap tool token in/out & swap fee fiat pricing.

We've had several incidents around inconsistencies of token in/out and fee prices due to various layers of cache where token out fiat value would appear greater than token in which should be impossible. We would also see prices that are incorrect/stale generally. The number of pricing sources is also a contributor to this.

While we've had longer-term efforts focused on solving this by reducing the number of pricing sources, caches and moving things to SQS, I've taken a detour to evaluate a shorter path to production.

The main goal has been to minimize the number of pricing queries needed to be made to the data sources so that we bypass caches and minimize the number of inconsistencies.

The approach that seems to be working well in local testing:
- getting a fiat price of token in
- getting price impact from quote (computed from spot price of pools in the route relative to execution prices)
- multiplying the fiat value of token in by the price impact to get the fiat value of token out
- using the swap fee from quote and multiplying it by the fiat price of token in on the client side

The downside of this approach is that the price impact only considers pools in the route. As a result, the final token out value would not take into account the broad updated prices of the market.

A side benefit of only considering pools in the route is that we are guaranteed to always display a smaller token out fiat value (assuming that price impact computation is correct). As a result, even if there is misrepresentation, it would be in user's favor.

It is worth noting the impossibility of getting a 100% correct price due to caches and asynchrony and this seems to be a marginally much better outcome than before.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/STABI-68/[swap-tool]-fix-pricing-issues-where-token-out-value-is-greater-than)

## Brief Changelog

- Exposed spot price from quote into the `useSwap` hook
- Removed obsolete pricing queries from swap swap router tRPC
   * These are now replaced by computations in the `useSwap` hook via the new algorithm
 - Implement and test `getTokenInFeeAmountFiatValue`
   * Uses the asset price from `useSwap` hook and the fee amount returned from quote to compute fee fiat value
 - Implement and test `getTokenOutFiatValue`
  * Uses the token out amount from quote, earlier computed token in fee value and the price impact from quote to compute the token out fee value
- Change slippage estimation logic by multiplying the previous results (if applicable) by the selected slippage to get a value ready for display

## Testing and Verifying

- Created unit tests for various getters, asserting the previously configured invariants
- Tested all previously called out bugs in QA

### Before and After Comparison

Before: https://app.osmosis.zone/?from=OSMO&to=ASTRO

image
- Note that token out fiat value is greater than token in fiat value which does not make sense

After:

image

### Suggestions For QA

- Test many different token pairs
- Focus on fiat values of token in, token out, token out after slippage.
- Please make sure that the magnitude of fiat values looks right
- Confirm that the fiat values are somewhat consistent with the market prices

## Summary by CodeRabbit

- **Refactor**
  - Simplified asset-related operations in the swap functionality by removing redundant fiat value calculations.
  - Enhanced the accuracy of displayed values in the `SwapTool` component by updating the calculation and display logic for fiat amounts.

- **New Features**
  - Introduced new functionalities for calculating fiat values of token transactions in `fiat-getters.ts`.
  - Added a new calculation method `mulPrice` to determine the fiat value of a coin based on its amount and price.

- **Tests**
  - Implemented new test cases to ensure the accuracy of fiat value calculations for token fees and amounts.
  - Added tests for the `mulPrice` function to validate correct multiplication results under various scenarios.

- **Bug Fixes**
  - Updated `useSwap` function to correctly compute total fees using new fiat value calculations, ensuring more precise transaction details.


packages/server/src/trpc-routers/__tests_e2e__/swap-router.test.ts
```diff
@@ -160,29 +153,6 @@ function assertValidQuote({
 
   // timeMs
   expect(isNumeric(quote.timeMs)).toBeTruthy();
-
-  // Token in fee amount fiat value
-  expect(quote.tokenInFeeAmountFiatValue).toBeInstanceOf(PricePretty);
```
p0mvn, line 165: Note to reviewer: these fields were removed to reduce the number of queries and caches. Instead, we find workarounds from other data fields in the `useSwap` hook

packages/web/components/swap-tool/index.tsx
```diff
@@ -90,17 +90,31 @@ export const SwapTool: FunctionComponent<SwapToolProps> = observer(
...
+
+      // Compute out fiat amount less slippage
+      const outFiatAmountLessSlippage = swapState.tokenOutFiatValue
+        ? new PricePretty(
+            DEFAULT_VS_CURRENCY,
+            swapState.tokenOutFiatValue?.toDec().mul(oneMinusSlippage)
+          )
+        : undefined;
+
+      return { outAmountLessSlippage, outFiatAmountLessSlippage };
```
p0mvn, line 111: Note to reviewer: similar to how we compute out amount after slippage by directly multiplying it with a slippage selection, we now also do the same with the fiat post-slippage amount

packages/web/hooks/use-swap.ts
```diff
@@ -417,9 +421,41 @@ export function useSwap({
...
+  // The price impact is computed directly from quote, ensuring most up-to-date state.
+  // This guarantees consistency between token in and token out fiat values.
+  let tokenOutFiatValue: PricePretty = useMemo(
+    () =>
+      getTokenOutFiatValue(
+        quote?.priceImpactTokenOut?.toDec(),
+        inAmountInput.fiatValue?.toDec()
+      ),
+    [inAmountInput.fiatValue, quote?.priceImpactTokenOut]
+  );
```
p0mvn, line 438: Note to reviewer: I think that we can refactor a lot of the webhooks in a similar manner to make them more testable. By extracting getters in such a way, we can define the invariants and cover them with unit tests to develop the foundation for making risky changes with more confidence.

Would appreciate feedback - maybe we could tackle some of these as part of stability?

packages/web/hooks/fiat-getters.ts
```diff
@@ -0,0 +1,71 @@
+import { Dec, DecUtils, Int, PricePretty } from "@keplr-wallet/unit";
```
JoseRFelix, line 1: I suggest we place this file within `packages/utils/src/swap.ts` as these methods will be useful for our Mobile app.

JoseRFelix (changes requested):
Thank you for this PR! Left some comments

packages/web/hooks/fiat-getters.ts
p0mvn: I have attempted this but ran into the following issues:
- There is no dependency on the `server` package from `utils`, moving this would require adding it for the following:
`import { type Asset, DEFAULT_VS_CURRENCY } from "@osmosis-labs/server";`
- It just feels awkward since these are meant to be called from hooks only so hooks feels like a better location.

I've concluded to keep it as is based on the above two points but mainly because of the first one.

Thanks for recommending and let me know if there are further objections to this

packages/web/hooks/queries/assets/use-coin-fiat-value.ts
```diff
@@ -9,13 +10,24 @@ export function useCoinFiatValue(
...
+}
 
-  return useMemo(() => {
-    if (!coin || !price) {
-      return undefined;
-    }
+// mulPrice multiplies the amount of a coin by its price to get the final value.
+// It returns the final value as a PricePretty object.
+// Returns undefined if either the amount or price is undefined.
+export function mulPrice(
```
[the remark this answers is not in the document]
p0mvn: Similar response to:
https://github.com/osmosis-labs/osmosis-frontend/pull/3070#discussion_r1558843007

I believe `FiatCurrency` would require import

packages/web/hooks/fiat-getters.ts
JoseRFelix: There seems to be some tech debt left after we created the `server` package. We can move both of these to the `utils` package. If you'd like, you can add these changes or I can create a PR. 

1) The `Asset` import should be a type called `MinimalAsset` over [`makeMinimalAsset`](https://github.com/osmosis-labs/osmosis-frontend/blob/59e4ed928dc53172e6ebc26cd58d5aeab04a3421/packages/utils/src/asset-utils.ts#L53). It would look like this 
```tsx
// In packages/types/src/asset-types.ts

export type MinimalAsset = Currency & {
  coinGeckoId: string | undefined;
  coinName: string;
  isUnstable: boolean;
  isVerified: boolean;
} 
```

```tsx
// In packages/utils/src/asset-utils.ts

import type { MinimalAsset } from "@osmosis-labs/types" 

/** Convert an asset list asset into an asset with minimal content and that
 *  is compliant with the `Currency` type. */
export function makeMinimalAsset(assetListAsset: Asset): MinimalAsset {
	...
}
```

2) Add a `packages/utils/src/currency.ts` file and export the `DEFAULT_VS_CURRENCY`


> It just feels awkward since these are meant to be called from hooks only so hooks feels like a better location.

Because these are standard TypeScript functions without reactivity, they can be invoked in any context, including on the server. Therefore, it's good to regard them as universal methods and store them in a utils folder. This approach also enhances their reusability across different platforms like mobile

I will approve this PR to not block these changes

packages/web/hooks/queries/assets/use-coin-fiat-value.ts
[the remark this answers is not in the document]
JoseRFelix: Since it's a type package, there's no worry to add it to the `utils` package, however, like my previous comment. I'll approve this PR and tentatively open a new PR with these improvements.

JoseRFelix (approved):
LGTM, thank you for these changes!

packages/web/components/swap-tool/index.tsx
```diff
@@ -90,17 +90,31 @@ export const SwapTool: FunctionComponent<SwapToolProps> = observer(
...
-                .mul(new Dec(1).sub(slippageConfig.slippage.toDec()))
-            )
-          : undefined,
-      [swapState.quote, swapState.toAsset, slippageConfig.slippage]
-    );
+          ? new IntPretty(swapState.quote.amount.toDec().mul(oneMinusSlippage))
+          : undefined;
+
+      // Compute out fiat amount less slippage
+      const outFiatAmountLessSlippage = swapState.tokenOutFiatValue
```
jonator, line 104: Thoughts on something like `outFiatValueLessSlippage`?

Removing the term "amount" and adding "value" would make the name a bit clearer IMO.

packages/web/hooks/fiat-getters.ts
```diff
@@ -0,0 +1,71 @@
...
+
+// getTokenInFeeAmountFiatValue returns the fiat value of the token in fee amount.
+// If any of the input values are undefined, it returns a PricePretty with 0 value.
+// The client must handle the case where the returned value is 0.
+// Returns zero on any truncation
+// - asset used to determine the precision exponent
+// - tokenInFeeAmount the fee amount in the token in asset (scaled according to the asset's precision)
+// - inAmountInputPrice the price of the token in asset in fiat (scaled according to the asset's precision)
+export function getTokenInFeeAmountFiatValue(
+  asset: Asset | undefined,
```
jonator, line 12: Also I think it's best the `Currency` (from "@osmosis-labs/types") type be used instead of `Asset`, it's a bit less constrictive and `Asset` conforms to `Currency`.

jonator (commented):
Some more small thoughts

packages/web/components/swap-tool/index.tsx
p0mvn: Attempted the change. The suggested naming would break consistency with another related identifier `outAmountLessSlippage`

Renaming them all introduces confusion, to address which takes a lot of the diff. I think we should pursue this separately if this is desired

packages/web/hooks/queries/assets/use-coin-fiat-value.test.ts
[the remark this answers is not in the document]
p0mvn: Yep, got auto-generated and didn't notice - apologies

packages/web/components/swap-tool/index.tsx
jonator: That's the amount less the slippage. This is the fiat value less slippage. I'm not sure I'm following?

p0mvn: `outFiatAmountLessSlippage` and `outAmountLessSlippage`

are 2 related identifiers so having a similar name improves readability. We always call `outAmount` as amount and not value. In fact, `outValue` feels more confusing.

For these reasons, I would prefer to keep the names as they are

packages/web/hooks/fiat-getters.ts
p0mvn: [9800e8b](https://github.com/osmosis-labs/osmosis-frontend/pull/3070/commits/9800e8ba53db4e33984ef04c486224f5ab2486fc)

packages/web/hooks/fiat-getters.ts
p0mvn: https://linear.app/osmosis/issue/FE-153/reduce-server-package-tech-debt

packages/web/hooks/queries/assets/use-coin-fiat-value.ts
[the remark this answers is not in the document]
p0mvn: https://linear.app/osmosis/issue/FE-153/reduce-server-package-tech-debt

kamal-sutra (approved):

CryptoAssassin1 (approved):

PR #3071, opened 2024-04-10
refactor: lower coin price cache and stale time from 5 to 3 seconds

p0mvn:

In addition to this cache client side, we also have it in SQS. We should incrementally lower the TTL in the client, to avoid the caching complexity and staleness problems around prices.

If this works well and performance is unaffected, we can evaluate further going down to 1-2 seconds

### DIFF
--- a/packages/web/hooks/queries/assets/use-coin-price.ts
+++ b/packages/web/hooks/queries/assets/use-coin-price.ts
@@ -9,8 +9,8 @@ export function useCoinPrice(coin?: CoinPretty) {
     },
     {
       enabled: Boolean(coin?.currency),
-      cacheTime: 1000 * 5, // 5 seconds
-      staleTime: 1000 * 5, // 5 seconds
+      cacheTime: 1000 * 3, // 3 second
+      staleTime: 1000 * 3, // 3 second
     }
   );
 


packages/web/hooks/queries/assets/use-coin-price.ts
```diff
@@ -9,8 +9,8 @@ export function useCoinPrice(coin?: CoinPretty) {
     },
     {
       enabled: Boolean(coin?.currency),
-      cacheTime: 1000 * 5, // 5 seconds
-      staleTime: 1000 * 5, // 5 seconds
+      cacheTime: 1000 * 3, // 3 second
+      staleTime: 1000 * 3, // 3 second
```
JoseRFelix, line 13: Please not that this only lowers the frequency at the FE fetches the prices, this won't affect the cached value in our tRPC function. Is this expected?

p0mvn: Yes, let me explain.

For displaying prices in the swap tool we used to have 2 pricing sources:
1. This
2. tRPC 


 [this PR](https://github.com/osmosis-labs/osmosis-frontend/pull/3070) removes the swap tool fiat pricing display dependency from tRPC.

Now, we only use 1). This PR simply makes it update more frequently

JoseRFelix (approved):

PR #3072, opened 2024-04-10
feat: Adjust Network Fees From Swap Amout

JoseRFelix:
### Description

This pull request is designed to modify the maximum balance input to account for the fees incurred. Specifically, if the gas fee is to be paid in the same token type as the transaction, it adjusts the transaction amount to ensure there is sufficient balance left to cover the gas fees. It's important to note, however, that due to fluctuations in the gas limit dust may remain. The primary goal of this update is to optimize transactions involving the maximum balance, rather than to minimize these dust amounts.

Error state when there is an error computing the fee or max balance quote: 

image

## Brief Changelog

- Add 'isLoading' state to UI button
- Subtract max balance tx fees to prevent tx failures.

## Testing and Verifying

- [ ] Can perform max balance transaction with OSMO as in token in swap 

## Summary by CodeRabbit

- **New Features**
	- Enhanced swap tool with support for `maxSlippage` parameter for improved precision.
	- Integrated tooltips to provide additional information in the swap tool interface.
	- Improved input handling in the swap tool by adding support for gas amount and enhancing validation.

- **Bug Fixes**
	- Removed redundant slippage calculation in the trade execution process.


packages/web/hooks/use-swap.ts
```diff
@@ -62,13 +64,16 @@ type SwapOptions = {
...
-}: SwapOptions = {}) {
+export function useSwap(
+  {
+    initialFromDenom = "ATOM",
+    initialToDenom = "OSMO",
+    useQueryParams = true,
+    useOtherCurrencies = true,
+    forceSwapInPoolId,
+    maxSlippage,
+  }: SwapOptions = { maxSlippage: undefined }
```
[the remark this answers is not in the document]
JoseRFelix: The current type definition for `maxSlippage` is `Dec | undefined`, highlighting a subtle, but important difference from `maxSlippage?: Dec`. The former requires the developer to explicitly include `maxSlippage` in their code, even if its value is `undefined`. This approach ensures the developer consciously includes this parameter, preventing it from being overlooked accidentally. In contrast, `maxSlippage?: Dec` suggests that `maxSlippage` is an optional parameter, allowing developers to omit it entirely without specifying a value

packages/web/hooks/use-faction-button-state.ts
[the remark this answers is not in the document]
JoseRFelix: I believe @Amosel wanted to separate the concerns. However, I agree with you that it's better to keep this inline

kamal-sutra (approved):

CryptoAssassin1 (approved):

packages/web/hooks/use-swap.ts
```diff
@@ -81,7 +85,50 @@ export function useSwap({
...
 
-  const inAmountInput = useAmountInput(swapAssets.fromAsset);
+  const [gasAmount, setGasAmount] = useState<CoinPretty>();
+  const inAmountInput = useAmountInput({
+    currency: swapAssets.fromAsset,
+    gasAmount: gasAmount,
+  });
+
+  const {
+    data: maxBalanceQuote,
```
Amosel, line 95: Not clear why `maxBalanceQuote` means? I am not seeing in the input anything to suggest that the max balance is passed as argument to the hook.

packages/web/hooks/use-swap.ts
```diff
@@ -81,7 +85,50 @@ export function useSwap({
...
+    !!inAmountInput.balance && !inAmountInput.balance?.toDec().isZero()
+  );
+
+  const {
+    data: maxBalanceNetworkFee,
+    isLoading: isLoadingMaxBalanceNetworkFee,
+    error: maxBalanceNetworkFeeError,
+  } = useEstimateTxFees({
+    chainId: chainStore.osmosis.chainId,
+    messages: maxBalanceQuote?.messages,
```
Amosel, line 115: Great to see this!

packages/web/hooks/use-swap.ts
```diff
@@ -81,7 +85,50 @@ export function useSwap({
...
+      !!inAmountInput.balance &&
+      !isMaxBalanceQuoteLoading,
+  });
+
+  const hasMaxBalanceError = useMemo(() => {
+    return !!maxBalanceNetworkFeeError || !!maxBalanceQuoteError;
+  }, [maxBalanceNetworkFeeError, maxBalanceQuoteError]);
+
+  useEffect(() => {
+    if (isNil(maxBalanceNetworkFee?.gasAmount)) return;
```
Amosel, line 127: Maybe it is better to move this effect to also be embedded in `useEstiamteTxFee` as optional callback, it will def reduce the clutter here.

packages/web/hooks/use-swap.ts
```diff
@@ -372,12 +308,15 @@ export function useSwap({
         inAmountInput.reset();
```
Amosel, line 308: If the promise resolves or rejects quickly, it's possible that inAmountInput.reset() could be called twice, leading to unexpected behavior. Move the inAmountInput.reset() call to the finally block to ensure it's only called once, regardless of the promise's outcome.

Amosel (changes requested):

packages/web/hooks/use-swap.ts
JoseRFelix: `maxBalanceQuote` is a quote using the account's max balance for tokenIn. In line [102](https://github.com/osmosis-labs/osmosis-frontend/pull/3072/files/3c7a635f8d2412d9b2dc0453ac10379b95594a89#diff-188329441517b71d6811aaaf32971da03955042040446222fe5fb397220e5bf9R102), we get the max balance from the `useAmountInput` hook. 

Perhaps naming could be better, any ideas?

packages/web/hooks/use-swap.ts
JoseRFelix: How about we create a separate hook, such as useSwapAmountInput, for this input flow? I'm reluctant to integrate the callback into useEstimateTxFee due to the added complexity of managing useEffect dependencies when they are nested

JoseRFelix:
@Amosel Thank you for the review 🙌  Addressed the feedback on my latest commits

packages/web/hooks/use-swap.ts
Amosel: currently it's called [quote](https://github.com/osmosis-labs/osmosis-frontend/blob/9acea4f0af4b841843c1717fed007b80696472d2/packages/web/hooks/use-swap.ts#L98C11-L98C16). It's a results of a query hook `useQueryRouterBestQuote` where the in amount value is `inAmountInput.balance?.toCoin().amount!` nothing else here tells me we are actually working with a `max value`, so it is confusing, and looks like a mistake, or it makes the expectation for a reader to know something they might not know. Maybe a better name is `quoteForCurrentBalance`?

packages/web/hooks/use-swap.ts
Amosel: Definitely a step in the right direction! Why not just allow  `useQueryRouterBestQuote` to chain `estimateTxFeesQueryFn` with an optional arg?

Amosel (changes requested):

packages/web/hooks/input/use-amount-input.ts
```diff
@@ -31,55 +39,62 @@ export function useAmountInput(currency?: Currency, inputDebounceMs = 500) {
...
   // clear fraction when user changes currency
   // and user has no balance
   useEffect(() => {
     if (isBalancesFetched && !rawCurrencyBalance) setFraction(null);
   }, [isBalancesFetched, rawCurrencyBalance, currency]);
 
-  /** Amount derived from user input or from a fraction of the user's balance. */
+  const isHalf = useMemo(() => fraction === 0.5, [fraction]);
+  const isMax = useMemo(() => fraction === 1, [fraction]);
+
```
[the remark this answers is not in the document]
JoseRFelix: I'm not completely sure if displaying the 'max' button as selected will effectively represent the current state, considering the maximum value can change. I think it might be better for it to simply reflect a dynamic state that updates with new fee estimations. Also, it’s quite rare for a user to hit the exact maximum number on their own. Furthermore, using estimates to display this value could mislead users in such a critical part of our process. In my opinion, we should treat this as a separate task with additional product considerations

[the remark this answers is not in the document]
Amosel: @JoseRFelix  @jonator I feel this unsatisfactory. That is I am not sure it is that rare for a user to hit the exact maximum number on their own (I personally keep doing that when I am trying things out) nor I am sure that it is not part of the app to provide assurances for. Additionally leaving a complex state without tests feels very risky for me.

However, I don't think it's a good idea to keep this PR waiting because it's pretty good. So I'm resolving this, although the concern is not.

Amosel (approved):

packages/web/hooks/input/use-amount-input.ts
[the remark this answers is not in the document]
JoseRFelix: I appreciate your input on this. I see you’ve already created a ticket for it in Linear. Let’s keep the conversation going and aim to resolve this in a separate PR using that ticket. Thank you 🙌

PR #3075, opened 2024-04-10
Optimize market assets tRPC procedure

jonator:
* Use prices from /tokens/all instead of batched queries to SQS prices

### DIFF
--- a/packages/server/src/queries/complex/assets/market.ts
+++ b/packages/server/src/queries/complex/assets/market.ts
@@ -78,9 +78,7 @@ export async function mapGetMarketAssets<TAsset extends Asset>({
 } & AssetFilter): Promise<(TAsset & AssetMarketInfo)[]> {
   if (!assets) assets = getAssets({ ...params }) as TAsset[];
 
-  return await Promise.all(
-    assets.map((asset) => getMarketAsset({ ...params, asset }))
-  );
+  return await Promise.all(assets.map((asset) => getMarketAsset({ asset })));
 }
 
 const assetMarketCache = new LRUCache<string, CacheEntry>(DEFAULT_LRU_OPTIONS);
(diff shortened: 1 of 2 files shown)


packages/server/src/queries/complex/assets/market.ts
```diff
@@ -27,24 +26,17 @@ export type AssetMarketInfo = Partial<{
...
     key: `market-asset-${asset.coinMinimalDenom}`,
     ttl: 1000 * 60 * 5, // 5 minutes
     getFreshValue: async () => {
-      const [currentPrice, marketCap, assetMarketActivity, coingeckoCoin] =
-        await Promise.all([
-          getAssetPrice({ assetLists, chainList, asset }).catch((e) =>
-            captureErrorAndReturn(e, undefined)
-          ),
+      const [marketCap, assetMarketActivity, coingeckoCoin] = await Promise.all(
+        [
```
p0mvn, line 39: Do we need to couple all 3 queries together and block them on each other? I think these could be split and fetched separately

jonator: This allows them to be fetched concurrently. WDYM separately?

p0mvn: Let's say one query times out for one or the other reason. Then, the entire operation/display is blocked.

By fetching data separately and displaying on-demand, we will maximize fault-tolerance and enable best-effort display

p0mvn: To be clear, this is outside of the scope of this PR as it would require larger changes. However, I think these are easy performance and fault-tolerance wins that we can tackle

jonator: The catch statements would still allow the promise to resolve, and the error would get logged in Sentry, and undefined would get returned. So, yes, `Promise.all` does throw if one of the promises throws, but since we use `catch()` on all the contained promises then we prevent them all from throwing.

p0mvn: My comment is not limited to `Promise.all` constraints. What I'm trying to point out is that, for example, we don't need to block displaying asset market cap on market activity data.

As a result, these should be fetched via separate operations as opposed to one

p0mvn (approved):
LGTM.

I still think we can evaluate [this](https://github.com/osmosis-labs/osmosis-frontend/pull/3075/files#r1559862874) separately

PR #3076, opened 2024-04-10
(Observability) Set Up Distributed Tracing and Send Only Swap Transactions to Sentry

JoseRFelix:
### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/STABI-88/[swap-tool][observability]-limit-sentry-traces-to-swap-tool)

### DIFF
--- a/packages/web/sentry.client.config.ts
+++ b/packages/web/sentry.client.config.ts
@@ -20,9 +20,8 @@ Sentry.init({
   replaysSessionSampleRate: 0.1,
 
   // Temporarily disable due to client-side noise coming from extensions, Cosmos Kit, etc.
-  enabled: false,
-  // enabled:
-  //   process.env.NODE_ENV !== "development" && process.env.NODE_ENV !== "test",
+  enabled:
+    process.env.NODE_ENV !== "development" && process.env.NODE_ENV !== "test",
 
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
   debug: false,
--- a/packages/web/sentry.client.config.ts
+++ b/packages/web/sentry.client.config.ts
@@ -20,8 +20,9 @@ Sentry.init({
   replaysSessionSampleRate: 0.1,
 
   // Temporarily disable due to client-side noise coming from extensions, Cosmos Kit, etc.
-  enabled:
-    process.env.NODE_ENV !== "development" && process.env.NODE_ENV !== "test",
+  enabled: false,
+  // enabled:
+  //   process.env.NODE_ENV !== "development" && process.env.NODE_ENV !== "test",
 
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
   debug: false,
--- a/packages/web/sentry.client.config.ts
+++ b/packages/web/sentry.client.config.ts
@@ -12,6 +12,9 @@ Sentry.init({
   environment:
     process.env.NEXT_PUBLIC_VERCEL_ENV || process.env.NODE_ENV || "development",
 
+  // Only send 25% of error events to Sentry
+  sampleRate: 0.25,
+
   // Adjust this value in production, or use tracesSampler for greater control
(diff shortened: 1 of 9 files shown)


packages/web/sentry.server.config.ts
```diff
@@ -11,11 +11,20 @@ Sentry.init({
...
-  tracesSampleRate: 0.05,
+  tracesSampleRate: 0.01,
 
-  enabled:
-    process.env.NODE_ENV !== "development" && process.env.NODE_ENV !== "test",
+  /**
+   * Disable server-side (Node.js serverless) tracing since we are only interested
+   * in tracing swap transactions, specifically including bridge transfer transactions.
+   */
+  enabled: false,
```
p0mvn, line 20: For my understanding, what components are encapsulated by this?

packages/web/sentry.client.config.ts
```diff
@@ -4,14 +4,30 @@
...
   dsn: "https://c696452bb7ce4cc98150142ebea1c32f@o4505285755600896.ingest.us.sentry.io/4505285757698048",
 
   environment:
     process.env.NEXT_PUBLIC_VERCEL_ENV || process.env.NODE_ENV || "development",
 
   // Adjust this value in production, or use tracesSampler for greater control
-  tracesSampleRate: 0.05,
+  tracesSampler: (samplingContext) => {
+    const validPaths = ["/", "/pool/[id]", "/assets/[denom]"];
+    const validTrpcRoutes = getValidSwapTRPCRoutesForSentry();
```
p0mvn, line 18: Why are tRPC routes relevant in the client?

packages/web/sentry.client.config.ts
```diff
@@ -4,14 +4,30 @@
...
 Sentry.init({
   dsn: "https://c696452bb7ce4cc98150142ebea1c32f@o4505285755600896.ingest.us.sentry.io/4505285757698048",
 
   environment:
     process.env.NEXT_PUBLIC_VERCEL_ENV || process.env.NODE_ENV || "development",
 
   // Adjust this value in production, or use tracesSampler for greater control
-  tracesSampleRate: 0.05,
+  tracesSampler: (samplingContext) => {
+    const validPaths = ["/", "/pool/[id]", "/assets/[denom]"];
```
p0mvn, line 17: Which one of these guarantees that `useSwap` hook is visible?

packages/web/sentry.server.config.ts
JoseRFelix: Currently, we only run the bridge transfer router in Node.js serverless functions because of crypto package dependency. Furthermore, The comment is not clear enough. I'll push a commit clarifying this.

packages/web/sentry.client.config.ts
JoseRFelix: In our FE, tRPC routers can run locally. For example, we have the [localRouter](https://github.com/osmosis-labs/osmosis-frontend/blob/1edcc9b92711303a65d3cb69415d9604b1d4832d/packages/web/server/api/root-router.ts#L13-L16) that includes all of these functions. Since they are local, they begin their trace in the browser.

packages/web/sentry.client.config.ts
JoseRFelix: The `useSwap` uses this file. All errors reported in the browser will be forwarded to Sentry.

p0mvn: We previously discussed browser causing too much noise. Just asking for context - have we gotten a chance to address this?

p0mvn (approved):
LGTM, nice work!

packages/web/sentry.client.config.ts
JoseRFelix: Thanks for reminding me. I'll reduce the error reporting sample rate to 25%. We'll revisit this once we make some improvements to reduce the noise coming from Cosmos-Kit

PR #3095, opened 2024-04-15
feat: remember previous trade in swap tool

JoseRFelix:
### Linear Task


packages/web/tests/test-utils.tsx
jonator, line 1: I'm wondering if this file should be in a __tests__ folder somewhere. It's nice, IMO, having peace of mind that all code that is used for testing is in __tests__ and otherwise code outside of those folders is used in prod.

jonator (commented):
Minor comment

CryptoAssassin1 (approved):

kamal-sutra (approved):

jonator (approved):

jonator (changes requested):
Was the requirement to remember the last _traded_ tokens? Or the last _selected_ tokens? I imagine it would be similarly useful to be able to go to/from the swap page and have the last tokens you selected still selected

packages/web/__tests__/test-utils.tsx
```diff
@@ -8,29 +8,37 @@ import { ReactNode } from "react";
...
-            url: "http://localhost:3000/trpc",
-          }),
-        ],
-      })}
-      queryClient={queryClient}
-    >
-      <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
-    </trpcReact.Provider>
-  </MultiLanguageProvider>
+  <StoreProvider>
```
Amosel, line 21: Would be nice to see this being more configurable in the future.

Amosel (approved):

JoseRFelix:
> Was the requirement to remember the last _traded_ tokens? Or the last _selected_ tokens? I imagine it would be similarly useful to be able to go to/from the swap page and have the last tokens you selected still selected

@jonator  The initial requirement was to keep track of the last traded tokens. We can plan to implement tracking the last selected tokens in a future PR. I wouldn't want to delay this feature

packages/web/__tests__/test-utils.tsx
JoseRFelix: Certainly. It's designed to be flexible enough to extend as issues arise during testing

packages/web/hooks/window/use-query-param-state.ts
```diff
@@ -11,8 +11,10 @@ type QueryParamValue =
   | null
   | undefined;
 
-/** Emulates `React.useState` but uses next/router to store state in URL query params.
- *  Only sets from `defaultValue` if it's not present in the query params. */
+/**
+ * Emulates `React.useState` but uses next/router to store state in URL query params.
+ * Only sets from `defaultValue` if it's not present in the query params.
+ */
 export function useQueryParamState<TValue extends QueryParamValue>(
```
DavideSegullo, line 18: Instead of using `useEffect` and manually manage query params, inside the pools table we're already using [`nuqs`](https://nuqs.47ng.com/docs/installation), do you think we can use it here too? It can simplify the code a bit, but mostly it avoids unnecessary rerendering

jonator (approved):

packages/web/hooks/window/use-query-param-state.ts
JoseRFelix: Thank you for the suggestion! To avoid expanding the scope and delaying the current PR, I have created an [issue](https://linear.app/osmosis/issue/FE-197/sync-swap-tools-token-query-params-with-nuqs) in Linear to address this in a separate PR.

PR #3097, opened 2024-04-16
Feat: add persistent queries cache

DavideSegullo:

The goal is to integrate a local cache using localstorage for the queries we make, this allows on page refresh to still have data and while doing so we invalidate it in the background if necessary based on the stale time configured for each query.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-188/implement-tanstack-persist-storage)

## Summary by CodeRabbit

- **New Features**
	- Enhanced data synchronization and persistence capabilities in the app using new libraries.
- **Chores**
	- Updated development dependencies and cleaned up unnecessary regular dependencies.


packages/web/utils/trpc.ts
jonator, line 1: I think we have a few direct uses of react query throughout the app. I'm assuming this logic won't apply to those uses?

Perhaps it's enough to just have this persistence logic for tRPC queries.

DavideSegullo: The logic will apply to all of them, I assume that trpc creates an instance of the provider that tanstack query requires, but I can verify that.

Do you mean can we exclude mutations?

packages/web/utils/trpc.ts
```diff
@@ -26,7 +29,27 @@ const getBaseUrl = () => {
...
+    const localStoragePersister = createSyncStoragePersister({
+      storage: typeof window !== "undefined" ? window.localStorage : undefined,
+      serialize: (client) => superjson.stringify(client),
+      deserialize: (cachedString) => superjson.parse(cachedString),
+    });
+
+    const queryClient = new QueryClient({
+      defaultOptions: {
+        queries: {
+          cacheTime: 1000 * 60 * 60 * 24, // 24 hours
```
JoseRFelix, line 41: Wouldn't this cache time affect other queries? This could unintentionally impact prices

DavideSegullo: This one is a global config yes, but it doesn't impact other queries in a bad way, because it's only the cacheTime:
https://tanstack.com/query/v4/docs/framework/react/plugins/persistQueryClient#how-it-works

By default all queries on tanstack, unless otherwise specified, have a[ `staleTime` equal to zero.](https://tanstack.com/query/v4/docs/framework/react/guides/caching#basic-example)
This means that when we load the page, if we find cached data we show it right away, but since the `staleTime` is equal to zero in the background revalidation of the data will take place and it will then be updated, in short as long as we don't have fresh data for up to 24 hours it shows the cached one.

It may not be a problem for the prices, however, it is definitely to be paid attention to during testing

packages/web/utils/trpc.ts
jonator: So, just to make sure, it will apply to these as well? tRPC is able to set our global react query setup?

https://github.com/osmosis-labs/osmosis-frontend/blob/45b3ad1e0650307edfdeedcac8d3aeea22fae900/packages/web/hooks/queries/cosmos/use-balances.ts#L19

JoseRFelix: That's right. Under the hood, the `queryClient` is propagated through a context so all `useQuery`s will leverage it.

DavideSegullo: Yes it works, thank to the tanstack query context:
Screenshot 2024-04-17 alle 08 50 26

CryptoAssassin1 (approved):

kamal-sutra (approved):

jonator (approved):
Nice. This makes the app appear faster for subsequent visits 🔥

DavideSegullo:
> Nice. This makes the app appear faster for subsequent visits 🔥. Why not set the value to `Infinity` for subsequent visits? Since we know it will always be re-fetched since it's stale?

The idea of the `cacheTime` is to determine after how long that data can be considered totally invalid, assuming the background fetch fails for more than 24 hours, before the end of the `cacheTime` we will continue to see the last available data, after which it will be removed from the cache and possibly if expected show an error.
This is kind of the approach of an offline app first.

In any case, whenever the refetch is successful the end date of the cacheTime gets longer, so if the request never fails it is as if the cache is infinite say

JoseRFelix (approved):

PR #3098, opened 2024-04-16
WIP: Add auto-login for swaps (DO NOT MERGE)

nimesh-notifi:
Allow login to Notifi to begin receiving updates immediately after a tx has occurred.

Periodically grab a nonce from Notifi in the background.
When a tx is executed, put the nonce in the memo.
Let the Notifi card context know that a new tx has just executed, with the hash so that it can complete auth flow.

## Testing and Verifying

WIP looking for feedback on if we should modify the root store wrap the StoreProvider with a Notifi context.

### DIFF
--- a/packages/web/stores/index.tsx
+++ b/packages/web/stores/index.tsx
@@ -2,7 +2,6 @@ import React, { FunctionComponent, useState } from "react";
 
 import { RootStore } from "~/stores/root";
 import { api } from "~/utils/trpc";
-
 const storeContext = React.createContext<RootStore | null>(null);
 
 /** Once data is invalidated, React Query will automatically refetch data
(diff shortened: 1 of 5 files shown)


packages/web/stores/index.tsx
```diff
@@ -22,7 +21,12 @@ export const StoreProvider: FunctionComponent = ({ children }) => {
       new RootStore({
         txEvents: {
           onBroadcastFailed: () => invalidateQueryData(apiUtils),
-          onFulfill: () => invalidateQueryData(apiUtils),
+          onFulfill: () => {
+            invalidateQueryData(apiUtils);
+            // NOTIFI QUESTION - We need a way to let the Notifi card know that the transaction has been fulfilled.
```
nimesh-notifi, line 26: @jonator we couldn't relay from here, as it'd require a Notifi context to wrap the StoreProvider

packages/web/stores/root.ts
```diff
@@ -85,6 +85,10 @@ export class RootStore {
 
   public readonly profileStore: ProfileStore;
 
+  // NOTIFI QUESTION - We didn't want to modify the store here, but didn't see a better place to have
```
nimesh-notifi, line 88: @jonator not ideal, but this was the only way we could come up with sharing bidirectional state with Notifi context, and any possible transactions that are being executed

jonator:
Let's pause work on this feature, can reopen

PR #3101, opened 2024-04-16
improvement: Fetch Token In Price Before Quote

JoseRFelix:

We currently fetch the price after generating a quote, which causes a delay visible to users on slow networks—they see a blank space where the price should be while it loads. To address this and improve efficiency, we will start fetching the price as soon as the swap tool loads or after a token is selected.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/STABI-106/on-swap-page-getting-the-base-asset-prices-should-be-fetched-in)

## Testing and Verifying

- [ ] Can perform swaps
- [ ] Token in and token out price loads

## Summary by CodeRabbit

- **New Features**
	- Enhanced the logic for price calculation based on input amount and currency in the input handling functionality.

### DIFF
--- a/packages/web/hooks/input/use-amount-input.ts
+++ b/packages/web/hooks/input/use-amount-input.ts
@@ -102,14 +102,14 @@ export function useAmountInput(currency?: Currency, inputDebounceMs = 500) {
     setDebounceInAmount(amount ?? null);
   }, [setDebounceInAmount, amount]);
 
-  let coin: CoinPretty | undefined;
+  let coinForPrice: CoinPretty | undefined;
   if (!isNil(amount)) {
-    coin = amount;
+    coinForPrice = amount;
   } else if (!isNil(currency)) {
-    coin = new CoinPretty(currency, 0);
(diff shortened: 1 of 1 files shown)


MaxMillington (approved):

packages/web/hooks/input/use-amount-input.ts
```diff
@@ -101,7 +102,14 @@ export function useAmountInput(currency?: Currency, inputDebounceMs = 500) {
     setDebounceInAmount(amount ?? null);
   }, [setDebounceInAmount, amount]);
 
-  const { price } = useCoinPrice(amount);
+  let coinForPrice: CoinPretty | undefined;
+  if (!isNil(amount)) {
+    coinForPrice = amount;
+  } else if (!isNil(currency)) {
+    coinForPrice = new CoinPretty(currency, 0);
+  }
```
p0mvn, line 110: nit: let's add a comment explaining why we are doing this please

p0mvn (approved):
🥪

PR #3123, opened 2024-04-19
Improve: token info page chart

DavideSegullo:

This PR improve some behaviours inside the token info page:
1) When the user is not hovering on the chart, now we show the sqs price of that asset, this ensures that there is no volatility problem.
2) Fixed a bug that allowed case sensitive symbols to be used, this broke some queries

### Linear Task

[FE-133](https://linear.app/osmosis/issue/FE-133/add-sqs-price-for-chart-price)
[FE-202](https://linear.app/osmosis/issue/FE-202/remove-case-sensitivity-from-token-info-page)

## Summary by CodeRabbit

- **New Features**
	- Enhanced asset information configuration to include support for `coinMinimalDenom` and optional `coingeckoId`, improving API query conditions and observable properties.
	- Updated asset pages to utilize minimal denominations and improved data handling in `TokenChartHeader`.

- **Refactor**
	- Simplified the logic for setting hover prices in `TokenChart`.
	- Removed outdated functions and variables, streamlining the asset information handling process.

- **Bug Fixes**
	- Adjusted `getStaticProps` to correctly handle new token denominations, ensuring accurate static data generation.

### DIFF
--- a/packages/web/hooks/ui-config/use-asset-info-config.ts
+++ b/packages/web/hooks/ui-config/use-asset-info-config.ts
@@ -4,17 +4,20 @@ import {
   type TokenHistoricalPrice,
 } from "@osmosis-labs/server";
 import dayjs from "dayjs";
-import { action, autorun, computed, makeObservable, observable } from "mobx";
+import { action, computed, makeObservable, observable } from "mobx";
 import { useEffect, useMemo } from "react";
 
 import { api } from "~/utils/trpc";
 
 export const useAssetInfoConfig = (
   denom: string,
-  queryDenom: string | null,
+  coinMinimalDenom?: string,
   coingeckoId?: string
 ) => {
-  const config = useMemo(() => new ObservableAssetInfoConfig(), []);
+  const config = useMemo(
+    () => new ObservableAssetInfoConfig(denom, coinMinimalDenom),
+    [denom, coinMinimalDenom]
+  );
 
   useEffect(
     () => () => {
@@ -67,13 +70,13 @@ export const useAssetInfoConfig = (
     isError,
   } = api.edge.assets.getAssetHistoricalPrice.useQuery(
     {
-      coinDenom: denom ?? queryDenom,
+      coinDenom: denom,
       timeFrame: {
         custom: customTimeFrame,
       },
     },
     {
-      enabled: Boolean(denom ?? queryDenom),
+      enabled: Boolean(denom),
       staleTime: 1000 * 60 * 3, // 3 minutes
       cacheTime: 1000 * 60 * 6, // 6 minutes
       trpc: {
@@ -212,6 +215,12 @@ export class ObservableAssetInfoConfig {
 
   protected _disposers: (() => void)[] = [];
 
+  @observable
+  denom: string;
+
+  @observable
+  coinMinimalDenom?: string;
+
   @action
   readonly setHistoricalData = (data: TokenHistoricalPrice[]) => {
     this._historicalData = data;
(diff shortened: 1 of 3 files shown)


fabryscript (approved):

CryptoAssassin1 (approved):

packages/web/pages/assets/[denom].tsx
```diff
@@ -421,35 +420,51 @@ const TokenChartSection = () => {
 const TokenChartHeader = observer(() => {
   const { assetInfoConfig } = useAssetInfoView();
 
-  const minimumDecimals = 2;
-  const maxDecimals = Math.max(
-    getDecimalCount(
-      (assetInfoConfig.hoverPrice?.toDec() ?? new Dec(0)).toString()
-    ),
-    minimumDecimals
+  const { data, isLoading } = api.edge.assets.getAssetPrice.useQuery(
```
Comment (jonator) [inline packages/web/pages/assets/[denom].tsx:423]: NIT: may be a bit more readable if we rename `data` to `data: assetPrice`

packages/web/pages/assets/[denom].tsx
```diff
@@ -616,32 +594,9 @@ export const getStaticProps: GetStaticProps<AssetInfoPageProps> = async ({
    * Lookup for the current token
    */
   const token = currencies.find(
-    (currency) =>
-      currency.coinDenom.toUpperCase() === tokenDenom.toLocaleUpperCase()
+    (currency) => currency.coinDenom.toUpperCase() === tokenDenom.toUpperCase()
```
Comment (jonator) [inline packages/web/pages/assets/[denom].tsx:597]: Is it possible to use `getAsset` here? Instead of processing the chain list? We could make this change later when we get to migrating the asset detail page fully to tRPC but it's worth a mention and your thoughts.

jonator (approved):

packages/web/pages/assets/[denom].tsx
Comment (DavideSegullo) [reply packages/web/pages/assets/[denom].tsx]: Done

packages/web/pages/assets/[denom].tsx
Comment (DavideSegullo) [reply packages/web/pages/assets/[denom].tsx]: Done

Comment (jonator) [reply packages/web/pages/assets/[denom].tsx]: Nice!

jonator (changes requested):
Let's run the loading state by the design team before merging this PR. To me it looks a bit off.

kamal-sutra (approved):

packages/web/hooks/ui-config/use-asset-info-config.ts
```diff
@@ -212,6 +216,12 @@ export class ObservableAssetInfoConfig {
 
   protected _disposers: (() => void)[] = [];
 
+  @observable
+  denom: string;
+
+  @observable
+  coinMinimalDenom?: string;
```
jonator, line 223: Does this need to be observable state, or can it just be readonly properties (defined statically as `readonly` in constructor params)

jonator (changes requested):

packages/web/hooks/ui-config/use-asset-info-config.ts
DavideSegullo: Good point, it can be an observable because we are also using a memo, so if they change we'll have a new instance of `ObservableAssetInfoConfig`.
I'll refactor it

DavideSegullo: Done

DavideSegullo:
> Let's run the loading state by the design team before merging this PR. To me it looks a bit off.

I improved it

jonator (approved):
Nice, loading looks better

packages/web/hooks/ui-config/use-asset-info-config.ts
```diff
@@ -281,19 +289,11 @@ export class ObservableAssetInfoConfig {
     return this._historicalRange;
   }
 
-  constructor() {
+  constructor(denom: string, coinMinimalDenom?: string) {
```
jonator, line 292: BTW it's possible to remove a few LOC elsewhere in the class if you add `readonly` in front of each param, then they can be accessed directly via `this.denom` (without having to set `this.denom = denom`)

PR #3125, opened 2024-04-19
Portfolio page: existing deposit/withdraw flows

jonator:

Only shows deposit/withdraw buttons when relevant (has a balance to withdraw). Opens external link modal for assets that prefer an external link transfer method.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-181/portfolio-depositwithdraw)

## Brief Changelog

* Append bridge info from asset list v2 to assets returned for portfolio page assets table
* Conditionally show deposit/withdraw buttons
* Use external link modal for external deposit/withdraw methods

## Summary by CodeRabbit

- **New Features**
	- Introduced a new modal behavior with an optional setting to force display without opt-out, improving user interaction flexibility.
	- Added functionality to append bridge information to assets, enhancing asset management capabilities.

- **Enhancements**
	- Updated asset filtering and mapping in the assets router to support new bridge asset functionality.
	- Enhanced the `AssetActionsCell` to handle external transfer URLs, facilitating smoother asset operations.


jonator:
- [ ] Ensure external URLs are working from "Continue to site" button in external link modal: test LUNC

packages/server/src/queries/complex/assets/bridge.ts
```diff
@@ -0,0 +1,30 @@
...
+/** Appends bridge info to a given asset. If asset is not found in asset list, empty bridge info will be returned.
+ *  @throws if a given asset is not found in asset list.
+ */
+export function getBridgeAsset<TAsset extends Asset>(
+  assetLists: AssetList[],
+  asset: TAsset
+): TAsset & BridgeAsset {
+  const assetListAsset = assetLists
+    .flatMap(({ assets }) => assets)
+    .find((a) => a.coinMinimalDenom === asset.coinMinimalDenom);
```
JoseRFelix, line 20: Nit:

```suggestion
  import { getAssetFromAssetList } from "@osmosis-labs/utils"

  const assetListAsset = getAssetFromAssetList({
     coinMinimalDenom: asset.coinMinimalDenom
  })
```

packages/server/src/queries/complex/assets/bridge.ts
```diff
@@ -0,0 +1,30 @@
...
+/** A bridgeable asset. */
+export type BridgeAsset = {
+  transferMethods: AssetListAsset["transferMethods"];
+  counterparty: AssetListAsset["counterparty"];
+};
+
+/** Appends bridge info to a given asset. If asset is not found in asset list, empty bridge info will be returned.
+ *  @throws if a given asset is not found in asset list.
+ */
+export function getBridgeAsset<TAsset extends Asset>(
```
JoseRFelix, line 14: What are your thoughts on including the `transferMethod` and `counterparty` to our `getAssetFromAssetList` instead of adding a new method?

packages/web/components/table/asset-balances.tsx
```diff
@@ -253,6 +257,12 @@ export const AssetBalancesTable: FunctionComponent<{
 
   return (
     <div className="w-full">
+      <ExternalLinkModal
+        url={externalUrl ?? ""}
+        isOpen={Boolean(externalUrl)}
+        onRequestClose={useCallback(() => setExternalUrl(null), [])}
```
JoseRFelix, line 263: Any reason why we are using a `useCallback`?  It becomes an overhead if the component does not perform reference checks which I believe `ExternalLinkModal` doesn't.

JoseRFelix (changes requested):
Looks good! Left some comments

packages/web/components/table/asset-balances.tsx
jonator: I'll admit, I thought React only re-rendered children when props changed, but it appears it only does that when we use memo. Otherwise it always re-renders children.

https://react.dev/reference/react/memo

packages/server/src/queries/complex/assets/bridge.ts
jonator: See other comment

packages/server/src/queries/complex/assets/bridge.ts
jonator: Also, to clarify, are you suggesting removing this function altogether? Or just using `getAssetFromAssetList` within it?

packages/server/src/queries/complex/assets/bridge.ts
JoseRFelix: Since we're not tying this to the `getAssetFromAssetList` method, adding the `getAssetFromAssetlist` here shouldn’t hinder our ability to switch to another data source later on. We’re including it mainly to take advantage of the `find` we already know works well. What do you think?

DavideSegullo (approved):

packages/server/src/queries/complex/assets/bridge.ts
JoseRFelix: I'm now suggesting we include the `getAssetFromAssetList` within

packages/server/src/queries/complex/assets/bridge.ts
jonator: True, but I like to avoid using utils such as those as it adds an unnecessary dependency that could be avoided. Also, as a new dev, I'm unaware of that `getAssetFromAssetList` does under the hood. Does it modify asset list data? I'm not sure, unless I jump into the implementation. With just a flatmap and a find, I know I'm just reading data from the asset list. It's more transparent and involves less indirection. I think we've discussed this before but that's my rationale.

JoseRFelix (approved):
LGTM

PR #3132, opened 2024-04-20
feat: Add `useAmountInput` Integration Tests, Take Into Account Manual Entry for Max/Half, and Change Tokens Automatically If Not Enough Balance To Pay Fees

JoseRFelix:

This PR aims to address all outstanding issues related to gas amount calculations in the Swap Tool. It introduces comprehensive integration tests for the amount input and now includes manual entries in calculations for maximum and half amounts. 

Additionally, now the `useEstimateTxFees` hook automatically switches the fee token if the remaining balance is insufficient to cover the fees, choosing from other available fee tokens associated with that account. This update should also eliminate the problem of residual 'dust' amounts left in a token when the maximum amount is used.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/STABI-125/useamountinput-gas-fees-does-not-take-in-account-manual-amount-entry)

## Testing and Verifying

- [ ] Can swap tokens
- [ ] `max` transactions are successful for OSMO, and some other random tokens

## Summary by CodeRabbit

- **New Features**
	- Added `InsufficientBalanceForFeeError` to improve error messaging for balance issues.
	- Enhanced `SwapTool` with updated tooltip content and button states based on balance and error conditions.
	- Introduced new test wallet functionality for better simulation and testing.

- **Enhancements**
	- Extended `AccountStore` functionalities with custom signing options and fee exclusion settings.
	- Improved fee estimation logic in `useEstimateTxFees` and related hooks to better handle various scenarios.

- **Bug Fixes**
	- Adjusted error handling in `useAmountInput` to incorporate `InsufficientBalanceForFeeError`.

- **Tests**
	- Added and updated tests for account management, input handling, and transaction fee estimation.

- **Documentation**
	- Updated error translation support for new error types in localization components.


DavideSegullo (approved):

packages/web/hooks/input/use-amount-input.ts
```diff
@@ -82,33 +104,37 @@ export function useAmountInput({
...
+        ? new Dec(amount.toCoin().amount).equals(
+            new Dec(rawCurrencyBalance).mul(new Dec(0.5))
+          )
+        : false),
+    [amount, fraction, rawCurrencyBalance]
+  );
+  const isMaxValue = useMemo(
+    () =>
+      fraction === 1 ||
+      (amount && maxAmount ? amount.toDec().equals(maxAmount.toDec()) : false),
```
jonator, line 135: What about a value that is not equal to max, but is still high enough to cause the out of balance gas error?

packages/stores/src/account/base.ts
```diff
@@ -1185,10 +1192,14 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
     gasLimit,
     chainId,
     address,
+    checkOtherFeeTokens = true,
```
jonator, line 1195: Is there a reason we wouldn't want the behavior by default?

packages/web/hooks/use-estimate-tx-fees.ts
```diff
@@ -91,6 +135,7 @@ export function useEstimateTxFees({
         accountStore,
```
jonator, line 135: NIT: on line 130 could we change the query key to be `"estimate-tx-fees"` instead of `"simulate-swap-tx"`

packages/web/hooks/use-estimate-tx-fees.ts
```diff
@@ -26,34 +26,70 @@ async function estimateTxFeesQueryFn({
...
   wallet: AccountStoreWallet<[OsmosisAccount, CosmosAccount, CosmwasmAccount]>;
   accountStore: AccountStore<[OsmosisAccount, CosmosAccount, CosmwasmAccount]>;
   messages: EncodeObject[] | undefined;
   apiUtils: ReturnType<typeof api.useUtils>;
+  sendToken?: {
+    balance: CoinPretty;
+    amount: CoinPretty;
+  };
 }): Promise<QueryResult> {
   if (!messages) throw new Error("No messages");
```
jonator, line 40: If we're going to throw an error immediately at runtime if `messages` is undefined, why don't we just change the `messages` type to not be capable of being undefined at build time?

packages/web/hooks/use-estimate-tx-fees.ts
jonator, line 135: `superjson.stringify(messages)`

Should we also stringify the `sendToken`, so if that changes it triggers a new query?

jonator (changes requested):

packages/web/hooks/input/use-amount-input.ts
JoseRFelix: This error is handled on [these](https://github.com/osmosis-labs/osmosis-frontend/pull/3132/files/f7b0c7533183a6333a78cd46c4189259c9664b8a#diff-c4bd127a6249c1fc41dc3ea0d6f89afb55d71b6bfd22f9e731e94c7cd1804c76R189-R191) lines. There is also a [test](https://github.com/osmosis-labs/osmosis-frontend/pull/3132/files/f7b0c7533183a6333a78cd46c4189259c9664b8a#diff-69ecd68b40ff73e084ebec4e17d6d2ee010c11d3535936e5a4b161abee678e72R153-R164)

packages/stores/src/account/base.ts
JoseRFelix: This is used for 1-Click trading. I accidentally included this from that PR. I'll remove this as we don't need it now

packages/web/hooks/input/use-amount-input.ts
jonator: Hmm. Like consider the scenario where a total tx amount is 10 OSMO, gas is 1 OSMO, and the user manually enters 9.1 OSMO as the amount.

They would be short 0.1 OSMO to pay the fee, and I'm not sure this logic would handle that unless the user entered 10 OSMO (either with max or manually)

JoseRFelix: Perhaps I'm overlooking something, but we would encounter an insufficient balance error if the user doesn't have enough funds for this transaction. Are you suggesting that there should be a different response or behavior?

jonator: Oh ok. I may not have looked at the logic close enough. Thanks for confirming.

packages/stores/src/account/base.ts
```diff
@@ -1043,12 +1045,14 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
...
     memo = "",
     signOptions = {},
+    excludedFeeMinimalDenoms = [],
   }: {
     wallet: AccountStoreWallet;
     messages: readonly EncodeObject[];
     initialFee?: Optional<TxFee, "gas">;
     memo?: string;
     signOptions?: SignOptions;
+    excludedFeeMinimalDenoms?: string[];
```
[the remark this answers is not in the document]
JoseRFelix: Not sure why it didn't pick up the inline doc, but it's on this [line](https://github.com/osmosis-labs/osmosis-frontend/pull/3132/files/9c628d042f9fad2bdaa3a5366fa42c41328c1764#diff-bb0fbc30c7bc07c03059acc0214134a21f144bbcb6d284f90097c54908c66ac5R1023)

kamal-sutra (approved):

jonator (approved):

PR #3135, opened 2024-04-22
Display compacted zeros on PriceChartHeader with compressZeros func…

fabryscript:
…tion

Display compact zeros on token chart header like CoinGecko

### Linear Task

### DIFF
--- a/packages/web/components/chart/compress-zeros.ts
+++ b/packages/web/components/chart/compress-zeros.ts
@@ -1,0 +1,82 @@
+const ZERO = "0";
+const DOUBLE_ZERO = "00";
+
+const countZeros = (decimalDigits: string) => {
+  // Count the leading zeros
+  let zeroCount = 0;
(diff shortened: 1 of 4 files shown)


DavideSegullo (approved):

packages/web/components/chart/compress-zeros.ts
jonator, line 1: Thoughts on adding this functionality to `formatPretty` so it can more easily be reused?

fabryscript: That would be difficult imo, as `formatPretty` returns a string and `compressZeros` returns an atomic-style object. That would mean that we should modify every `formatPretty` call.
The "problem" is that `compressZeros` is meant to be paired together with a `<sub />` element or any other form of styling for the values returned by the function itself

kamal-sutra (approved):

CryptoAssassin1 (approved):

PR #3141, opened 2024-04-22
feat: Display Pool Tokens in Swap Tool

JoseRFelix:

There was a regression in the swap tool behavior on the pool details page following #3097. This PR resolved this by hiding the token select drawer. However, For StableSwap pools with >2 tokens, they will be able to see the token drawer.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/STABI-127/fix-swap-tool)

## Brief Changelog

- Hide drawer when pools assets are less than 3
- Show drawer when pool assets are greater than 2
- Add integration test verifying this behavior
- Divide swap tool props to give devs better control of desired behavior on the different pages. 

## Testing and Verifying

- [ ] `/pool/908` should display the token select drawer without search and recommended assets as it has three assets
- [ ] Swap tool in other pages should be unaffected

## Summary by CodeRabbit

- **New Features**
	- Added filtering of assets based on pool ID in asset retrieval queries.
	- Introduced optional search box and recommended token toggles in the token selection components.
	- Enhanced the Swap Tool with new properties to support fixed width, usage of other currencies, and handling query parameters.
	- Modified the Trade Tokens modal to support transactions in multiple currencies.

- **Bug Fixes**
	- Updated the Swap Tool to ensure correct initial token denomination is set based on the asset page being viewed.

- **Tests**
	- Added mock implementation for IntersectionObserver to improve testing capabilities.

- **Documentation**
	- No visible changes for end-users.

### DIFF
--- a/packages/web/components/swap-tool/__tests__/swap-tool.spec.tsx
+++ b/packages/web/components/swap-tool/__tests__/swap-tool.spec.tsx
@@ -1,0 +1,231 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { Dec, PricePretty } from "@keplr-wallet/unit";
+import {
+  createCallerFactory,
+  DEFAULT_VS_CURRENCY,
+  FilteredPoolsResponse,
+  NumPoolsResponse,
+} from "@osmosis-labs/server";
+import { getAssetFromAssetList } from "@osmosis-labs/utils";
+import { screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { rest } from "msw";
+import mockRouter from "next-router-mock";
+
+import { server, trpcMsw } from "~/__tests__/msw";
+import { renderWithProviders } from "~/__tests__/test-utils";
+import { SwapTool } from "~/components/swap-tool";
+import { AssetLists } from "~/config/generated/asset-lists";
+import { ChainList } from "~/config/generated/chain-list";
+import { appRouter } from "~/server/api/root-router";
+
+jest.mock("next/router", () => jest.requireActual("next-router-mock"));
+
+const createCaller = createCallerFactory(appRouter);
+const caller = createCaller({
+  /**
+   * The asset list and chain list used here are snapshots of the
+   * production asset list. In case of discrepancies due to outdated
+   * assets or chains, consider updating these mocks.
+   */
+  assetLists: AssetLists,
+  chainList: ChainList,
+});
+
+beforeEach(() => {
+  server.use(
+    trpcMsw.edge.assets.getUserAssets.query(async (req, res, ctx) => {
+      return res(
+        ctx.status(200),
+        ctx.data(await caller.edge.assets.getUserAssets(req.getInput()))
+      );
+    }),
+    trpcMsw.edge.assets.getAssetPrice.query((_req, res, ctx) => {
+      return res(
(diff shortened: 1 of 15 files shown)


jonator (changes requested):
What are your thoughts on handling this force pool ID case on the client, by querying the pool on the client and returning it's reserve coins as the selectable assets, instead of modifying the getUserAssets procedure? I think it's OK if the user balances are not available in the dropdown in the direct swap, as the dropdown is only available in the edge case of a > 2 asset stable pool anyways.

JoseRFelix:
> What are your thoughts on handling this force pool ID case on the client, by querying the pool on the client and returning it's reserve coins as the selectable assets, instead of modifying the getUserAssets procedure? I think it's OK if the user balances are not available in the dropdown in the direct swap, as the dropdown is only available in the edge case of a > 2 asset stable pool anyways.

I've considered this as well, but I believe it would complicate the code and make testing more challenging because it would depend on reactivity rather than a function's output. Additionally, it's simpler to write a unit test for the `mapGetAssetsWithUserBalances` function, which already calculates the balances. Perhaps I'm overlooking something. Do you think it would be unwise to include the pool id as a filter option solely for the `getUserAssets` function?

packages/web/components/swap-tool/__tests__/swap-tool.spec.tsx
```diff
@@ -0,0 +1,231 @@
...
+      );
+    })
+  );
+});
+
+afterEach(() => {
+  mockRouter.setCurrentUrl("/");
+});
+
+it("should return only pool assets when sending a pool id and useOtherCurrencies is true", async () => {
```
jonator, line 56: I'm wondering, should we test this kind of functionality at the UI level or at the function level?

I think tests may be easier to maintain if we test these complex flows at the TS function level, then for UI tests we mock the complex flows from these functions then just test the fact that the UI properly displays it.

I think a good principle for tests is to try to only test one layer at a time, and the other layers can be carefully mocked. This violates that principle as it tests the UI as well as the complex server function.

jonator (approved):
Actually, I do see a use case of passing a pool to filter getting assets only for those in that pool. 2 ideas to improve

* Adding a comment documenting the use of the pool ID
* Consider passing an array of pool IDs so we could filter user assets across multiple pools (could be over engineering)
* See comments on what layer the tests should live (test strategy). I think we don't need to necessarily rewrite the test but I think it's an idea that may simplify and increase maintainability of tests.

packages/web/components/swap-tool/__tests__/swap-tool.spec.tsx
jonator: BTW don't let this block the PR

JoseRFelix: Thank you for the review! Overall, this is a solid approach and a good idea for our general testing strategy.

In this case, I wanted to focus on how much confidence the tests give us rather than just their coverage. I feel more assured when we minimize our use of mocks. For the swap tool, in particular, I want us to be very confident that everything will work seamlessly with the client as long as the network provider is up and running. I expect these tests to be infrequent and reserved only for very critical workflows, such as the Swap Tool.

CryptoAssassin1 (approved):

PR #3144, opened 2024-04-23
change N/D on token info to be '-'

sunnya97:
### Linear Task

## Summary by CodeRabbit

- **Documentation**
	- Updated the display text for "noData" from "N.D." to "-" in English localization settings.
	- Updated various localization files to remove the "noData" key or translation, impacting the display of "no data" messages in multiple languages.

### DIFF
--- a/packages/web/components/token-details/token-details.tsx
+++ b/packages/web/components/token-details/token-details.tsx
@@ -290,7 +290,7 @@ const TokenStats: FunctionComponent<TokenStatsProps> = observer(
             {t("tokenInfos.marketCapRank")}
           </p>
           <h5 className="text-xl font-h5 leading-8">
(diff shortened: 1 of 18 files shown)


jonator (commented):
Should this be in translations? Or removed from translations and installed in the view (where this key was formerly used)?

DavideSegullo (approved):

DavideSegullo:
> Should this be in translations? Or removed from translations and installed in the view (where this key was formerly used)?

We're using it inside the token info page

jonator (dismissed):
@DavideSegullo true but I don't think "-" needs to be translated, therefore it shouldn't be in the translation files?

DavideSegullo:
> @DavideSegullo true but I don't think "-" needs to be translated, therefore it shouldn't be in the translation files? 

Ye we can get rid of It from translations

PR #3151, opened 2024-04-23
SQS routes: add check for additional error message

jonator:
https://linear.app/osmosis/issue/STABI-129/fe-client-for-sqs-routes-handle-no-candidate-routes-found-as

### DIFF
--- a/packages/server/src/queries/sidecar/router.ts
+++ b/packages/server/src/queries/sidecar/router.ts
@@ -70,7 +70,10 @@ export class OsmosisSidecarRemoteRouter implements TokenOutGivenInRouter {
 
       const errorMessage = error.data?.message;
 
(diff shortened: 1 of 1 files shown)


PaddyMc (approved):
We need to add this check as Astro port pools don't have enough liquidity but are available on the as listed tokens on the frontend, we need to add this check to ensure that users get correct feedback on pool routing.

Run query below to see reason:
```bash
https://sqsprod.osmosis.zone/router/quote?tokenIn=1000000ibc%2F498A0751C798A0D9A389AA3691123DADA57DAA4FE165D5C75894505B876BA6E4&tokenOutDenom=ibc%2FB8C608CEE08C4F30A15A7955306F2EDAF4A02BB191CABC4185C1A57FD978DA1B
```

Thanks @jonator

PR #3155, opened 2024-04-24
improve the price format on assets page

sunnya97:
### Linear Task

## Summary by CodeRabbit

- **Enhancements**
	- Improved the display formatting of asset prices in the `AssetsInfoTable` for clearer and more detailed financial information.


jonator (commented):
Seems trailing 0s are not being accounted for:
cc @DavideSegullo 
Screenshot 2024-04-24 at 9 37 16 PM

sunnya97:
No, that's on purpose, cause it should always have minimum 3 sig figs.

However, we should maybe switch this to use the "compacted zeros" from here:  https://github.com/osmosis-labs/osmosis-frontend/pull/3135/files

jonator:
@sunnya97 do you still want to use compacted 0s? That seems to be the move now, I think CoinGecko does that

sunnya97:
We can switch it to the compacted 0s, once the compacted 0s fix the trailing 0s issue

PR #3156, opened 2024-04-24
feat: Take Into Account Spread Factor for Token Out Value

JoseRFelix:

This PR modifies the token out value by subtracting the swap fee, so users can see the true expected outcome of their swap.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/STABI-113/token-out-fiat-value-does-not-account-for-spread-factor)

## Testing and Verifying

- [ ] Subtracts swap fee from token out and token out fiat value

## Summary by CodeRabbit

- **New Features**
	- Updated the `SwapTool` component to enhance the accuracy of displayed values by incorporating swap fees into the calculation of token out amounts.
- **Tests**
	- Added tests for the new `getTokenOutMinusSwapFee` function to ensure reliable calculation of token amounts post-swap fees.
- **Refactor**
	- Refined the logic in the `use-swap` hook to improve the consistency and precision in fee and token value calculations.

### DIFF
--- a/packages/web/components/swap-tool/index.tsx
+++ b/packages/web/components/swap-tool/index.tsx
@@ -103,8 +103,10 @@ export const SwapTool: FunctionComponent<SwapToolProps> = observer(
 
       // Compute out amount less slippage
       const outAmountLessSlippage =
-        swapState.quote && swapState.toAsset
-          ? new IntPretty(swapState.quote.amount.toDec().mul(oneMinusSlippage))
+        swapState.tokenOutAmountMinusSwapFee && swapState.toAsset
+          ? new IntPretty(
+              swapState.tokenOutAmountMinusSwapFee.toDec().mul(oneMinusSlippage)
+            )
           : undefined;
 
       // Compute out fiat amount less slippage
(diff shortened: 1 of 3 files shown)


jonator (commented):
Wait, why doesn't SQS return an amount with the swap fee included?

JoseRFelix:
> Wait, why doesn't SQS return an amount with the swap fee included?

I've also thought about this. Because we need the `tokenOutMinusSwapFee` to calculate `tokenOutFiatValue`, moving this to SQS is not straightforward. This PR serves as a temporary solution until we can transition to SQS, which should happen in about two weeks when Roman and I return from vacation. Once back, we plan to tackle the price caching issues, which initially prompted moving the `tokenOutFiatValue` calculation to the frontend. Once this is resolved, we'll be able to calculate the `tokenOutFiatValue` and the token out amount minus the swap fee using SQS, allowing us to remove this code

JoseRFelix:
> Wait, why doesn't SQS return an amount with the swap fee included?

I agree with this, and it's the correct solution. Unfortunately, the swap tool is currently not delivering accurate data to users, and I don't have the bandwidth or context to implement the necessary changes in SQS. This PR is intended as a temporary fix until we can move this to the server. I plan to make this move in about two weeks, once Roman is back from vacation.

jonator:
@JoseRFelix this is very surprising to me. I feel like spread factor should be included when swaps are simulated 🤔

CryptoAssassin1 (approved):

jonator (approved):

PR #3189, opened 2024-04-30
Mattupham/fe 346 fix table overflow behavior on mobile

mattupham:

- design QA blocking / non blocking tasks

### Linear Task

[fix table overflow behavior on mobile](https://linear.app/osmosis/issue/FE-346/fix-table-overflow-behavior-on-mobile)
[fix warp and jank with drawer slide transitions](https://linear.app/osmosis/issue/FE-343/fix-warp-and-jank-with-drawer-slide-transitions)
[reduce page title size on mobile](https://linear.app/osmosis/issue/FE-345/reduce-page-title-size-on-mobile)
[Flip images to be on right side of price, make column widths 1/3](https://linear.app/osmosis/issue/FE-361/flip-images-to-be-on-right-side-of-price-make-column-widths-13)

## Brief Changelog

- changes for the above tickets

## Testing and Verifying

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/0e027039-95ae-40d2-bb29-1aae9c7ab0c4

Screenshot 2024-04-30 at 3 58 32 PM

### DIFF
--- a/packages/server/src/trpc-routers/assets-router.ts
+++ b/packages/server/src/trpc-routers/assets-router.ts
@@ -420,10 +420,6 @@ export const assetsRouter = createTRPCRouter({
             coinMinimalDenom: asset.coinMinimalDenom,
           });
 
-          if (listingDate) {
-            console.log("listingDate", asset.coinDenom, listingDate);
-          }
-
           return {
             ...asset,
             listingDate,


packages/web/components/transactions/transaction-content.tsx
```diff
@@ -48,7 +48,7 @@ export const TransactionContent = ({
   return (
     <div className="flex w-full flex-col pb-16">
       <div className="flex w-full justify-between pt-8 pb-4">
-        <h1 className="text-h3">{t("transactions.title")}</h1>
+        <h1 className="text-h3 md:text-h5">{t("transactions.title")}</h1>
```
jonator, line 51: To me it's notable here we have an h1 tag (that gets styles applied from globals.css), with text for h3 and h5 also applied as tailwind classes (but not the fonts for h3 and h5). Thoughts?

jonator: Yeah I've heard that from nass as well, I think it's on track. 

But, we need to find a way to also apply the font classes for h* styles as well. Perhaps as we did with `body*`, `subtitle*`, etc. classes in globals.css, we should also make classes that re-apply the text and font tailwind classes that are getting applied to the h* elements themselves.

jonator (approved):

PR #3193, opened 2024-04-30
Revert "Publish Stage"

MaxMillington:
Reverts osmosis-labs/osmosis-frontend#3191


jonator:
FYI: we reverted the last deployment via Vercel, causing the same effect of this PR, but without delay.

I believe publishing stage should promote a new deployment to production over the rolled back deployment.

PR #3203, opened 2024-05-02
Update DYM and NIM Deposit Overrides

JeremyParish69:

To update DYM and NIM to use the correct deposit URL for the Dym portal

## Brief Changelog

- Update ibc-overrides.ts depositURL for DYM and NIM

## Testing and Verifying

Validation successful--DYM Deposit now goes to the correct interface.


jonator (approved):
Soon, you don't need to do this 🙏

PR #3204, opened 2024-05-02
Swap Amplitude event: use valid number in swap event

jonator:
* Use valid number in swap event property (remove "$" by using `toDec()`)
* Include property in swap started event
* Change swap button disabled to more explicitly be if there's a quote available, instead of if it's loading (more accurate)

### DIFF
--- a/packages/web/components/swap-tool/index.tsx
+++ b/packages/web/components/swap-tool/index.tsx
@@ -201,9 +201,6 @@ export const SwapTool: FunctionComponent<SwapToolProps> = observer(
           router: swapState.quote?.name,
         },
       ]);
-      console.log({
-        valueUsd: swapState.tokenOutFiatValue?.toDec().toString(),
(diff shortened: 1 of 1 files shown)


packages/web/components/swap-tool/index.tsx
```diff
@@ -188,16 +188,22 @@ export const SwapTool: FunctionComponent<SwapToolProps> = observer(
...
       logEvent([
         EventName.Swap.swapStarted,
         {
           ...baseEvent,
           quoteTimeMilliseconds: swapState.quote?.timeMs,
           router: swapState.quote?.name,
-          page,
         },
       ]);
+      console.log({
```
DavideSegullo, line 204: Do we need this?

DavideSegullo (changes requested):

DavideSegullo (approved):

PR #3211, opened 2024-05-06
feat: add token chart queries by denom

DavideSegullo:

Migrate to the new numia service to get the chart, instead of using the token symbol we can now use the denom, this should reduce the errors of searching for a token.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-310/migrate-from-tokensv2symbol-to-tokensv2denom)

### DIFF
--- a/packages/server/src/queries/data-services/token-historical-chart.ts
+++ b/packages/server/src/queries/data-services/token-historical-chart.ts
@@ -42,14 +42,20 @@ export async function queryTokenHistoricalChart({
   coinDenom,
   timeFrameMinutes,
 }: {
-  /** Major (symbol) denom to fetch historical price data for. */
+  /**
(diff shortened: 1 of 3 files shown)


packages/server/src/queries/data-services/token-historical-chart.ts
```diff
@@ -42,14 +42,20 @@ export async function queryTokenHistoricalChart({
   coinDenom,
   timeFrameMinutes,
 }: {
-  /** Major (symbol) denom to fetch historical price data for. */
+  /**
+   * Major (symbol) denom to fetch historical price data for.
+   *
+   * Note: this can be both a symbol or a denom (coinMinimalDenom)
```
jonator, line 48: Good comment

jonator (approved):

kamal-sutra (approved):

PR #3221, opened 2024-05-07
Cherry pick counterparty fix

jonator:
Unblock asset listings immediately


jonator:
Unrelated build errors

PR #3235, opened 2024-05-08
feat: upgrade cosmos kit

DavideSegullo:

This PR upgrade cosmos-kit deps, it's a required upgrade to support new cosmiframe for DAODAO and it also bring some fixes

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-431/upgrade-cosmos-kit-to-support-cosmiframe-for-daodao)


jonator:
Let's make sure to test this thoroughly

DavideSegullo:
> Let's make sure to test this thoroughly

Yep, I already asked for a deep QA

packages/stores/src/account/base.ts
```diff
@@ -948,10 +947,7 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
       ? wallet.client.signDirect(
           wallet.chainId,
           signerAddress,
-          {
-            ...signDoc,
-            accountNumber: Long.fromString(signDoc.accountNumber.toString()),
-          },
+          signDoc,
```
jonator, line 950: Was there a guide for this? If you used a guide a link in the PR would be good for record keeping

If not, disregard

jonator (approved):

NoahSaso:
tested and works with latest DAO DAO iframe upgrade!

packages/stores/src/account/base.ts
DavideSegullo: No there isn't a guide I was just getting ts errors, I think this happens because now they use the latest version of cosmjs

DavideSegullo:
@JoseRFelix @jonator I am testing but deposit and withdraw are broken because of this upgrade, there are infinite loading

jonator (changes requested):
I'm getting some sort of wild jitter on transfer modals (M1 Macbook pro)

It may be due to an infinite loop causing wallet connected state to keep toggling

https://github.com/osmosis-labs/osmosis-frontend/assets/4606373/5249e85a-3a9f-4886-9960-638c09916133

DavideSegullo:
> I'm getting some sort of wild jitter on transfer modals (M1 Macbook pro)
> 
> It may be due to an infinite loop causing wallet connected state to keep toggling
> 
>  Screen.Recording.2024-05-14.at.11.13.27.AM.mov

@jonator @JoseRFelix The tricky part is here: https://github.com/osmosis-labs/osmosis-frontend/pull/3235/commits/4eb7c0cd13386c0a56d2f40bf70f4ea87335239c#diff-bb0fbc30c7bc07c03059acc0214134a21f144bbcb6d284f90097c54908c66ac5L234

Somehow this part changed now inside cosmoskit and these callbacks are executed differently and at different times than we expected them to be, this caused constant state refreshes.

With this change it's working but tell me you who know the store better if it makes sense, I'll ask the QA team to do a few more rounds

jonator (approved):

DavideSegullo:
@jonator there are some errors, qa team gave me new feedback.... Keplr works fine but cosmostatiom doesnt work

CryptoAssassin1:
> @jonator there are some errors, qa team gave me new feedback.... Keplr works fine but cosmostatiom doesnt work

@DavideSegullo there are some issues with Keplr too please check the slack thread

CryptoAssassin1 (approved):

kamal-sutra (approved):

PR #3245, opened 2024-05-09
Mattupham/fe 350 table row bg on select

mattupham:

- when row is selected, apply background state to row

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-350/table-row-bg-on-hover)

## Brief Changelog

- update hover state
- update background state
- update x button to not scroll with page
- simplify selected transaction type to use undefined instead of null as default
- add delay for slideover close for a smoother transition

## Testing and Verifying

https://github.com/osmosis-labs/osmosis-frontend/assets/30577966/919a20f7-6711-4c72-96c0-da38bcac90a2

### DIFF
--- a/packages/web/components/transactions/transaction-row.tsx
+++ b/packages/web/components/transactions/transaction-row.tsx
@@ -1,5 +1,4 @@
 import { CoinPretty, PricePretty } from "@keplr-wallet/unit";
-import { FormattedTransaction } from "@osmosis-labs/server";
 import classNames from "classnames";
 import { FunctionComponent } from "react";
 
@@ -16,8 +15,7 @@ export type TransactionStatus = "pending" | "success" | "failed";
 type Effect = "swap" | "deposit" | "withdraw";
 
 interface Transaction {
-  selectedTransaction?: FormattedTransaction;
-  hash: string;
+  isSelected?: boolean;
   status: TransactionStatus;
   /** At a high level- what this transaction does. */
   effect: Effect;
@@ -44,7 +42,7 @@ interface Transaction {
 }
 
 export const TransactionRow: FunctionComponent<Transaction> = ({
-  hash,
+  isSelected = false,
   status,
   effect,
   title,
@@ -52,7 +50,6 @@ export const TransactionRow: FunctionComponent<Transaction> = ({
   tokenConversion,
   transfer,
   onClick,
-  selectedTransaction,
 }) => {
   const effectIconId = effect === "swap" ? "swap" : "down-arrow";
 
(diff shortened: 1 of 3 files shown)


packages/web/pages/transactions.tsx
```diff
@@ -105,8 +105,11 @@ const Transactions: React.FC = observer(() => {
     ctas: [],
   });
 
-  const [selectedTransaction, setSelectedTransaction] =
-    useState<FormattedTransaction | null>(null);
+  const [selectedTransaction, setSelectedTransaction] = useState<
+    FormattedTransaction | undefined
+  >(undefined);
+
+  console.log("selectedTransaction: ", selectedTransaction);
```
jonator, line 112: Log

packages/web/components/transactions/transaction-row.tsx
```diff
@@ -15,6 +16,8 @@ export type TransactionStatus = "pending" | "success" | "failed";
 type Effect = "swap" | "deposit" | "withdraw";
 
 interface Transaction {
+  selectedTransaction?: FormattedTransaction;
```
jonator, line 19: We could simplify this type by adding an `isSelected` boolean flag instead of requiring the whole object. Thoughts?

packages/web/components/transactions/transaction-rows.tsx
```diff
@@ -10,11 +10,13 @@ import { useAmplitudeAnalytics, useTranslation } from "~/hooks";
 
 export const TransactionRows = ({
   transactions,
+  selectedTransaction,
```
jonator, line 13: Similar to prev comment, thought on just using the hash as the ID for the selected tx? We access and compare the hash anyways, this makes it a bit easier to set the selected transaction: perhaps, in the scenario some piece of code only has the hash and not the whole object.

jonator (dismissed):

DavideSegullo (approved):

packages/web/pages/transactions.tsx
```diff
@@ -105,8 +105,9 @@ const Transactions: React.FC = observer(() => {
     ctas: [],
   });
 
-  const [selectedTransaction, setSelectedTransaction] =
-    useState<FormattedTransaction | null>(null);
+  const [selectedTransaction, setSelectedTransaction] = useState<
+    FormattedTransaction | undefined
```
jonator, line 109: IDK if you've seen this comment in other PRs, but IMO things would be simplified and more flexible if we stored IDs (in this case tx hashes) instead of the objects themselves

jonator (approved):

packages/web/pages/transactions.tsx
jonator: We could use `find(hash)` to return the selected tx, and only render the popover if that is true, which is safer as we aren't at risk for memory leaks, etc.

PR #3247, opened 2024-05-10
Move bridge providers out of `@osmosis-labs/web`

jonator:

Increases modularity by moving bridge providers into it's own package: `@osmosis-labs/bridge`.

The bridge providers are subject to change with the new deposit/withdraw project.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-360/move-bridge-providers-out-of-osmosis-labsweb)

## Brief Changelog

* Copied in some dependencies from web into bridge folder
* Renamed to help clarify
* Updated README to provide more documentation and examples
* Remove some files by moving some modules around
* Dependency inject some queries needed by providers (to avoid circular dependency with server package and to increase testability and portability)

This only includes one change to logic. All other changes are (mainly) just moving symbols:
* Move quote price querying out of bridge provider and into tRPC procedure
  * This increases fairness and accuracy by ignoring the USD fiat value that bridge providers return (Squid has been known to use old price data)

## TODO
- [ ] Investigate the 400kb balloon in base bundle

## Testing and Verifying

Tests pass.


packages/bridge/src/axelar/transfer-status.ts
```diff
@@ -1,27 +1,27 @@
-import { ITxStatusReceiver, ITxStatusSource } from "@osmosis-labs/stores";
+import { poll } from "@osmosis-labs/utils";
 
 import type {
-  BridgeProviderContext,
   BridgeTransferStatus,
+  Environment,
```
JoseRFelix, line 5: Thoughts on renaming this to something more descriptive like `BridgeEnvironment`?

packages/bridge/package.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "keywords": [
+    "osmosis",
+    "bridge"
+  ],
+  "license": "Apache-2.0",
+  "sideEffects": [
+    "./build/utils/superjson.js",
+    "./build/utils/dayjs.js"
+  ],
```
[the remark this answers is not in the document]
jonator: Good catch

packages/bridge/src/axelar/transfer-status.ts
jonator: I can

But since the context of the package is for bridging, wouldn't prepending everything with `Bridge` be redundant? Then the caller, if they choose (or need to clarify their context) can prepend bridge to the type upon import

JoseRFelix: This is a good point. I believe this could improve import autocomplete and enhance code readability. However, I'm fine with whatever you consider is best

JoseRFelix (approved):

PR #3250, opened 2024-05-10
Remove price query from providers

jonator:
Targets branch containing progress to extract bridge providers from web package.

Query prices in tRPC procedure from tokens returned by bridge providers instead of querying for prices within each provider.

This:
* Reduces repeated logic
  * No need to parse raw fiat value strings in some providers
* Encourages fair fiat prices from bridge providers
* Removes a query dependency from bridge providers (in prep for move to separate package)


JoseRFelix (approved):
LGTM!

PR #3252, opened 2024-05-10
Attempt smaller amount to swap in for spot price with no amount entered

p0mvn:

Context: When no amount is entered into swap tool, the spot price between token in and out is estimated by using a quote of 1 unit of token in.

With high-value tokens like WBTC, this causes a severe price impact and a price that is quite off.

To mitigate this, we added a temporary hack to divide the amount in by 10. This seems to have helped the WBTC and USDC pair.

### Linear Task


p0mvn:
Tried another change

jonator (approved):

CryptoAssassin1 (approved):

kamal-sutra (approved):

p0mvn:
I think e2e test is failing on stage. Given 3 ACKs and tests done, I'm merging this

PR #3271, opened 2024-05-22
feat: remove token holding filter and minor ui bug

DavideSegullo:
removed token holding filter if we're using yout strategy tab.

Inside the earn page we are going to remove "token holding" filter if you're inside "your strategies" tab, because the data is already filtered for the tokens you have.

 
Screenshot 2024-05-22 alle 10 26 23

This PR also ensure that we render `<SearchBox />` using `outline` variant everywhere inside the earn page.

### Linear Task

### DIFF
--- a/packages/web/components/earn/filters/top-filters.tsx
+++ b/packages/web/components/earn/filters/top-filters.tsx
@@ -123,14 +123,18 @@ export const TopFilters = ({
   return (
     <div className="flex flex-col gap-5 px-10 py-8 1.5xs:px-7 1.5xs:py-7">
       <div className="flex flex-wrap items-center justify-between gap-7 2xl:gap-10 1.5xl:gap-4 lg:hidden">
-        <RadioWithOptions
-          disabled={tokenHolderSwitchDisabled}
-          mode="primary"
-          variant="large"
-          value={tokenHolder}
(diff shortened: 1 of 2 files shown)


packages/web/components/earn/filters/top-filters.tsx
```diff
@@ -123,14 +123,18 @@ export const TopFilters = ({
...
+          <RadioWithOptions
+            disabled={tokenHolderSwitchDisabled}
+            mode="primary"
+            variant="large"
+            value={tokenHolder}
+            onChange={(value) => setFilter("tokenHolder", value)}
+            options={tokenFilterOptions}
+          />
+        ) : (
+          false
```
fabryscript, line 136: Why not use `!tokenHolderSwitchDisabled && <Component />`?

fabryscript (approved):

packages/web/components/earn/filters/top-filters.tsx
DavideSegullo: `&&` operator can lead unexpected behaviours:
https://www.crocoder.dev/blog/react-conditional-rendering/

MaxMillington: interesting.

CryptoAssassin1 (approved):

PR #3281, opened 2024-05-28
bump code id for astroport pcl pools

sunnya97:
### Linear Task


p0mvn (approved):

packages/server/src/queries/complex/pools/env.ts
```diff
@@ -2,7 +2,7 @@ import { IS_TESTNET } from "../../../env";
 
 /** Cosmwasm Code Ids confirmed to be transmuter pools in current env. */
 const TransmuterPoolCodeIds = IS_TESTNET ? ["3084"] : ["148"];
-const AstroportPclPoolCodeIds = IS_TESTNET ? ["5005"] : ["580"];
+const AstroportPclPoolCodeIds = IS_TESTNET ? ["8611"] : ["773"];
```
MaxMillington, line 5: If these are going to change frequently, we probably want something like this:

```
let AstroportPclPoolCodeIds;

if (process.env.IS_TESTNET) {
    AstroportPclPoolCodeIds = process.env.NEXT_PUBLIC_TESTNET_ASTROPORT_ID || process.env.TESTNET_ASTROPORT_ID;
} else {
    AstroportPclPoolCodeIds = process.env.NEXT_PUBLIC_ASTROPORT_ID || process.env.ASTROPORT_ID;
}
```

And then put the NEXT_PUBLIC ids in env variables in vercel and the regular env variables in the .env file for third parties.

MaxMillington (approved):
Approving in case we want to get this out quick

PR #3286, opened 2024-05-30
feat: improve build speed

DavideSegullo:

- add bun package manager.
- get rid of `pre-commit` it's a very old dep (published 7 years ago), instead now we can use [husky](https://typicode.github.io/husky/).

### Linear Task


jonator:
@DavideSegullo do you think these conflicts are fixable? Or should we revisit this later?

DavideSegullo:
> @DavideSegullo do you think these conflicts are fixable? Or should we revisit this later?

Maybe we can revisit this PR later, also today/tomorrow bun will get a very important update so it will be much easier to integrate inside a monorepo

PR #3298, opened 2024-06-04
Gas estimation fixes

jonator:
* Return eligible fee token in order returned from chain or config (base asset first, AKA OSMO)
* Fix issue with extracting sent coins from gas simulation events, add test case


PaddyMc (approved):
lgtm

PR #3307, opened 2024-06-07
Use query params for limit tool tabs and implement new designs

fabryscript:

Implement new designs for the tab system in the new Home page with "Buy", "Sell" and "Swap" options and manage it with query params.
### Linear Task

[Linear Task URL 1](https://linear.app/osmosis/issue/LIM-105/implement-tab-system-with-nuqs)
[Linear Task URL 2](https://linear.app/osmosis/issue/LIM-106/implement-new-tabs-design)

### DIFF
--- a/packages/web/components/trade-tool/index.tsx
+++ b/packages/web/components/trade-tool/index.tsx
@@ -26,6 +26,7 @@ export const TradeTool: FunctionComponent<TradeToolProps> = () => {
             case SwapToolTab.SELL:
               return <PlaceLimitTool orderDirection={OrderDirection.Ask} />;
             case SwapToolTab.SWAP:
+            default:
               return (
                 <SwapTool useOtherCurrencies useQueryParams page="Swap Page" />
               );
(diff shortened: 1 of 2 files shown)


fabryscript:
Currently local testing is available only through `localosmosis`

packages/web/components/trade-tool/index.tsx
```diff
@@ -11,25 +12,27 @@ import { OrderDirection } from "~/hooks/limit-orders";
 
 export interface TradeToolProps {}
 
-export const TradeTool: FunctionComponent<TradeToolProps> = observer(() => {
-  const [tab, setTab] = useState<SwapToolTab>(SwapToolTab.BUY);
+export const TradeTool: FunctionComponent<TradeToolProps> = () => {
+  const [tab, setTab] = useQueryState("tab", parseAsInteger.withDefault(0));
```
crnbarr93, line 16: nit: Do you think we should swap this to use a string instead of an integer?

i.e. `?tab=swap` instead of `?tab=0`

crnbarr93: We can adjust the enum as necessary

crnbarr93 (approved):

packages/web/components/trade-tool/index.tsx
fabryscript: Of course! I will do that asap

PR #3310, opened 2024-06-08
(Deposit/Withdraw) Support External Url Providers

JoseRFelix:

Providers often have their own user interfaces (UIs) that support their unique IDs, such as Satellite for Axelar. We could query each provider using the counterparty chain and token to obtain a URL that preselects those assets, along with Osmosis as the source or destination.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-471/support-external-url-providers)

## Testing and Verifying

- [ ] Tests are passing


jonator (commented):
Should we add support for tfm URLs in IBC provider? I recall they have deep IBC support, for CW20s as well since we're removing support for that.

JoseRFelix:
> Should we add support for tfm URLs in IBC provider? I recall they have deep IBC support, for CW20s as well since we're removing support for that.

@jonator Added an IBC external url function for TFM on my latest commit

jonator (approved):
Approved. But what are your thoughts on adding this as a method on the bridge provider interface?

JoseRFelix:
> Approved. But what are your thoughts on adding this as a method on the bridge provider interface?

I considered it, but I'm hesitant to include it to avoid having to manage the bundle size

PR #3311, opened 2024-06-10
feat: Review place limit order modal

crnbarr93:
These changes add another step to the process of placing a limit order. Upon attempting to place a limit order the user must go through a review modal summarising the order they are about to place.

### Linear Task

[LIM-103: Review Limit Order](https://linear.app/osmosis/issue/LIM-103/review-limit-order)

## Brief Changelog
- Added `baseAssetPrice`, `quoteAssetPrice` and `paymentFiatValue` to `PlaceLimitState`
- Added a new modal, `ReviewLimitOrderModal`
- Updated review order button disabled state


fabryscript (approved):
Small tips:
- Use `justify-between` instead of `place-content-between`
- Be careful about HTML semantics, you can use `div`s with text styles, but it's better to use a `h`-tag (or `p`) as a wrapper with layout properties while the `span`s inside of it inherit the text styles from the parent tag

PR #3316, opened 2024-06-11
feat: normalised price values against USD for limit orders

crnbarr93:
Previous calculations when placing a limit order assumed the quote asset to be a stablecoin. These changes are to adjust for the current quote asset's fiat price for the case where the quote asset is not a stablecoin.

### Linear Task
[LIM-109: Adjust Limited Order to Account for Quote Asset Price](https://linear.app/osmosis/issue/LIM-109/adjust-limit-order-placing-to-account-for-quote-asset-price)

## Brief Changelog
- Adjusted `usePlaceLimit` calculations to account for `quoteAssetPrice` (the current fiat value of the quote denom)

### DIFF
--- a/packages/web/hooks/limit-orders/use-place-limit.ts
+++ b/packages/web/hooks/limit-orders/use-place-limit.ts
@@ -108,8 +108,8 @@ export const usePlaceLimit = ({
 
   /**
    * Determines the fiat amount the user will pay for their order.
-   * In the case of an Ask the fiat amount is the amount of tokens the user will sell times the currently selected price.
-   * In the case of a Bid the fiat amount is the amount of tokens the user will buy times the current price of the quote asset.
+   * In the case of an Ask the fiat amount is the amount of tokens the user will sell multiplied by the currently selected price.
+   * In the case of a Bid the fiat amount is the amount of quote asset tokens the user will send multiplied by the current price of the quote asset.
    */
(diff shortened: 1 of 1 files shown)


packages/web/hooks/limit-orders/use-place-limit.ts
```diff
@@ -58,59 +58,87 @@ export const usePlaceLimit = ({
...
+   * @returns The amount of tokens to be sent with the order in base asset amounts for an Ask and quote asset amounts for a Bid.
+   */
+  const paymentTokenValue = useMemo(() => {
+    // The amount of tokens the user wishes to buy/sell
+    const baseTokenAmount =
+      inAmountInput.amount ?? new CoinPretty(baseAsset, new Dec(0));
+    if (orderDirection === OrderDirection.Ask) {
+      // In the case of an Ask we just return the amount requested to sell
+      return baseTokenAmount;
+    } else {
```
fabryscript, line 83: note for readability: is this else block necessary if you are guard-clausing with the if -> return in the condition before?

fabryscript (approved):

packages/web/hooks/limit-orders/use-place-limit.ts
crnbarr93: The else block isn't necessary, was just how I wrote it as I thought through the logic. Will remove it.

PR #3321, opened 2024-06-12
feat: added maker fee query for orderbooks

crnbarr93:
These changes add a tRPC router for orderbooks that exposes a single query (for now). The exposed query is to retrieve the maker fee for the orderbook.

### Linear Task

[LIM-120: Maker Fee Smart Contract Query](https://linear.app/osmosis/issue/LIM-120/[services]-maker-fee-smart-contract-query)

## Brief Changelog
- Added `orderbookRouter` to the server package
- Added `orderbookRouter` to the `localRouter` for tRPC
- Added `makerFee` and `useMakerFee` to `useOrderbookPool`
- Updated review modal to use maker fee from the selected orderbook


packages/server/src/trpc-routers/orderbook-router.ts
jonator, line 1: It seems your branch is quite out of date from stage. We created a `@osmosis-labs/trpc` package where the routers now live.

packages/web/hooks/limit-orders/use-orderbook-pool.ts
```diff
@@ -1,15 +1,43 @@
...
+      contractOsmoAddress: orderbookAddress,
+    });
+
+  const makerFee = useMemo(() => {
+    if (isLoading) return new Dec(0);
+    return new Dec(makerFeeData?.makerFee ?? 0);
+  }, [isLoading, makerFeeData]);
+  return {
+    makerFee,
+    isLoading,
```
jonator, line 41: Why not do this dec conversion in the tRPC router? It would likely be useful to other users of that function in the future. tRPC is set up with superjson so it can send Dec, PricePretty, CoinPretty, Int, etc. over the wire.

packages/server/src/queries/complex/parameter-types.ts
```diff
@@ -7,3 +7,7 @@ export type UserOsmoAddress = z.infer<typeof UserOsmoAddressSchema>;
 export const UserOsmoAddressSchema = z.object({
   userOsmoAddress: z.string().startsWith("osmo").optional(),
 });
+export type ContractOsmoAddress = z.infer<typeof ContractOsmoAddressSchema>;
+export const ContractOsmoAddressSchema = z.object({
+  contractOsmoAddress: z.string().startsWith("osmo").optional(),
+});
```
jonator, line 13: These are the same. It's a nit, but perhaps we can rename and generalize `UserOsmoAddressSchema` > `OsmoAddressSchema` into a single schema.

packages/server/src/trpc-routers/orderbook-router.ts
```diff
@@ -0,0 +1,18 @@
...
+    .query(async ({ input, ctx }) => {
+      const { contractOsmoAddress } = input;
+      const { data } = await queryOrderbookMakerFee({
+        orderbookAddress: contractOsmoAddress,
+        chainList: ctx.chainList,
+      });
+      return {
+        makerFee: data,
+      };
+    }),
```
jonator, line 17: To benefit from caching, our convention currently is to create a function in the complex folder. Then, wrap the function you want to cache in cachified.

In this case, I'd create a "orderbooks" folder under complex directory (with an index.ts), put this in "maker-fee.ts" file. And there I'd create an instance of an in memory cache and use that to wrap this in cachified. Choose a ttl that makes sense to you. This is a low overhead way to prevent some needless queries to the chain and improve user UX.

jonator (changes requested):
Looking good!

packages/web/hooks/limit-orders/use-orderbook-pool.ts
crnbarr93: https://github.com/osmosis-labs/osmosis-frontend/pull/3321/commits/371934ab9171d5c79d1bc900cbc9e13056217acf

packages/server/src/queries/complex/parameter-types.ts
crnbarr93: https://github.com/osmosis-labs/osmosis-frontend/pull/3321/commits/0c9ed586bc14c2fbcbeea55426accdc973088b6a

packages/server/src/trpc-routers/orderbook-router.ts
crnbarr93: https://github.com/osmosis-labs/osmosis-frontend/pull/3321/commits/43817c6ba60b9e5d722df68b07c77e0dda5e8755

packages/server/src/trpc-routers/orderbook-router.ts
crnbarr93: Will make sure to make any updates when we're ready to merge to stage

packages/web/hooks/limit-orders/use-orderbook-pool.ts
```diff
@@ -1,15 +1,43 @@
...
+
 export const useOrderbookPool = ({
   baseDenom,
   quoteDenom,
 }: {
   baseDenom: string;
   quoteDenom: string;
 }) => {
+  const [contractAddress] = useState<string>(
+    "osmo1svmdh0ega4jg44xc3gg36tkjpzrzlrgajv6v6c2wf0ul8m3gjajs0dps9w"
```
jonator, line 14: This will eventually come from some config or registry value?

jonator (approved):

packages/web/hooks/limit-orders/use-orderbook-pool.ts
crnbarr93: I've started moving this already on [this](https://github.com/osmosis-labs/osmosis-frontend/tree/connor/use-orderbook-hook) branch

PR #3330, opened 2024-06-13
feat: alloyed assets composition section

DavideSegullo:

This PR adds a section detailing the composition of an alloyed asset, specifically on the token detail page.

An example can be consulted here: `/assets/allUSDT`

What I need to happen on the user side is for the data loading skeleton to appear and for the data to be rendered correctly, each item in the composition must be clickable and take you back to the related detail page

Screenshot 2024-06-14 alle 12 51 11

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-511/alloyed-assets-composition-section)


CryptoAssassin1 (approved):

packages/server/src/queries/complex/pools/transmuter.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      const {
+        data: { total_pool_liquidity = [] },
+      } = await queryTransmuterTotalPoolLiquidity({
+        contractAddress,
+        chainList,
+      });
+
+      const totalLiquidity = total_pool_liquidity.reduce((acc, asset) => {
+        return acc.add(new Dec(parseInt(asset.amount, 10)));
```
jonator, line 44: Is this `parseInt` needed? Dec accepts strings, and using JS numbers could result in precision/MAXINT issues

jonator (approved):

packages/server/src/queries/complex/pools/transmuter.ts
DavideSegullo: Not required, if Dec support strings I'll just remove `parseInt`, good suggestion! Thank you

sunnya97 (approved):

packages/server/src/queries/complex/pools/transmuter.ts
DavideSegullo: done

PR #3333, opened 2024-06-13
Bridge Providers: implement `getSupportedAssets`

jonator:

Implements methods for each bridge providers that return the source assets to be sent to a given to bridge and asset. Swaps are excluded and only assets of the same variant are returned. This will be used to present the user with a list of assets and networks to deposit from.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-451/add-supportedvariantassets-method-to-bridge-providers)

## Brief Changelog

* Implement `getAvailableSourceAssetVariants` for each bridge provider

## Testing and Verifying

* Tests pass. UI will be implemented later.


packages/bridge/src/interface.ts
```diff
@@ -41,6 +41,24 @@ export interface BridgeProvider {
...
+   *
+   * @param toChain The destination chain.
+   * @param toAsset The destination coin.
+   * @returns A promise that resolves to an array of source coins with source identifiers.
+   * @throws Unexpected errors.
+   */
+  getAvailableSourceAssetVariants(
+    toChain: BridgeChain,
+    toAsset: BridgeAsset
+  ): Promise<(BridgeChain & BridgeAsset)[]>;
```
JoseRFelix, line 60: Nit: In line with our other methods, I think we should make the parameters a schema or an object. This will make it easier to understand and manage the parameters

packages/bridge/src/skip/__tests__/skip-bridge-provider.spec.ts
```diff
@@ -431,4 +540,109 @@ describe("SkipBridgeProvider", () => {
...
+
+      expect(sourceVariants[0]).toEqual({
+        address: "uusdc",
+        chainId: "noble-1",
+        chainType: "cosmos",
+        decimals: 6,
+        denom: "USDC",
+        sourceDenom: "uusdc",
+      });
+    });
```
JoseRFelix, line 621: Wasn't this tested in the previous test? If we want to emphasize it, how about adding a comment or an assertion there?

packages/bridge/src/skip/__tests__/skip-bridge-provider.spec.ts
```diff
@@ -431,4 +540,109 @@ describe("SkipBridgeProvider", () => {
...
+
+      expect(sourceVariants[1]).toEqual({
+        address: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
+        chainId: 1,
+        chainType: "evm",
+        decimals: 6,
+        denom: "USDC",
+        sourceDenom: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
+      });
+    });
```
JoseRFelix, line 646: Same as above

JoseRFelix (changes requested):

packages/bridge/src/axelar/types.ts
```diff
@@ -32,10 +32,12 @@ export const AxelarChainIds_SourceChainMap: (env: BridgeEnvironment) => {
...
 
+/** Any bridge that we want supported by Axelar provider needs to be added here. */
 const TestnetCosmosChainIds_AxelarChainIds: Partial<Record<string, string>> = {
   "osmo-test-5": "osmosis-7",
 };
 
+/** Any bridge that we want supported by Axelar provider needs to be added here. */
 const MainnetCosmosChainIds_AxelarChainIds: Partial<Record<string, string>> = {
   "osmosis-1": "osmosis",
 };
```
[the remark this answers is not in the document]
jonator: Tracking
https://linear.app/osmosis/issue/FE-560/axelar-provider-use-endpoint-instead-of-hardcoded-chain-id-map

packages/bridge/src/interface.ts
jonator: IMO it's more succinct to have a set of composable types that are single purpose that can be combined like this when necessary. In this case, it returns a bridge chain and asset together (the combo)

jonator: In general I think the codebase is simpler when there's overall fewer symbols/functions/defs vs when there's more. If we can avoid a def, then it's easier to grok/navigate

JoseRFelix: Maybe we can inline then these parameters if we want to avoid creating a schema. 

```suggestion
  getAvailableSourceAssetVariants(params: {
    toChain: BridgeChain,
    toAsset: BridgeAsset
  }): Promise<(BridgeChain & BridgeAsset)[]>;
```

JoseRFelix (approved):
LGTM!

PR #3343, opened 2024-06-15
feat: Current orderbook spot price hook

crnbarr93:
This change adds spot price data from the orderbook smart contract. The price is determined by the provided smart contract address and the token in/out pairing (this will swap depending on the order direction).

**Spot price is cached for 2 minutes**

### Linear Task

[ LIM-123: Current Orderbook Spot Price](https://linear.app/osmosis/issue/LIM-123/[services]-current-orderbook-spot-price)

## Brief Changelog
- Added `useOrderbookSpotPrice` hook that returns spot price data from the provided orderbook smart contract
- Added loading state for spot price loading condition

### DIFF
--- a/packages/server/src/queries/complex/orderbooks/spot-price.ts
+++ b/packages/server/src/queries/complex/orderbooks/spot-price.ts
@@ -1,0 +1,35 @@
+import { Chain } from "@osmosis-labs/types";
+import cachified, { CacheEntry } from "cachified";
+import { LRUCache } from "lru-cache";
+
+import { DEFAULT_LRU_OPTIONS } from "../../../utils/cache";
+import { queryOrderbookSpotPrice } from "../../osmosis";
+
+const orderbookSpotPriceCache = new LRUCache<string, CacheEntry>(
+  DEFAULT_LRU_OPTIONS
+);
+
+export function getOrderbookSpotPrice({
+  orderbookAddress,
+  chainList,
+  quoteAssetDenom,
+  baseAssetDenom,
+}: {
+  orderbookAddress: string;
+  quoteAssetDenom: string;
+  baseAssetDenom: string;
+  chainList: Chain[];
+}) {
+  return cachified({
+    cache: orderbookSpotPriceCache,
+    key: `orderbookSpotPrice-${orderbookAddress}-${quoteAssetDenom}-${baseAssetDenom}`,
+    ttl: 1000 * 60 * 2, // 2 minutes
+    getFreshValue: () =>
+      queryOrderbookSpotPrice({
+        orderbookAddress,
+        chainList,
+        quoteAssetDenom,
+        baseAssetDenom,
+      }).then(({ data }) => data.spot_price),
+  });
+}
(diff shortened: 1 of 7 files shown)


packages/web/hooks/limit-orders/use-place-limit.ts
```diff
@@ -50,7 +51,15 @@ export const usePlaceLimit = ({
   const quoteAsset = swapAssets.toAsset;
   const baseAsset = swapAssets.fromAsset;
 
-  const priceState = useLimitPrice();
+  const tokenIn =
+    orderDirection === OrderDirection.Bid ? quoteAsset : baseAsset;
+  const tokenOut =
+    orderDirection === OrderDirection.Bid ? baseAsset : quoteAsset;
```
fabryscript, line 57: maybe this is `orderDirection === OrderDirection.Ask`?

fabryscript (approved):

packages/server/src/queries/complex/orderbooks/spot-price.ts
```diff
@@ -0,0 +1,35 @@
...
+    cache: orderbookSpotPriceCache,
+    key: `orderbookSpotPrice-${orderbookAddress}-${quoteAssetDenom}-${baseAssetDenom}`,
+    ttl: 1000 * 60 * 2, // 2 minutes
+    getFreshValue: () =>
+      queryOrderbookSpotPrice({
+        orderbookAddress,
+        chainList,
+        quoteAssetDenom,
+        baseAssetDenom,
+      }).then(({ data }) => data.spot_price),
```
jonator, line 33: NIT: Since this is in the complex folder, where we can make modifications / do processing on raw queries, what are your thoughts on adding the `Dec` return type here instead of in the tRPC procedure

packages/web/components/place-limit-tool/index.tsx
```diff
@@ -51,7 +51,11 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
       }
```
jonator, line 51: I can't comment on the useState hooks since they're not modified in this PR, but what if we initialize them as null then set them in the effects when the query loads? I'm assuming they will always need to be initialized from the result of the query. If not, please ignore this suggestion

Also, IMO it's safer to use chain denoms (coinMinimalDenoms) instead of token symbols as they're more reliable asset identifiers.

packages/web/hooks/limit-orders/use-place-limit.ts
jonator: Personally, with typescript I prefer string literals over enums (in TS I see enums as quite antiquated)

Example:
`type OrderDirection = "bid" | "ask";`

crnbarr93: Can change this if it's preferred. Gotten used to working in Rust and I use enums quite a lot so probably a bad habit to carry over to TS.

packages/server/src/queries/complex/orderbooks/spot-price.ts
crnbarr93: Sure

packages/web/components/place-limit-tool/index.tsx
crnbarr93: > Also, IMO it's safer to use chain denoms (coinMinimalDenoms) instead of token symbols as they're more reliable asset identifiers.

I've been slowly transitioning this on other branches

crnbarr93:
https://github.com/osmosis-labs/osmosis-frontend/pull/3352

Moved this PR to here to include usage of the spot price

PR #3366, opened 2024-06-19
feat: pagination for limit order queries

crnbarr93:
These changes alter the active order queries to include pagination.

## Brief Changelog
- Altered `getAllActiveOrders` and `getActiveOrders` to include pagination

### DIFF
--- a/packages/trpc/src/orderbook-router.ts
+++ b/packages/trpc/src/orderbook-router.ts
@@ -165,23 +165,26 @@ export const orderbookRouter = createTRPCRouter({
           const { contractAddresses, userOsmoAddress } = input;
           if (contractAddresses.length === 0 || userOsmoAddress.length === 0)
             return [];
-          let allOrders: MappedLimitOrder[] = [];
-          for (let i = 0; i < contractAddresses.length; i++) {
-            const contractOsmoAddress = contractAddresses[i];
-            const resp = await getOrderbookActiveOrders({
-              orderbookAddress: contractOsmoAddress,
-              userOsmoAddress: userOsmoAddress,
-              chainList: ctx.chainList,
-            });
+          const promises = contractAddresses.map(
+            async (contractOsmoAddress) => {
+              const resp = await getOrderbookActiveOrders({
+                orderbookAddress: contractOsmoAddress,
+                userOsmoAddress: userOsmoAddress,
+                chainList: ctx.chainList,
+              });
 
-            if (resp.orders.length === 0) continue;
-            const mappedOrders = await getTickInfoAndTransformOrders(
-              contractOsmoAddress,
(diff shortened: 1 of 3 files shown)


packages/web/components/place-limit-tool/index.tsx
```diff
@@ -76,6 +79,11 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
 
     const account = accountStore.getWallet(accountStore.osmosisChainId);
 
+    const { orders } = useOrderbookAllActiveOrders({
+      userAddress: account?.address ?? "",
+    });
+    console.log(orders);
```
MaxMillington, line 85: remove log

packages/trpc/src/orderbook-router.ts
```diff
@@ -114,57 +122,75 @@ export const orderbookRouter = createTRPCRouter({
...
-          userOsmoAddress: userOsmoAddress,
-          chainList: ctx.chainList,
-        });
+      return maybeCachePaginatedItems({
+        getFreshItems: async () => {
+          const { contractAddresses, userOsmoAddress } = input;
+          if (contractAddresses.length === 0 || userOsmoAddress.length === 0)
+            return [];
+          let allOrders: MappedLimitOrder[] = [];
+          for (let i = 0; i < contractAddresses.length; i++) {
```
MaxMillington, line 169: `forEach` might be a little more idiomatic here

MaxMillington (approved):

packages/trpc/src/orderbook-router.ts
crnbarr93: I'm not sure `forEach` supports async operations does it? I could map it to an array of promises and run each loop asynchronously but not sure how that works with caching.

packages/web/components/place-limit-tool/index.tsx
crnbarr93: https://github.com/osmosis-labs/osmosis-frontend/pull/3366/commits/6e089c02c9c8aef11b15a1bb818dc857776daf9a

packages/trpc/src/orderbook-router.ts
crnbarr93: https://github.com/osmosis-labs/osmosis-frontend/pull/3366/commits/e592dbbc08c38284b2545f103ad69791da36c753

Because the loop is using individual contracts per loop I can't see a caching conflict so I altered it to use promises instead (and `.map`)

MaxMillington: You are right that `forEach` will not wait for each block in the case of an async function. I like your map solution better 👍

fabryscript (approved):

PR #3367, opened 2024-06-19
refactor: replace OrderDirection enum with string literal

crnbarr93:
Replaced `OrderDirection` enum with a `type` that can be either `"bid"` or `"ask"`

## Brief Changelog
- `OrderDirection` is now a `type` instead of an `enum`


fabryscript (approved):

jonator (approved):
Nice

PR #3375, opened 2024-06-19
fix: :bug: transmuter pool liquidity percentage calc

DavideSegullo:

Right now there is a bug on production, the percentage we show in the alloyed asset is miscalculated because it does not take into account that each asset has different decimals.

### Linear Task

### DIFF
--- a/packages/server/src/queries/complex/pools/transmuter.ts
+++ b/packages/server/src/queries/complex/pools/transmuter.ts
@@ -1,4 +1,4 @@
-import { CoinPretty, Dec, RatePretty } from "@keplr-wallet/unit";
+import { CoinPretty, Dec, DecUtils, RatePretty } from "@keplr-wallet/unit";
 import { AssetList, Chain } from "@osmosis-labs/types";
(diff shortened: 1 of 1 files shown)


packages/server/src/queries/complex/pools/transmuter.ts
```diff
@@ -40,8 +40,21 @@ export async function getCachedTransmuterTotalPoolLiquidity(
...
+      const totalLiquidity = total_pool_liquidity.reduce((acc, coin) => {
+        const asset = captureIfError(() =>
+          getAsset({ assetLists, anyDenom: coin.denom })
+        );
+
+        if (asset) {
+          const coinPretty = new CoinPretty(
+            asset,
+            coin.amount
+          ).moveDecimalPointLeft(asset.coinDecimals);
```
jonator, line 52: Idea: instead of using CoinPretty you could simply multiply the amount with `DecUtils.getExponentN(coinDecimals)`

DavideSegullo: Done!

PR #3376, opened 2024-06-19
(Deposit/Withdrawal) Canonical Asset Selection

JoseRFelix:

image

image

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-453/ui-canonical-asset-selection)


packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,272 @@
...
+import { BridgeFlowProvider } from "~/hooks/bridge";
+import { useAmplitudeAnalytics } from "~/hooks/use-amplitude-analytics";
+import { useDisclosure } from "~/hooks/use-disclosure";
+import { FiatRampKey } from "~/integrations";
+import { ModalCloseButton } from "~/modals";
+import { FiatOnrampSelectionModal } from "~/modals/fiat-on-ramp-selection";
+import { FiatRampsModal } from "~/modals/fiat-ramps";
+import { api } from "~/utils/trpc";
+
+enum ImmersiveBridgeScreens {
```
jonator, line 23: Since screen prop is a string key anyways, wouldn't it be slightly less verbose to just use string literal type here?

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+export const ImmersiveBridgeFlow = ({
+  Provider,
+  children,
+}: PropsWithChildren<BridgeFlowProvider>) => {
+  const [isVisible, setIsVisible] = useState(false);
+  const [step, setStep] = useState<ImmersiveBridgeScreens>(
+    ImmersiveBridgeScreens.Asset
+  );
+  const [type, setType] = useState<"deposit" | "withdraw">("deposit");
```
jonator, line 41: IMO "type" is too generic of a term to describe something more accurately, like "transfer direction" or just "direction"

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const { logEvent } = useAmplitudeAnalytics();
+
+  const apiUtils = api.useUtils();
+
+  const [assetInOsmosis, setAssetInOsmosis] = useState<MinimalAsset>();
```
jonator, line 46: What's the plan for handling multi variant assets? Will this be handled within the amount screen?

jonator (changes requested):
It would be nice to keep the input centered, shift the token denom "USDC" left, and make the input text smaller as more characters are input (Apple style).
Screenshot 2024-06-24 at 7 45 26 AM

packages/web/components/bridge/immersive/asset-select-screen.tsx
```diff
@@ -0,0 +1,226 @@
...
+          className="my-4 flex-shrink-0"
+          placeholder={t("transfer.assetSelectScreen.searchAssets")}
+          size="full"
+        />
+
+        <div className="flex flex-col gap-1">
+          {isLoading ? (
+            <>
+              {new Array(7).fill(undefined).map((_, i) => (
+                <SkeletonLoader key={i} className="h-[80px] w-full" />
```
jonator, line 140: Is this what was used in the design? IMO the feel of the skeleton loader feels a bit jarring

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,272 @@
...
+            }
+          };
+
+          const asset = await fetchAssetWithRetry();
+
+          if (!asset) {
+            return;
+          }
+
+          setAssetInOsmosis(asset);
```
jonator, line 119: Is this async stuff with the retry logic and catch statements necessary? Would it be possible to set a state var with the denom, then pass that as input to a query hook? Then, as long as it returns a valid asset from the query `isVisible` would return true

packages/web/components/bridge/immersive/bridge-network-select.tsx
```diff
@@ -4,54 +4,53 @@ import React, { useMemo, useState } from "react";
...
   const canLoadMore = !isLoading && !isFetchingNextPage && hasNextPage;
 
   return (
     <ModalBase
-      title="Select network"
+      title={t("transfer.bridgeNetworkSelect.title")}
       className="!max-w-[30rem]"
       {...modalProps}
     >
+      {/* TODO: Add translation */}
```
jonator, line 47: This can be cleaned up

packages/web/components/complex/assets-page-v1.tsx
```diff
@@ -60,13 +60,13 @@ export const AssetsPageV1: FunctionComponent = observer(() => {
       {
         label: t("assets.table.depositButton"),
         onClick: () => {
-          startBridge("deposit");
+          startBridge({ direction: "deposit" });
```
jonator, line 63: I'm not opposed to objects as params, but is change actually necessary to make? I find the params being used are pretty straightforward and obvious, and there's no use of any optional/default params.

packages/web/components/bridge/immersive/immersive-bridge.tsx
jonator: We may want to pass the Osmosis asset variants as an array here since we do some custom filtering in the `getImmersiveBridgeAssets` tRPC procedure in the component before this.

I'm thinking then, in this component (or in a hook here), we do the logic of aggregating the bridge provider data returned for each Osmosis variant (selectable chains + assets)

JoseRFelix:
>It would be nice to keep the input centered, shift the token denom "USDC" left, and make the input text smaller as more characters are input (Apple style).

Yeah, I talked with Nass about this. He created a [Code Sandbox](https://codepen.io/nassdonald/pen/RwmMMzK) demonstrating this behavior 🙌 . I created an [issue](https://linear.app/osmosis/issue/FE-581/scale-crypto-amount-input-according-to-content) to track it. I'll open a new PR because I don't want to get distracted from my connection work

packages/types/src/asset-types.ts
```diff
@@ -179,3 +179,12 @@ export interface Asset {
   /** Denom key of variant of asset this is grouped with. */
   variantGroupKey?: string;
 }
+
+export type MinimalAsset = Currency & {
```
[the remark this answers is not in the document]
JoseRFelix: Right, I meant to use the `KeplrBaseCurrency`

packages/web/components/bridge/immersive/immersive-bridge.tsx
JoseRFelix: While it's true, achieving type safety would otherwise be difficult and verbose. We'd have to supply the union type to every Screen component. I think an enum can solve this problem elegantly. What do you think?

packages/web/components/bridge/immersive/immersive-bridge.tsx
JoseRFelix: I agree. I'll update this

packages/web/components/bridge/immersive/immersive-bridge.tsx
JoseRFelix: The thing about variants is that they are not necessarily assets in Osmosis. I don't want to complicate the state flow or cause unnecessary re-renders, as these variants come from providers. However, I agree that we should aggregate this data somewhere. Creating a hook makes sense, and that's my current plan. I'll most likely keep them in a separate state to make it easier to manage

packages/web/components/bridge/immersive/asset-select-screen.tsx
JoseRFelix: Yes, this can be improved. My plan is to finish the flow and then do a design check with the team for some final refinements

packages/types/src/asset-types.ts
```diff
@@ -179,3 +179,12 @@ export interface Asset {
...
+  coinGeckoId: string | undefined;
+  coinName: string;
+  isUnstable: boolean;
+  isVerified: boolean;
+  sourceDenom: string;
```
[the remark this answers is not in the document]
JoseRFelix: I'll update this as soon as your PR is merged

packages/web/components/complex/assets-page-v1.tsx
JoseRFelix: What I like about using object parameters is that they are explicit and much easier to extend. If we ever need to add another property to the function, it's very straightforward to update it throughout the code

packages/web/components/bridge/immersive/immersive-bridge.tsx
jonator: By variants I mean the variants on Osmosis. When the user clicks USDC, we're going to pass Noble USDC to each bridge provider `getSupportedAssets`, but we're also going to be passing USDC.axl (and any other variants on Osmosis) to `getSupportedAssets` separately. Then, at the end we try to swap whatever variant the user owns into our canonical version (Noble USDC, an alloy, etc)

JoseRFelix: You're right! I'll fix this

packages/server/src/queries/complex/assets/index.ts
```diff
@@ -48,6 +39,38 @@ export function getAsset({
...
+}: {
+  assetLists: AssetList[];
+  anyDenom: string;
+}) {
+  const asset = getAsset({
+    assetLists,
+    anyDenom,
+  });
+
+  if (!asset) throw new Error(anyDenom + " not found in asset list");
```
jonator, line 60: This should not be necessary. `getAsset` will throw on it's own

packages/types/src/asset-types.ts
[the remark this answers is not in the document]
jonator: Don't forget to clean this up

packages/types/src/bridge-types.ts
```diff
@@ -0,0 +1 @@
+export type BridgeTransactionDirection = "deposit" | "withdraw";
```
jonator, line 1: Can we possibly remove this type?

packages/types/src/asset-types.ts
```diff
@@ -179,3 +179,12 @@ export interface Asset {
   /** Denom key of variant of asset this is grouped with. */
   variantGroupKey?: string;
 }
+
+export type MinimalAsset = KeplrBaseCurrency & {
```
jonator, line 183: If I were to implement this, given our divorce from Keplr, I'd just inline the members here as we had and avoid importing their types

packages/web/components/bridge/immersive/immersive-bridge.tsx
jonator: Since you're using string indexes, thoughts on using a const enum instead? Then, TS will remove it at bulid time and use indexes under the hood

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,257 @@
...
+export const ImmersiveBridgeFlow = ({
+  Provider,
+  children,
+}: PropsWithChildren<BridgeFlowProvider>) => {
+  const [isVisible, setIsVisible] = useState(false);
+  const [step, setStep] = useState<ImmersiveBridgeScreens>(
+    ImmersiveBridgeScreens.Asset
+  );
+  const [direction, setDirection] =
+    useState<BridgeTransactionDirection>("deposit");
```
jonator, line 42: From other comment, IMO it's slightly more readable and less redirected if we just directly use the `"deposit" | "withdraw"` type literal

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+  const {
+    isOpen: isFiatOnrampSelectionOpen,
+    onOpen: onOpenFiatOnrampSelection,
+    onClose: onCloseFiatOnrampSelection,
+  } = useDisclosure();
+
+  // const { isConnected, address } = useEvmWalletAccount();
+  // const { onOpenWalletSelect } = useWalletSelect();
+  // const { disconnect } = useDisconnectEvmWallet();
```
jonator, line 72: Let's clean these up

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,257 @@
...
+                aria-label="Go Back"
+              />
+            )}
+
+            <div className="flex h-full w-full max-w-[30rem] flex-col gap-10 py-12">
+              <StepProgress
+                className="w-full"
+                steps={[
+                  {
+                    displayLabel: "Asset",
```
jonator, line 162: Do these need to be localized?

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,257 @@
...
+                    <AmountScreen
+                      type={direction}
+                      assetsInOsmosis={canonicalAssetsWithVariants}
+                    />
+                  )}
+                </Screen>
+                <Screen screenName={ImmersiveBridgeScreens.Review}>
+                  {({ goBack }) => (
+                    <div>
+                      <h6>Step 3: Review</h6>
```
jonator, line 205: Also, localized

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,257 @@
...
+                      type={direction}
+                      onSelectAsset={(asset) => {
+                        setCurrentScreen(ImmersiveBridgeScreens.Amount);
+                        setSelectedAssetDenom(asset.coinDenom);
+                      }}
+                    />
+                  )}
+                </Screen>
+                <Screen screenName={ImmersiveBridgeScreens.Amount}>
+                  {() => (
```
jonator, line 195: Is it necessary to wrap these screens in a callback?

packages/web/components/bridge/immersive/immersive-bridge.tsx
```diff
@@ -0,0 +1,257 @@
...
+                        walletType: "evm",
+                      },
+                    ],
+                    layout: "list",
+                  })
+                }
+              >
+                Connect EVM Wallet
+              </Button>
+            )} */}
```
jonator, line 232: Let's clean this up

packages/web/components/bridge/immersive/bridge-network-select.tsx
```diff
@@ -4,45 +4,43 @@ import React, { useMemo, useState } from "react";
 import { SearchBox } from "~/components/input";
 import { Intersection } from "~/components/intersection";
 import { SkeletonLoader, Spinner } from "~/components/loaders";
+import { useTranslation } from "~/hooks/language";
 import { ModalBase, ModalBaseProps } from "~/modals";
 import { api } from "~/utils/trpc";
 
 export const BridgeNetworkSelect = (modalProps: ModalBaseProps) => {
```
jonator, line 11: Think we could append `Modal` to this symbol so it's more clear it's a modal

packages/web/components/bridge/immersive/amount-screen.tsx
```diff
@@ -0,0 +1,472 @@
...
+import { trimPlaceholderZeros } from "~/utils/number";
+import { api } from "~/utils/trpc";
+
+interface AmountScreenProps {
+  type: "deposit" | "withdraw";
+
+  /**
+   * Includes both the canonical asset and its variants.
+   */
+  assetsInOsmosis: MinimalAsset[] | undefined;
```
jonator, line 30: Nit: what if we call this `assetWithVariants`?

packages/web/components/bridge/immersive/asset-select-screen.tsx
```diff
@@ -0,0 +1,226 @@
...
+} from "~/config/generated/asset-lists";
+import { useTranslation } from "~/hooks";
+import { useShowPreviewAssets } from "~/hooks/use-show-preview-assets";
+import { ActivateUnverifiedTokenConfirmation } from "~/modals/activate-unverified-token-confirmation";
+import { useStore } from "~/stores";
+import { UnverifiedAssetsState } from "~/stores/user-settings/unverified-assets";
+import { formatPretty } from "~/utils/formatter";
+import { api, RouterOutputs } from "~/utils/trpc";
+
+const variantsNotToBeExcluded = ["WBTC", "BTC"] satisfies (
```
jonator, line 27: Getting a build error here. I think asset list switched to min denoms

jonator (approved):
Approved, with some more nits. Also, we should make this branch compatible with new asset list now

packages/web/components/bridge/immersive/immersive-bridge.tsx
JoseRFelix: This is a temp value so I think we can skip it

packages/web/components/bridge/immersive/amount-screen.tsx
JoseRFelix: Hmm, my initial thought with this name is that it suggests an asset object with nested variants. Since it's an array, perhaps it could be called `canonicalAndVariantAssets`

packages/types/src/asset-types.ts
```diff
@@ -179,3 +179,20 @@ export interface Asset {
...
```
jonator, line 183: Let's not forget to remove `KeplrBaseCurrency` here

packages/types/src/asset-types.ts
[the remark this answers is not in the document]
jonator: What PR?

jonator (changes requested):

packages/types/src/asset-types.ts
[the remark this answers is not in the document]
JoseRFelix: #3389

jonator (approved):

PR #3383, opened 2024-06-20
[Order History]: Add base reference headers

fabryscript:

Add reference table headers
Screenshot 2024-06-20 alle 13 49 50

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/LIM-132/[ui]-add-base-reference-headers)


MaxMillington (approved):
lgtm

PR #3392, opened 2024-06-21
feat: aprs should be shown as a range

DavideSegullo:
### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-527/aprs-should-be-shown-as-a-range)


jonator (changes requested):
I'm no design expert, but IMO we should show 0 decimals to reduce the required width
Screenshot 2024-06-24 at 6 43 03 PM

sunnya97:
These all seem flipped –

should be `lower bound - upper bound`

CleanShot 2024-06-26 at 11 51 41@2x

sunnya97:
Also, why do some pools not have both?

CleanShot 2024-06-26 at 11 52 28@2x

sunnya97:
Also, why do some pools not have both?

CleanShot 2024-06-26 at 11 53 02@2x

DavideSegullo:
> These all seem flipped. Should be `lower bound - upper bound`
> 
> Also, why do some pools not have both?
> 
> CleanShot 2024-06-26 at 11 53 02@2x

It's already like that, maybe they're swapped on backend side. Lower bound is higher then upper bound.

if we do not show a range it is because lower and upper bound are equal

CryptoAssassin1 (approved):

jonator (approved):

PR #3403, opened 2024-06-25
feat: asset info page redo stats section

DavideSegullo:

This PR add the new stats section inside the token info pages and other minor updates:
1) New twitter section style
2) New about section style

### Linear Task

[Redo stats section](https://linear.app/osmosis/issue/FE-518/asset-info-page-redo-stats-section)
[Add osmosis Dex volume](https://linear.app/osmosis/issue/FE-502/asset-info-pages-add-osmosis-dex-volume-as-a-basic-stat)


jonator:
Is it possible to install tradingview bundles from yarn instead of checking out in repo?

DavideSegullo:
> Is it possible to install tradingview bundles from yarn instead of checking out in repo?

They didn't publish a library on npm, we can still use npm using github url, but the repo isn't public... So if we wanna do that for vercel I don't know if it's possible to add a github token for auth

sunnya97:
I'm getting "24 hr volume" as blank for every asset

CleanShot 2024-06-25 at 11 02 53@2x

DavideSegullo:
> I'm getting "24 hr volume" as blank for every asset
> 
> CleanShot 2024-06-25 at 11 02 53@2x

Oh right, I pinged Niccolo, he'll update the price endpoint asap to support this new endpoint

kamal-sutra (approved):

sunnya97:
LOVE the embedded pools table!

Could we have it sorted by volume, not liquidity?

DavideSegullo:
> LOVE the embedded pools table!
> 
> Could we have it sorted by volume, not liquidity?

Done, we also have 24h volume

sunnya97:
Are we waiting on endpoints for adding:
- Supply on Osmosis
- Liquidity on Osmosis

packages/trpc/src/pools.ts
```diff
@@ -146,14 +146,11 @@ export const poolsRouter = createTRPCRouter({
               })
               .filter((pool): pool is NonNullable<typeof pool> => !!pool);
 
-            // won't sort if searching
-            if (search) return marketIncentivePools;
```
jonator, line 150: This is now enforced on the client? I think it would be good for sorting to be possible while searching, but at least for the pools table in the pools page, having a rank-order fuzzy search result + sorting can yield what appear to be inaccurate results, especially for lower cap tokens (that have a strong search match but get sorted to the bottom)

packages/web/components/pages/asset-info-page/token-details.tsx
```diff
@@ -177,59 +173,175 @@ const _TokenDetails = ({ className }: TokenDetailsProps) => {
...
-  const { coingeckoCoin, isLoadingCoingeckoCoin } = useAssetInfo();
+  const {
+    coinGeckoId,
+    coingeckoCoin,
+    isLoadingCoingeckoCoin = true,
+    tokenDenom,
+  } = useAssetInfo();
+
+  const { data: tokenMarket, isLoading: isLoadingTokenMarket } =
+    api.edge.assets.getUserMarketAsset.useQuery(
```
jonator, line 197: Let's create a `getMarketAsset` that doesn't include the support for user address to simplify things a bit. It would be oriented around asset market data

packages/web/components/complex/pools-table.tsx
```diff
@@ -111,11 +111,7 @@ export const PoolsTable = (props: PropsWithChildren<PoolsTableProps>) => {
     children,
   } = props;
 
-  /** Won't sort when searching is happening. */
-  const sortKey = useMemo(
-    () => (Boolean(filters.searchQuery) ? undefined : sortParams.allPoolsSort),
-    [filters.searchQuery, sortParams.allPoolsSort]
-  );
+  const sortKey = sortParams.allPoolsSort;
```
jonator, line 114: Even though the tRPC procedures supports searching and sorting at the same time, I think it could yield confusing results, especially with lower cap tokens. 

Just for this table. I think it's OK to sort + search with the few items on the token page

I was getting complaints about this when I implemented this new pools table

However, I was also getting complaints about not being able to sort + search. But, we agreed that when first searching the sort would be disabled. Then after the search they could sort if they want. That would help make it clearer that searching + sorting is happening at the same time. I'm not saying you should do this, it's just some background on what we haven't gotten to doing yet

jonator (changes requested):
Looking good

packages/web/components/complex/pools-table.tsx
DavideSegullo: Ye makes sense, I'll add a different field for search by denoms because it's a different use case if compared to search + sort

packages/trpc/src/pools.ts
DavideSegullo: YE, I'll roll it back and implement it in a better way so we don't touch the current behaviour

DavideSegullo:
> Are we waiting on endpoints for adding:
> 
> * Supply on Osmosis
> * Liquidity on Osmosis

Actually we have liquidity, I just used a wrong array slice... build in pending

packages/web/components/complex/pools-table.tsx
jonator: Is that necessary? What about just not passing sort params to the tRPC procedure while the user is sorting? Or WDYM by field?

DavideSegullo: I need it because inside the token info page we need pools filtered by denom but also sorted by volume not by liquidity

packages/server/src/queries/complex/pools/index.ts
```diff
@@ -105,6 +107,19 @@ export async function getPools(
...
+    denomPools = search(
+      denomPools,
+      ["coinDenoms"],
+      {
+        query,
+      },
+      0.0 // Exact match
+    );
+  }
+
```
jonator, line 122: Instead of adding `denoms` to `PoolFilterSchema`, didn't @mattupham update the pools search to match denoms explicitly? This may not be necessary is what I was trying to say earlier.

jonator (changes requested):

packages/server/src/queries/complex/pools/index.ts
DavideSegullo: But it doesn't work with an exact match, if you search just for "SOL" you'll also get "SOLAR"

jonator: OK we can keep this then. Should we do a find instead of use `search` then just to be more explicit?

DavideSegullo: Yep, I'll add it so we don't use `search` function

DavideSegullo: Done!

sunnya97 (approved):

PR #3423, opened 2024-06-28
Add alt fee token feature to wallets

jonator:

Some wallets don't support alt fee tokens. This adds an `"alt-fee-tokens"` flag that can be specified in the wallet registry for determining if a wallet accepts a token other than the default gas fee token on chains.

### DIFF
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -1273,6 +1273,10 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
           })
         : undefined;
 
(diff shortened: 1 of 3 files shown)


DavideSegullo (approved):

jonator:
Awaiting confirmation, @JoseRFelix says that wallets should be able to accept alt fee tokens

echel0n-HX:
Any reason why this was not merged? it would resolve the issue with our PR.

MaxMillington:
cc @jonator ^^

PR #3402, opened 2024-06-24
(Deposit/Withdrawal) Fetch Bridge Provider Quote

JoseRFelix:
### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-575/fetch-bridge-providers-quote)

PR #3434, opened 2024-07-02
(Deposit/Withdrawal) Review Screen

jonator:

This implements the UI for the review screen for the deposit/withdrawal project. NOTE: should not be merged until the commits from https://github.com/osmosis-labs/osmosis-frontend/pull/3402 are merged.

### Linear Task


packages/web/components/bridge/immersive/quote-detail.tsx
```diff
@@ -0,0 +1,244 @@
...
+  isRefetchingQuote,
+}) => (
+  <QuoteDetailRow label={t("transfer.provider")} isLoading={isRefetchingQuote}>
+    <BridgeProviderDropdown
+      selectedQuote={selectedQuote}
+      quotes={successfulQuotes}
+      onSelect={(bridgeId) => setSelectedBridgeProvider(bridgeId)}
+    />
+  </QuoteDetailRow>
+);
```
JoseRFelix, line 34: I believe we should be cautious about passing all state values to a component. Any reference change can trigger a re-render, even if we use a memo function. In my opinion, we should only pass the necessary props to a component. This approach simplifies the code, making it easier to read and maintain while also reducing re-renders when required

packages/web/components/bridge/immersive/quote-detail.tsx
```diff
@@ -0,0 +1,244 @@
...
+) {
+  let totalCost = quote.gasCostFiat;
+
+  if (quote.transferFeeFiat) {
+    if (!totalCost) totalCost = quote.transferFeeFiat;
+    else totalCost = totalCost.add(quote.transferFeeFiat);
+  }
+
+  return totalCost;
+}
```
JoseRFelix, line 130:
```suggestion
export function calcSelectedQuoteTotalFee(
  quote: BridgeQuotesDetails["selectedQuote"]
) {
  let totalCost = quote.gasCostFiat;

  if (quote.transferFeeFiat) {
    totalCost = totalCost ? totalCost.add(quote.transferFeeFiat) : quote.transferFeeFiat;
  }

  return totalCost;
}
```

packages/web/components/bridge/immersive/use-bridge-quotes.ts
```diff
@@ -26,7 +27,17 @@ import { api, RouterInputs } from "~/utils/trpc";
 
 const refetchInterval = 30 * 1000; // 30 seconds
 
-export const useBridgeQuote = ({
+export type BridgeQuotes = ReturnType<typeof useBridgeQuotes>;
```
JoseRFelix, line 30:
```suggestion
export type BridgeQuote = ReturnType<typeof useBridgeQuotes>;
```

JoseRFelix (changes requested):

packages/web/components/bridge/immersive/amount-and-review-screen.tsx
```diff
@@ -122,16 +130,40 @@ export const AmountAndConfirmationScreen = observer(
...
-              <Button onClick={onClose}>Close</Button>
-            </div>
+            <>
+              {fromChain &&
+                toChain &&
+                fromAddress &&
+                toAddress &&
+                fromWalletIcon &&
+                toWalletIcon &&
+                sourceAsset &&
```
JoseRFelix, line 146: What if we add this conditional within the review screen, similar to what I did in the amount screen? However, instead of displaying a loading indicator, we throw an error because it shouldn't be possible to proceed without this info

jonator: I think the component code & props are much cleaner and less error prone within the review component if we validate that those are present before passing them to the component. Perhaps Instead of showing nothing we can display a text error message in the UI?

JoseRFelix: Yes, we should display an error message. I believe we should throw the error in the component so we can catch it with an ErrorBoundary and share the fallback between components

JoseRFelix (approved):
LGTM!

PR #3438, opened 2024-07-02
Split Prod and Preview tests

yury-dubinin:

The idea is to reduce a spam from deployment events in Github and allow ability to re-run failed jobs.
So, instead of waiting for a random event and skipping now workflow will wait for a specific Vercel deployment.
Workflow will fail if deployment fails.
Workflow starts on push

- Added separate Prod tests workflow (only swaps for now) on deployment
- Added Portfolio tests to frontend e2e tests 
- Frontend e2e tests will not alert in slack (deployment is visible in a Pull request)
Screenshot 2024-07-04 at 11 38 12


.github/await_deployment.py
```diff
@@ -0,0 +1,46 @@
...
+import requests
+import os
+
+
+def wait_for_deployment(timeout):
+    branch_name = os.getenv('BRANCH_NAME')
+    print(f"Wait for a deployment for a branch name: {branch_name}")
+    time.sleep(30)
+    bearer_token = os.getenv('VERCEL_TOKEN')
+    project_id = "prj_80UzRn6hG8nvUJSwujWUqqho3osT"
```
DavideSegullo, line 11: Should we add it inside the env vars?

yury-dubinin: yes, we can. Better to keep it sealed :)

yury-dubinin: updated!

DavideSegullo (approved):

PR #3439, opened 2024-07-02
feat: added pagination to order history

crnbarr93:
These changes add infinite scroll based pagination to the order history page. To adjust for the headers of each section the virtualization logic includes an extra row for each header (computed based on if it needs to render the header or not).

### Linear Task

[LIM-135: Add Pagination](https://linear.app/osmosis/issue/LIM-135/[ui]-add-pagination)

## Brief Changelog
- Added the use of `useWindowVirtualizer` for order history pagination
- Fixed an error case where `place_timestamp` for historical data may be null
- Fixed an issue with swap page rendering


packages/web/components/complex/orders-history/index.tsx
```diff
@@ -156,54 +211,111 @@ export const OrderHistory = observer(() => {
...
-                      {flexRender(
-                        cell.column.columnDef.cell,
-                        cell.getContext()
-                      )}
-                    </td>
-                  ))}
-                </tr>
-              ))}
+              <h6 className="h-[84px] pb-4 pt-8">Past</h6>
+              {virtualRows
```
MaxMillington, line 271: This was a little hard for me to follow. 

Suggestion:


```
const filledOrdersCount = filledOrders.length;
const pendingOrdersCount = pendingOrders.length;
const hasFilledOrders = filledOrdersCount > 0 ? 1 : 0;
const hasPendingOrders = pendingOrdersCount > 0 ? 1 : 0;
const additionalRows = hasFilledOrders + hasPendingOrders;

const startIndex = filledOrdersCount + pendingOrdersCount + additionalRows;
const endIndex = orders.length + additionalRows;

{virtualRows
  .filter((row) => row.index > startIndex && row.index <= endIndex)
  .map((virtualRow) => {
    const rowIndex = virtualRow.index - (1 + additionalRows);
    const row = rows[rowIndex];

    return (
      <tr key={row.id} className="h-[84px]">
        {row.getVisibleCells().map((cell) => (
          <td key={cell.id} className="!px-0 !text-left">
            {flexRender(cell.column.columnDef.cell, cell.getContext())}
          </td>
        ))}
      </tr>
    );
  })}
```

crnbarr93: I'd considered filtering the actual sub arrays in `useMemo` to make it cleaner

crnbarr93: It's a bit of a hacky solution so curious if there's a general cleaner approach

crnbarr93: Rewrote this to help with readability

fabryscript (approved):

PR #3442, opened 2024-07-02
(Deposit/Withdraw) IBC provider fix gas estimation

jonator:

Currently investigating why timeout height is rejected when simulating IBC transfers on some chains.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-611/ibc-provider-fix-gas-estimation-on-other-chains)


packages/web/components/bridge/immersive/use-bridge-quotes.ts
```diff
@@ -560,14 +533,14 @@ export const useBridgeQuotes = ({
...
-      });
-    }
+    const tx =
+      transactionRequest.type === "evm"
+        ? handleEvmTx({ ...quote, transactionRequest })
+        : handleCosmosTx({ ...quote, transactionRequest });
+
+    await tx.catch((e) => {
+      console.error(transactionRequest.type, "transaction failed", e);
+    });
```
JoseRFelix, line 543: I believe we should also throw the error here so it can be caught at the consumer level

JoseRFelix (approved):

PR #3444, opened 2024-07-03
feat: added sqs info for orderbooks

crnbarr93:
These changes incorporate SQS data when interacting with on chain orderbooks.

Also includes a fix for market order inputs.

## Brief Changelog
- Added `queryOrderbookPools` query
- Altered `useOrderbooks` to use the above query
- Fixed market order spot price reference


packages/server/src/queries/complex/orderbooks/pools.ts
```diff
@@ -0,0 +1,41 @@
...
+export function getOrderbookPools() {
+  return cachified({
+    cache: orderbookPoolsCache,
+    key: `orderbookPools`,
+    ttl: 1000 * 60 * 60, // 1 hour
+    getFreshValue: () =>
+      queryPools().then((data) => {
+        return data
+          .filter((pool) =>
+            orderBookCodeIds.includes((pool as any).chain_model.code_id)
```
fabryscript, line 29: Do we not have any typing for this yet?

packages/server/src/queries/complex/orderbooks/pools.ts
```diff
@@ -0,0 +1,41 @@
...
+          )
+          .map((pool) => {
+            return {
+              baseDenom: (pool as any).balances[1].denom,
+              quoteDenom: (pool as any).balances[0].denom,
```
fabryscript, line 34: just a question, is by design the quote denom the first element and the base the second?

packages/web/components/place-limit-tool/index.tsx
```diff
@@ -129,7 +129,11 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
               <span className="body2 text-osmoverse-300">
                 {swapState.baseAsset?.coinDenom}{" "}
                 {t("assets.table.price").toLowerCase()} ≈{" "}
-                {formatPretty(swapState.priceState.spotPrice ?? new Dec(0))}{" "}
+                {formatPretty(
+                  orderDirection === "bid"
```
fabryscript, line 133: I might be wrong here, so you want to display the ask spot price when the direction is bid?

packages/web/hooks/limit-orders/use-orderbook.ts
```diff
@@ -10,34 +10,12 @@ import { useSwapAsset } from "~/hooks/use-swap";
 import { useStore } from "~/stores";
 import { api } from "~/utils/trpc";
 
-interface Orderbook {
-  baseDenom: string;
-  quoteDenom: string;
-  contractAddress: string;
-}
-
 const USDC_DENOM = process.env.NEXT_PUBLIC_IS_TESTNET
```
fabryscript, line 13: Aren't these the same minimal denoms?

fabryscript (changes requested):

packages/server/src/queries/complex/orderbooks/pools.ts
crnbarr93: Yes, it's what's returned from the contract

packages/web/components/place-limit-tool/index.tsx
crnbarr93: A market order is run against the price of the opposite direction

packages/web/hooks/limit-orders/use-orderbook.ts
crnbarr93: Yea forgot to paste in the testnet denom again, good spot!

fabryscript (approved):

PR #3445, opened 2024-07-03
(Deposit/Withdraw) Add auto scale input, and match figma

jonator:

This re-uses our existing react autosize input 

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-581/scale-crypto-amount-input-according-to-content)


packages/web/components/bridge/immersive/crypto-fiat-input.tsx
```diff
@@ -0,0 +1,226 @@
...
+            width={16}
+            height={16}
+          />
+        </span>
+      </button>
+    </div>
+  );
+};
+
+function calcTextSizeClass(numChars: number): string {
```
MaxMillington, line 212: something like this might be more elegant:

```
const calcTextSizeClass = (numChars: number): string => {
  const sizeMapping: { [key: number]: string } = {
    8: "text-4xl",
    10: "text-3xl",
    12: "text-2xl",
    18: "text-xl",
    24: "text-lg",
  };

  for (const [key, value] of Object.entries(sizeMapping)) {
    if (numChars <= Number(key)) {
      return value;
    }
  }

  return "text-md";
};
```

packages/web/components/bridge/immersive/crypto-fiat-input.tsx
```diff
@@ -0,0 +1,226 @@
...
+  } else if (numChars <= 12) {
+    return "text-2xl";
+  } else if (numChars <= 18) {
+    return "text-xl";
+  } else if (numChars <= 24) {
+    return "text-lg";
+  }
+
+  return "text-md";
+}
```
JoseRFelix, line 226: I wonder how scalable this is. Nass previously created a [code sandbox](https://codepen.io/nassdonald/pen/RwmMMzK) that scales the input based on the content. I believe that solution would be better than hardcoding the text classes and would also be smoother during size transitions. Additionally, I'm a bit concerned about the autosize library. It seems we need many workarounds to get it to function properly

JoseRFelix: I think we could use GPT to convert it to a react component and modify it promptly for this case

jonator: I agree that works really smoothly, but for sake of time can we run with this then adapt the codepen solution later if we decide to?

MaxMillington: agree with @jonator

jonator: This existing solution, re-using our already-imported react-autosize-input package, already took several hours to get working smoothly in a way that matches the Figma designs.

JoseRFelix (approved):

PR #3447, opened 2024-07-03
(Deposit/Withdraw) Subtract gas from input amount

jonator:

If the gas cost in the selected quote exceeds the available balance when summed with the input amount, it is subtracted from the fiat and crypto input amounts. In effect, a new quote is loaded with an input that leaves enough balance to pay fees.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-613/amount-input-subtract-gas)

### DIFF
--- a/packages/web/components/bridge/immersive/amount-and-review-screen.tsx
+++ b/packages/web/components/bridge/immersive/amount-and-review-screen.tsx
@@ -2,7 +2,7 @@ import { CoinPretty } from "@keplr-wallet/unit";
 import { BridgeChain } from "@osmosis-labs/bridge";
 import { isNil, noop } from "@osmosis-labs/utils";
 import { observer } from "mobx-react-lite";
-import { useState } from "react";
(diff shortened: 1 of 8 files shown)


packages/web/components/bridge/immersive/amount-and-review-screen.tsx
```diff
@@ -39,6 +39,12 @@ export const AmountAndReviewScreen = observer(
     const [cryptoAmount, setCryptoAmount] = useState<string>("0");
     const [fiatAmount, setFiatAmount] = useState<string>("0");
 
+    // if the chains change, reset the amounts
+    useEffect(() => {
```
JoseRFelix, line 43: I prefer predictable changes when it comes to state management. I think it's a better idea to handle this change in the `onSelectChain` function on the amount screen

JoseRFelix (changes requested):

JoseRFelix (approved):
LGTM!

PR #3448, opened 2024-07-04
(Deposit/Withdrawal) Display Chain Icons

JoseRFelix:

Display the recently added chain logo URIs in the chain selector. Additionally, simplify the chain object and its setter to eliminate the use of useChainDisplayInfo, and share this simplified object with the review screen

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-446/chain-display-info)


packages/web/server/api/routers/bridge-transfer.ts
```diff
@@ -20,6 +20,7 @@ import { CacheEntry } from "cachified";
 import { LRUCache } from "lru-cache";
 import { z } from "zod";
 
+import { BridgeChainWithDisplayInfo } from "~/components/bridge/immersive/amount-and-review-screen";
```
jonator, line 23: IMO we should define this type here, then import type into the components. Seems bad to import stuff from the view layer into the server layer

packages/web/components/bridge/immersive/review-screen.tsx
```diff
@@ -101,7 +102,7 @@ export const ReviewScreen: FunctionComponent<ConfirmationScreenProps> = ({
           direction === "withdraw"
             ? "transfer.confirmWithdrawTo"
             : "transfer.confirmDepositTo",
-          { chain: quote.toChainInfo?.prettyName ?? "" }
+          { chain: toChain.prettyName ?? "" }
```
jonator, line 105: NIT: may be able to remove `??` operator, since we always know we'll have it.

jonator (changes requested):
Seems there's wrapping issues when the name is long. Maybe we can truncate with ellipsis?
Screenshot 2024-07-04 at 12 01 56 PM

Also on the preview, it seems the icons are not loading. Only EVM icons load.
Screenshot 2024-07-04 at 12 02 39 PM

JoseRFelix:
@jonator Icons are not visible because the Asset List PR has not been merged yet. I'll sync with @JeremyParish69 to merge it

packages/utils/src/ethereum.ts
```diff
@@ -64,94 +64,132 @@ const createEthereumChainInfo = <
...
     clientChainId: "Filecoin - Mainnet",
+    color: "#0494fc",
+    relativeLogoUrl: "/networks/filecoin.svg",
   }),
   "Filecoin Hyperspace": mapChainInfo({
     chain: filecoinHyperspace,
     axelarChainName: "Filecoin Hyperspace",
     clientChainId: "Filecoin Hyperspace",
+    color: "#0494fc",
+    relativeLogoUrl: "/networks/filecoin.svg",
```
jonator, line 193: A problem with this is it may not work on other apps, like mobile for instance. I wonder if there's a way to map these images within web package?

jonator (approved):

packages/utils/src/ethereum.ts
JoseRFelix: My thinking here is that since it's an object, mobile can either overwrite the values or follow the convention and create a folder with the same relative path

PR #3450, opened 2024-07-04
(Deposit/Withdraw) Fix IBC status provider

jonator:

Fixed some issues with IBC status provider (it's tx query), as well as invalidating and preventing queries when txs are sent. It seems there's still some rare edge case bugs with balances for the selected asset being incorrect.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-621/fix-ibc-status-provider)

## Brief Changelog

* Fix /txs/ endpoint in IBC status provider
* Fix deserializing tx events in IBC status provider (some nodes return the tx.event array base64 encoded, in which case the raw_log will be present with decoded events array when parsed as JSON. If raw_log is not present, the tx.event array should be decoded and used)
* Invalidate user account queries used by bridge flow when a tx is confirmed
* Prevent subtracting into a negative value when balance is less than the gas fee. Set to 0 instead.

## Testing and Verifying

Transfers should work and be reflected in transfer history table and toasts.


packages/web/components/bridge/immersive/amount-and-review-screen.tsx
```diff
@@ -103,8 +105,12 @@ export const AmountAndReviewScreen = observer(
           ? fromAsset?.supportedVariants[toAsset?.address ?? ""]
           : toAsset?.supportedVariants[fromAsset?.address ?? ""],
       onTransfer: () => {
+        setToAsset(undefined);
+        setFromAsset(undefined);
         setCryptoAmount("0");
         setFiatAmount("0");
+
+        refetchUserQueries(apiUtils);
```
jonator, line 113: Makes sure balances are accurate when restarting the d/w flow

packages/web/components/bridge/immersive/amount-and-review-screen.tsx
```diff
@@ -103,8 +105,12 @@ export const AmountAndReviewScreen = observer(
           ? fromAsset?.supportedVariants[toAsset?.address ?? ""]
           : toAsset?.supportedVariants[fromAsset?.address ?? ""],
       onTransfer: () => {
+        setToAsset(undefined);
+        setFromAsset(undefined);
```
jonator, line 109: Resets state when restarting the flow

packages/web/components/bridge/immersive/use-bridge-quotes.ts
```diff
@@ -160,6 +167,7 @@ export const useBridgeQuotes = ({
         },
         {
           enabled:
+            !isTxPending &&
```
jonator, line 170: Prevents queries from running while the tx is confirming/just confirmed

packages/web/components/bridge/immersive/use-bridge-quotes.ts
```diff
@@ -377,28 +385,22 @@ export const useBridgeQuotes = ({
           `${providerId}${JSON.stringify(params)}`,
           inputCoin.trim(true).toString(),
           isWithdraw,
-          toAddress ?? "" // use osmosis account (destinationAddress) for account keys (vs any EVM account)
+          (isWithdraw ? fromAddress : toAddress) ?? "" // use osmosis account (destinationAddress) for account keys (vs any EVM account)
```
jonator, line 388: Transfer history is keyed and filtered by Osmosis account that is currently connected

packages/web/stores/index.tsx
```diff
@@ -26,6 +26,8 @@ function refetchUserQueries(apiUtils: ReturnType<typeof api.useUtils>) {
   apiUtils.edge.pools.getUserSharePool.invalidate();
   apiUtils.edge.pools.getSharePoolBondDurations.invalidate();
   apiUtils.local.balances.getUserBalances.invalidate();
+  apiUtils.local.bridgeTransfer.getSupportedAssetsBalances.invalidate();
+  apiUtils.edge.assets.getImmersiveBridgeAssets.invalidate();
```
jonator, line 30: Added these to ensure old balance data is not displayed in the d/w flow after a transfer is started / completed. @JoseRFelix LMK if I'm missing any

yarn.lock
jonator, line 1: Not sure why yarn lock was changed when installing deps on stage

packages/web/components/bridge/immersive/amount-and-review-screen.tsx
JoseRFelix: Nit: I believe we should add this same GH comment as a code comment for this line

packages/web/components/bridge/immersive/use-bridge-quotes.ts
JoseRFelix: Let's add this comment here too

JoseRFelix (approved):
LGTM!

PR #3453, opened 2024-07-05
(Deposit/Withdraw) Allow Selecting a Custom Withdraw Address

JoseRFelix:
## Description

This PR introduces the ability to add a custom address for withdrawals. It also makes wallet connection and switching entirely predictable and imperative by not relying on `useEffect`s.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-619/allow-selecting-a-custom-withdraw-address)


packages/web/components/bridge/immersive/amount-screen.tsx
```diff
@@ -538,6 +535,8 @@ export const AmountScreen = observer(
       setFiatAmount("0");
     };
 
+    console.log(manualToAddress);
```
jonator, line 538: log

packages/web/components/bridge/immersive/crypto-fiat-input.tsx
```diff
@@ -46,6 +50,7 @@ export const CryptoFiatInput: FunctionComponent<{
   const inputRef = useRef<HTMLInputElement>(null);
 
   const [isMax, setIsMax] = useState(false);
+  const [hasSubtractedAmount, setHasSubtractedAmount] = useState(false);
```
jonator, line 53: Wouldn't it be possible to derive this value instead? We could see if the input amount is greater then the max available amount and set `hasSubtractedAmount` to false if that's not the case. This could subject us to less invalid state / race conditions

jonator (changes requested):
Some initial comments. BTW for some reason when I select an asset to withdraw, the page freezes and chrome suggests killing the page. Perhaps it's stuck in an infinite loop or something?

packages/web/components/bridge/immersive/crypto-fiat-input.tsx
JoseRFelix: I considered this, but I'm not sure it's possible since we're changing the input value. Without a reference point, unless I keep a state value, it would be difficult to manage

JoseRFelix: Resolved in my latest commit. It's much more streamlined, thank you!

jonator (approved):

PR #3462, opened 2024-07-08
feat: implemented max button for limit inputs

crnbarr93:
These changes implement the "max" button for limit/market order inputs. The implementation of this button currently assumes the quote denom  is a stablecoin as it is directly connected to the fiat input. Future iterations should remove this assumption once non-stablecoin quote denoms are enabled.

### Linear Task

[LIM-164: Wire up Limit Input Max Button](https://linear.app/osmosis/issue/LIM-164/wire-up-limit-input-max-button)

## Brief Changelog

<!-- _(for example:)_
- Added a `toggleMax` callback to the limit input that sets the relevant input amount to its max amount


fabryscript (approved):
simplified some logic concerning the `toggleMax` function and updated custom max label to be `rust-300` when the `insufficientBalance` state is triggered

PR #3484, opened 2024-07-10
feat: add amplitude events

JoseRFelix:
### Linear Task


jonator (approved):
LGTM

Wonder if we can add a user property `isOneClickTradingEnabled` so we a session can be flagged as 1CT or not. Might be useful to bring up with Rafa

PR #3488, opened 2024-07-10
(Deposit/Withdraw) Chain order (+ polishing)

jonator:
### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-716/amount-screen-adjust-initial-chain-and-chains-order)

[Padding and background are incorrect on alternative asset dropdown](https://linear.app/osmosis/issue/FE-643/padding-and-background-are-incorrect-on-alternative-asset-dropdown)
[Modal titles are the incorrect Header size](https://linear.app/osmosis/issue/FE-644/modal-titles-are-the-incorrect-header-size)

### DIFF
--- a/packages/web/components/bridge/amount-screen.tsx
+++ b/packages/web/components/bridge/amount-screen.tsx
@@ -921,8 +921,7 @@ export const AmountScreen = observer(
                                 <button
                                   className={classNames(
                                     "flex items-center gap-3 rounded-lg py-2 px-3 text-left data-[active]:bg-osmoverse-800",
-                                    isSelected && "bg-osmoverse-700",
-                                    !isSelected && "bg-osmoverse-800"
+                                    isSelected && "bg-osmoverse-700"
                                   )}
                                   onClick={onClick}
                                 >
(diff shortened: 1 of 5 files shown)


packages/web/components/bridge/use-bridges-supported-assets.ts
```diff
@@ -56,14 +56,6 @@ export const useBridgesSupportedAssets = ({
...
             // query that the user can use.
             retry: false,
-
-            // prevent batching so that fast routers can
-            // return requests faster than the slowest router
-            trpc: {
-              context: {
-                skipBatch: true,
-              },
-            },
```
jonator, line 66: I realized there was a race condition when the initial from chain was getting set. We wanted to pick EVM chains first, but if a query landed with just cosmos chains, the first cosmos chain would be picked.
Hopefully, if we send all the supported assets queries as a single batch, we'd more reliably get the chains we want to sort and set initially

packages/web/components/bridge/amount-screen.tsx
```diff
@@ -371,27 +371,37 @@ export const AmountScreen = observer(
...
+            !isNil(fromChain) &&
+            !isNil(supportedSourceAssets) &&
+            supportedSourceAssets.length > 0,
 
           select: (data) => {
             let nextData: typeof data = data;
 
             // Filter out assets with no balance
-            if (nextData) {
+            if (nextData && nextData.length) {
```
jonator, line 404: This shouldn't happen, but we should check that the balances query returned is not empty before using `[0]`

DavideSegullo (approved):

PR #3495, opened 2024-07-11
(Deposit/Withdraw) Add external url UI if asset not quotable

jonator:

This PR introduces Bitcoin and Solana as networks, as well as BTC and SOL as assets.

For BTC and SOL (high prio assets), we don't currently have providers that can actually carry out the transfer natively within the app. Instead, we need to direct users to external UIs to carry out this transfer.

This PR adds the Nomic (BTC) and Wormhole (SOL) providers and just implements the supported assets and external URL methods. If the user selects an Osmosis asset variant that is bridged by Nomic or Wormhole, the providers would return BTC and SOL respectively as options, including Bitcoin and Solana networks. These then become selectable options in the UI as networks and assets.

However, in the client, we have a notion of "quotable bridges" that are passed to the `useBridgesQuotes` hook. Only some providers (the original ones) can actually provide quotes. So, if an asset/network is selected that is not quotable, the `useBridggesQuotes` hook returns an `enabled: false` flag indicating that. We can then rely on that to replace the Amount UI with this external link UI that directs focus to the external URLs available for this chain + asset.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-743/add-external-url-chains-and-external-ui)

## Brief Changelog

* Add nomic and wormhole bridge providers, throw exceptions in the quote related methods. Returns BTC and SOL and Bitcoin and Solana networks as supported asstes
* Pass bridges to external URL providers to only get external URLs for the selected from asset
* Add Solana and Bitcoin network display info

## Testing and Verifying

Select SOL or nBTC, observe Solana and Bitcoin networks are available. Observe the amount UI being replaced with the external link UI.


packages/web/components/bridge/amount-screen.tsx
```diff
@@ -351,23 +356,36 @@ export const AmountScreen = observer(
...
+              : {
+                  // Note: for now, these will always return 0 balances since bitcoin and solana are not supported
+                  type: fromChain?.chainType ?? "bitcoin",
+                  assets:
+                    (supportedSourceAssets as Extract<
+                      SupportedAsset,
+                      { chainType: "bitcoin" }
+                    >) ?? [],
+                },
+        },
```
JoseRFelix, line 388: Thoughts on using a self-calling function to simplify the conditional flow so it's more readable? 

```suggestion
const assetDetails = (() => {
  if (fromChain?.chainType === "evm") {
    return {
      type: "evm",
      assets: supportedSourceAssets.filter(asset => asset.chainType === "evm"),
      userEvmAddress: evmAddress,
    };
  } else if (fromChain?.chainType === "cosmos") {
    return {
      type: "cosmos",
      assets: supportedSourceAssets.filter(asset => asset.chainType === "cosmos"),
      userCosmosAddress: fromCosmosCounterpartyAccount?.address,
    };
  } else {
    return {
      type: fromChain?.chainType ?? "bitcoin",
      assets: supportedSourceAssets.filter(asset => asset.chainType === "bitcoin") ?? [],
    };
  }
})();
```

packages/web/server/api/routers/bridge-transfer.ts
```diff
@@ -335,6 +341,24 @@ export const bridgeTransferRouter = createTRPCRouter({
...
+              color: "#F7931A",
+            } as Extract<BridgeChainWithDisplayInfo, { chainType: "bitcoin" }>;
+          } else if (chainType === "solana") {
+            return {
+              prettyName: "Solana",
+              chainId: "solana",
+              chainName: "Solana",
+              chainType,
+              logoUri: "/networks/solana.svg",
+              color: "#9945FF",
```
JoseRFelix, line 360: How about we add this static data to our @osmosis-utils package? Similar to Ethereum, I believe it could be useful to share

JoseRFelix (approved):
LGTM! I left some nits

PR #3501, opened 2024-07-12
Davidesegullo/fe 549 create edge function to call blockaid

DavideSegullo:
### Linear Task


packages/server/src/queries/blockaid/index.ts
```diff
@@ -0,0 +1,10 @@
+export * from "./transaction";
+
+export const BASE_URL = process.env.BLOCKAID_BASE_URL;
```
jonator, line 3: Could we access all process env vars in the same place in the root env file of this package?

packages/server/src/queries/blockaid/index.ts
```diff
@@ -0,0 +1,8 @@
+export * from "./transaction";
+
+export const blockaidAuthHeaders: HeadersInit | undefined = process.env
+  .BLOCKAID_API_KEY
+  ? {
+      "x-api-key": process.env.BLOCKAID_API_KEY,
+    }
+  : undefined;
```
MaxMillington, line 8: If the API Key is undefined, I'm assuming it won't cause a breaking error? I'm thinking about third party developers who download the app and want to run it locally but wont have the env variable.

MaxMillington (approved):
One question but otherwise looks good to me.

packages/server/src/queries/blockaid/index.ts
DavideSegullo: Ye they'll get errors, but it should not break anything because we don't await for this promise inside the sign and broadcast steps

packages/stores/src/account/base.ts
```diff
@@ -1272,6 +1272,16 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
           })
         : undefined;
 
+      apiClient("/api/transaction-scan", {
```
jonator, line 1275: Is this just logging? Why not await? Is it validating the tx, and returning an error HTTP status if invalid?

PaddyMc (approved):

packages/stores/src/account/base.ts
PaddyMc: It's just to send data to Blockaid so they can train their model, likely soon well need to do something with the data

PR #3508, opened 2024-07-12
(Deposit/Withdraw) Improve ux for picking best quote

jonator:
### Linear Task

### DIFF
--- a/packages/bridge/src/skip/index.ts
+++ b/packages/bridge/src/skip/index.ts
@@ -1,7 +1,8 @@
 import { fromBech32, toBech32 } from "@cosmjs/encoding";
 import { Registry } from "@cosmjs/proto-signing";
 import { ibcProtoRegistry } from "@osmosis-labs/proto-codecs";
-import { estimateGasFee } from "@osmosis-labs/tx";
+import { queryRPCStatus } from "@osmosis-labs/server";
+import { calcAverageBlockTimeMs, estimateGasFee } from "@osmosis-labs/tx";
 import { CosmosCounterparty, EVMCounterparty } from "@osmosis-labs/types";
 import {
   EthereumChainInfo,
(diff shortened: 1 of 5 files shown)


packages/bridge/src/skip/index.ts
```diff
@@ -621,7 +614,77 @@ export class SkipBridgeProvider implements BridgeProvider {
...
+    const toRpc = toChain?.apis.rpc[0]?.address;
+
+    const [fromBlockTimeMs, toBlockTimeMs] = await Promise.all([
+      fromRpc
+        ? queryRPCStatus({ restUrl: fromRpc }).then(calcAverageBlockTimeMs)
+        : this.getFinalityTimeForEvmChain(fromChainId) * 1000,
+      toRpc
+        ? queryRPCStatus({ restUrl: toRpc }).then(calcAverageBlockTimeMs)
+        : this.getFinalityTimeForEvmChain(toChainId) * 1000,
+    ]);
```
JoseRFelix, line 667: If we're unsure whether the RPC address is defined, we should handle the case where it's a Cosmos chain and the address is undefined. Otherwise, we could send the wrong chain id to getFinalityTimeForEvmChain.

JoseRFelix (changes requested):
Looking good! I left one comment

packages/bridge/src/skip/index.ts
jonator: In that case perhaps we can use a hardcoded value? Like 7.5 seconds for a cosmos chain

jonator: But also, an RPC should be defined with pretty high confidence in the chain registry though

JoseRFelix: I agree that this is unlikely to occur, however for insurance a 7.5 seconds like you mention could be a good fallback

JoseRFelix (approved):

PR #3509, opened 2024-07-12
(Deposit/Withdraw) Handle Approve Txs in Review Screen

JoseRFelix:

Currently, the amount screen continue button shows "approve transaction" if the provider returns a tx approval request. Clicking on that button text advances you to the review screen where clicking "confirm" then shows the approval tx popup.

We should change this logic to not show "approve transaction" on the amount screen, and instead handle the approval immediately followed by the transfer tx in the review screen.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-739/handle-approve-txs-in-the-review-screen)

## Brief Changelog

- Match max button design with Figma
- Fix issue where wallet was not prompting user to change network when the first network was different than selected.


packages/web/components/bridge/use-bridge-quotes.ts
```diff
@@ -481,6 +501,7 @@ export const useBridgeQuotes = ({
       setTransferInitiated(true);
     } catch (e) {
       const error = e as BaseError;
+      console.dir(e);
```
jonator, line 504: Console log

jonator (approved):
One nit

PR #3521, opened 2024-07-15
(1CT) Exclude other fee tokens for 1CT transactions

JoseRFelix:

There's an unexpected error coming from the chain when using other fee tokens while broadcasting a 1CT transaction. We've identified this as an issue on the chain that will be resolved in v26. As a temporary measure, we'll exclude alternate fee tokens. However, our current gas estimation logic doesn't prioritize OSMO over other tokens, which means we'll consistently get tokens other than OSMO. If we enable onlyDefaultFeeDenom, we'll get OSMO as a fee token, but this doesn't mean we want to exclude other tokens entirely. We still want them as a fallback for wallet signing, so we can't use that setting. Therefore, we need to update our gas logic to support this case

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-782/exclude-other-fee-tokens-for-1ct-transactions)

## Brief Changelog

- Excludes fee tokens from 1CT transactions
- Prioritize OSMO to pay gas

## Testing and Verifying

- [ ] Can perform 1CT txs without unexpected errors


packages/tx/src/gas.ts
jonator, line 1: Let's make sure tests are updated to account for these changes in picking of fee tokens

jonator (changes requested):

CryptoAssassin1 (approved):

kamal-sutra (approved):

PaddyMc (approved):
Ace, ty

PR #3531, opened 2024-07-16
feat: error states for limit/market orders

crnbarr93:
These changes add better error feedback for users for market/limit orders. They also include some alterations to inputs to accommodate for better decimal place calculations.

## Brief Changelog
- Added `error` field to `usePlaceLimit`
- Added `error` field to `useOrderbook`
- Limited inputs to account for maximum decimal places
- Altered fiat/crypto calculations to include decimal places and trim any unnecessary trailing "0"s


fabryscript (approved):
I think we can avoid some redundancy related to parsing the inputs

PR #3558, opened 2024-07-18
feat: limit order amplitude events

crnbarr93:
This change adds amplitude analytics events for limit orders. The following events are included:
- Place Limit Started
- Place Limit Completed
- Place Limit Failed
- Claim All Orders Started
- Claim All Orders Failed
- Order History Page Viewed
- Buy/Sell/Swap tab selected
- Limit/Market selected

The current events for swaps were also added for market orders.

### Linear Task

[LIM-178: Add Amplitude Analytics Events](https://linear.app/osmosis/issue/LIM-178/add-amplitude-analytic-events)

## Brief Changelog
- Added limit order amplitude analytics event names
- Added event logs for limit order related events

### DIFF
--- a/packages/web/components/complex/orders-history/index.tsx
+++ b/packages/web/components/complex/orders-history/index.tsx
@@ -52,6 +52,10 @@ export const OrderHistory = observer(() => {
       await claimAllOrders();
       logEvent([EventName.LimitOrder.claimOrdersCompleted]);
     } catch (error) {
+      if (error instanceof Error && error.message === "Request rejected") {
+        // don't log when the user rejects in wallet
+        return;
+      }
       const { message } = error as Error;
       logEvent([
         EventName.LimitOrder.claimOrdersFailed,
--- a/packages/web/components/complex/orders-history/index.tsx
+++ b/packages/web/components/complex/orders-history/index.tsx
@@ -24,7 +24,9 @@ import { useStore } from "~/stores";
 export type Order = ReturnType<typeof useOrderbookAllActiveOrders>["orders"][0];
 
 export const OrderHistory = observer(() => {
-  const { logEvent } = useAmplitudeAnalytics();
+  const { logEvent } = useAmplitudeAnalytics({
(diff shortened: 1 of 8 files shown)


packages/web/components/complex/orders-history/index.tsx
```diff
@@ -41,6 +46,24 @@ export const OrderHistory = observer(() => {
...
+      logEvent([EventName.LimitOrder.claimOrdersCompleted]);
+    } catch (error) {
+      if (error instanceof Error && error.message === "Request rejected") {
+        // don't log when the user rejects in wallet
+        return;
+      }
+      const { message } = error as Error;
+      logEvent([
+        EventName.LimitOrder.claimOrdersFailed,
+        { errorMessage: message },
```
jonator, line 62: is this going to be useful via amplitude? To me, it seems it would be preferable to use Sentry or DataDog (which I believe we're now using) instead to capture error messages. Then, we'd get stack trace info and otherwise more info useful for troubleshooting

I don't think Amplitude is the tool to use for error messages, IMO

packages/web/components/place-limit-tool/limit-trade-details.tsx
```diff
@@ -20,18 +19,9 @@ interface LimitTradeDetailsProps {
...
-      ).mul(swapState.makerFee);
-
-      const totalFeeFiat = makerFeeFiat;
-      return {
-        makerFeeFiat,
-        totalFeeFiat,
-      };
-    }, [swapState.makerFee, swapState.paymentFiatValue]);
+    const makerFeeFiat = useMemo(() => {
+      return swapState.feeUsdValue;
```
jonator, line 23: Nit: this memo could be removed

jonator (approved):
Some comments

packages/web/components/complex/orders-history/index.tsx
crnbarr93: We can move this to Datadog if that's preferred. Do you have any examples of the best way to do that?

PR #3571, opened 2024-07-19
Add blockaid base URL to .env

jonator:

This will make it available by default, and override-able in deployments


packages/web/.env
```diff
@@ -11,6 +11,7 @@ NEXT_PUBLIC_TFM_API_BASE_URL=https://api.tfm.com
 SENTRY_IGNORE_API_RESOLUTION_ERROR=1
 NEXT_PUBLIC_SQUID_INTEGRATOR_ID=osmosis-api
 NEXT_PUBLIC_SIDECAR_BASE_URL=https://sqs.osmosis.zone
+BLOCKAID_BASE_URL=http://api.blockaid.io:80
```
DavideSegullo, line 14: Actually it is already defined on line 50 on .env file

DavideSegullo (changes requested):

packages/web/.env
jonator: Ah it just needs to be uncommented 👍

DavideSegullo (approved):

PR #3578, opened 2024-07-22
[Limit Orders]: Design QA Changes

crnbarr93:
These changes are for minor design adjustments across the board. The majority of changes were in the denom selection and limit input components. An entire list of the changes can be found in the linear task.

A few tasks are left as bold as they will be part of other PRs in future.

### Linear Task

[FE-810: Design QA](https://linear.app/osmosis/issue/FE-810/design-qa)

## Brief Changelog
- Several design changes to denom selection
- Several design changes to limit input
- Limit input validation changes


packages/web/components/layouts/main.tsx
```diff
@@ -60,7 +56,6 @@ export const MainLayout = observer(
         </div>
         <NavBar
           className="ml-sidebar md:ml-0"
-          title={selectedMenuItem?.label ?? ""}
```
fabryscript, line 63: i think we don't need to remove every title on every page, you can use the `useNavbar` hook inside the selected page to manipulate the current title

packages/web/components/navbar/index.tsx
```diff
@@ -185,7 +184,7 @@ export const NavBar: FunctionComponent<
       <>
         <div
           className={classNames(
-            "fixed z-[60] flex h-navbar w-[calc(100vw_-_14.58rem)] place-content-between items-center bg-osmoverse-900 px-8 shadow-md lg:gap-5 md:h-navbar-mobile md:w-full md:place-content-start md:px-4",
```
fabryscript, line 188: you can pass something like `!shadow-none` in the className prop in order to do this, no need to edit the global navbar

packages/web/components/navbar/index.tsx
```diff
@@ -254,9 +253,6 @@ export const NavBar: FunctionComponent<
             </Popover>
           </div>
           <div className="flex shrink-0 grow items-center gap-9 lg:gap-2 md:place-content-between md:gap-1">
-            <h4 className="md:text-h6 md:font-h6">
```
fabryscript, line 257: same as above

packages/web/components/place-limit-tool/index.tsx
```diff
@@ -120,7 +140,12 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
     };
 
     const buttonText = useMemo(() => {
-      if (swapState.error) {
+      if (
+        swapState.error &&
+        swapState.error !== "errors.zeroAmount" &&
+        swapState.error !== "errors.emptyAmount" &&
+        swapState.error !== "limitOrders.insufficientFunds"
```
fabryscript, line 148: can't you invert this whole chain of "!==" into one expression of "===" with the only leftover condition?

fabryscript (changes requested):

packages/web/components/place-limit-tool/index.tsx
crnbarr93: Isn't exclusion simpler here? If we add more errors to the `swapState` we would have to update this every time right?

fabryscript: I do not get if this boolean chains aims to isolate a specific error or not

crnbarr93: We don't want to display error text for certain errors right? Like Zero amount input/Empty input/Insufficient balance

crnbarr93:
```typescript
const NON_DISPLAY_ERRORS = [
  "errors.zeroAmount",
  "errors.emptyAmount",
  "limitOrders.insufficientFunds",
];

const buttonText = useMemo(() => {
  if (swapState.error && !NON_DISPLAY_ERRORS.includes(swapState.error)) {
    return t(swapState.error);
  } else {
    return orderDirection === "bid"
      ? t("portfolio.buy")
      : t("limitOrders.sell");
  }
}, [orderDirection, swapState.error, t]);
```

How about something like this?

crnbarr93:
Closing, covered in other PRs post redesign.

PR #3580, opened 2024-07-22
(Deposit/Withdraw) Add direction & asset query params

jonator:

Store asset and direction in query params. The default direction will be deposit if only the asset is given.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-793/store-denom-and-direction-in-query-params)

### DIFF
--- a/packages/web/components/bridge/immersive-bridge.tsx
+++ b/packages/web/components/bridge/immersive-bridge.tsx
@@ -67,7 +67,7 @@ export const ImmersiveBridgeFlow = ({
       setIsVisible(true);
       setStep(ImmersiveBridgeScreen.Amount);
     }
-  }, [direction, selectedAssetDenom, isVisible, setIsVisible, setDirection]);
+  }, [direction, selectedAssetDenom, isVisible, setDirection]);
 
   const [fiatRampParams, setFiatRampParams] = useState<{
     fiatRampKey: FiatRampKey;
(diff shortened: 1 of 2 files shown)


DavideSegullo (approved):

packages/web/components/bridge/immersive-bridge.tsx
```diff
@@ -36,15 +37,41 @@ export const ImmersiveBridgeFlow = ({
...
+    parse: (value) => {
+      if (typeof value === "string") return value;
+      return null;
+    },
+  });
 
-  const [selectedAssetDenom, setSelectedAssetDenom] = useState<string>();
+  useEffect(() => {
+    if (!isNil(direction) && !isVisible) setIsVisible(true);
+    if (isNil(direction) && isVisible) setDirection("deposit"); // default direction
```
JoseRFelix, line 69: I think we can set this in the `useQueryState` directly. defaultValue: "deposit" and parse returns "deposit" with invalid value

JoseRFelix (changes requested):

packages/web/components/bridge/immersive-bridge.tsx
jonator: I tried this, but then the state being set causes the flow to become visible. How do we differentiate that the value was the default or was set from query params?

JoseRFelix (approved):

PR #3581, opened 2024-07-22
(Deposit/Withdraw) Add Error Boundary and Fallback

JoseRFelix:

Display generic error screen when there is an unhandled error on the bridge flow. So, users do not get stuck in infinite loading if there is a non-recoverable error.  

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-837/add-error-boundary-and-fallback)


packages/web/components/error/__tests__/error-boundary.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+    render(
+      <ErrorBoundary FallbackComponent={FallbackComponent}>
+        <ChildComponent />
+      </ErrorBoundary>
+    );
+
+    expect(screen.getByText("Something went wrong")).toBeInTheDocument();
+  });
+
+  it("should call Sentry.captureException when an error is thrown", () => {
```
jonator, line 36: NIT BTW we're migrating away from sentry, may be good to use some sort of generic error capturing proxy to help decouple us from the observability provider. Should already have one

jonator (approved):

PR #3587, opened 2024-07-23
(Deposit/Withdraw) Avoid persistent cache for bridge balance query

JoseRFelix:

Outdated balances can sometimes appear in the deposit/withdraw flow, especially on the assets page. This PR disables the persistent cache for bridge-supported assets' balance queries to avoid receiving stale data on subsequent requests, even after a reload.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-831/prevent-stale-balances-from-showing)


packages/web/utils/trpc.ts
```diff
@@ -60,6 +60,25 @@ export const api = createTRPCNext<AppRouter>({
     persistQueryClient({
       queryClient,
       persister: localStoragePersister,
+      dehydrateOptions: {
+        shouldDehydrateQuery: (query) => {
+          const [key] = query.queryKey as [string[]];
+          if (Array.isArray(key)) {
+            const trpcKey = key.join(".") as RouterKeys;
+            const excludedKeys: RouterKeys[] = [
+              "local.bridgeTransfer.getSupportedAssetsBalances",
```
jonator, line 69: This seems leaky, can we not specify this somehow at the place the query is used? Aren't there custom query opts or something?

jonator (changes requested):

packages/web/utils/trpc.ts
JoseRFelix: Yeah, we can set the cacheTime and staleTime to 0. However, that will only work for that specific call. My idea is to remove the cache for all subsequent queries

jonator: I'm wondering if there's one place where we can specify queries that involve balances, somehow

In the stores context (where we construct the Root store), we list them so that they can get invalidated after txs, but a similar concept applies here.

I imagine it'd be somewhere in the web package, since that's where the routes get defined. Or, I wonder if tRPC has some way we can decorate procedures to flag it as a balances procedure...

jonator (approved):

PR #3592, opened 2024-07-23
Reduce Build Time by ~%62.58

JoseRFelix:

There have been complaints about build times and out-of-memory issues during the build process. These problems started after we integrated Wormhole. It's not surprising, given that the library is 86 MB. It's very demanding to build another app on top of our own. This PR resolves the issue by loading Wormhole through the script with unpkg, deferring the script loading to the client rather than during the build process.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-851/fix-build-times-and-oom-issues)

### Testing

- [ ] Wormhole page works as prod


DavideSegullo (approved):

packages/web/pages/wormhole.tsx
```diff
@@ -176,24 +354,57 @@ const Wormhole: FunctionComponent = () => {
...
+      )}
+      <div
+        id="wormhole-connect"
+        data-config={JSON.stringify(config)}
+        data-theme={JSON.stringify(customTheme)}
+        style={{ display: scriptLoaded ? "block" : "none" }}
+      ></div>
+
+      <link
+        rel="stylesheet"
```
DavideSegullo, line 403: Just for security reason can we use the [`integrity`](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) attribute so we make sure the file hash is valid?

JoseRFelix: Good catch! I'll open a PR adding it

PR #3595, opened 2024-07-24
(Deposit/Withdraw) Allow deposit/withdraw of canonical assets in asset info page

jonator:

In short, this PR removes the requirement that there be some transfer methods for an asset in order for d/w buttons to be enabled.

It also comes with changes to file names, symbol names, component props/type decorations, to make things a bit more consistent and readable (i.e. we now call it "asset info page" instead of "token info page"

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-852/allow-depositwithdraw-in-asset-detail-page-for-canonical-assets)

## Brief Changelog

* Remove requirement that the be at least one transfer method for d/w buttons to be enabled in asset info page
* Clean up some symbols in asset info page components

## Testing and Verifying

Assets like OSMO and USDT(alloy) should now have transfer methods available.


packages/utils/src/string.ts
```diff
@@ -2,14 +2,14 @@ import * as cosmjsEncoding from "@cosmjs/encoding";
...
-export function truncateString(str: string, num = 8) {
+export function truncate(str: string, num = 8) {
   if (str.length <= num) {
     return str;
   }
   return str.slice(0, num) + "...";
 }
 
-export function getShortAddress(
+export function shorten(
```
JoseRFelix, line 12: Thoughts on renamin this to  `shortenAddress` since. the latter doesn't convey what is being shortened?

packages/web/components/pages/asset-info-page/balance.tsx
```diff
@@ -5,30 +5,28 @@ import { observer } from "mobx-react-lite";
...
-interface YourBalanceProps {
-  className?: string;
-}
-
-export const YourBalance = observer(({ className }: YourBalanceProps) => {
+export const AssetBalance = observer(({ className }: CustomClasses) => {
   const { chainStore, accountStore } = useStore();
   const { bridgeAsset } = useBridge();
-  const { token } = useAssetInfo();
+  const { asset: asset } = useAssetInfo();
```
JoseRFelix, line 20:
```suggestion
  const { asset } = useAssetInfo();
```

JoseRFelix (changes requested):
Looks good! Left some comments

packages/utils/src/string.ts
jonator: I made it more generic since it can be used for any string, like how it's used in asset info page to allow user to copy IBC hash (or base denom)

DavideSegullo (approved):

JoseRFelix (approved):

PR #3601, opened 2024-07-25
Portfolio performance chart

mattupham:

- add portfolio chart data over time

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-783/portfolio-add-chart-for-portfolio-performance)

## Brief Changelog

note - will add responsiveness / expand, minimize in separate PR

- [x] Loading states
- [x] Derive Range from AvailablePriceRanges // challenging to share between client / server
- [x] Add chart color
- [x] Add day / date conversion
- [x] Add price calc
- [x] Add Up / Down arrow + colors
- [x] Create tRPC query
- [x] Pull data in on FE
- [x] Map Data
- [x] Render Skeleton
- [x] Render Chart
- [x] Add interval Select

## Testing and Verifying

https://github.com/user-attachments/assets/d0b8bfed-b165-414a-a975-404e18148523

Screenshot 2024-07-26 at 9 24 25 AM

### DIFF
--- a/packages/server/src/queries/complex/portfolio/allocation.ts
+++ b/packages/server/src/queries/complex/portfolio/allocation.ts
@@ -70,7 +70,7 @@ export function calculatePercentAndFiatValues(
   categories: Categories,
   assetLists: AssetList[],
   category: "total-assets" | "user-balances",
-  allocationLimit: number
+  allocationLimit = 5
 ) {
   const totalAssets = categories[category];
   const totalCap = new Dec(totalAssets.capitalization);
--- a/packages/web/server/api/routers/bridge-transfer.ts
+++ b/packages/web/server/api/routers/bridge-transfer.ts
@@ -96,6 +96,14 @@ export const bridgeTransferRouter = createTRPCRouter({
       /** If the bridge takes longer than 10 seconds to respond, we should timeout that quote. */
       const quote = await timeout(quoteFn, 10 * 1000)();
 
+      // Basic circuit breaker to validate some invariants
+      // from input + given quote
+      if (input.fromAsset.address !== quote.input.address) {
+        throw new Error(
+          `Invalid quote: Expected fromAsset address ${input.fromAsset.address} but got ${quote.input.address} in quote`
+        );
+      }
+
       /**
        * Since transfer fee is deducted from input amount,
        * we overwrite the transfer fee asset to be the input


packages/server/src/queries/data-services/portfolio-over-time.ts
```diff
@@ -0,0 +1,27 @@
...
+  address: string;
+  range: string;
+}): Promise<PortfolioOverTimeResponse[]> {
+  const url = new URL("/users/portfolio/over_time", NUMIA_BASE_URL);
+
+  url.searchParams.append("address", address);
+  url.searchParams.append("range", range);
+
+  const headers = {
+    Authorization: `Bearer ${process.env.NEXT_PUBLIC_NUMIA_API_KEY}`,
```
jonator, line 23: Like we did with `NUMIA_BASE_URL`, could we import `NEXT_PUBLIC_NUMIA_API_KEY` from that file as well?

Also, for an API key, we'd likely want to remove `NEXT_PUBLIC_` as that would leak the env var into the client (that's the signifier next.js uses to make env vars available to client)

packages/web/components/complex/portfolio/get-started-with-osmosis.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <div className="flex max-w-sm flex-col gap-8">
+      <p className="body1 text-osmoverse-400">{t("portfolio.connectWallet")}</p>
+      <Button
+        className="flex !h-11 w-fit items-center gap-2 !rounded-full !py-1"
+        onClick={() => {
+          onOpenWalletSelect({
+            walletOptions: [
+              { walletType: "cosmos", chainId: chainStore.osmosis.chainId },
```
jonator, line 22: We're considering `chainStore` to be deprecated; a part of the Keplr architecture we're trying to migrate away from. In the meantime, could we use the osmosis chain ID from the `accountStore`? I believe it's just `accountStore.osmosisChainId`

jonator (changes requested):
Since this is full screen only, could this button be hidden somehow?
Screenshot 2024-07-29 at 2 02 47 PM

Also, I noticed the loading animation is fake ghost-like chart data, but I noticed it doesn't take up the full width of the chart view port. Would be cleaner if we could get that to take up the viewport more naturally:
Screenshot 2024-07-29 at 2 03 58 PM

jonator (approved):

jonator:
Apologies for delayed approval

PR #3609, opened 2024-07-26
(Deposit/Withdraw) Reduce thrashing in loading

jonator:

Update UI only when all quotes are loaded, instead of the best quote being loaded while others are still loading. This causes less thrash in the UI since, as the best quote is being selected, the loading state does not constantly toggle on and off. This improves the UX by reducing the amount of changes in the UI.

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-864/amount-screen-improve-loading-ux)


jonator:
Going with a different strategy. We decided to allow the faster quotes to load and we will use that to update gas and/or allow the user to proceed

PR #3614, opened 2024-07-26
Flip USDC by default

mattupham:

- from: Can we make the execution price for trade history always default to the USDC one if one asset is USDC?

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-881/transaction-usdc-default)

## Brief Changelog

- if denominator is USDC (but not numerator) - flip the execution price by default

## Testing and Verifying

https://github.com/user-attachments/assets/601afcc5-8f69-4792-a641-92b52dfe3f84

### DIFF
--- a/packages/web/components/transactions/transaction-details/transaction-details-content.tsx
+++ b/packages/web/components/transactions/transaction-details/transaction-details-content.tsx
@@ -70,8 +70,8 @@ export const TransactionDetailsContent = ({
   // if USDC, toggle conversion - temporary until stablecoin logic is implemented
   useEffect(() => {
     if (
-      conversion.numerator.denom !== "USDC" &&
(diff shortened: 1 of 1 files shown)


packages/web/components/transactions/transaction-details/transaction-details-content.tsx
```diff
@@ -67,6 +67,16 @@ export const TransactionDetailsContent = ({
     );
   }, [conversion.numerator, conversion.denominator]);
 
+  // if USDC, toggle conversion - temporary until stablecoin logic is implemented
+  useEffect(() => {
```
jonator, line 71: Couldn't this order be determined in the tRPC procedure, also, the conversion rate could be calculated there as well?

yury-dubinin (approved):

PR #3617, opened 2024-07-26
(Deposit/Withdraw) Support Ethereum key signing

JoseRFelix:

Gives us support for Injective, Dymension, Evmos deposits

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-885/support-ethereum-key-signing)

## Testing and Verifying

- [ ] Can transfer Kava, Dymension, etc.


packages/stores/src/account/utils.ts
```diff
@@ -90,6 +90,32 @@ export function makeSignDocAmino(
...
+  chainFeatures,
+}: {
+  chainId: string;
+  chainFeatures: string[];
+}) {
+  const chainIsInjective = chainId.startsWith("injective");
+  const chainIsStratos = chainId.startsWith("stratos");
+  const useEthereumSign = chainFeatures.includes("eth-key-sign");
+
+  if (!useEthereumSign) {
```
jonator, line 104: Is this ! intentional?

JoseRFelix: Yes, this is the default typeUrl. All non `eth-key-sign` transactions require this

jonator (approved):

PR #3634, opened 2024-07-30
(Deposit/Withdraw) Crypto Fiat Improvements

JoseRFelix:

- Decimals shouldn't be reduced further if the input type is inactive. The same # of decimals should be visible for crypto input whether active or inactive
- If fiat is inactive, display <$0.01 if value of crypto input is worth <1¢. Similar to above, users can only input 2 decimals if fiat is active, but if crypto is active and it's valued less than 1¢, we should show an inequality symbol in the fiat input. This means if they switch to fiat input with a <1¢ crypto input, it would show ".00"

### Linear Task

[Linear Task URL](https://linear.app/osmosis/issue/FE-879/crypto-fiat-input-improvements)


packages/web/components/control/crypto-fiat-input.tsx
```diff
@@ -185,6 +207,8 @@ export const CryptoFiatInput: FunctionComponent<{
     canSetMax,
   ]);
 
+  console.log(fiatInputRaw);
```
jonator, line 210: console.log

jonator (approved):

jonator (changes requested):
IDK if your PR introduced this, but it appears max button is not setting balance anymore:
Screenshot 2024-07-30 at 6 03 35 PM

jonator (approved):

PR #3635, opened 2024-07-30
Use to SQS pools provider

jonator:

Now that SQS pools query returns capitalization (TVL), we can avoid the performance issue of sending additional SQS price queries to calculate it within the edge function. This should greatly increase performance and allow us to remove the Numia indexer as a major dependency here. As a secondary consequence, pool detail pages should be more performant since SQS supports querying by individual pool ID(s), which greatly reduces the load of getting individual pools (since all pools don't need to be queried)

### Linear Task

[FE-909 : Use SQS pools](https://linear.app/osmosis/issue/FE-909/use-sqs-pools)

## Brief Changelog

* Add liquidity capitalization to SQS pools response, and remove price queries to get those values

## Testing and Verifying

Tested manually locally

### DIFF
--- a/packages/server/src/queries/complex/pools/providers/sidecar.ts
+++ b/packages/server/src/queries/complex/pools/providers/sidecar.ts
@@ -21,18 +21,23 @@ const poolsCache = new LRUCache<string, CacheEntry>(DEFAULT_LRU_OPTIONS);
 export function getPoolsFromSidecar({
   assetLists,
   poolIds,
+  minLiquidityUsd,
 }: {
   assetLists: AssetList[];
   chainList: Chain[];
   poolIds?: string[];
+  minLiquidityUsd?: number;
 }): Promise<Pool[]> {
   return cachified({
     cache: poolsCache,
-    key: poolIds ? `sidecar-pools-${poolIds.join(",")}` : "sidecar-pools",
+    key:
+      (poolIds ? `sidecar-pools-${poolIds.join(",")}` : "sidecar-pools") +
+      minLiquidityUsd,
     ttl: 5_000, // 5 seconds
     getFreshValue: async () => {
       const sidecarPools = await timeout(
-        () => queryPools({ poolIds }),
+        () =>
+          queryPools({ poolIds, minLiquidityCap: minLiquidityUsd?.toString() }),
         9_000, // 9 seconds
         "sidecarQueryPools"
       )();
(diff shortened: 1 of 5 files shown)


p0mvn (commented):
Do you think we need pagination?

Could we also filter out the pools based on some huristic such as liquidity?

We probably don't need al pools all but serializing them all would be super expensive

jonator:
> Do you think we need pagination?

Like in SQS?

> Could we also filter out the pools based on some huristic such as liquidity?

Yeah. I'll filter them in the provider. Though it could be useful to have a min_liquidity flag as a param within SQS to avoid sending the data in the first place, as we filter out 0 liq pools in the edge function regardless

> We probably don't need al pools all but serializing them all would be super expensive

I think the above point will help with this, which I agree is an issue given how many pools there are now.

p0mvn:
> Like in SQS?

Yeah

> Yeah. I'll filter them in the provider. Though it could be useful to have a min_liquidity flag as a param within SQS to avoid sending the data in the first place, as we filter out 0 liq pools in the edge function regardless

Agreed to do it in SQS.

 I think we should add these abilities in SQS before launching to avoid performance degradations (filter point being more important) - do you want to collaborate on this?

Happy to deploy today/tomorrow if you make PRs

DavideSegullo (approved):

p0mvn (changes requested):
Blocking until we deplot this change:
https://github.com/osmosis-labs/sqs/pull/418

packages/server/src/queries/complex/pools/providers/sidecar.ts
```diff
@@ -20,20 +20,24 @@ const poolsCache = new LRUCache<string, CacheEntry>(DEFAULT_LRU_OPTIONS);
...
-    key: poolIds ? `sidecar-pools-${poolIds.join(",")}` : "sidecar-pools",
+    key:
+      (poolIds ? `sidecar-pools-${poolIds.join(",")}` : "sidecar-pools") +
+      minLiquidityUsd,
     ttl: 5_000, // 5 seconds
     getFreshValue: async () => {
       const sidecarPools = await timeout(
-        () => queryPools({ poolIds }),
+        () =>
+          queryPools({ poolIds, minLiquidityCap: minLiquidityUsd?.toString() }),
```
p0mvn, line 40: oh nice it's already here

ValarDragon:
The per pool page is so much faster! (I think its still working by the edge function querying all pools though, so probably still something to optimize more later)

This is a significant speedup!

packages/server/src/queries/complex/pools/providers/sidecar.ts
jonator: Yeah, and it should be back compat so we can release this now then once SQS is released, the var will start being respected on the SQS side

jonator:
@ValarDragon also, I'm anticipating the pool data to be more up to date than Numia, so hopefully the issues with adding/increasing CL positions will be alleviated with more up-to-date CL pool data

PR #3643, opened 2024-07-31
Delete notifi code to save bundle size

ValarDragon:
This deletes the Notifi code, saving us almost 10% of bundle size. (2.4MB compressed -> 2.2MB compressed)

### Linear Task

## Testing and Verifying

FE works as before


jonator (approved):
So glad I told them to put everything possible in integrations folder

Let's make sure any packages they used that we don't use area also removed from web/package.json to also speed up yarn install times

jonator:
Also, removing FF would be good (in use-feature-flags)

ValarDragon:
> So glad I told them to put everything possible in integrations folder

Great call on this!

PR #3653, opened 2024-08-02
(Deposit/Withdraw) Handle disabled assets

jonator:

This PR refactors our fallback to providing external URLs. It now checks for certain conditions for the selected from asset/chain (if withdrawing) or the to asset/chain (if depositing) to see if it's disabled, has no supported assets, or no quotable bridges, and shows the dedicated external URL UI if any of those conditions are met. This provides us with a nice fallback for users if any issues occur with an Osmosis asset that prevent it from being bridged via our integration, or if our providers don't generate any supported counterparty assets/chains, or if a counterparty asset/chain is not quotable since the only providers that returned that asset/chain do not have quoting implemented (wormhole, nomic).

I'm still seeing errors with wallets since I don't check if its disabled before we attempt to connect to wallets. I believe the errors are caught so users wouldn't be impacted by them, but, I could see us potentially checking if an asset is disabled before attempting to connect to wallets.

### Linear Task

[FE-922 : Use external methods UI if native integration not supported](https://linear.app/osmosis/issue/FE-922/use-external-methods-ui-if-native-integration-not-supported)

## Brief Changelog

- Added disabled check for the canonical asset in amount screen
- After we check if things are loading, we add a check for these conditions. If true, show the dedicated external URL UI
- In order for providers to offer more URLs, I made to/from asset/chains optional in the getExternalUrl interface
  - I could see an argument for not doing this at the provider level, but just in the getExternalUrl tRPC procedure. Then in the procedure we check if any of the params are missing and only query providers if they aren't missing. Then we can still fallback to the asset list external transfer methods since they don't fill in URLs from those params regardless.
- In more deposit options UI, added more localized strings to tolerate when the to/from asset/chain is not known

## Testing and Verifying

Tested locally


packages/types/src/asset-types.ts
```diff
@@ -191,7 +191,11 @@ export type MinimalAsset = {
    */
   coinGeckoId: string | undefined;
   coinName: string;
+  /** Transfers are allowed, but unstable. */
   isUnstable: boolean;
+  /** Transfers are NOT allowed. */
+  isDisabled: boolean;
```
JoseRFelix, line 197: What are your thoughts on renaming this to `areTransfersDisabled` instead? The term `isDisabled` is too broad and could imply that the asset can't be selected or skipped altogether

JoseRFelix (changes requested):

JoseRFelix (approved):

PR #3654, opened 2024-08-02
[Limit Orders - V3]: Small QA Round 2 Changes

crnbarr93:
These changes are for several small functionality changes as part of QA feedback.

## Brief Changelog
- Moved error display in buy/sell to above input
- Added autofocus input on swapping input type
- Removed gas fee estimation from trade details
- Display base denom value in buy/sell tab below inputs
- Added trade details state for no routes found
- Removed "Add Funds" state from buy/sell button
- Limited percentage input for price to 3 decimals


packages/web/components/swap-tool/trade-details.tsx
```diff
@@ -91,42 +82,15 @@ export const TradeDetails = observer(
...
-            "inline-flex items-center gap-1 text-osmoverse-100",
-            { "animate-pulse": isGasLoading }
-          )}
-        >
-          <Icon id="gas" width={16} height={16} />
-          {gasAmount && gasAmount.toString()}
-        </span>
-      );
-    }, [gasAmount, isGasLoading, gasFeeError, t]);
+    useEffect(() => {
```
DavideSegullo, line 85: We can avoid that useEffect, instead we can do the following things:

1) const [outAsBase, setOutAsBase] = useState(tab === "buy");
2) When we use `TradeDetails` component inside `PlaceLimitTool`, you can pass a prop to `TradeDetails` that is `key={tab}`, you already have the tab value on the parent

So we avoid side effects, and just re-render the component if the tab change, it's a suggested behaviour from react docs (for example for form reset)

packages/web/components/swap-tool/trade-details.tsx
```diff
@@ -362,8 +297,18 @@ export const TradeDetails = observer(
                               </GenericDisclaimer>
                               <div className="flex items-center gap-1 text-wosmongton-300">
                                 <span className="body2">
-                                  {routes?.length}{" "}
-                                  {routes?.length === 1 ? "route" : "routes"}
+                                  {!!routes && routes.length > 0 ? (
+                                    <>
+                                      {routes?.length}{" "}
+                                      {routes?.length === 1
+                                        ? "route"
```
DavideSegullo, line 304: Translations are missing

DavideSegullo (changes requested):

DavideSegullo (approved):

PR #3658, opened 2024-08-02
(Deposit/Withdraw) Update Deposit/Withdraw Toasts

JoseRFelix:

image

### Linear Task

https://linear.app/osmosis/issue/FE-918/update-depositwithdraw-toasts


packages/web/components/bridge/use-bridge-quotes.ts
```diff
@@ -422,23 +424,42 @@ export const useBridgeQuotes = ({
...
+        transferHistoryStore.pushTxNow({
+          prefixedKey: `${providerId}${JSON.stringify(params)}`,
+          amount: inputCoin.trim(true).toString(),
+          amountLogo: isWithdraw ? toAsset?.imageUrl : fromAsset.imageUrl,
           isWithdraw,
-          (isWithdraw ? fromAddress : toAddress) ?? "" // use osmosis account (destinationAddress) for account keys (vs any EVM account)
-        );
+          chainPrettyName:
+            direction === "deposit"
+              ? fromChain?.chainName ?? ""
```
jonator, line 444: I think if these are the display chain type, we should use the `prettyName` instead so we can have capitalization

packages/web/components/bridge/amount-and-review-screen.tsx
```diff
@@ -222,7 +224,9 @@ export const AmountAndReviewScreen = observer(
       },
     });
 
-    if (!selectedAssetDenom) return;
+    if (!selectedAssetDenom) {
+      throw new Error("Selected asset denom is undefined.");
```
jonator, line 228: We should be careful with this. Perhaps we just close the flow. This caused the page to crash when I went back/forward in the browser (and I believe the query params were updated)

packages/web/components/bridge/review-screen.tsx
```diff
@@ -227,9 +228,11 @@ const AssetBox: FunctionComponent<{
               height={16}
             />
           )}
-          <div className="text-wosmongton-200">
-            {shorten(address, { prefixLength: 12 })}
-          </div>
+          <Tooltip maxWidth="500px" content={address}>
```
jonator, line 231: especially since the cursor is pointer, what if we allow the user to copy the address when they click? Then the tooltip could be momentarily changed to "copy". 

If you don't want to bother, then lets remove the cursor pointer style

jonator (changes requested):

packages/web/components/bridge/amount-and-review-screen.tsx
JoseRFelix: This will show the error boundary and let the user restart the flow. I don’t think we should close the flow, as that could be unexpected and jarring for the user

jonator: I think I noticed that it showed the global fallback screen, at least in the preview. It was from going back or forward in the browser history (I don't remember which)

jonator: Try going to amount screen then going back in browser. Do you think we could set state directly instead of relying on exception handling for happy path functionality ?

Screenshot 2024-08-02 at 3 51 35 PM

jonator (approved):

PR #3660, opened 2024-08-02
Portfolio allocation

mattupham:

- add portfolio allocation for user

## Brief Changelog

- add trpc query
- add logic to format response
- add frontend component / breakdown
- add tabs component

## Testing and Verifying

https://github.com/user-attachments/assets/f731758d-87bc-449b-9b21-ca3fef1f4cac

https://github.com/user-attachments/assets/98db47af-6de2-4f5a-a705-225b2dda925f

### DIFF
--- a/packages/server/src/queries/complex/portfolio/__tests__/allocation.spec.ts
+++ b/packages/server/src/queries/complex/portfolio/__tests__/allocation.spec.ts
@@ -121,7 +121,8 @@ describe("Allocation Functions", () => {
       const result = await calculatePercentAndFiatValues(
         MOCK_DATA.categories,
         assetLists,
-        "total-assets"
+        "total-assets",
+        5
       ).map((allocation) => ({
         ...allocation,
         percentage: allocation.percentage.toString(),
@@ -156,7 +157,8 @@ describe("Allocation Functions", () => {
       const result = await calculatePercentAndFiatValues(
         MOCK_DATA.categories,
         assetLists,
-        "user-balances"
+        "user-balances",
+        5
       ).map((allocation) => ({
         ...allocation,
         percentage: allocation.percentage.toString(),


jonator (dismissed):
Initial review

packages/server/src/queries/data-services/index.ts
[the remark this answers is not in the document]
jonator: How is that related to the structure of the queries folder? To keep things organized and discoverable, IMO it's a good idea to co-locate all queries from sidecar in the sidecar folder

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,184 @@
...
+  assets: FormattedAllocation[];
+  available: FormattedAllocation[];
+}
+
+async function getAll(categories: Categories) {
+  const userBalancesCap = +categories["user-balances"].capitalization;
+  const stakedCap = +categories["staked"].capitalization;
+  const unstakingCap = +categories["unstaking"].capitalization;
+  const unclaimedRewardsCap = +categories["unclaimed-rewards"].capitalization;
+  const pooledCap = +categories["pooled"].capitalization;
```
jonator, line 32: I think it would be a good idea to Dec types to do calculations. It uses the `BigInt` type under the hood. It's a 1-1 mapping of the Dec type used by SQS.

I looked at the query response, and the capitalization responses contain 18 decimals, which may cause precision issues if used directly with normal JS number type.

Also, these Dec types are directly passable to CoinPretty, which would be a better more useful type to return the coin amounts as.

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,184 @@
...
+export async function getAllocation({
+  address,
+  assetLists,
+}: {
+  address: string;
+  assetLists: AssetList[];
+}): Promise<GetAllocationResponse> {
+  return await cachified({
+    cache: allocationCache,
+    ttl: 1000 * 60 * 0.25, // 15 seconds since a user can transact quickly
```
jonator, line 164: I'd avoid adding caching at all to this query. A main reason it was added to SQS was that it contained all user data and needed the cache miss performance gains. I think it would be a good idea to completely avoid the risk of old balances appearing after the queries are invalidated in the client after a transaction.

Then, they'd only later update if they become stale on the client, which could take a while.

packages/web/server/api/edge-router.ts
```diff
@@ -17,5 +18,6 @@ export const edgeRouter = createTRPCRouter({
   earn: earnRouter,
   transactions: transactionsRouter,
   chains: chainsRouter,
+  portfolio: portfolioRouter,
```
jonator, line 21: IMO, we should add this to the local router. Simply remove it from here and import and add to local-router.ts file in same folder.

There's no benefit, as far as I can tell, to first send a request to the edge since we don't want to cache user data. Instead, the user's browser will directly send a request to SQS, avoiding a hop through the edge function. 

It should notably increase client side performance.

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,184 @@
...
+  const assets = top5AccountCoinsResults.map((asset: AccountCoinsResult) => {
+    const assetFromAssetLists = getAsset({
+      assetLists,
+      anyDenom: asset.coin.denom,
+    });
+
+    return {
+      key: assetFromAssetLists.coinDenom,
+      percentage: (+asset.cap_value / totalCap) * 100,
+      amount: +asset.cap_value,
```
jonator, line 135: `amount` is normally used for crypto amounts, like `CoinPretty` objects.

Thoughts on calling it `fiatValue` so it's more clear that it's a USD value.

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,184 @@
...
+  const top5AccountCoinsResults = sortedAccountCoinsResults.slice(0, 5);
+
+  const assets = top5AccountCoinsResults.map((asset: AccountCoinsResult) => {
+    const assetFromAssetLists = getAsset({
+      assetLists,
+      anyDenom: asset.coin.denom,
+    });
+
+    return {
+      key: assetFromAssetLists.coinDenom,
```
jonator, line 133: Thoughts on coinDenom as the name of this member, to better indicate that it's the denom of a coin?

packages/server/src/queries/complex/portfolio/allocation.ts
jonator, line 135: It could simplify all future client side logic if we directly return `percentage` as `RatePretty` and `fiatValue` as `PricePretty`

Later, we may want a server side mechanism to convert prices to other fiat currencies.

Also, the client side implementation is simpler.

In general, from my experience the more complexity we can push to the server, the better.

jonator (changes requested):
Added a review. Mostly around increasing consistency with existing server logic, and reducing client side complexity.

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,184 @@
...
+    (sum: number, asset: AccountCoinsResult) => sum + +asset.cap_value,
+    0
+  );
+  const otherPercentage = (otherAmount / totalCap) * 100;
+
+  const other = {
+    key: "Other",
+    percentage: otherPercentage,
+    amount: otherAmount,
+    color: "bg-osmoverse-500",
```
jonator, line 107: Thoughts on only having client UI logic in client? For mobile, we may not use tailwind, so, in that sense to me only data should be returned from server package. We should avoid leaking client side logic/handling.

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    return {
+      key: assetFromAssetLists.coinDenom,
+      percentage: (+asset.cap_value / totalCap) * 100,
+      amount: +asset.cap_value,
+    };
+  });
+
+  const otherAssets = sortedAccountCoinsResults.slice(5);
+  const otherAmount = otherAssets.reduce(
```
jonator, line 97: This may be a logical error. Why are we slicing the other coins then reducing that into an amount that we use to calculate %?

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,184 @@
...
+    getFreshValue: async () => {
+      const data = await queryAllocation({
+        address,
+      });
+
+      const categories = data.categories;
+
+      const all = await getAll(categories);
+      const assets = await getAssets(categories, assetLists);
+      const available = await getAvailable(categories, assetLists);
```
jonator, line 175: Nit: I don't think these need to be async functions. Could clean things up a bit.

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,184 @@
...
+    },
+    {
+      key: "pooled",
+      percentage: (pooledCap / totalCap) * 100,
+      amount: pooledCap,
+    },
+  ];
+}
+
+async function getAssets(categories: Categories, assetLists: AssetList[]) {
```
jonator, line 70: It seems these functions are highly similar.

Thoughts on only keeping one of these and calling it something like `calcCategoryPercentages`? Then, it could take in another parameter that is "user-balances" or "total-assets" to determine which one to calculate for.

packages/server/src/queries/complex/portfolio/allocation.ts
jonator, line 1: Let's create some tests with mocked response data to ensure the calculations are correct.

If I were you, I'd use chatGPT to generate the initial tests to save time. It's a huge time save.

jonator (changes requested):
Another review, moreso around logic, testing, and reducing code redundancy

jonator (commented):
One other thing I noticed

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -0,0 +1,152 @@
...
+        percentage: new RatePretty(new Dec(asset.cap_value).quo(totalCap)),
+        fiatValue: new PricePretty(
+          DEFAULT_VS_CURRENCY,
+          new Dec(asset.cap_value)
+        ),
+      };
+    }
+  );
+
+  const otherAssets = sortedAccountCoinsResults.slice(5);
```
jonator, line 103: NIT: let's try to avoid hardcoding with magic numbers. Thoughts on creating a number parameter on function perhaps called allocationLimit or something, that is defaulted to 5? Then it could be passed to both slices

jonator (approved):

PR #3666, opened 2024-08-03
implement getNativeAssetPrice

p0mvn:
This PR implements `getNativeAssetPrice`, which retrieves prices from the sidecar with no other fallbacks.

This use case helps decouple 90% of the native chain pricing needs while retaining the original abstraction for bridging needs.

## Testing

Upon approval, I will request QA.


packages/server/src/queries/complex/assets/price/index.ts
```diff
@@ -12,20 +12,39 @@ import { getPriceFromSidecar } from "./providers/sidecar";
...
+    throw new Error("coinDenom is required");
+  }
+
+  return cachified({
+    key: `asset-price-nat-${coinMinimalDenom}`,
+    cache: pricesCache,
+    ttl: 1000 * 10, // 10 seconds
+    getFreshValue: () => getPriceFromSidecar(coinMinimalDenom),
+  });
+}
```
p0mvn, line 35: @jonator instead of doing the factory pattern that we discussed, I simply introduced another helper.

PR #3667, opened 2024-08-04
fix: frontend bug breaking weighted LP

p0mvn:
FE used to depend on amino encoding of pools.

But SQS returns app encoding.

The difference wasn't accounted for in the transition.

This PR fixes the bug around inconsistency => we need to do a proper move away from amino

### DIFF
--- a/packages/pools/src/weighted.ts
+++ b/packages/pools/src/weighted.ts
@@ -133,7 +133,7 @@ export class WeightedPool implements SharePool, RoutablePool {
 
   /** LBP pool */
   get smoothWeightChange(): SmoothWeightChangeParams | undefined {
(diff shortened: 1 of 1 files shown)


packages/pools/src/weighted.ts
```diff
@@ -133,7 +133,7 @@ export class WeightedPool implements SharePool, RoutablePool {
 
   /** LBP pool */
   get smoothWeightChange(): SmoothWeightChangeParams | undefined {
-    if (this.raw.pool_params.smooth_weight_change_params !== null) {
```
p0mvn, line 136: Currently returns "undefined" and panics here due to strict null check

czarcas7ic (approved):
utACK

CryptoAssassin1 (approved):

PR #3673, opened 2024-08-05
[Limit Price - V3]: Price Input Default Values

crnbarr93:
Reworked limit price input to function as follows:
- Default input is price
- Price will update with spot price updating
- If user inputs a value price no longer updates
- If user swaps to percentage order price adjusts with spot price and percentage input
- If user then swaps back to price input the price no longer updates


packages/web/components/place-limit-tool/limit-price-selector.tsx
```diff
@@ -45,32 +47,54 @@ export const LimitPriceSelector: FC<LimitPriceSelectorProps> = ({
...
-      const percentAdjusted = manualPrice
-        .quo(priceState.spotPrice)
-        .sub(new Dec(1))
-        .mul(new Dec(100));
+    if (inputMode === InputMode.Price && !priceState.priceLocked) {
+      const formattedPrice = formatPretty(
+        priceState.priceFiat,
+        getPriceExtendedFormatOptions(priceState.priceFiat.toDec())
+      )
+        .replace("$", "")
```
DavideSegullo, line 70: Can we do it in a single regex?

packages/web/components/place-limit-tool/limit-price-selector.tsx
```diff
@@ -45,32 +47,54 @@ export const LimitPriceSelector: FC<LimitPriceSelectorProps> = ({
...
+    }
 
-      priceState._setPercentAdjustedUnsafe(
-        percentAdjusted.isZero() || manualPrice.isZero()
-          ? ""
-          : formatPretty(percentAdjusted.abs(), {
-              maxDecimals: 3,
-            }).toString()
+    if (inputMode === InputMode.Percentage) {
+      console.log(
```
DavideSegullo, line 77: Do we need it for prod?

DavideSegullo (changes requested):

packages/web/components/place-limit-tool/limit-price-selector.tsx
crnbarr93: Good spot, was using it to ensure the price was updating (it can take a while)

DavideSegullo (approved):

PR #3680, opened 2024-08-06
switch pool aprs and fees to sqs

p0mvn:

Pool APRs and fees are now available in SQS. As a result, we are migrating and simplifying the tRPC accordingly.

We don't expect this to bring an immediate and direct improvement. The improvement stems from reducing the caching/data aggregation complexity as well as eliminating the blocking behavior on cache misses. 

## Testing

- Tested manually
- Requesting QA of stage as a whole


packages/server/src/queries/complex/pools/providers/sidecar.ts
```diff
@@ -86,9 +103,117 @@ function makePoolFromSidecarPool({
       DEFAULT_VS_CURRENCY,
       sidecarPool.liquidity_cap
     ),
+    ...data,
+  };
+}
+
+function getMarketIncentivesData(
```
p0mvn, line 110: nit: docs

packages/server/src/queries/complex/pools/providers/sidecar.ts
```diff
@@ -1,43 +1,57 @@
...
   return cachified({
     cache: poolsCache,
     key:
       (poolIds ? `sidecar-pools-${poolIds.join(",")}` : "sidecar-pools") +
-      minLiquidityUsd,
-    ttl: 5_000, // 5 seconds
+      minLiquidityUsd +
+      withMarketIncentives.toString(),
+    ttl: 6_000, // 6 seconds
+    staleWhileRevalidate: 3_500, // 3.5 seconds
```
p0mvn, line 46: Do you recall if this is safe to do in the edge environment? @jonator 

I remember us having issues with the flag in edge but unsure if it was because Imperator was taking overly long to load

packages/server/src/queries/sidecar/pools.ts
```diff
@@ -48,6 +48,38 @@ export type ChainPool =
   | ChainConcentratedPool
   | ChainCosmwasmPool;
 
+export type SQSAprData = {
```
p0mvn, line 51: nit: docs

packages/server/src/queries/sidecar/pools.ts
```diff
@@ -48,6 +48,38 @@ export type ChainPool =
...
+    upper?: number;
+    lower: number;
+  };
+  total_apr: {
+    upper?: number;
+    lower?: number;
+  };
+};
+
+export type SQSPoolFeesData = {
```
p0mvn, line 74: nit: docs

packages/server/src/queries/sidecar/pools.ts
```diff
@@ -74,7 +114,13 @@ export function queryPools({
...
     params.append("min_liquidity_cap", minLiquidityCap);
   }
+  if (withMarketIncentives) {
+    params.append("with_market_incentives", "true");
+  }
 
   url.search = params.toString();
+
+  console.log({ s: url.search });
+
```
p0mvn:
```suggestion
```

packages/web/components/complex/pools-table.tsx
```diff
@@ -180,17 +181,19 @@ export const PoolsTable = (props: PropsWithChildren<PoolsTableProps>) => {
     [poolsPagesData]
   );
 
+  console.log(poolsData[0]);
+
```
p0mvn:
```suggestion
```

p0mvn (commented):
@jonator 

I can help with addressing the nits shortly.

My biggest open question is whether stale while revalidate is safe in edge.

Also, I think that we can set the flag to always query with market incentives data by default per your recommendation.

packages/server/src/queries/complex/pools/providers/sidecar.ts
```diff
@@ -1,43 +1,56 @@
...
   minLiquidityUsd?: number;
+  withMarketIncentives?: boolean;
 }): Promise<Pool[]> {
   return cachified({
     cache: poolsCache,
     key:
       (poolIds ? `sidecar-pools-${poolIds.join(",")}` : "sidecar-pools") +
-      minLiquidityUsd,
+      minLiquidityUsd +
+      withMarketIncentives.toString(),
```
p0mvn, line 44: We should consider refactoring this cache longer-term.

Ref:
https://osmosis-network.slack.com/archives/C07FJ04P013/p1723937101276429

jonator (approved):

p0mvn:
Discussed offline with Jon and agreed to merge given that both of us tested + we will request QA of stage as a whole

PR #3702, opened 2024-08-08
(Deposit/Withdraw) Support trx and trx.rt tokens

JoseRFelix:

TRX.rt and Alloyed TRX (as canonical TRX) have been listed. This PR adds a provider to redirect users to an external bridge UI so users can deposit/withdraw these assets. 

### Linear Task

https://linear.app/osmosis/issue/FE-954/support-trx-and-trxrt

## Brief Changelog

- Add Nitro bridge provider
- Cleanup unsupported quote bridges logic

## Testing and Verifying

- [ ] Tron assets (USDT, TRX, TRX.rt) redirect to external interfaces to bridge


packages/bridge/src/nitro/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  static readonly ID = "Nitro";
+  readonly providerName = NitroBridgeProvider.ID;
+
+  constructor(protected readonly ctx: BridgeProviderContext) {}
+
+  async getQuote(): Promise<BridgeQuote> {
+    throw new Error("Nitro quotes are currently not supported.");
+  }
+
+  async getSupportedAssets({
```
jonator, line 32: Should we catch and warn and return [] incase anything unexpected happens? Or, will client just tolerate that

packages/bridge/src/nitro/index.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    if (toChain) {
+      url.searchParams.set("toChain", String(toChain.chainId));
+    }
+
+    const setTokenParam = (
+      asset: NonNullable<
+        | GetBridgeExternalUrlParams["fromAsset"]
+        | GetBridgeExternalUrlParams["toAsset"]
```
jonator, line 110: Nit: what if we just use BridgeAsset type directly?

jonator (approved):

PR #3719, opened 2024-08-09
(Datadog) Add OTEL instrumentation

JoseRFelix:

Enable tracing in Datadog

### Linear Task

https://linear.app/osmosis/issue/PLAT-323/vercel

## Brief Changelog

- Remove Sentry
- Add tracing in trpc


p0mvn (commented):
How does trace id header propagation to servers work?

For example, wouldn't we need to set headers explicitly when making a request to SQS?

packages/web/instrumentation.ts
```diff
@@ -0,0 +1,14 @@
...
+import { getOpentelemetryServiceName } from "~/utils/service-name";
+
+export function register() {
+  registerOTel({
+    serviceName: getOpentelemetryServiceName(),
+    instrumentationConfig: {
+      fetch: {
+        propagateContextUrls: ["*"],
+      },
+    },
```
JoseRFelix, line 12: @p0mvn Otel will automatically propagate the context on fetch requests. However, we have to make sure the service is compatible with the header

packages/web/utils/error.ts
```diff
@@ -1,8 +1,6 @@
-import * as Sentry from "@sentry/nextjs";
-
 export function captureError(e: any) {
   if (e instanceof Error) {
-    Sentry.captureException(e);
+    // Sentry.captureException(e);
```
jonator, line 3: Q: will exceptions be captured some other way?

jonator (approved):

packages/web/utils/error.ts
JoseRFelix: Yes, we'll use Opentelemetry. I updated the one on the `server` package but missed this one. Good catch! My latest commit fixes this

PR #3734, opened 2024-08-13
[Limit Orders]: Query Hotfixes

crnbarr93:
These changes include some small limit order hotfixes:
- Reworded claim and close to just close when all claimable amount of a partially filled order is claimed
- Updated balance query for `usePlaceLimit`
- Wired denom selection search in to marketstate for buy/sell tabs


packages/web/hooks/limit-orders/use-place-limit.ts
```diff
@@ -378,33 +378,39 @@ export const usePlaceLimit = ({
...
-        select: (balances) =>
-          balances.filter(
-            ({ denom }) =>
-              denom === baseAsset?.coinMinimalDenom ||
-              denom === quoteAsset?.coinMinimalDenom
+        select: ({ items }) =>
+          items.filter(
+            ({ coinMinimalDenom }) =>
+              coinMinimalDenom === baseAsset?.coinMinimalDenom ||
+              coinMinimalDenom === quoteAsset?.coinMinimalDenom
```
JoseRFelix, line 392: Maybe it's a good idea to use toLowerCase to avoid case-sensitive mismatches. Thoughts?

crnbarr93: For both sides of the equality check?

Don't see much downside personally and should be safer

JoseRFelix (approved):

PR #3749, opened 2024-08-16
Add Transmuter Contract Codes

JohnnyWyles:
Adding the 3 other contract uploads for Transmuter pools so that these can be filtered out in the pools list if wanted.

Transmuter 1.0 already listed
Transmuter 2.0 was never instantiated
This adds 3.0, 3.1 and 3.2 which are the alloyed asset pools - but also are transmuters and often dominate the volume charts.

### DIFF
--- a/packages/server/src/queries/complex/pools/env.ts
+++ b/packages/server/src/queries/complex/pools/env.ts
@@ -1,7 +1,7 @@
 import { IS_TESTNET } from "../../../env";
 
 /** Cosmwasm Code Ids confirmed to be transmuter pools in current env. */
-const TransmuterPoolCodeIds = IS_TESTNET ? ["3084"] : ["148"];
+const TransmuterPoolCodeIds = IS_TESTNET ? ["3084"] : ["148", "814", "867", "996"];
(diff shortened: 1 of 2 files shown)


jonator (changes requested):
I believe this is connected to legacy query code.

It's preferred to use this now:

https://github.com/osmosis-labs/osmosis-frontend/blob/2291cffffb5f5ef2a41ad747bd51ae6072a2becd/packages/server/src/queries/complex/pools/env.ts#L4

jonator (approved):

jonator (commented):
@JohnnyWyles if we could add `// @deprecated use packages/server/src/queries/complex/pools/env.ts` comment above the old one, that would help prevent this confusion in the future

MaxMillington:
@JohnnyWyles is there any update on this? need any help?

JohnnyWyles:
> @JohnnyWyles is there any update on this? need any help?

I think I did the change that Jon requested?
Not sure what's missing from this

JohnnyWyles:
Please let me know what I'm missing here - I added a comment and the new code IDs to the alternative location as recommended

EnricoBarbieri1997 (approved):

DavideSegullo (approved):

PR #3754, opened 2024-08-18
move legacy pools query to local from edge

p0mvn:

Moves legacy pools query from edge to local

### DIFF
--- a/packages/web/pages/api/pools/[id].ts
+++ b/packages/web/pages/api/pools/[id].ts
@@ -50,5 +50,5 @@ export default async function pools(req: Request) {
 }
 
 export const config = {
-  runtime: "edge",
+  runtime: "local",
 };


jonator (changes requested):
Note this may still be in use by legacy mobx pool query stores (for individual pool vs all pools)

So we should remove all uses of those before we can merge this

https://github.com/osmosis-labs/osmosis-frontend/blob/683a00e7e6420e307af7d8dfaee1152a656c6aba/packages/stores/src/queries-external/pools/pool.ts#L340

p0mvn:
I see - but we can probably move it to local from edge for now

p0mvn:
@jonator restored query and moved to local from edge

jonator (changes requested):
I don't believe this is going to move the runtime to the browser (as is possible with tRPC)

This config export is simply a config used by _Vercel_ to specify the runtime (serverless or edge) to be used when deploying this edge function. However, I'm not aware of Next.js supporting this feature as a way to move a query into browser. If so, that's news to me.

p0mvn:
> I don't believe this is going to move the runtime to the browser (as is possible with tRPC)
> 
> This config export is simply a config used by _Vercel_ to specify the runtime (serverless or edge) to be used when deploying this edge function. However, I'm not aware of Next.js supporting this feature as a way to move a query into browser. If so, that's news to me.

I see - thanks for breaking it down. Let's then wait for portfolio v2 launch to remove the last use of it so that we can remove it in general

PR #3755, opened 2024-08-18
Move getPool to local router from edge

p0mvn:

Given that we rely on SQS pools, we can leverage the SQS regional co-location with the FE. As a result, we begin moving the pools router from edge to local.

This move will also allow us to eventually reduce complexity of some of the complex pool tRPCs.

For scope limitation reasons, we move tRPC by tRPC as to minimize the impact of accidental regressions..


jonator:
If we branch from `roman/switch-aprs-sidecar` perhaps we'll have less conflicts?

p0mvn:
> If we branch from `roman/switch-aprs-sidecar` perhaps we'll have less conflicts?

I think we should QA that change before merging. However, this can be merged without QA upon approval.

I propose not blocking this PR on QA of the APRs - wdyt?

I will resolve conflicts

jonator:
@p0mvn we renamed the getMarketIncentivePools procedure to just getPools once that data became available in SQS. We'll prob want to migrate that to local as well. IMO this PR is low prio enough to allow it to be blocked by that PR as it gets QA'd, it's in the same class of changes anyways.

packages/trpc/src/pools.ts
```diff
@@ -149,3 +143,13 @@ export const poolsRouter = createTRPCRouter({
       )
     ),
 });
+
+// Local router queries for pools.
+export const poolsLocalRouter = createTRPCRouter({
+  getSharePool: publicProcedure
+    .input(z.object({ poolId: z.string() }))
+    .query(({ input: { poolId }, ctx }) => getSharePool({ ...ctx, poolId })),
```
p0mvn, line 151: Had to move `getSharePool` as well since it depended on `getPool`

p0mvn:
> @p0mvn we renamed the getMarketIncentivePools procedure to just getPools once that data became available in SQS. We'll prob want to migrate that to local as well. IMO this PR is low prio enough to allow it to be blocked by that PR as it gets QA'd, it's in the same class of changes anyways.

We merged the APRs so this should be unblocked now

JoseRFelix (approved):

JoseRFelix:
@p0mvn Is this PR still relevant?

jonator (approved):

PR #3756, opened 2024-08-18
Remove some legacy pools query usage

jonator:
Removes some uses of legacy pools query in CL flow.

Remaining uses
* Assets (now called portfolio) page v1: for filtering pools by show dust setting
*


p0mvn (approved):

p0mvn:
Tested that adding and removing CL liquidity works from this preview.

Will request more QA of stage - merging

PR #3757, opened 2024-08-18
Unwire the legacy router

p0mvn:

- Unwire the legacy router
- Always return default Sidecar router
- Remove indexer pools
- Remove old feature flags

Next PR:
- Remove the actual implementation of the legacy router


packages/server/src/queries/complex/swap-routers.ts
jonator, line 1: IMO, we should go further and just call the SQS out-given-in query function and not even conform to the `TokenOutGivenInRouter` interface. We should be able to delete the pools package entirely.

jonator (approved):

packages/server/src/queries/complex/swap-routers.ts
p0mvn: OK I will try

kamal-sutra (approved):

jonator (approved):
This is a good cleanup step.

Without spending tons of time digging into the code, I think the next step to evetually delete the pools package (or help us be ready to) would be to copy over any imports directly from the pools package into /packages/server/src/queries/sidecar/router.ts (like error types, the router interface, etc).

Basically removing references into the pools package. Then when the tRPC migration is complete and all legacy pools queries are deleted, we can just delete that package in entirety.

p0mvn:
> This is a good cleanup step.
> 
> Without spending tons of time digging into the code, I think the next step to evetually delete the pools package (or help us be ready to) would be to copy over any imports directly from the pools package into /packages/server/src/queries/sidecar/router.ts (like error types, the router interface, etc).
> 
> Basically removing references into the pools package. Then when the tRPC migration is complete and all legacy pools queries are deleted, we can just delete that package in entirety.

Thanks for suggestions - tracked here: https://github.com/osmosis-labs/osmosis-frontend/issues/3764

p0mvn:
Merging - appreciate the reviews

PR #3760, opened 2024-08-18
Added search for portfolio tests

yury-dubinin:

Due to portfolio page change added a search for a token for 2 reasons:
- make sure search works
- make sure token gets to a view

### DIFF
--- a/packages/web/e2e/pages/portfolio-page.ts
+++ b/packages/web/e2e/pages/portfolio-page.ts
@@ -9,6 +9,7 @@ export class PortfolioPage extends BasePage {
   readonly viewMore: Locator;
   readonly portfolioLink: Locator;
   readonly viewTransactions: Locator;
+  readonly searchInput: Locator;
 
   constructor(page: Page) {
     super(page);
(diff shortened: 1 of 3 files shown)


packages/web/e2e/pages/portfolio-page.ts
```diff
@@ -61,4 +64,10 @@ export class PortfolioPage extends BasePage {
       await this.page.waitForTimeout(1000);
     }
   }
+
+  async searchForToken(tokenName: string) {
+    await this.searchInput.fill(tokenName);
+    // we expect that after 2 seconds tokens are loaded and any failure after this point should be considered a bug.
+    await this.page.waitForTimeout(2000);
```
p0mvn, line 71: Would there be a clear feedback about why the test failed? If not, is there anywhere we could add a comment or a log?

packages/web/e2e/tests/portfolio.wallet.spec.ts
```diff
@@ -58,33 +58,46 @@ test.describe("Test Portfolio feature", () => {
   });
 
   test("User should be able to see native balances", async () => {
+    await portfolioPage.searchForToken("OSMO");
```
p0mvn, line 61: Nit: we could probably put these in a loop to minimize cognitive overhead stemming from code duplication

p0mvn (approved):
LGTM - thank you

packages/web/e2e/tests/portfolio.wallet.spec.ts
yury-dubinin: agree, I just need to do it. Moved to parameteried tests approach.

PR #3772, opened 2024-08-20
Use timeout_height from signed amino doc

sam-utila:

In institutional cases, the signature process may take a while.

This is why we override:

```
  sequence: signedSequence
```

Since maybe the sequence number has changes since the dApp sent its requested sign doc to the wallet - the wallet may update the sequence number to a new value, and the dApp should use the returned value - which was signed on.

The same reasoning should follow for `timeout_height` which is a newer addition to the protocol than sequence number.

### Linear Task

N/A

## Brief Changelog

Use timeout_height from signed amino doc

## Testing and Verifying

This change has been tested locally using our own extension wallet that modifies timeout height

### DIFF
--- a/packages/stores/src/account/base.ts
+++ b/packages/stores/src/account/base.ts
@@ -1111,7 +1111,7 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
         return res;
       }),
       memo: signed.memo,
-      timeoutHeight: BigInt(signDoc.timeout_height ?? "0"),
+      timeoutHeight: BigInt(signed.timeout_height ?? "0"),
     });
 
     const signedGasLimit = Int53.fromString(String(signed.fee.gas)).toNumber();


jonator:
@sam-utila how did you notice this?

sam-utila:
Basically timeout_height bytes are part of the `body` not the `auth_info` (where `sequence` is located) - and generally they should not be touched by the wallet.

But `memo` is also in `body`.

So if there are multiple signers for the transactions - the wallet is not allowed to touch `memo` or `timeout_height`.

However - if I'm the only signer - why not?
I want to update the `timeout_height` to be something close to the point of the actual signing of the transaction.
In institutional wallets (like Utila) the TX may live for a while before it is approved and signed - so editing the timeout_height should be okay.

kamal-sutra:
@sam-utila can you update to the latest branch of stage please

PR #3773, opened 2024-08-20
In Given Out

crnbarr93:
These changes add in-given-out functionality to the trade modal.

### Linear Task

[LIM-212](https://linear.app/osmosis/issue/LIM-212/fe-exact-amount-out-implemention)

## Brief Changelog

- Added support for "in-given-out" quotes to the `useSwap` hook
- Added implementations for `MsgSwapExactAmountOut`
- Added a feature flag for "in-given-out"
- Added a dynamic slippage handler for "in-given-out"


packages/stores/src/account/osmosis/index.ts
```diff
@@ -1205,6 +1206,80 @@ export class OsmosisAccountImpl {
...
+              //     tokenIn.currency.coinMinimalDenom ||
+              //   routes
+              //     .flatMap(({ pools }) => pools)
+              //     .find(
+              //       (pool) =>
+              //         pool.tokenOutDenom === bal.currency.coinMinimalDenom
+              //     )
+              // ) {
+              //   bal.waitFreshResponse();
+              // }
```
JoseRFelix, line 1267: If we want to delete this, we need to determine in this PR whether it's necessary. We should test it with the pool details page. Otherwise, let's leave it uncommented

packages/web/components/place-limit-tool/index.tsx
```diff
@@ -201,55 +217,85 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
...
+          ? swapState.marketState.quote
+            ? new IntPretty(
+                swapState.marketState.quote.amount.toDec().mul(oneMinusSlippage)
+              )
+            : undefined
+          : new IntPretty(
+              swapState.marketState.inAmountInput?.amount
+                ?.toDec()
+                .mul(onePlusSlippage) ?? new Dec(0)
+            );
```
JoseRFelix, line 237: Nested ternary operators make it complex to read this variable. Thoughts on using a self-calling function with if statements to improve this?
```suggestion
      const amountWithSlippage = (() => {
  if (quoteType === "out-given-in") {
    return swapState.marketState.quote
      ? new IntPretty(
          swapState.marketState.quote.amount.toDec().mul(oneMinusSlippage)
        )
      : undefined;
  } else {
    return new IntPretty(
      swapState.marketState.inAmountInput?.amount
        ?.toDec()
        .mul(onePlusSlippage) ?? new Dec(0)
    );
  }
})();
```

packages/web/components/place-limit-tool/index.tsx
```diff
@@ -426,6 +503,72 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
...
+          );
+    }, [
+      focused,
+      swapState.marketState.inAmountInput.inputAmount,
+      swapState.marketState.outAmountInput.inputAmount,
+      type,
+      tab,
+      fiatAmount,
+      swapState.inAmountInput.inputAmount,
+    ]);
```
JoseRFelix, line 547: Same as above
```suggestion
    const nonFocusedDisplayAmount = useMemo(() => {
  const formatInputAsPrice = (inputAmount: string | undefined) =>
    formatFiatPrice(
      new PricePretty(DEFAULT_VS_CURRENCY, inputAmount || "0")
    );

  const getTrimmedAmount = (inputAmount: string | undefined) =>
    trimPlaceholderZeros(inputAmount ?? "") ?? "";

  const handleMarketTab = () => {
    if (tab === "sell") {
      return focused === "fiat"
        ? getTrimmedAmount(swapState.marketState.inAmountInput.inputAmount)
        : formatInputAsPrice(swapState.marketState.outAmountInput.inputAmount);
    } else {
      return focused === "fiat"
        ? getTrimmedAmount(swapState.marketState.outAmountInput.inputAmount)
        : formatInputAsPrice(swapState.marketState.inAmountInput.inputAmount);
    }
  };

  if (type === "market") {
    return handleMarketTab();
  } else {
    return focused === "fiat"
      ? swapState.inAmountInput.inputAmount
      : formatInputAsPrice(fiatAmount);
  }
}, [
  focused,
  swapState.marketState.inAmountInput.inputAmount,
  swapState.marketState.outAmountInput.inputAmount,
  type,
  tab,
  fiatAmount,
  swapState.inAmountInput.inputAmount,
]);

```

JoseRFelix (changes requested):

packages/web/components/place-limit-tool/index.tsx
crnbarr93: I refactored this for some changes and added it to a hook in `useSwap`.

packages/web/hooks/use-swap.tsx
```diff
@@ -1124,45 +1461,81 @@ function useQueryRouterBestQuote(
...
+
+            // prevent batching so that fast routers can
+            // return requests faster than the slowest router
+            trpc: {
+              context: {
+                skipBatch: true,
+              },
+            },
+          }
+        );
```
JoseRFelix, line 1538: It goes against the rule of hooks to conditionally render hooks. Is this working?

JoseRFelix (changes requested):

packages/web/hooks/use-swap.tsx
crnbarr93: It is, what's the advised solution here?

crnbarr93: For reference the call to this hook is duplicated, one for each quote type.

JoseRFelix: If this is the case, I suggest modifying the params themselves using a ternary operator

JoseRFelix (approved):

PR #3784, opened 2024-08-22
[In Given Out]: Default input adjustments

crnbarr93:
These changes fix a bug with placing limit orders if the market input is empty. The default input for the sell tab is now fiat and the available balance display changes with the input selected.

### DIFF
--- a/packages/web/components/place-limit-tool/index.tsx
+++ b/packages/web/components/place-limit-tool/index.tsx
@@ -125,13 +125,15 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
 
     const setBase = useCallback((base: string) => set({ from: base }), [set]);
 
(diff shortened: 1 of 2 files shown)


packages/web/components/place-limit-tool/index.tsx
```diff
@@ -117,20 +118,22 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
     const [isSendingTx, setIsSendingTx] = useState(false);
 
     const [focused, setFocused] = useState<"fiat" | "token">(
-      tab === "buy" ? "fiat" : "token"
+      !featureFlags.inGivenOut && tab === "sell" ? "token" : "fiat"
```
MaxMillington, line 121: Nit: it would probably be better for these to be constants

MaxMillington (approved):

PR #3788, opened 2024-08-22
fix: remove logs and correct limit input amount

crnbarr93:
These changes correct the input value for limit orders on the buy tab


p0mvn (approved):
Reviewed offline with Connor

PR #3801, opened 2024-08-28
sqs pools provider: handle empty poolIDs array

jonator:
### Linear Task

We were passing user pool IDs to the SQS pools provider to only fetch the pools for the user.

However, if the user had no pools this would cause an empty array to be passed to SQS pools provider.

The problem arose because SQS would return all pools since the pool IDs list would be empty. This would cause pools to be returned as if they were user pools.

## Brief Changelog

- Explicitly return empty array if poolIDs array is present with no elements

## Testing and Verifying

Tested locally


ValarDragon (approved):
Yikes, thanks for fixing these

PR #3803, opened 2024-08-28
[Trade Tool]: Small QA Fixes

crnbarr93:
These changes adjust the following for the buy/sell tab:
- Non focused token amount restricted to 10 decimals (only when non-focused)
- Sell tab fiat input now rounds up to avoid payment fiat value not matching fiat input amount for limit orders


yury-dubinin (approved):
lgtm

PR #3808, opened 2024-08-29
fix: limit toast arrow button action

crnbarr93:
These changes fix the following:
- Clicking the toast arrow button did not redirect the user
- The close button background was incorrect


MaxMillington (approved):
lgtm

PR #3828, opened 2024-09-05
[Limit Orders]: SQS Active Order Query

crnbarr93:
These changes are primarily aimed at moving the active orders query to SQS via a passthrough endpoint. Overall this includes the following:
- A new `orderbookRouter` query for `getAllOrdersSQS` which retrieves active orders from the provided SQS endpoint
- Moving the `orderbookRouter` to be local as it no longer uses node queries
- Updates to caching lengths for order queries
- An additional check to see if claimable orders need to be queried
- Claimable orders now uses the same query as active orders with an additional filter

**NOTE: These changes need to be load tested and SHOULD NOT BE MERGED to `master` until verified**


packages/server/src/queries/complex/orderbooks/active-orders.ts
```diff
@@ -16,6 +17,46 @@ import type { MappedLimitOrder, OrderStatus } from "./types";
...
+  userOsmoAddress,
+  assetList,
+}: {
+  userOsmoAddress: string;
+  assetList: AssetList[];
+}) {
+  return cachified({
+    cache: activeOrdersCache,
+    key: `orderbookActiveOrders-sqs-${userOsmoAddress}`,
+    ttl: 5000, // 5 seconds
```
jonator, line 30: I have a personal rule of never caching user account queries just to be safe. But, it's a bit extreme and I think it's fine with low ttl queries

jonator (approved):

packages/server/src/queries/complex/orderbooks/active-orders.ts
crnbarr93: I think we want to be strict with this query in terms of refetch/cache times and loosen it once we're comfortable it won't cause any issues with SQS

PR #3849, opened 2024-09-19
Use Skip API key with tx status endpoints

jonator:
### Linear Task

This PR refactors the Skip TransferStatusProvider to dep-inject the tx status and track endpoints so that we can create serverless functions that utilize the credentialed SkipApiClient on the server.

Also, I refactored the trackTransaction POST call to skip to be within a catch clause instead of a try catch statement. This makes it slightly more readable as there's less jumping around when interpreting the logic related to prompting Skip to track transactions before continuing the polling.

## Brief Changelog

- Added "SkipStatusProvider" as param to Skip's transfer status provider
- Added 2 serverless functions for Skip's tx status and track API calls. Utilizes the existing SkipApiClient which makes use of the Skip API key

## Testing and Verifying

Tested locally with Skip transfers in the deposit/withdraw flow, verifying that conclusive statuses can be received from Skip.

### DIFF
--- a/packages/bridge/src/skip/transfer-status.ts
+++ b/packages/bridge/src/skip/transfer-status.ts
@@ -18,14 +18,14 @@ type Transaction = {
   env: BridgeEnvironment;
 };
 
-export type SkipStatusProvider = {
+export interface SkipStatusProvider {
   transactionStatus: ({
     chainID,
     txHash,
     env,
   }: Transaction) => Promise<SkipTxStatusResponse>;
   trackTransaction: ({ chainID, txHash, env }: Transaction) => Promise<void>;
-};
+}
 
 /** Tracks (polls skip endpoint) and reports status updates on Skip bridge transfers. */
 export class SkipTransferStatusProvider implements TransferStatusProvider {
--- a/packages/bridge/src/skip/transfer-status.ts
+++ b/packages/bridge/src/skip/transfer-status.ts
@@ -9,8 +9,27 @@ import type {
   TransferStatusProvider,
   TransferStatusReceiver,
 } from "../interface";
-import { SkipApiClient } from "./client";
 import { SkipBridgeProvider } from "./index";
+import { SkipTxStatusResponse } from "./types";
+
+type Transaction = {
+  chainID: string;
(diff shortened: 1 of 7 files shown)


packages/bridge/src/skip/transfer-status.ts
```diff
@@ -40,39 +60,38 @@ export class SkipTransferStatusProvider implements TransferStatusProvider {
...
+        };
+
+        const txStatus = await this.skipStatusProvider
+          .transactionStatus(tx)
+          .catch(async (error) => {
+            if (error instanceof Error && error.message.includes("not found")) {
+              // if we get an error that it's not found, prompt skip to track it first
+              // then try again
+              await this.skipStatusProvider.trackTransaction(tx);
+              return this.skipStatusProvider.transactionStatus(tx);
```
JoseRFelix, line 76: Wouldn't this call short-circuit the `poll` since this is returning a promise without a catch statement?

jonator: WDYM without a catch statement? I believe what would happen here is it would be transformed into the same promise that is awaited at the initial call above

JoseRFelix: It wouldn’t, since this isn't a recursive call and it's using the dependency-injected method. If we want to share the catch statement, we’d likely need to create an inline function and call it inside the catch method

jonator: I'm not sure I follow, how would this look?

JoseRFelix: Actually, I think we could avoid recursion by just adding a catch 

```suggestion
              return this.skipStatusProvider.transactionStatus(tx).catch(() => {});
```

CryptoAssassin1 (approved):

packages/bridge/src/skip/transfer-status.ts
jonator: I don't think recursion is being used, we're just calling the same given function twice

JoseRFelix (approved):

PR #3852, opened 2024-09-21
fix: amino signing for withdraw position message

PaddyMc:

Due to the v0.50 upgrade there were some changes with how signing works,

This only effects the proto fields defined as `"cosmossdk.io/math.LegacyDec"`

Before the FE was signing something like "123.123456789123456789" where it should be signing "123123456789123456789".

The PR fixes the most used messages so end users can now use ledgers to sign withdraw position messages on the FE.

### How to test

- create a cl position in any cl pool
- withdraw a position :tada: 

Other tests of note:
- please test ledger as well :pray:

### DIFF
--- a/packages/proto-codecs/src/codegen/osmosis/concentratedliquidity/v1beta1/tx.ts
+++ b/packages/proto-codecs/src/codegen/osmosis/concentratedliquidity/v1beta1/tx.ts
@@ -1025,10 +1025,11 @@ export const MsgWithdrawPosition = {
     if (message.sender !== "") {
       writer.uint32(18).string(message.sender);
     }
+    // NOTE: Withdraw postion will break if we regenerate protos
+    // if you see this line being removed in a PR, flag it!
+    // TODO: Abstract to an override function.
     if (message.liquidityAmount !== "") {
-      writer
-        .uint32(26)
-        .string(Decimal.fromUserInput(message.liquidityAmount, 18).atomics);
+      writer.uint32(26).string(message.liquidityAmount.replace(".", ""));
     }
     return writer;
   },
--- a/packages/stores/src/account/amino-converters.ts
+++ b/packages/stores/src/account/amino-converters.ts
@@ -1,4 +1,8 @@
 import type { AminoMsgTransfer } from "@cosmjs/stargate";
+import type {
+  MsgWithdrawPosition,
+  MsgWithdrawPositionAmino,
+} from "@osmosis-labs/proto-codecs/build/codegen/osmosis/concentratedliquidity/v1beta1/tx";
 import type { MsgTransfer } from "cosmjs-types/ibc/applications/transfer/v1/tx";
 import Long from "long";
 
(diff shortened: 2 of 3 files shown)


CryptoAssassin1 (approved):

packages/proto-codecs/src/codegen/osmosis/concentratedliquidity/v1beta1/tx.ts
```diff
@@ -1025,10 +1025,11 @@ export const MsgWithdrawPosition = {
...
       writer.uint32(18).string(message.sender);
     }
+    // NOTE: Withdraw postion will break if we regenerate protos
+    // if you see this line being removed in a PR, flag it!
+    // TODO: Abstract to an override function.
     if (message.liquidityAmount !== "") {
-      writer
-        .uint32(26)
-        .string(Decimal.fromUserInput(message.liquidityAmount, 18).atomics);
+      writer.uint32(26).string(message.liquidityAmount.replace(".", ""));
```
JoseRFelix, line 1032: Has this been flagged Telescope? Furthermore, there's an [option](https://github.com/cosmology-tech/telescope#typings-and-formating) called `prototypes.typingsFormat.customTypes.useCosmosSDKDec`. Perhaps that could help us with this too?

JoseRFelix: I believe the amino-converter changes should already cover this. Is this not the case?

PaddyMc: > I believe the amino-converter changes should already cover this. Is this not the case?

no they don't, for some reason the encoding is adding extra bytes, so the signature is incorrect when the node compare e.g 

```
[0A][8C][01]
:/osmosis.concentratedliquidity.v1beta1.MsgWithdrawPositionN[08][F1][92][FE][03]+osmo1sh8lreekwcytxpql6lxmw5cl7kdrflpt2ujlvz795950193821780327758658024[18][C9][A7][8A][0A]
```
txBody2 CLI:
```
[0A][8C][01]
:/osmosis.concentratedliquidity.v1beta1.MsgWithdrawPositionN[08][F1][92][FE][03]+osmo1sh8lreekwcytxpql6lxmw5cl7kdrflpt2ujlvz795950193821780327758658024
```

PaddyMc: > Has this been flagged Telescope? Furthermore, there's an [option](https://github.com/cosmology-tech/telescope#typings-and-formating) called prototypes.typingsFormat.customTypes.useCosmosSDKDec. Perhaps that could help us with this too?

No it hasn't, but I think we already have this flag on the FE => https://github.com/osmosis-labs/osmosis-frontend/blob/stage/packages/proto-codecs/scripts/codegen.ts#L83-L91

what's the best way to reach out to Telescope?

JoseRFelix: > No it hasn't, but I think we already have this flag on the FE => [stage/packages/proto-codecs/scripts/codegen.ts#L83-L91](https://github.com/osmosis-labs/osmosis-frontend/blob/stage/packages/proto-codecs/scripts/codegen.ts?rgh-link-date=2024-09-23T15%3A40%3A46Z#L83-L91)

Maybe we can set `useCosmosSDKDec` to false?

PaddyMc:
```
diff --git a/packages/proto-codecs/src/codegen/osmosis/concentratedliquidity/v1beta1/tx.ts b/packages/proto-codecs/src/codegen/osmosis/concentratedliquidity/v1beta1/tx.ts
index aae9510e0..80c56ad87 100644
--- a/packages/proto-codecs/src/codegen/osmosis/concentratedliquidity/v1beta1/tx.ts
+++ b/packages/proto-codecs/src/codegen/osmosis/concentratedliquidity/v1beta1/tx.ts
@@ -1,6 +1,4 @@
 //@ts-nocheck
-import { Decimal } from "@cosmjs/math";
-
 import { BinaryReader, BinaryWriter } from "../../../binary";
 import {
   Coin,
@@ -594,9 +592,7 @@ export const MsgCreatePositionResponse = {
       writer.uint32(26).string(message.amount1);
     }
     if (message.liquidityCreated !== "") {
-      writer
-        .uint32(42)
-        .string(Decimal.fromUserInput(message.liquidityCreated, 18).atomics);
+      writer.uint32(42).string(message.liquidityCreated);
     }
     if (message.lowerTick !== BigInt(0)) {
       writer.uint32(48).int64(message.lowerTick);
@@ -627,10 +623,7 @@ export const MsgCreatePositionResponse = {
           message.amount1 = reader.string();
           break;
         case 5:
-          message.liquidityCreated = Decimal.fromAtomics(
-            reader.string(),
-            18
-          ).toString();
+          message.liquidityCreated = reader.string();
           break;
         case 6:
           message.lowerTick = reader.int64();
@@ -1025,11 +1018,8 @@ export const MsgWithdrawPosition = {
     if (message.sender !== "") {
       writer.uint32(18).string(message.sender);
     }
-    // NOTE: Withdraw postion will break if we regenerate protos
-    // if you see this line being removed in a PR, flag it!
-    // TODO: Abstract to an override function.
     if (message.liquidityAmount !== "") {
-      writer.uint32(26).string(message.liquidityAmount.replace(".", ""));
+      writer.uint32(26).string(message.liquidityAmount);
     }
     return writer;
   },
@@ -1051,10 +1041,7 @@ export const MsgWithdrawPosition = {
           message.sender = reader.string();
           break;
         case 3:
-          message.liquidityAmount = Decimal.fromAtomics(
-            reader.string(),
-            18
-          ).toString();
+          message.liquidityAmount = reader.string();
           break;
         default:
           reader.skipType(tag & 7);
```

This is the diff, I don't think this solves the issue either, wdyt?

JoseRFelix: The latter did not work. We'll have to reach Cosmology

JoseRFelix (approved):

PR #3864, opened 2024-10-02
filter out bad pool per request from adam and johnny

MaxMillington:

Filtering out pool 2159

### Linear Task

https://linear.app/osmosis/issue/POL-1363/filter-out-pool-2159

## Brief Changelog

- Filter out bad pool

## Testing and Verifying

### DIFF
--- a/packages/web/components/complex/pools-table.tsx
+++ b/packages/web/components/complex/pools-table.tsx
@@ -36,8 +36,6 @@ import { api, RouterOutputs } from "~/utils/trpc";
 
 import { Tooltip } from "../tooltip";
 
-const FILTERABLE_IDS = ["2159"];
-
 export type Pool = RouterOutputs["edge"]["pools"]["getPools"]["items"][number];
 /** UI doesn't support cosmwasm pools as first class so exclude it from list of filter options. */
 export type PoolTypeFilter = Exclude<Pool["type"], "cosmwasm">;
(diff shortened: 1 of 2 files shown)


packages/web/components/complex/pools-table.tsx
```diff
@@ -36,6 +36,8 @@ import { api, RouterOutputs } from "~/utils/trpc";
 
 import { Tooltip } from "../tooltip";
 
+const FILTERABLE_IDS = ["2159"];
```
jonator, line 39: Perhaps we can add a ternary operator that creates an array for testnet & mainnet.

Like:
`const FILTERABLE_IDS = IS_TESTNET ? [] : ["2159"];`

jonator:
I commented in slack, but I'm also no longer seeing the Cosmo pool in prod

packages/web/components/complex/pools-table.tsx
```diff
@@ -176,10 +178,14 @@ export const PoolsTable = (props: PropsWithChildren<PoolsTableProps>) => {
...
-    [poolsPagesData]
-  );
+  const poolsData = useMemo(() => {
+    const allItems =
+      poolsPagesData?.pages.flatMap((page) => {
+        return page?.items.filter((item) => !FILTERABLE_IDS.includes(item.id)); // Filter out ids in FILTERABLE_IDS
+      }) ?? [];
+
+    return allItems;
+  }, [poolsPagesData]);
```
JoseRFelix, line 188: Do we only want to filter this pool in the pools table? If not, I suggest modifying https://github.com/osmosis-labs/osmosis-frontend/blob/288c023d8100418b8f5a2b607f532dd6e0014112/packages/server/src/queries/complex/pools/index.ts#L89 instead, so the changes propagate to all components that rely on pools.

JoseRFelix (approved):

raphaellafar:
I reiterate the request to unblock liquidity pool 2159. As confirmed in the thread https://forum.osmosis.zone/t/dao-proposal-establish-transparent-rules-for-token-and-liquidity-pool-blocking/3233 , this pool should not be blocked.

PR #3866, opened 2024-10-03
Fix: Unbond from weighted pools TX

JoseRFelix:

This PR resolves an issue with unbonding from a weighted pool by updating the proto codecs. The update ensures that the coins array is excluded when it’s empty, as the new version of the SDK does not support empty arrays. The bug was introduced following the upgrade to Cosmos SDK after the Osmosis v26 upgrade.

### Linear Task

https://linear.app/osmosis/issue/FE-1143/fix-pool-creation-and-unboding-weighted-pools-tx

## Testing and Verifying

- [ ] Can unbond from weighted pools


DavideSegullo (approved):

PaddyMc (approved):
gg

CryptoAssassin1 (approved):

jonator (approved):

PR #3867, opened 2024-10-04
feat: Improve codebase Onboarding

JoseRFelix:

This PR improves the onboarding process by addressing issues related to the initial build when environment variables have not been set up. Specifically, we replaced the hardcoded Osmosis API URL, which requires an API key, with a fallback to the free Coingecko API. While the Coingecko API has rate limits, it is sufficient for initial onboarding and short-term contributions to the code base.

Additionally, this PR updates the default feature flags to display a more up-to-date UI that aligns closely with the current production version. The README has also been updated to streamline the first installation and dev server setup. With these improvements, setting up the code base is now as simple as following the included instructions.

### Linear Task

https://linear.app/osmosis/issue/FE-1146/improve-osmosis-fe-onboarding

## Brief Changelog

- Update README
- Use free Coingecko API when the Osmosis API Key is not available
- Update default Feature Flags to match current production UI

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,13 @@ npx turbo link
 ...press y (yes) and choose "OsmoLabs" as the Vercel build scope...
 ```
 
-2.  Run local server at [`localhost:3000`](localhost:3000)
+2. Run an initial build to create packages build artifacts:
+
+```bash
+yarn build
+```
+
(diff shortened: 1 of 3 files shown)


MaxMillington (approved):
Any special env variables needed?

PR #3868, opened 2024-10-04
fix: LegacyDec transactions

JoseRFelix:

This PR resolves transaction issues related to Decimal changes introduced in Osmosis v26. The fix involves encoding messages to the proto format, which normalizes data types like Decimals. Afterward, the messages are converted to their amino representations.

The following transactions are now correctly handled:

- Unbonding weighted pool shares
- Withdrawing concentrated liquidity positions
- Creating all types of pools

### Linear Task

https://linear.app/osmosis/issue/FE-1130/fix-create-pools-tx

## Testing and Verifying

- [ ] Can unbond weighted pool shares
- [ ] Can withdraw concentrated liquidity positions
- [ ] Can create all pool types
- [ ] Swaps work


packages/stores/src/account/base.ts
```diff
@@ -1104,16 +1128,10 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
           signDoc
         ));
 
-    const registry = await this.getRegistry();
+    console.log(normalizedMessages, msgs);
```
PaddyMc, line 1131: should this be here?

jonator (approved):

packages/stores/src/account/base.ts
JoseRFelix: Oh, I missed this. Good catch

kamal-sutra (approved):

CryptoAssassin1 (approved):

PR #3871, opened 2024-10-07
Mattupham/fe 725 porfolio v3 variants detected popup

mattupham:

- we need to prompt a user to convert their alloyed asset variants if their portfolio includes alloyed assets

### Linear Task

https://linear.app/osmosis/issue/FE-725/porfolio-v3-variants-detected-popup

## Brief Changelog

- Add variant detection logic
- Add toast for Alloyed Assets variants
- Add local storage logic to show / not show alloyed assets variants
- Add query for alloyed assets variants
- i18n, integrate feature flag
- add static icons for assets + asset variants to sprite
- only show on desktop

## Testing and Verifying
add jest test cases for
- has asset variants
- does not have asset variants
- has no assets 

screenshot
Screenshot 2024-10-07 at 2 16 33 PM

example on any page

https://github.com/user-attachments/assets/5a6baa7e-ec4e-4b19-8c1c-96a945f87540

local storage persistence example

https://github.com/user-attachments/assets/6434af48-5704-4fa8-a6e5-4bfb5eda8272

### DIFF
--- a/packages/server/src/queries/complex/portfolio/allocation.ts
+++ b/packages/server/src/queries/complex/portfolio/allocation.ts
@@ -173,7 +173,7 @@ export async function getAllocation({
       (result) => result.coin.denom
     ) ?? [];
 
-  // check for alloyed asset variants
+  // check for asset variants, alloys and canonical assets such as USDC
   const hasVariants = checkHasAssetVariants(
     userBalanceDenoms,
     assetLists.flatMap((list) => list.assets)


packages/web/components/layouts/main.tsx
```diff
@@ -34,6 +39,8 @@ export const MainLayout = observer(
       ({ selectionTest }) => selectionTest?.test(router.pathname) ?? false
     );
 
+    useHasAssetVariants();
```
jonator, line 42: Is it necessary to check for this on every page? May affect performance for other pages. Why not where actionable, like on portfolio page? May be a question for nass

packages/web/hooks/use-has-asset-variants.tsx
```diff
@@ -0,0 +1,57 @@
...
+    AlloyedAssetsToastDoNotShowKey,
+    false
+  );
+
+  const enabled =
+    alloyedAssets &&
+    !doNotShowAgain &&
+    Boolean(wallet?.isWalletConnected && wallet?.address);
+
+  api.local.portfolio.getHasAssetVariants.useQuery(
```
jonator, line 29: From looking at implementation of `useHasAssetVariants`

What if we just run that function inside `getAllocation` since we operate on that same data that we query elsewhere anyways? Then `useHasAssetVariants` here would call `getAllocation` looking for that `hasAssetVariants: boolean` flag instead when deciding to show the toast.

Pros:
* Can remove a redundant tRPC procedure call (this one, `getHasAssetVariants`) in `portfolioRouter`
* Prevent redundant API call to SQS for `passthrough/portfolio-assets`, reducing load on SQS services and avoiding an unnecessary query from the browser
* If we check for convertible assets on every page anyways, we can reuse that allocation query data when they navigate to the portfolio page, increasing navigation performance

Cons:
* May be hard to know that flag is there. In other words, the purpose of `getAllocation` has expanded a bit to involve alloyed assets logic

Overall, I think it's worth it, as we can delete a bunch of code. Which means less chance for error, and less to test.

jonator (changes requested):

packages/web/hooks/use-has-asset-variants.tsx
jonator: I'm seeing redundant calls to portfolio-assets from different initiator call stacks.

I think it could easily be avoided, and could save on bandwidth to SQS:

Screenshot 2024-10-07 at 6 03 02 PM

jonator (changes requested):
Noticed another issue

packages/web/hooks/use-has-asset-variants.tsx
jonator: @mattupham 

Check this out
https://github.com/osmosis-labs/osmosis-frontend/pull/3872

packages/server/src/queries/complex/portfolio/allocation.ts
```diff
@@ -168,11 +168,12 @@ export async function getAllocation({
     new Dec(categories["total-assets"].capitalization)
   );
 
-  // Check for variants
   const userBalanceDenoms =
     categories["user-balances"]?.account_coins_result?.map(
       (result) => result.coin.denom
     ) ?? [];
+
+  // check for alloyed asset variants
```
jonator, line 176: Giga Nit: variants include canonical assets such as USDC, not just alloys

jonator (approved):

jonator (changes requested):
I noticed nothing happens when I click convert:
https://github.com/user-attachments/assets/f2445a3a-34c5-47f7-9f26-1ad786f20d23

jonator (approved):
I see now, to be done in future PR

https://github.com/osmosis-labs/osmosis-frontend/blob/daad653f88f511a74c2dd02c5e36d18c5f46a823/packages/web/components/alert/toast.tsx#L245

MaxMillington (approved):

PR #3873, opened 2024-10-07
(Deposit/Withdraw) Add Pending Nomic Deposits

JoseRFelix:
### Linear Task

https://linear.app/osmosis/issue/FE-1136/display-pending-nomic-transfers

## Brief Changelog

- Display Pending Nomic Deposits

## Testing and Verifying

- [ ] Can see pending Nomic deposits


packages/bridge/src/nomic/index.ts
```diff
@@ -158,4 +164,22 @@ export class NomicBridgeProvider implements BridgeProvider {
       url,
     };
   }
+
+  async getPendingDeposits({ address }: { address: string }) {
```
jonator, line 168: Why not implement this as a transfer status provider by Nomic? Would match other providers. Also, would display pending transfers in the pending transfers table.

If you just want Nomics, we could filter the pending transfers by a specific provider ID, perhaps

JoseRFelix: Unfortunately, we cannot determine when a transaction is pushed because the user is not using a wallet extension but rather a different interface. To track the transaction, we need to first query Nomic’s chain for pending transfers. If any are found, our UI will display them in a table without tooltips, which could disrupt the current transfer history flow. While it’s possible to implement a transfer status provider, this would introduce a lot of flags and glue code to make it work with our store. Since most pending transfers will require a case-by-case approach, I prefer to keep it simple by adding specific functions for Nomic.

jonator (commented):
Noticed unit tests are failing

PR #3880, opened 2024-10-08
fix: Tests

JoseRFelix:


jonator (approved):
What is the main bundle size? Do we still get the tree shaking benefits?

jonator:
I see from build it's still 1.77, nice

JoseRFelix:
> What is the main bundle size? Do we still get the tree shaking benefits?

Yes, since Babel config is only used in Jest

PR #3888, opened 2024-10-14
Filter out additional incorrect TVL pools

JohnnyWyles:
Similar to https://github.com/osmosis-labs/osmosis-frontend/pull/3864
This pool was noticed to be filtered and so 2 new ones were made with the same singlesided tactic, skewing displayed TVL.

## Brief Changelog

Adds 2 pools to list

## Testing and Verifying

This change has not been tested locally


raphaellafar:
I strongly protest, there is no reason to block these pools. Moreover, introducing a filter at this level is a crime.

PR #3894, opened 2024-10-16
Use market cap from tokens endpoint

jonator:

Numia recently added market cap to the tokens endpoints.

This unlocks the ability for FE to not query mcap list separately, reducing the # of network requests and improving performance.

### Linear Task

[FE-1168: Server pkg: remove /mcap query, use mcaps from new tokens query](https://linear.app/osmosis/issue/FE-1168/server-pkg-remove-mcap-query-use-mcaps-from-new-tokens-query)

## Brief Changelog

- Add `coingecko_mcap` to tokens response type from Numia
- Use that value instead of querying /mcap endpoint
- Remove some fallback queries to CoinGecko

## Testing and Verifying

Tested locally


MaxMillington (approved):
I love red PRs

packages/server/src/queries/data-services/token-data.ts
```diff
@@ -17,6 +17,7 @@ export interface TokenData {
   price_7d_change: number | null;
   exponent: number;
   display: string;
+  coingecko_mcap: number | null;
```
niccoloraspa, line 20: There is also a new field called `coingecko_id`. 
Numia added it just in case it could be useful. 

Not sure if we want to add it in the interface to match the response

jonator:
Marking drafts since the mcap field has not been added to Numia prod, only stage

https://data.numia-stage.osmosis.zone/tokens/v2/all

jonator:
I see mcaps in local dev when hardcoding root var

Will make sure they're in previews so we can test
Screenshot 2024-10-16 at 12 23 18 PM

jonator:
Numia should be deploying to prod soon. Then, we can confirm they're visible here before merging

jonator:
Prod is updated and data can be seen in these previews. However, it seems the mcap data in new endpoint is much more sparse than old /mcap endpoint from imperator. Investigating now on Numia side. Assumption is it's because we're not including coin market cap data like imperator may have been.

PR #3896, opened 2024-10-16
Add stratos pubkey support

Xiong-stratos:

Add stratos pubkey support

## Brief Changelog

-

### DIFF
--- a/packages/keplr-stores/src/account/cosmos.ts
+++ b/packages/keplr-stores/src/account/cosmos.ts
@@ -566,6 +566,10 @@ export class CosmosAccountImpl {
                     return "/injective.crypto.v1beta1.ethsecp256k1.PubKey";
                   }
 
+                  if (this.chainId.startsWith("stratos")) {
+                    return "/stratos.crypto.v1.ethsecp256k1.PubKey";
+                  }
+
                   return "/ethermint.crypto.v1.ethsecp256k1.PubKey";
                 })(),
                 value: PubKey.encode({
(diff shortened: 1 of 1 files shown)


packages/keplr-stores/src/account/cosmos.ts
```diff
@@ -566,6 +566,10 @@ export class CosmosAccountImpl {
                     return "/injective.crypto.v1beta1.ethsecp256k1.PubKey";
```
JoseRFelix, line 566: Thank you for opening this PR. Is this code addressing a specific issue within the app?  I believe we no longer use this store for signing transactions. If so, could you please point it out so we can ensure proper QA? Also, we already include Stratos in our [`getPublicKeyTypeUrl`](https://github.com/osmosis-labs/osmosis-frontend/blob/109ba2b43e477a9dbfda2ecb2017985152c83ea3/packages/stores/src/account/utils.ts#L93) which is used in all of our signing methods.

JoseRFelix (changes requested):

packages/keplr-stores/src/account/cosmos.ts
BoThe1K: @JoseRFelix we have this issue during integration https://github.com/osmosis-labs/assetlists/pull/1946#issuecomment-2417400492 . We have resolved this in keplr wallet, but on ui looks like it use own logic and broadcast without keplr.

JoseRFelix: Do you know if this is working locally?

PR #3907, opened 2024-10-24
fix(disable-link-on-pools-card): disable link so when user clicks car…

mattupham:
…d, it navigates to the pool instead of the asset

- users would frequently click an asset in their position card, which would navigate them to the assets page instead of the position

### Linear Task

https://linear.app/osmosis/issue/FE-926/if-i-click-on-the-solusdc-text-instead-of-being-taken-to-the-asset

## Brief Changelog

- since we reuse PoolAssetsName frequently, add a withLink prop that defaults to true, and removes the link if explicitly set to false

## Testing and Verifying

- example of navigating to pool instead of asset when withLink is set to false

https://github.com/user-attachments/assets/c10a49ee-b333-4885-a175-8b4faf1d34c8

- example of navigating to assets when withLink is true by default

https://github.com/user-attachments/assets/fb433813-2129-4e17-8aba-2ce72623146b

### DIFF
--- a/packages/web/components/assets/pool-assets-name.tsx
+++ b/packages/web/components/assets/pool-assets-name.tsx
@@ -7,8 +7,8 @@ export const PoolAssetsName: FunctionComponent<{
   size?: "sm" | "md";
   assetDenoms?: string[];
   className?: string;
-  withLink?: boolean;
-}> = ({ size = "md", assetDenoms, className, withLink = true }) => {
+  withAssetInfoLink?: boolean;
+}> = ({ size = "md", assetDenoms, className, withAssetInfoLink = true }) => {
   const formatAssetName = useMemo(() => {
     return (asset: string) => {
       if (asset.startsWith("ibc/")) return truncate(asset);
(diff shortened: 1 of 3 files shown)


jonator (approved):

packages/web/components/assets/pool-assets-name.tsx
```diff
@@ -1,13 +1,22 @@
...
 import Link from "next/link";
-import { Fragment, FunctionComponent } from "react";
+import { Fragment, FunctionComponent, useMemo } from "react";
 
 export const PoolAssetsName: FunctionComponent<{
   size?: "sm" | "md";
   assetDenoms?: string[];
   className?: string;
-}> = ({ size = "md", assetDenoms, className }) => {
+  withLink?: boolean;
```
JoseRFelix, line 10: Thoughts on renaming this to something more explicit like withAssetInfoLink?

JoseRFelix (changes requested):

PR #3917, opened 2024-10-27
Increase base fee multiplier from 1.5 to 1.65

p0mvn:

We have experienced instabilities with the base fee query. To avoid debugging its stability
for the sake of time, we have opted in to increase the base fee multiplier to 1.65 from the original
value of 1.5 which was equal to the gas multiplier. The update was applied on 2024-10-27.

More context: https://osmosis-network.slack.com/archives/C06QAMXGTJ5/p1730045591060839

### DIFF
--- a/packages/tx/src/__tests__/gas.spec.ts
+++ b/packages/tx/src/__tests__/gas.spec.ts
@@ -13,6 +13,7 @@ import { ApiClientError } from "@osmosis-labs/utils";
 import type { Any } from "cosmjs-types/google/protobuf/any";
 
 import {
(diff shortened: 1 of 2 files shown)


jonator (approved):

czarcas7ic (approved):

kamal-sutra (approved):

CryptoAssassin1 (approved):

yury-dubinin (approved):
@p0mvn ignore e2e fails, they are due to a new wallet being used.

PR #3920, opened 2024-10-29
(Deposit/Withdraw) Handle automatic conversion from nBTC to allBTC with IBC

JoseRFelix:
### Linear Task

https://linear.app/osmosis/issue/FE-1135/handle-automatic-conversion-from-nbtc-to-allbtc-with-ibc-hooks

## Brief Changelog

- Construct IBC hook memo to swap nBTC for BTC when funds arrive at Osmosis

## Testing and Verifying

- [ ] Can deposit nBTC and BTC


packages/tx/src/message-composers/complex/swap.ts
```diff
@@ -244,3 +244,63 @@ export async function getSwapMessages({
 
   throw new Error(`Unsupported quote type ${quoteType}`);
 }
+
+export function makeSkipIbcHookSwapMemo({
```
jonator, line 248: Giga nit: are IBC hooks exclusive to Skip? Is that why it's in the name?

packages/bridge/src/nomic/index.ts
```diff
@@ -111,7 +120,7 @@ export class NomicBridgeProvider implements BridgeProvider {
       | Awaited<ReturnType<typeof getRouteTokenOutGivenIn>>
       | undefined;
 
-    if (fromAsset.address !== this.nBTCMinimalDenom) {
+    if (fromAsset.address === this.allBtcMinimalDenom) {
```
jonator, line 123: Perhaps we should toLower to be safe

packages/tx/src/message-composers/complex/swap.ts
JoseRFelix: No, we could have different types of hooks, but since this is a Skip smart contract, we need to provide specific parameters that are compatible only with this contract hence why “Skip” is included in the name

kamal-sutra (approved):

PR #3931, opened 2024-11-01
Variants conversion flow

jonator:

Completes variant conversion flow. Due to limitations on chain, the user must convert each asset in individual txs.

Also, when selecting conversions from portfolio asset list, if the user has not seen the modal before, it opens the modal. Otherwise, the swap tx is directly requested in wallet.

Also, this PR makes some various improvements to naming/file organization, etc to hopefully make things easier to navigate and long term more maintainable.

### Linear Task

Primarily closes: [FE-593: Portfolio v3 - Convert to Alloy UX](https://linear.app/osmosis/issue/FE-593/portfolio-v3-convert-to-alloy-ux)
Closes: [FE-295: Alloyed assets amplitude events](https://linear.app/osmosis/issue/FE-295/alloyed-assets-amplitude-events)
Fixes: [FE-1075: Fix responsive issues in portfolio positions view charts](https://linear.app/osmosis/issue/FE-1075/fix-responsive-issues-in-portfolio-positions-view-charts)
Improves: [FE-1190: General refactor + performance improve](https://linear.app/osmosis/issue/FE-1190/general-refactor-performance-improve)

## Changelog

- Completes Matt's work on variant conversion flow
- Fixes a bug I noticed where transactions query would not get invalidated after tx
- Changes portfolio asset dropdown to jump to swap page instead of 
- Adds Amplitude events
- Simulates variants swap, shows error message if failed, logs ticker of unavailable variant to Amplitude

## Testing and Verifying

Tested locally


DavideSegullo (approved):

CryptoAssassin1 (approved):

packages/web/modals/variants-conversion.tsx
```diff
@@ -0,0 +1,363 @@
...
+          </>
+        )}
+        <span className="text-osmoverse-500">
+          {showConvertFee
+            ? `(${quote.swapFee.maxDecimals(2)})`
+            : quote.swapFee.maxDecimals(2).toString()}
+        </span>
+      </>
+    );
+  };
```
JoseRFelix, line 251: We should probably move this to a separate component to avoid unnecessary re-renders.

packages/web/pages/index.tsx
```diff
@@ -43,11 +44,13 @@ const Home = () => {
...
-              previousTrade={previousTrade}
-              setPreviousTrade={setPreviousTrade}
-            />
+            <ClientOnly>
+              <TradeTool
+                page="Swap Page"
+                previousTrade={previousTrade}
+                setPreviousTrade={setPreviousTrade}
+              />
+            </ClientOnly>
```
JoseRFelix, line 53: Could you provide some context on why this is needed now? Maybe a comment would be great too

JoseRFelix (changes requested):
Great work, looking very clean! Left some minor comments

packages/web/pages/index.tsx
jonator: I'm not sure. It was originally wrapping within TradeTool component, but I moved it back to here since I started having intermittent hydration errors in dev server. (I believe the errors are silenced in prod). I didn't want to take the time now to debug

JoseRFelix (approved):

PR #3940, opened 2024-11-12
(Deposit/Withdraw) Handle ui pending state when estimated time reaches 0

JoseRFelix:

During withdrawals or deposits, our toast currently shows the exact time remaining, even down to 0 seconds. This isn’t helpful, as users may end up waiting longer than expected due to relayer instability. Instead, we’ll display a range like “About 5s remaining” to set a more realistic expectation that the transaction will complete soon. This provides more useful feedback than displaying “0s remaining.”

image

### Linear Task

https://linear.app/osmosis/issue/FE-1208/change-ui-to-pending-state-when-estimated-time-reaches-0

## Testing and Verifying

- [ ] Displays "About 5 seconds remaining" on deposit or withdraw


packages/web/stores/transfer-history.tsx
```diff
@@ -422,12 +422,18 @@ const PendingTransfer: FunctionComponent<{
...
+      const humanizedTime = humanizeTime(date);
       if (progressRef.current) {
         // DANGER: We update the HTML directly because react-toastify is having issues while handling react state changes
-        progressRef.current.textContent = `${t("estimated")} ${
-          humanizedTime.value
-        } ${t(humanizedTime.unitTranslationKey)} ${t("remaining")}`;
+        progressRef.current.textContent =
+          date.diff(dayjs(), "seconds") < 5
+            ? t("aboutSecondsRemaining", {
+                seconds: "5 " + t("timeUnits.seconds"),
```
jonator, line 432: Wait so we just hardcode to 5 seconds while it's still pending past estimated time?

JoseRFelix: Yes, the idea is that users won’t see the exact count once it reaches a certain amount, allowing some flexibility for time estimate inaccuracies

PR #3949, opened 2024-11-14
(Asset listing) Fix preview assets flag

JoseRFelix:
### Linear Task

https://linear.app/osmosis/issue/FE-1218/osmosis-zone-preview-assets-flag-not-working

## Testing and Verifying

- [ ] Can see preview assets like Aave after adding `?show_preview_assets=true`


jonator (approved):
I think localization of this feature is low/no prio, but you already did it so LGTM

greg-nagy:
@JoseRFelix this PR has introduced an infinite rerender loop 

The problem is on `use-amount-input.ts:68`
The issue starts when you enter an amount on the swap page.

PR #3954, opened 2024-11-15
🤗 Embrace the 1-Click: Add 1-Click enable and modify to swap review

greg-nagy:
~~⚠️ Note: An infinite rerender issue exists when an amount is entered into the swap input. This was introduced in https://github.com/osmosis-labs/osmosis-frontend/pull/3949 and will affect this PR~~ Solved by #3956 

Adding the 1 click trading enable and parameter settings to the review order modal to promote the feature.

<details>
<summary>Design from figma</summary>

CleanShot 2024-11-14 at 17 34 58

</details>

## Notes/TBD
- ~~After enabling 1CT through the `ReviewOrder` modal the actual swap needs to be signed as well~~
  - ~~this is a bit quirky, but it seemed better than abusing our newfound power and executing a more expensive swap that was shown to the user (even though the usd amount of the change is negligible)~~
- The profile settings could be reworked so that part also uses `useOneClickTradingSessionManager` but wanted to keep the existing code paths for now
  - this feature also could be a good fit for using Zustand but didn't want to do that without buy-in

## Brief Changelog

- reworked the parameter setting modal to support "in flow" usage 
  - this need wasn't explicitly stated in the requirements, but keeping the existing behavior would have contradicted some user stories
- reworked `useOneClickTradingSession` and `useOneClickTradingParams` to properly propagate changes to the 1CT session coming from the account store
- extend `ReviewOrder` modal to implement the design and features
- create `useOneClickTradingSessionManager` for managing states 
- made the trpc logs configurable
- added short time units 

the commit messages are fairly well structured and descriptive

## Testing and Verifying

- enable 1ct through the reivew order modal
  - play with changing + saving/discarding the config
- when 1ct check is enabled, modify the params and approve the changes
- check the profile state shows the proper settings
- change 1ct in through the profile menu and check review order modal
- set the spending limit to 0usd and check the `Exceeded` state 

More nuanced things can be tested, but these cover the main paths. 

## Demo
CleanShot 2024-11-14 at 17 10 34


packages/web/components/one-click-trading/screens/session-period-screen.tsx
```diff
@@ -14,12 +14,12 @@ export function getSessionPeriodTranslationKey(
...
-    "30min": "oneClickTrading.settings.sessionPeriodScreen.periods.30min",
-    "1hour": "oneClickTrading.settings.sessionPeriodScreen.periods.1hour",
-    "3hours": "oneClickTrading.settings.sessionPeriodScreen.periods.3hours",
-    "12hours": "oneClickTrading.settings.sessionPeriodScreen.periods.12hours",
+    "5min": "oneClickTrading.sessionPeriods.5min",
+    "10min": "oneClickTrading.sessionPeriods.10min",
+    "30min": "oneClickTrading.sessionPeriods.30min",
+    "1hour": "oneClickTrading.sessionPeriods.1hour",
+    "3hours": "oneClickTrading.sessionPeriods.3hours",
+    "12hours": "oneClickTrading.sessionPeriods.12hours",
```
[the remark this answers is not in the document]
greg-nagy: these does exist https://github.com/greg-nagy/osmosis-frontend/blob/5727b9d20d24d162b90cf2a9416243f7fb0f3014/packages/web/localizations/en.json#L400-L407

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -39,90 +45,146 @@ export function getParametersFromOneClickTradingInfo({
...
+  next,
+}: {
+  prev?: OneClickTradingTransactionParams;
+  next?: OneClickTradingTransactionParams;
+}): OneClickTradingParamsChanges {
+  let changes: OneClickTradingParamsChanges = [];
+
+  if (prev?.spendLimit.toString() !== next?.spendLimit.toString()) {
+    changes.push("spendLimit");
+  }
```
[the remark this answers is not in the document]
greg-nagy: `PricePretty` does not have `.equals()` and this comparision should be fine as this check was used previously

packages/web/components/one-click-trading/one-click-trading-settings.tsx
```diff
@@ -105,11 +129,16 @@ export const OneClickTradingSettings = ({
   hasExistingSession,
   onEndSession,
   onClose,
+  standalone = true,
+  externalChanges = [],
+  setExternalChanges,
```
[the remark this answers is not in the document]
greg-nagy: this was intentional because standalone is the default and you need to put in extra work to setup without standalone

packages/web/components/one-click-trading/one-click-trading-settings.tsx
```diff
@@ -268,46 +309,51 @@ export const OneClickTradingSettings = ({
...
-                          checked={
-                            transaction1CTParams?.isOneClickEnabled ?? false
+                  {standalone && (
+                    <SettingRow
+                      title={t("oneClickTrading.settings.enableTitle")}
+                      onClick={() => {
+                        if (hasExistingSession) onEndSession?.();
+                        setTransaction1CTParams((params) => {
+                          if (!params)
+                            throw new Error("1CT Params is undefined");
```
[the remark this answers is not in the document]
greg-nagy: I think this was intentional previously 

I have only hidden this element in standalone mode

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -39,90 +45,146 @@ export function getParametersFromOneClickTradingInfo({
...
  *
- * The hook also provides a mechanism to reset the parameters to their initial values, which
- * can be either the custom parameters passed at initialization or the fetched default parameters.
+ * If `oneClickTradingInfo` needs time to stabilize before we can set reliably the initial value for
+ * `transaction1CTParams`, we can set `readyToInitialize` to `true` when it's stable.
+ *
+ * The hook provides a mechanism to reset the parameters to their initial values and tracks changes
+ * to the parameters when `trackChanges` is true, comparing them against the initial values.
  *
  * This hook is primarily intended for the one click trading settings modal.
```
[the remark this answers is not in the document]
greg-nagy: 151807e97c7a

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
```diff
@@ -0,0 +1,271 @@
...
+          : undefined,
+      },
+      {
+        onError: (e) => {
+          const error = e as Error;
+          rollbackCreateSession();
+          if (!isRejectedTxErrorMessage({ message: error?.message })) {
+            displayErrorRemovingSessionToast();
+          }
+        },
```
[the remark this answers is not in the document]
greg-nagy: this felt ok and it was used originally to signal the failure in `profile-one-click-trading.tsx`

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
```diff
@@ -0,0 +1,271 @@
...
+        ...(initialTransactionParams ?? transactionParams),
+        isOneClickEnabled: true,
+      });
+    };
+
+    if (!info) {
+      displayErrorRemovingSessionToast();
+      rollbackRemoveSession();
+      throw new Error("oneClickTradingInfo is undefined");
+    }
```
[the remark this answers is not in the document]
greg-nagy: this seemed intentional in the original code and made sense

packages/web/hooks/use-swap.tsx
```diff
@@ -408,6 +408,17 @@ export function useSwap(
...
+          //
+          // When 1CT is enabled via the swap modal, there's a timing mismatch between
+          // the local state (isOneClickTradingEnabled) and the account store state.
+          //
+          // As a result:
+          // 1. The first swap after enabling 1CT will require manual signing
+          // 2. Subsequent swaps will use 1CT automatically
+          //
+          // This occurs because 1CT transactions require more gas, which makes
+          // our current quote's gas estimation incorrect and low. This results in out of gas error.
```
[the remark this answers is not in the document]
greg-nagy: the suggested identation is not good

packages/web/modals/review-order.tsx
```diff
@@ -98,7 +105,38 @@ export function ReviewOrder({
...
+    },
+  });
+
+  const wouldExceedSpendLimit = useMemo(
+    () =>
+      transaction1CTParams?.spendLimit
+        ?.sub(amountSpent1CT ?? new Dec(0))
+        ?.sub(fiatAmountWithSlippage ?? new Dec(0))
+        ?.toDec()
+        .isNegative(),
```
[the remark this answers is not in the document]
greg-nagy: 4231622285f4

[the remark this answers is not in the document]
greg-nagy: the change was bad and removed it with a force push

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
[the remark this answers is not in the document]
greg-nagy: 151807e97c7a fixed

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
```diff
@@ -0,0 +1,271 @@
...
+    transactionParams,
+  ]);
+
+  const commitSessionChange = useCallback(() => {
+    if (!shouldSend1CTTx) {
+      cleanUpAndCommit();
+      return;
+    }
+
+    if (!transactionParams) return;
```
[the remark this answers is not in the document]
greg-nagy: it's better not to do this

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
```diff
@@ -0,0 +1,267 @@
...
+        userOsmoAddress: account?.address ?? "",
+        publicKey: oneClickTradingInfo?.publicKey ?? "",
+      },
+      {
+        enabled: enabled && shouldFetchExistingSessionAuthenticator,
+        cacheTime: 15_000, // 15 seconds
+        staleTime: 15_000, // 15 seconds
+        retry: false,
+      }
+    );
```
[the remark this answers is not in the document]
greg-nagy: trpc error wrapper should already handle this

packages/web/modals/review-order.tsx
```diff
@@ -210,481 +248,682 @@ export function ReviewOrder({
...
+                  t={t}
+                  transactionParams={transaction1CTParams}
+                  onClick={() =>
+                    setTransaction1CTParams((prev) => {
+                      if (!prev) return;
+
+                      return {
+                        ...prev,
+                        isOneClickEnabled: !prev.isOneClickEnabled,
+                      };
```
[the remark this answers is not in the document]
greg-nagy: the only scenario where this can be undefined is when we can't set default value because we haven't got the default param from the local trpc resolver `api.local.oneClickTrading.getParameters`

so we can't set a default value -> if was undefined we want to keep it undefined

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
```diff
@@ -0,0 +1,267 @@
...
+    createSession,
+    spendLimitTokenDecimals,
+    accountStore,
+    chainStore.osmosis.chainId,
+    sessionAuthenticator,
+    cleanUpAndCommit,
+    setTransactionParams,
+    initialTransactionParams,
+    logEvent,
+  ]);
```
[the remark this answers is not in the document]
greg-nagy: I don't have the error states for this, but adding something along these lines seem like a good idea

JoseRFelix:
@greg-nagy Thank you for this PR! I’ll review it soon. By the way, about this

> ⚠️ Note: An infinite rerender issue exists when an amount is entered into the swap input. This was introduced in https://github.com/osmosis-labs/osmosis-frontend/pull/3949 and will affect this PR

I can’t replicate this issue in production. Could you provide some steps to reproduce it? If not, it might be related to the changes in this PR

kamal-sutra:
@greg-nagy Clicking on 'Increase Spend Limit' doesn't pop-up anything
image

packages/web/utils/formatter.ts
```diff
@@ -392,3 +393,9 @@ export function calcFontSize(numChars: number, isMobile: boolean): string {
 
   return "16px";
 }
+
+export function formatSpendLimit(spendLimit: PricePretty | undefined) {
```
jonator, line 397: Nit: thoughts on a more generic name for readability? A name that describes what it's doing more than what data it's displaying: like trimmedPrice or something

packages/web/modals/review-order.tsx
```diff
@@ -98,7 +105,38 @@ export function ReviewOrder({
...
+    resetParams: reset1CTParams,
+    commitSessionChangeIsLoading: commit1CTSessionChangeIsLoading,
+  } = useOneClickTradingSessionManager({
+    onCommit: () => {
+      confirmAction();
+      onClose();
+    },
+  });
+
+  const wouldExceedSpendLimit = useMemo(
```
jonator, line 130: thoughts on lifting this calculation into `useOneClickTradingSessionManager`? If you believe it would be useful elsewhere or in the future? Perhaps as a function if outside data is needed

jonator (commented):
LGTM at a glance, but would prefer review from @JoseRFelix as he's the original author of 1CT.

Also, I'm wondering- is it possible to then execute the swap via 1CT instead of via wallet since they just enabled it?

greg-nagy:
@kamal-sutra thanks, I missed this edge case and didn't handle it properly.  (I didn't consider that some token values are not counted in the spending limit)

greg-nagy:
> Also, I'm wondering- is it possible to then execute the swap via 1CT instead of via wallet since they just enabled it?

@jonator, I also wanted to do this, but I wanted to know if it's okay to increase the gas amount and how to do it before the authenticator is properly registered. So, I went with the simpler solution for now.

packages/web/modals/review-order.tsx
greg-nagy: hm, good thought

originally didn't want to mix the current swap concerns with session manager, but it also makes sense to encapsulate this logic
let me see if I can find a good place for it

packages/web/utils/formatter.ts
greg-nagy: f8273fb95988

thanks,

packages/web/components/one-click-trading/one-click-trading-settings.tsx
```diff
@@ -170,11 +201,22 @@ export const OneClickTradingSettings = ({
...
-          nextParams: nextParams!,
-        })
+        Array.from(
+          new Set([
+            ...changes,
+            ...compare1CTTransactionParams({
+              prevParams: initialTransaction1CTParams!,
+              nextParams: nextParams!,
+            }),
+          ])
```
JoseRFelix, line 218: What situations are we preventing here? Could you add a comment explaining why we need to spread the `changes` state?

packages/web/utils/formatter.ts
```diff
@@ -392,3 +393,9 @@ export function calcFontSize(numChars: number, isMobile: boolean): string {
 
   return "16px";
 }
+
+export function trimmedPriceWithSymbol(price: PricePretty | undefined) {
+  return `${price?.symbol}${trimPlaceholderZeros(
+    price?.toDec().toString(2) ?? ""
+  )}`;
+}
```
JoseRFelix, line 401: I prefer not to introduce another general formatting option since we already have formatPretty. This function, originally named formatSpendLimit, should keep its name because it is specifically used for displaying the spend limit. I don’t foresee any future use cases for an alternative. WDYT?

packages/web/components/one-click-trading/one-click-trading-settings.tsx
```diff
@@ -147,7 +176,9 @@ export const OneClickTradingSettings = ({
   useEffect(() => {
     if (!transaction1CTParams || initialTransaction1CTParams) return;
     setInitialTransaction1CTParams(transaction1CTParams);
-  }, [initialTransaction1CTParams, transaction1CTParams]);
+    setChanges(externalChanges);
+    setInitialChanges(externalChanges);
```
JoseRFelix, line 180: To avoid maintaining two sources of truth, I recommend using controlled inputs to override the local changes state when necessary. We have a handy hook called useControlledState that simplifies this process. For documentation, refer to: [Chakra UI - useControllable Hook](https://v2.chakra-ui.com/docs/hooks/use-controllable).

packages/web/hooks/one-click-trading/__tests__/use-one-click-trading-params.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+    new Dec(DEFAULT_SPEND_LIMIT)
+  ),
+  spendLimitTokenDecimals: DEFAULT_USDC_DECIMALS,
+  networkFeeLimit: OneClickTradingMaxGasLimit,
+  sessionPeriod: {
+    end: DEFAULT_SESSION_PERIOD,
+  },
+};
+
+// tsw-trpc did not deserialize the spendLimit prop, so we mock it here
```
JoseRFelix, line 62:
```suggestion
// msw-trpc did not deserialize the spendLimit prop, so we mock it here
```

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -39,90 +45,143 @@ export function getParametersFromOneClickTradingInfo({
...
-        })
-      : undefined
+  >();
+  const [current, setCurrent] = useState<
+    OneClickTradingTransactionParams | undefined
+  >();
+
+  const [changes, setChanges] = useState<OneClickTradingParamsChanges>([]);
+
+  const sessionOrDefault = useMemo(
```
JoseRFelix, line 72:
```suggestion
  const sessionOrDefaultParams = useMemo(
```

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -39,90 +45,143 @@ export function getParametersFromOneClickTradingInfo({
...
+  );
+
+  const sessionOrDefaultWithEnabled = useMemo(
+    () =>
+      sessionOrDefault && {
+        ...sessionOrDefault,
+        isOneClickEnabled: defaultIsOneClickEnabled,
+      },
+    [sessionOrDefault, defaultIsOneClickEnabled]
   );
```
JoseRFelix, line 90: I suggest merging this with sessionOrDefault since we can simply avoid using `isOneClickEnabled`

```suggestion
```

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -39,90 +45,143 @@ export function getParametersFromOneClickTradingInfo({
...
-    data: defaultTransaction1CTParams,
-    isLoading,
-    isError,
-  } = api.local.oneClickTrading.getParameters.useQuery();
+} = {}) {
+  const { data: defaultParams, isLoading } =
+    api.local.oneClickTrading.getParameters.useQuery();
 
-  const [transaction1CTParams, setTransaction1CTParams] = useState<
+  const [draft, setDraft] = useState<
```
JoseRFelix, line 63:
```suggestion
  const [draftParams, setDraftParams] = useState<
```

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -39,90 +45,143 @@ export function getParametersFromOneClickTradingInfo({
...
-    oneClickTradingInfo
-      ? getParametersFromOneClickTradingInfo({
-          oneClickTradingInfo,
-          defaultIsOneClickEnabled,
-        })
-      : undefined
+  >();
+  const [current, setCurrent] = useState<
+    OneClickTradingTransactionParams | undefined
+  >();
```
JoseRFelix, line 68:
```suggestion
  const [currentParams, setCurrentParams] = useState<
    OneClickTradingTransactionParams | undefined
  >();
```

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -39,90 +45,143 @@ export function getParametersFromOneClickTradingInfo({
...
+  if (prev?.isOneClickEnabled !== next?.isOneClickEnabled) {
+    changes.push("isEnabled");
+  }
+
+  if (prev?.networkFeeLimit !== next?.networkFeeLimit) {
+    changes.push("networkFeeLimit");
+  }
+
+  return changes;
+}
```
JoseRFelix, line 187: I believe a similar method already exists in one-click-trading-settings. I recommend reusing it

packages/web/hooks/one-click-trading/use-one-click-trading-session.ts
```diff
@@ -35,7 +35,10 @@ export const useOneClickTradingSession = ({
     };
 
     if (!account?.address) {
-      return defaultReturn;
+      // Return a promise that never resolves so loading is kept true
+      // until we can decide what to share
+      // without address 1CT session should not exist
+      return new Promise<typeof defaultReturn>(() => {});
     }
```
JoseRFelix, line 42: It doesn’t seem logical to return isLoading as true when we know for certain that the session is `undefined`. Is this intended for a specific view? If so, we can simply set `isLoading` to true in that context when the session is `undefined`.

packages/web/hooks/one-click-trading/use-one-click-trading-session.ts
```diff
@@ -97,7 +100,7 @@ export const useOneClickTradingSession = ({
   return {
     oneClickTradingInfo: featureFlags.oneClickTrading ? value?.info : undefined,
     isOneClickTradingEnabled: featureFlags.oneClickTrading
-      ? value?.isEnabled
+      ? value?.isEnabled ?? false
```
JoseRFelix, line 103: With the previous suggested comment, this would no longer be necessary

packages/web/hooks/use-swap.tsx
JoseRFelix, line 421: I wonder if, when enabling the 1CT session in the swap tool, we could wait for the 1CT session to become active before opening a second transaction modal

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
```diff
@@ -0,0 +1,267 @@
...
+    transactionParams,
+    amountSpent,
+    remainingSpendLimit,
+    setTransactionParams,
+    commitSessionChange,
+    resetParams,
+  };
+}
+
+export function useRemainingSpendLimit({
```
JoseRFelix, line 211:
```suggestion
export function useOneClickRemainingSpendLimit({
```

JoseRFelix (changes requested):
Thank you for this PR! I left some comments

packages/web/utils/formatter.ts
greg-nagy: makes sense

this was not a good place for this formatter

ty

packages/web/hooks/one-click-trading/use-one-click-trading-session.ts
greg-nagy: 122440450425

packages/web/utils/formatter.ts
greg-nagy: 122440450425

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
greg-nagy: 122440450425

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
greg-nagy: 122440450425

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
```diff
@@ -0,0 +1,267 @@
...
+  const { data: amountSpentData } =
+    api.local.oneClickTrading.getAmountSpent.useQuery(
+      {
+        authenticatorId: oneClickTradingInfo?.authenticatorId!,
+        userOsmoAddress: oneClickTradingInfo?.userOsmoAddress!,
+      },
+      {
+        enabled: enabled && !!oneClickTradingInfo,
+      }
+    );
```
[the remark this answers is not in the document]
greg-nagy: 122440450425

packages/web/components/one-click-trading/one-click-trading-settings.tsx
greg-nagy: ok, this was the most difficult request so far :)

packages/web/hooks/use-swap.tsx
greg-nagy: based on the fact that both of you have asked for this I will assume that the changes in the gas fee is not a problem 

the naive approach of just waiting a bit did not work for me previously, but I will give it a more through check

packages/web/hooks/one-click-trading/use-one-click-trading-params.ts
```diff
@@ -39,90 +46,115 @@ export function getParametersFromOneClickTradingInfo({
...
-  }, [
-    defaultIsOneClickEnabled,
-    defaultTransaction1CTParams,
-    initialTransaction1CTParams,
-    oneClickTradingInfo,
-  ]);
+      );
+    },
+    [currentParams, reset, draftParams]
+  );
```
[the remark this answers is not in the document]
greg-nagy: resetting is not an error path but a safety measure and a recovery method here 
no need for extra logs

packages/web/components/one-click-trading/one-click-trading-settings.tsx
greg-nagy: Context: I have a habit of leaving details and explanations in commit messages for the given change so it's immediately obvious clicking on a line why a change was made.

I would like to understand if this is a good practice on this project or should I prefer inline comments for similar explanations?

packages/web/components/one-click-trading/one-click-trading-settings.tsx
greg-nagy: Wow, this is a nice pattern! I will be using this in the future, for sure!

But it does not work in this particular case. These are not two sources of truth but a draft and a draft of a draft. 
We create a draft, the `transaction1CTParams`, with its changes. This stores the value set by the user until the confirm button is pressed. 

When the user opens the edit modal, another draft is created, which can be saved by pressing back and discarded by pressing `X`.  This lets the user edit the parameters multiple times before sending the actual tx. 

The implementation is optimistic — we edit the main draft during the process by default but revert the changes and the main draft to the snapshot they were in when the modal was opened if they user decides to discard the changes.

CleanShot 2024-11-19 at 19 33 32


If I misunderstood what you meant and the `useControlledState` makes sense, please let me know and I will try to make it work

packages/web/hooks/use-swap.tsx
greg-nagy: Ok, so solving this ended up being a bigger thing than I hoped it to be, but it was lots of fun to tackle :) 

I have a simple wait solution in 70258a9060a0 but that solution felt uncomfortably brittle so I ended up reworking the mobx <> 1CT session synchronization in 2758eef306bc and 747dbc3c406c

packages/web/modals/review-order.tsx
greg-nagy: found this a nice and cozy new home in 675bb30deac0

greg-nagy:
@CryptoAssassin1 This feature should be hidden on mobile — the same as the current version. 

I assume this hiding logic is encoded in the feature flag from launch darkly, but I can't be sure.

greg-nagy:
@kamal-sutra, thanks for catching the spend limit issue 

I have removed that warning from the swap-tool because the session can now be updated through the review order modal.

packages/web/hooks/one-click-trading/use-one-click-trading-session.ts
```diff
@@ -67,12 +86,15 @@ export const useOneClickTradingSession = ({
     const timeoutId = setTimeout(() => {
       if (!value?.info) return;
 
-      setIsExpired(true);
+      // trigger mobx store to update when session expires
+      // (isExpired is a computed value in the store)
+      runInAction(() => {});
+
```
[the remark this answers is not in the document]
greg-nagy: it would introduce more problems than it would solve

Storing the expired flag in the store would trigger reactivity but would be used only as a trigger since the isExpired needs to be calculated. This would introduce a prop that's persisted and tracked without any real usage.

packages/stores/src/account/base.ts
```diff
@@ -216,13 +219,15 @@ export class AccountStore<Injects extends Record<string, any>[] = []> {
     makeObservable(this);
 
     autorun(async () => {
-      const isOneClickTradingEnabled = await this.getShouldUseOneClickTrading();
+      const isOneClickTradingEnabled = await this.isOneCLickTradingEnabled();
```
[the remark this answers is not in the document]
greg-nagy: nice 

425b49f5b4a0

greg-nagy:
I have this cloudflare issue with the `/api/edge-trpc-assets/assets.getAssetPrice`  from sidecar. I don't see any problem in the code, so I assume something has changed in the sidecar. 

I did my best to test the latest changes, but because of this issue, I am not sure if I could test everything. 

```
GET /api/edge-trpc-assets/assets.getAssetPrice?batch=1&input=%7B%220%22%3A%7B%22json%22%3A%7B%22coinMinimalDenom%22%3A%22uosmo%22%7D%7D%7D 200 in 747ms
Fetch Error. Info: {
  endpoint: 'https://stage-proxy-data-api.osmosis-labs.workers.dev/tokens/v2/historical/uosmo/chart?tf=60',
  config: { method: 'GET', body: undefined },
  status: 404,
  exception:  [Error: JSON parse error: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON] {
  status: 404,
  data: {
  url: 'https://stage-proxy-data-api.osmosis-labs.workers.dev/tokens/v2/historical/uosmo/chart?tf=60'
},
```

packages/web/modals/review-order.tsx
```diff
@@ -210,481 +257,684 @@ export function ReviewOrder({
...
-              <div className="flex w-full justify-between gap-3 pt-3">
+                <span className="subtitle1 w-full">
+                  {t("limitOrders.quoteUpdated")}
+                </span>
                 <Button
                   mode="primary"
-                  onClick={confirmAction}
-                  disabled={isConfirmationDisabled}
-                  className="body2 sm:caption !rounded-2xl"
+                  onClick={() => restart}
```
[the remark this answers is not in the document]
greg-nagy: adecb3e1b719

JoseRFelix:
Ending session from standalone modal is displaying the save changes button which is not expected. See prod. 

This PR 
image

Prod: 

image

packages/web/hooks/one-click-trading/use-one-click-trading-session.ts
```diff
@@ -22,42 +21,62 @@ export const useOneClickTradingSession = ({
...
+        }
+        setLoading(false);
+      },
+      {
+        fireImmediately: true,
+      }
+    );
+
+    return () => disposer();
+  }, [account?.address, accountStore]);
```
JoseRFelix, line 76: I prefer the previous solution because it doesn’t rely entirely on MobX. Since we plan to migrate the account store from MobX to React, retaining the previous syntax will be highly beneficial as it can help us avoid additional refactoring in the future. Do you think it's possible to keep that syntax or is using a `reaction` required?

packages/web/hooks/use-swap.tsx
```diff
@@ -371,11 +379,44 @@ export function useSwap(
...
+          // After retries, check if still loading
+          if (isLoadingNetworkFeeRef.current) {
+            return reject(
+              new Error("Network fee is still loading after 1 second")
+            );
+          }
+
+          if (networkFeeErrorRef.current) {
+            return reject(new Error(networkFeeErrorRef.current.message));
+          }
```
JoseRFelix, line 418: This works, but it feels a bit hacky. I wonder if we could use a global state with Zustand to check if the review modal’s 1CT is enabled? Then, we could construct a multi-message within the swap hooks. I have a hunch this approach would be more accurate and require less code. What do you think?

JoseRFelix (changes requested):
Thank you for addressing the previous comments! I’ve left a few more suggestions. We’re getting very close now

packages/web/hooks/one-click-trading/use-one-click-trading-session.ts
greg-nagy: It's possible to keep this syntax and make the feature work see 70258a9060a0

I have left a note about how that flow is triggered in 2d4225083a5b 
and reverted the related changes (7f5e878bf785, 4fc31a0268d8) 

I wanted to create a robust fix because the wait loop felt hacky.
However, I also had concerns about whether that was the direction you wanted to take.
Thanks for clearing this up!

packages/web/hooks/use-swap.tsx
greg-nagy: > This works, but it feels a bit hacky. 

Indeed :) I don't like it either. 

>  I wonder if we could use a global state with Zustand to check if the review modal’s 1CT is enabled? Then, we could construct a multi-message within the swap hooks.

This could work in the normal user flow and allow the user to `Continue Anyway` with the swap, even if the 1CT session has an overspend limit error. I like this idea a lot, but if we were to do it, I would prefer to tackle it in another PR. 

TL;DR;
This is me being extra cautious to prevent swap tx failure after enabling 1CT through the review order flow. 
It is rarely necessary, and we could revert the whole change if it's too much/too hacky.  

Full context: 

Using 1CT needs more gas, and the swap tx would fail with the old gas estimate after 1CT is enabled.   

When 1CT is enabled through the review order flow, the useSwap hook needs to update its internal state while the `createSession.mutate` and its `onSuccess` callback are running.

We wait for the session to update in the `onSuccess,` but the network fee invalidation in `useSwap` starts after the session update  (ecd25aa9ab83). 

In the `onSuccess`,  we don't have any info about when the `useSwap` will be ready. 

So we either
1. Take the "readiness signal" from `SwapTool > swapState` -> `ReviewOrder` -> `useOneClickTradingSessionManager` 
2. Wait for an arbitrary amount of time and hope that gas estimation will be ready after that. 
3. Make `SwapTool > sendSwapTx` smart enough to wait 
4. Make the `useSwap > sendTradeTokenInTx` smart enough to wait 
5. Ditch the safety mechanism 

I chose 3. to implement this protection with the rest of the other checks to ensure that the swap tx can be executed. 

What do you think would be a good next step?

packages/web/modals/review-order.tsx
```diff
@@ -210,481 +256,684 @@ export function ReviewOrder({
...
+          onClick={onClick}
+          size="xsm"
+          className="text-wosmongton-500 rounded-3xl"
+        >
+          {t("oneClickTrading.reviewOrder.edit")}
+        </UIButton>
+        <GenericDisclaimer
+          title={t("oneClickTrading.reviewOrder.exceededTooltip.title")}
+          body={t("oneClickTrading.reviewOrder.exceededTooltip.body")}
+        >
```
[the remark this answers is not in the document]
greg-nagy: not sure if accessibility is a concern for us

packages/web/modals/review-order.tsx
```diff
@@ -210,481 +256,684 @@ export function ReviewOrder({
...
+  onClick: () => void;
+  onParamsChange: () => void;
+}) => {
+  return (
+    <>
+      <div className="flex flex-col my-3">
+        <div
+          className="flex gap-4 rounded-2xl bg-osmoverse-alpha-800 px-4 py-3"
+          onClick={onClick}
+        >
```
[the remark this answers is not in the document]
greg-nagy: not sure if accessibility is a concern for us

packages/web/components/alert/prettify.ts
```diff
@@ -27,6 +27,24 @@ const regexRejectedTx = /Request rejected/;
...
+  const match = message.match(regexOverspendError);
+  if (!match) return;
+
+  const [, wouldSpendTotal, limit] = match;
+
+  if (!wouldSpendTotal || !limit) return;
+
+  return {
+    wouldSpendTotal: new Dec(wouldSpendTotal, 6),
+    limit: new Dec(limit, 6),
```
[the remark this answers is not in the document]
greg-nagy: hm, `uusdc` has 6 decimals and this felt like a safe bet, but sure if this is how it should be handled

packages/web/hooks/one-click-trading/use-one-click-trading-session-manager.ts
```diff
@@ -0,0 +1,315 @@
...
+                  )
+                );
+              } else {
+                retries++;
+                setTimeout(checkIsEnabled, 100);
+              }
+            };
+            checkIsEnabled();
+          });
+          onCommitRef.current();
```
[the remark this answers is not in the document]
greg-nagy: f8ec361d3dd1

greg-nagy:
> Ending session from standalone modal is displaying the save changes button which is not expected.

776b70b9b735

packages/web/hooks/use-swap.tsx
MaxMillington: imo we should keep this as is and address this in a separate PR

jonator (changes requested):
@greg-nagy build is failing

greg-nagy:
@jonator I had a linter error — fixed it now. Sorry about that 

Other than that, I can't validate if the build works well because the `https://stage-proxy-data-api.osmosis-labs.workers.dev/tokens/v2/all` throws an error on my machine 

```
@osmosis-labs/web:build:  ✓ Compiled successfully
   Collecting page data  ..Fetch Error. Info: {
@osmosis-labs/web:build:   endpoint: 'https://stage-proxy-data-api.osmosis-labs.workers.dev/tokens/v2/all',
@osmosis-labs/web:build:   config: { method: 'GET', body: undefined },
@osmosis-labs/web:build:   status: 404,
@osmosis-labs/web:build:   exception: a [Error]: JSON parse error: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
@osmosis-labs/web:build:       at /Users/greg/dev/osmosis/osmosis-frontend/packages/web/.next/server/chunks/1896.js:7:437527
@osmosis-labs/web:build:       at Generator.throw (<anonymous>)
@osmosis-labs/web:build:       at r (/Users/greg/dev/osmosis/osmosis-frontend/packages/web/.next/server/chunks/1896.js:7:436021)
@osmosis-labs/web:build:       at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
@osmosis-labs/web:build:     status: 404,
@osmosis-labs/web:build:     data: {
@osmosis-labs/web:build:       url: 'https://stage-proxy-data-api.osmosis-labs.workers.dev/tokens/v2/all'
@osmosis-labs/web:build:     },
@osmosis-labs/web:build:     response: Response {
@osmosis-labs/web:build:       [Symbol(realm)]: null,
@osmosis-labs/web:build:       [Symbol(state)]: [Object],
@osmosis-labs/web:build:       [Symbol(headers)]: [HeadersList]
@osmosis-labs/web:build:     }
@osmosis-labs/web:build:   }
```

CleanShot 2024-11-25 at 16 00 40

MaxMillington:
+100

PR #3955, opened 2024-11-15
Move pagination + sorting + query filtering of /pools endpoint to SQS

deividaspetraitis:

This page indroduces changes to /pools page. Until now, FE would fetch all available pools and will do relevant manipulations to display this data correctly to end user. This approach resulted in high load times, slowed down rendering of pools page. Introduces changes moves all relevant manipulations and mutations of /pools page to SQS to keep FE fast and lightweight.

### Linear Task

[BE-636](https://linear.app/osmosis/issue/BE-636/add-pagination-sorting-query-filtering-for-pools-endpoint)

## Brief Changelog

FE is no longer responsible for:

- Pagination
- Filtering
- Sorting
- Searching

## Testing and Verifying

This change has been tested locally by rebuilding the website and verified content and links are expected including [QA](https://osmosis-network.slack.com/archives/C03JT90769K/p1732725079083929) session.

You can access deployed instance here: https://osmosis-frontend-dev-git-be-636-osmo-labs.vercel.app/


packages/server/src/queries/complex/pools/index.ts
```diff
@@ -91,72 +117,20 @@ export async function getPool({
...
-    }
-  }
-
-  return denomPools;
+): Promise<{
+  items: Pool[];
+  total: number;
+  nextCursor: number | undefined;
+}> {
+  params.notPoolIds = FILTERABLE_IDS;
```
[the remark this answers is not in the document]
DavideSegullo: I agree, if this is still something we need to apply this suggestion or `notPoolIds` can't be overrided

packages/server/src/queries/complex/pools/providers/sidecar.ts
```diff
@@ -27,51 +33,83 @@ const ExcludedExternalBoostPools: string[] =
...
+  sort?: SortType;
+}): Promise<PoolProviderResponse> {
+  if (poolIds && !poolIds.length) {
+    return Promise.resolve({ data: [], total: 0, nextCursor: undefined });
+  }
 
   return cachified({
     cache: poolsCache,
     key:
+      JSON.stringify(assetLists) +
```
DavideSegullo, line 64: I think this can create some memory issues, because assetList is a very large array of object, if we need this maybe is better to just store asset unique identifier

DavideSegullo (changes requested):

packages/server/src/queries/complex/pools/providers/sidecar.ts
deividaspetraitis: Good catch! No, it's not needed.

packages/server/src/queries/complex/pools/index.ts
```diff
@@ -91,72 +117,20 @@ export async function getPool({
...
-    }
-  }
-
-  return denomPools;
+): Promise<{
+  items: Pool[];
+  total: number;
+  nextCursor: number | undefined;
+}> {
+  params.notPoolIds = [...(params.notPoolIds ?? []), ...FILTERABLE_IDS];
```
DavideSegullo, line 125: Maybe we can just do here:
```ts
params.notPoolIds = !params.notPoolIds ? FILTERABLE_IDS : params.notPoolIds
```

This ensure we set `FILTERABLE_IDS` as default value, but then you can totally override it, does it makes sense?

deividaspetraitis: Yea, it makes sense! Adjusted as per you suggestion, thank you!

DavideSegullo (approved):

deividaspetraitis:
> Thank you! I will leave this open for now until we merge SQS related changes and deploy to stage, merging now would result in not working pools page.

PR #3957, opened 2024-11-17
(DevX) Create lock file to manage asset list image cache

JoseRFelix:

Previously, we disabled caching to avoid token image conflicts when switching environments. However, since waiting for image downloads to start the server is time-consuming and tedious, we propose creating a lock file inside the generated folder. This lock file will store the current asset list hash and override the cache if the hash changes.

### Linear Task

https://linear.app/osmosis/issue/FE-1228/create-lock-file-to-manage-asset-list-images-cache

## Testing and Verifying

- [ ] Asset list is generated correctly

### DIFF
--- a/packages/web/config/generate-lists.ts
+++ b/packages/web/config/generate-lists.ts
@@ -19,6 +19,7 @@ import type {
   IbcTransferMethod,
 } from "@osmosis-labs/types";
 import { isNil } from "@osmosis-labs/utils";
+import * as fs from "fs";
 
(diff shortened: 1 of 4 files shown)


JoseRFelix:
>How are token images persisted across builds with a lockfile? They're still downloaded each time, just from the locked asset list?

In the production build, images will always be downloaded. In the development environment, images will only be downloaded if the asset list commit has changed or if the file is missing

PR #3958, opened 2024-11-18
Convert variants: check for large difference in amount

jonator:

Add more user protection imposing a limit on price impact when converting non-alloy variants, such as USDC.

If the _amounts_ in vs out differ by more than 5% in the quote, block the conversion tx from being possible. Since we're converting like kinds, we can assume the amounts (with respective decimals in place) should be about the same. This is instead of comparing USD values, which may not be accurate if there's issues getting prices.

### Linear Task

Hot fix.

## Brief Changelog

- Include in/out amount diff slippage check in isError flag

## Testing and Verifying

Tested small change locally

### DIFF
--- a/packages/web/components/table/portfolio-asset-balances.tsx
+++ b/packages/web/components/table/portfolio-asset-balances.tsx
@@ -572,7 +572,7 @@ const AssetActionsCell: AssetCellComponent<{
       {needsActivation && (
         <Button
           variant="secondary"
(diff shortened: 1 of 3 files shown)


DavideSegullo (approved):

JoseRFelix (changes requested):
Looks great! I believe we just need a unit test

CryptoAssassin1 (approved):

JoseRFelix (approved):

PR #3960, opened 2024-11-19
(Pool creation) Supercharged Pool creation is missing several assets

JoseRFelix:

The supercharged pool creation interface currently lacks several assets due to the absence of the `onlyVerified` and `showPreviewAssets` inputs. This PR addresses the issue by ensuring assets are displayed according to the user’s preferences.

### Linear Task

https://linear.app/osmosis/issue/FE-964/supercharged-pool-creation-is-missing-several-assets

## Testing and Verifying

- [ ] Can see preview and unverified assets based on users' preference in the create supercharged pool asset select

### DIFF
--- a/packages/trpc/src/concentrated-liquidity.ts
+++ b/packages/trpc/src/concentrated-liquidity.ts
@@ -106,34 +106,42 @@ export const concentratedLiquidityRouter = createTRPCRouter({
   getLiquidityPerTickRange: publicProcedure
     .input(z.object({ poolId: z.string() }))
     .query(({ ctx, input }) => getLiquidityPerTickRange({ ...ctx, ...input })),
-  getBaseTokens: publicProcedure.query(({ ctx }) =>
-    getAssets({
-      assetLists: ctx.assetLists,
-      onlyVerified: true,
-    })
-      .map((asset) => {
-        const assetListAsset = getAssetFromAssetList({
-          assetLists: ctx.assetLists,
-          coinMinimalDenom: asset.coinMinimalDenom,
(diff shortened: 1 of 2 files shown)


packages/trpc/src/concentrated-liquidity.ts
```diff
@@ -106,34 +106,42 @@ export const concentratedLiquidityRouter = createTRPCRouter({
...
-            },
-            0
-          ).currency,
-        };
-      })
-      .filter((asset): asset is NonNullable<typeof asset> => Boolean(asset))
-  ),
+          return {
+            chainName: assetListAsset.rawAsset.chainName,
+            token: new CoinPretty(
```
jonator, line 132: Why not just return the currency object directly instead of use CoinPretty?

packages/web/components/complex/pool/create/cl/set-base-info.tsx
```diff
@@ -43,12 +45,24 @@ export const SetBaseInfos = observer(
   }: SetBaseInfosProps) => {
     const { t } = useTranslation();
 
-    const { accountStore } = useStore();
+    const { accountStore, userSettings } = useStore();
```
jonator, line 48: Nice

jonator (approved):

packages/trpc/src/concentrated-liquidity.ts
JoseRFelix: Ha ha, that's right. I'm not sure what happened there. I'll push a commit

PR #3961, opened 2024-11-20
(Swap Tool) Add top gainers

JoseRFelix:

https://github.com/user-attachments/assets/a08c5479-fc08-4d22-bc6a-050f20a0a4a9

### Linear Task

https://linear.app/osmosis/issue/FE-1229/add-trending-tab-to-the-trade-page

## Testing and Verifying

- [ ] Can see top gainers below swap tool

### DIFF
--- a/packages/web/components/table/asset-info.tsx
+++ b/packages/web/components/table/asset-info.tsx
@@ -19,6 +19,7 @@ import {
   useMemo,
   useState,
 } from "react";
+import { useMount } from "react-use";
 
 import { HighlightsCategories } from "~/components/assets/highlights-categories";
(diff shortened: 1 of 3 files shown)


packages/web/components/table/asset-info.tsx
```diff
@@ -97,6 +97,15 @@ export const AssetsInfoTable: FunctionComponent<{
     [selectedCategory]
   );
 
+  useMount(() => {
```
DavideSegullo, line 100: I think we have nuqs installed, maybe we can use that instead on this? That would be good so we remove location search params sync logic and useState ecc.

jonator (approved):

PR #3977, opened 2024-11-28
Dev/small speedup to test if we changed right thing

ValarDragon:
Start speeding up IntPretty/PricePretty constructors which are shown to be quite expensive in prod.

This is trying to first initially validate if this is editing the right spot (as profiles look fairly different on vercel preview vs prod), before going for larger wins.


ValarDragon:
Confirmed this is speeding up the right spot. That first commit dropped the IntPretty constructor time down by 8%. Will proceed going for more significant reductions

ValarDragon:
Its hard to apples to apples compare across feature branches (And that the numbers are way different than what I get on stage) but at least on a profile before and after the last commit, i got what appears to be a 3x speedup on the IntPretty load times. (Many IntPretty clones were called before, but now we save expensive dec logic in each)

If its possible would like to get to prod and measure whats the impact. I'll continue with some more slight improvements around, but I think this is likely an 80-20 for impact.

Of the 4.4s refresh time I saw on main, this should eliminate at least 2s. I'd like to rebenchmark to see if more was removed

PR #3992, opened 2024-12-05
Remove transaction scan

jonator:
### Linear Task

Removes transaction scan. See issue for details


packages/web/components/place-limit-tool/index.tsx
```diff
@@ -562,6 +562,7 @@ export const PlaceLimitTool: FunctionComponent<PlaceLimitToolProps> = observer(
       swapState.isMarket,
       swapState.marketState.networkFeeError,
       swapState.marketState.isSlippageOverBalance,
+      account?.isWalletConnected,
```
DavideSegullo, line 565: Why do we also remove this here? Is it related to something?

DavideSegullo (approved):

packages/web/components/place-limit-tool/index.tsx
jonator: It fixed a lint error

DavideSegullo: Oh got it, thank you!

PR #4016, opened 2024-12-17
Show last chart data as non-hover price on token info page

fabryscript:


EnricoBarbieri1997 (approved):

DavideSegullo (approved):

packages/web/components/pages/asset-info-page/chart.tsx
```diff
@@ -188,9 +188,7 @@ const ChartHeader = observer(() => {
       data = decHoverPrice;
     } else {
       if (assetInfoConfig.dataType === "price") {
-        data =
-          marketAsset?.currentPrice?.toDec() ??
-          new Dec(assetInfoConfig.lastChartData?.close ?? 0);
+        data = new Dec(assetInfoConfig.lastChartData?.close ?? 0);
```
JoseRFelix, line 191: Have we verified whether this price reflects the most current value? I’m concerned that discrepancies might appear throughout the app

JoseRFelix (changes requested):

packages/web/components/pages/asset-info-page/chart.tsx
EnricoBarbieri1997: There has been a discussion going on with Numia talking about that. As of now data from Numia and SQS differ so the last point on the chart and the price displayed when no hover is performed are different and inconsistent. They tried to make the SQS value the same as Numia but they ultimately decided to just use the Numia value in this page

KamaIOps (approved):

CryptoAssassin1 (approved):

packages/web/components/pages/asset-info-page/chart.tsx
alvarogar4: Hey! The discussion mentioned above was related to the portfolio page, but the problem seems to be similar in this case

However, there's a small but important difference:
- In the portfolio endpoint the last value comes from the LCD endpoints to ensure real time values
- In the asset price charts, the last datapoint comes from our database because its computationally heavy to do it on request, so with the current state we can't assure real time data. We could potentially try to find a solution if you really need it (or at least try to reduce the delays as much as possible)

The delays overall shouldn't be too long in any case, but I just want to make sure we are all aligned before pushint anything to prod. 

Let me know if you want to explore any potential improvements

PR #4026, opened 2024-12-28
feat: twinned swap input hook

crnbarr93:

These changes refactor the buy/sell tab inputs to use a new hook: `useTwinnedSwapInput`. This hook takes both a fiat and token input along with a price and adjusts the values of each depending on the input of the other in relation to the price. For example at a token price of $0.50 if the user input $1 then the token amount would be 2. This change was to make the code more readable/maintainable and to also prevent infinite rerendering caused by the use of `useEffect`. This hook now primarily makes use of callbacks instead of `useEffect`, with the exception of the price updating as that can occur with no input.

### Linear Task

[FE-1273](https://linear.app/osmosis/issue/FE-1273/fix-infinite-render-bug-in-limit-orders-tab)

## Brief Changelog

- Added a new `useTwinnedSwapInput` hook
- Refactored several sections of `PlaceLimitTool`
- Moved several number functions out of `PlaceLimitTool`

## Testing and Verifying
Manually tested and added unit tests for the new hook.


packages/web/hooks/use-twinned-swap-input.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  const handleTokenAmountChange = useCallback(
+    (value: string) => {
+      if (isEmptyString(value)) {
+        reset();
+        return;
+      }
+
+      if (price.isNegative()) {
+        console.error("INPUT WARNING: Price is negative");
```
crnbarr93, line 99: Should this throw an error?

jonator (changes requested):
I'd love to get this in, but it seems we're having CORS issues with whatever SQS endpoint is in use here:

Screenshot 2025-04-10 at 10 13 28 AM

PR #4027, opened 2025-01-06
Add MoonPay onramp option

EnricoBarbieri1997:

Add MoonPay as an option to buy assets with FIAT

### Linear Task

https://linear.app/osmosis/issue/FE-1270/review-moonpay-proposal-and-consider-implementation

## Brief Changelog

- Add MoonPay brandin
- Add MoonPay Localization

## Testing and Verifying

This change has not been tested locally, because some functionalities are available on production environment only


packages/web/integrations/moonpay/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+    )
+  ).signature;
+}
+
+export const Moonpay: FunctionComponent<
+  { assetKey: string } & Pick<ModalBaseProps, "isOpen" | "onRequestClose">
+> = observer(({ assetKey, isOpen }) => {
+  const { chainStore, accountStore } = useStore();
+
+  const prefersDark = useMemo(
```
DavideSegullo, line 34: We have `react-use` inside our codebase we can use their hook for this media query:
https://streamich.github.io/react-use/?path=/story/sensors-usemedia--docs

DavideSegullo (changes requested):

DavideSegullo (approved):

PR #4029, opened 2025-01-14
Show 0 liquidity pools in the pools page

deividaspetraitis:

This PR enables including 0 liquidity pools in the pools page.
### Linear Task

[FE-1308](https://linear.app/osmosis/issue/FE-1308/display-0-liquidity-pools-on-frontend-for-user-visibility)

## Brief Changelog

- Updates `minLiquidityUsd` param from `1_000` to `0` to be sent to SQS.

## Testing and Verifying

### DIFF
--- a/packages/web/components/complex/pools-table.tsx
+++ b/packages/web/components/complex/pools-table.tsx
@@ -153,7 +153,7 @@ export const PoolsTable = (props: PropsWithChildren<PoolsTableProps>) => {
             direction: sortParams.allPoolsSortDir,
           }
         : undefined,
-      minLiquidityUsd: 1_000,
+      minLiquidityUsd: 0,
     },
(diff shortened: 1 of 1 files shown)


JeremyParishTheBuilder (commented):
What I'm seeing in the build is that the Pools page shows pools starting at $0 liquidity (which perhaps should be required to be explicitly enabled as a toggle in settings??), and it may have inconsistent ordering.
But, also, that once clicking on one of these pools, the Pools Detail page automatically redirects to Celatone, which means the flow to add liquidity to these pools isn't shown. Clicking the browser's back Button once just goes back to Celatone (because back one time would be the pool detail page, which would just redirect you again); so you have to click Back twice to actually get back to Osmosis Zone.

PR #4031, opened 2025-01-21
Assets table: reduce market asset cache

jonator:

We use prices from the Numia tokens endpoint only for the assets table. This helps keep the table performant since we don't also have to send batch price queries to SQS in addition to the asset info query.

This reduces the cache time so that prices are kept more up to date during high volatility.

## Brief Changelog

- 5mins > 1min

## Testing and Verifying

Tested locally

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,6 +4,7 @@
   },
   "editor.formatOnSave": true,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "biome.enabled": false,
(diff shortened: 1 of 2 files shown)


greg-nagy (approved):

jonator:
@greg-nagy I'm currently mid battle with my biome extension and our current eslint/prettier setup

greg-nagy:
you have no idea how much I can relate ... :) 

in desperate times I have found this to be helpful https://marketplace.visualstudio.com/items?itemName=tombonnike.vscode-status-bar-format-toggle

PR #4035, opened 2025-01-27
Updated Next versions and added lint:workspace

yury-dubinin:

- Updated Next versions 
- Updated tanstack version
- Added lint:workspace job


packages/web/pages/api/broadcast-transaction.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line unused-imports/no-unused-imports
```
DavideSegullo, line 1: Why do we need this?

DavideSegullo (changes requested):

packages/web/pages/api/broadcast-transaction.ts
yury-dubinin: lint fails, no idea how to fix it. This did not help.

greg-nagy: `yarn build` should be run in the `/packages/web` to generate the definition this line imports, otherwise we would get a lint error

yury-dubinin: yes, I was running it from web.

yury-dubinin: reverted and working now.

DavideSegullo (approved):

PR #4040, opened 2025-02-07
Add Galaxy Station wallet

echel0n-HX:

Add support for Galaxy Station extension and mobile


JohnnyWyles:
Hey, sorry for the delay - would just need rebasing from stage due to the cosmoskit dependency needing to be bumped since the last time - looks good apart from that

echel0n-HX:
> Hey, sorry for the delay - would just need rebasing from stage due to the cosmoskit dependency needing to be bumped since the last time - looks good apart from that

@JohnnyWyles we've merged stage branch and fixed any conflicts, please review again, thanks!

JohnnyWyles:
@echel0n-HX  If you want to check it all works while you have the wallet installed, there's a deployment now above

echel0n-HX:
@JohnnyWyles everything looks good from what I can see, thanks!

JohnnyWyles (approved):

JohnnyWyles:
HI @echel0n-HX, after discussing with the rest of the team, we aren't onboarding new wallets at the moment. Each adds additional overhead for ongoing support. This was why this hadn't been reviewed earlier. Apologies.

echel0n-HX:
@JohnnyWyles appreciate the response.

Could you clarify what specific ongoing overhead you're referring to? For most integrations we're part of, we handle the maintenance on our side, and for upstream (e.g. cosmos-kit), we already maintain our integration there as well.

From our perspective, the incremental effort for Osmosis should be minimal.

Also, with Leap Wallet sunsetting, there’s likely going to be a gap in supported wallet options across the ecosystem. We’re actively maintaining and supporting our integration, so we’d be happy to help fill that where it makes sense.

If onboarding isn’t possible right now, is this something you’d be open to revisiting in the near future? Happy to align on any requirements or criteria needed.

JohnnyWyles:
> @JohnnyWyles appreciate the response.
> 
> Could you clarify what specific ongoing overhead you're referring to? For most integrations we're part of, we handle the maintenance on our side, and for upstream (e.g. cosmos-kit), we already maintain our integration there as well.
> 
> From our perspective, the incremental effort for Osmosis should be minimal.
> 
> Also, with Leap Wallet sunsetting, there’s likely going to be a gap in supported wallet options across the ecosystem. We’re actively maintaining and supporting our integration, so we’d be happy to help fill that where it makes sense.
> 
> If onboarding isn’t possible right now, is this something you’d be open to revisiting in the near future? Happy to align on any requirements or criteria needed.

Leap wallet's sunsetting was one reason I moved this forward. I agree that we might need alternatives available, but this wasn't the view of the wider team during our review phase before release.

It's not just the version bump, which is low-effort; it's the additional research needed on wallets and their security. Since we're allowing connections via a wallet, we could be seen as promoting its use, which means we would have to continually keep up to date with upgrades in case of supply chain issues.

PR #4044, opened 2025-02-10
Enrico/fe-1322-put-onramp-options-behind-feature-flag

EnricoBarbieri1997:

Reintegrate moonpay and put each onramp option behind a Feature Flag

### DIFF
--- a/packages/web/modals/fiat-on-ramp-selection.tsx
+++ b/packages/web/modals/fiat-on-ramp-selection.tsx
@@ -4,7 +4,7 @@ import React, { FunctionComponent } from "react";
 
 import { Icon } from "~/components/assets";
 import { Button } from "~/components/ui/button";
-import { MultiLanguageT, useTranslation } from "~/hooks";
+import { MultiLanguageT, useFeatureFlags, useTranslation } from "~/hooks";
 import { useBridgeStore } from "~/hooks/bridge";
 import { FiatRampDisplayInfos, FiatRampKey } from "~/integrations";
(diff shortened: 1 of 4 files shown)


packages/web/modals/fiat-on-ramp-selection.tsx
```diff
@@ -50,54 +56,63 @@ export const FiatOnrampSelectionModal: FunctionComponent<
...
-                  <div className="flex items-center gap-2">
-                    <h6>{displayName}</h6>
-                    <span className="text-xs text-white-mid">
-                      🌎{" "}
-                      {t(
-                        "components.fiatOnrampSelection.moonpayCountrySupport"
-                      )}
-                    </span>
+          ({ rampKey, displayName, initialAsset, logoId, subtitle }) => {
+            if (rampKey === "layerswapcoinbase" && !flags.layerswapcoinbase)
```
DavideSegullo, line 60: nit: What if we just do:
```ts
if (!flags[rampKey]) return null
```

So we don't need to specify it for each type

DavideSegullo (approved):

packages/web/modals/fiat-on-ramp-selection.tsx
EnricoBarbieri1997: I thought about it but it would mean:
- either we extend AvailableFeatureFlags with FiatOnRampOptions to keep them aligned or duplicate definition and keep track of it
- Features flag would be dependent on another definition instead of being standalone

aside from that I can easily be done

yury-dubinin (approved):

PR #4056, opened 2025-03-07
Fix: sBTC Deposits Failing Due to Low Precision Fee Tokens

JoseRFelix:

Deposits from sBTC are failing via the IBC provider because our gas estimation logic skips fee tokens with low precision. This was originally implemented to prevent gas errors caused by high volatility in expensive tokens like WBTC.

This PR resolves both issues by ensuring that the minimum fee amount is at least “1”. As a result:
	•	Extremely expensive tokens can still be selected, even if their precision is low after applying the fee ratio.
	•	sBTC deposits now succeed, as sBTC can be properly used as a fee token.

### Linear Task

https://linear.app/osmosis/issue/FE-1338/sbtc-deposit-fails-due-to-insufficient-alternative-fees-on-side-chain

## Brief Changelog

- Change "unknown time remaining" copy for a more human readable alternative "Deposit/Withdrawal is taking longer than expected"
- Allow using Low Precision Fee Tokens with a minimum fee of at least "1"

## Testing and Verifying

- [ ] Can use allBTC, and wBTC as fee tokens
- [ ] Can deposit sBTC using IBC provider


packages/e2e/pages/base-page.ts
```diff
@@ -1,86 +1,86 @@
...
 
 export class BasePage {
-  readonly page: Page;
-  readonly connectWalletBtn: Locator;
-  readonly kepltWalletBtn: Locator;
-  readonly portfolioLink: Locator;
-  readonly poolsLink: Locator;
-  readonly walletBalance: Locator;
-  readonly connectedWalletBtn: Locator;
+  readonly page: Page
```
greg-nagy, line 4: is this biome vs prettier? Should we reformat the whole codebase in one go?

greg-nagy (approved):

DavideSegullo (approved):

packages/e2e/pages/base-page.ts
JoseRFelix: This is prettier. I'm not sure why only the e2e consistently need prettier format. I'll investigate when possible

PR #4060, opened 2025-03-16
Have managed pool creation redirect to magma vaults page

sunnya97:
## Brief Changelog

-


packages/server/package.json
```diff
@@ -36,6 +36,8 @@
     "@osmosis-labs/types": "^1.0.0",
     "@osmosis-labs/unit": "0.10.24-ibc.go.v7.hot.fix",
     "@osmosis-labs/utils": "^1.0.0",
+    "@sentry/core": "^9.5.0",
+    "@sentry/node": "^9.5.0",
```
jonator, line 40: What are these used for?

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -80,6 +80,14 @@ export const AddConcLiquidity: FunctionComponent<
       poolId,
     });
 
+    const getMagmaUrl = (pool?: Pool) => {
+      if (!pool) return "";
+      const assetDenoms = pool.reserveCoins
+        .map((coin) => coin.currency.coinMinimalDenom)
```
jonator, line 86: Consider encoding denoms, esp those containing slashes, by wrapping with `encodeURIComponent` call

packages/web/localizations/en.json
```diff
@@ -22,7 +22,7 @@
     "step2Title": "Step 2: Add Liquidity",
     "step2TitleManaged": "Step 2: Select Provider",
     "managed": "Managed",
-    "managedDescription": "Enlist one of our providers to manage your position. An entirely hands off experience.",
+    "managedDescription": "Enlist a community liquidity manager to actively manage your position through Magma vaults.",
```
jonator, line 25: BTW this update isn't reflected in other languages. 

It's somewhat tedious to update since we need to remove all this entry from all files, add it to en.json, then run `yarn machine-translate` to generate updated ones in all files.

But this may not be high prio enough to do.

jonator (approved):
Some nits, but LGTM

packages/web/components/complex/add-conc-liquidity.tsx
```diff
@@ -108,12 +117,9 @@ export const AddConcLiquidity: FunctionComponent<
...
             case "add_managed":
-              return (
-                <AddConcLiqManaged
-                  quasarVaults={quasarVaults}
-                  addLiquidityConfig={addLiquidityConfig}
-                />
-              );
+              window.open(getMagmaUrl(pool), "_blank");
+              addLiquidityConfig.setModalView("overview");
+              return null;
```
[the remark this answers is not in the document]
greg-nagy: using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams might be the best here

packages/web/components/complex/add-conc-liquidity.tsx
greg-nagy: using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams might be also good

greg-nagy (approved):
LGTM

packages/server/package.json
sunnya97: Uh, I'm not sure.  My yarn dev wasn't working without them.  Warp added those to fix it

PR #4074, opened 2025-04-24
feat: fastusdc bridge provider

samsiegart:
## Brief Changelog

-


packages/bridge/src/fast-usdc/client.ts
```diff
@@ -0,0 +1,302 @@
...
+import cachified from "cachified";
+
+import { BridgeProviderContext } from "../interface";
+
+export const AGORIC_API_URL = "https://main.api.agoric.net";
+export const NOBLE_TO_AGORIC_CHANNEL = "channel-21";
+export const NETWORK_CONFIG_URL = "https://main.agoric.net/network-config";
+export const FAST_USDC_SERVICE_URL =
+  "https://fastusdc-map.agoric-core.workers.dev/store";
+export const NOBLE_API_URL = "https://noble-api.polkachu.com";
```
JoseRFelix, line 12: nit: We use PascalCase for our const variables

packages/bridge/src/fast-usdc/transfer-status.ts
```diff
@@ -0,0 +1,97 @@
...
+    })
+      .catch((e) => console.error(`Polling Fast USDC has failed`, e))
+      .then((s) => {
+        if (s) this.receiveConclusiveStatus(sendTxHash, s);
+      });
+  }
+
+  makeExplorerUrl(_snapshot: TxSnapshot): string {
+    return "";
+  }
```
JoseRFelix, line 78: We should include a link so users can view the transaction. Do you know a good URL we can use for that?

JoseRFelix (changes requested):
This is looking great! I’ve left a few minor comments. While those are being addressed, I’m going ahead and sending this to QA.

samsiegart (commented):
Thanks for reviewing. We aligned with Aaron that this could be made the default. How do you suggest we address that in this PR? A few ideas:

- We hardcode `bestQuote` to prioritize the fast usdc provider ID if available
- We change `bestQuote` to prioritize the fastest quote instead of the cheapest (similar to https://go.skip.build/ which seems to default to fastest)
- We change `bestQuote` to balance speed and price based on some formula. Basically, formalize in some way that it's better to pay a little extra to go from 18 minutes to 1 minute.

JoseRFelix:
@samsiegart Thanks for the suggestions and for checking in! We discussed this internally with management and the team, and for now, we’ve decided to stick with the current approach. While we see the potential benefits of prioritizing speed, making it the default introduces trade-offs we’re not ready to take on at this stage.

PR #4080, opened 2025-05-06
Fix Incorrect Amount When Swapping Exact Amount Out

JoseRFelix:

When doing a swap exact amount out (specifying the exact token amount to receive), the router was incorrectly using `in_amount` instead of `out_amount` for the initialAmount property in the split response. This caused incorrect calculations, resulting in small amounts received.

### Linear Task

https://linear.app/osmosis/issue/BE-733/frontend-issue-with-incorrect-amounts-when-swapping-to-alldym


jonator (approved):
🥲

DavideSegullo (approved):

PR #4082, opened 2025-05-15
BE-740 | Fix pricing discrepancies is the swap tool

deividaspetraitis:
### Linear Task

[BE-740](https://linear.app/osmosis/issue/BE-740/log-ticket-for-previously-reported-pricing-issues)

## Brief Changelog

This PR attemps to fix pricing discrepancies is the swap tool.
Main source of the changes stems from the fact that price impact can be negative. Negative impact means negative impact where user would receve less and positive impact represents arb opportunity.

## Testing and Verifying

This change has been tested by fixing existing and adding new relevant test cases

### DIFF
--- a/packages/pools/src/router/types.ts
+++ b/packages/pools/src/router/types.ts
@@ -43,7 +43,7 @@ export type Quote = {
   afterSpotPriceOutOverIn?: Dec;
   effectivePriceInOverOut?: Dec;
   effectivePriceOutOverIn?: Dec;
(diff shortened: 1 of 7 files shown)


DavideSegullo (approved):
Looks good, but I don't have a lot of context here, so idk if this could introduce side effects, cc @fabryscript @JoseRFelix @crnbarr93  what do you think?

jonator (approved):

PR #4102, opened 2025-06-29
Fix broken link to TradingView Widget methods documentation

VolodymyrBg:
Replaced the outdated URL to the TradingView Widget methods documentation in charting_library.d.ts with the current, working link. This ensures that users and developers are directed to the correct resource for up-to-date information on widget methods.


jonator:
Thank you!

PR #4103, opened 2025-06-30
Improving token listing automation FE Changes

fabryscript:
FE Changes required for ["Improving Token Listing Automation"](https://www.notion.so/osmosiszone/Improving-Token-Listing-Automation-1e49e6a7a3c180589492ff46c7130131?source=copy_link)

### DIFF
--- a/packages/server/src/queries/complex/assets/user.ts
+++ b/packages/server/src/queries/complex/assets/user.ts
@@ -1,5 +1,5 @@
 import { AssetList, Chain, MinimalAsset } from "@osmosis-labs/types";
-import { CoinPretty, PricePretty } from "@osmosis-labs/unit";
+import { CoinPretty, Dec, PricePretty } from "@osmosis-labs/unit";
 import {
   aggregateCoinsByDenom,
   isNil,
@@ -16,7 +16,13 @@ import {
   getUserTotalDelegatedCoin,
   getUserTotalUndelegations,
 } from "../staking/user";
-import { AssetFilter, calcSumCoinsValue, getAsset, getAssets } from ".";
+import {
+  AssetFilter,
+  calcSumCoinsValue,
+  getAsset,
+  getAssetMarketActivity,
+  getAssets,
+} from ".";
 import { DEFAULT_VS_CURRENCY } from "./config";
 import { calcAssetValue } from "./price";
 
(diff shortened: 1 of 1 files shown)


JeremyParish69 (changes requested):
Using FBTC.axl (asset with no liquidity on Osmosis)
(Good) I am able to manually enter the URL for this asset and can still see it's page render properly.
(Good) When I search for "FBTC" on the Assets page, I can no longer find FBTC.axl in the results.
(Bad) When I search for "FBTC" on the Trade page, I can still find FBTC.axl in the results, which should not be the case. (What is the point in asking SQS to route a token that we know has no liquidity?)
(Bad) When I search for "FBTC" on the Portfolio page, I can still find FBTC.axl in the results, which should not be the case.
(Unsure) When I search for FBTC in the Deposit flow, I cannot see the asset--this is expected. However I also cannot find it on production, even though I'd expect to be able to... and I don't understand why it's not there but other BTC variants are there.... Possible due to not having any price??

fabryscript:
> Using FBTC.axl (asset with no liquidity on Osmosis)
> 
> (Good) I am able to manually enter the URL for this asset and can still see it's page render properly.
> 
> (Good) When I search for "FBTC" on the Assets page, I can no longer find FBTC.axl in the results.
> 
> (Bad) When I search for "FBTC" on the Trade page, I can still find FBTC.axl in the results, which should not be the case. (What is the point in asking SQS to route a token that we know has no liquidity?)
> 
> (Bad) When I search for "FBTC" on the Portfolio page, I can still find FBTC.axl in the results, which should not be the case.
> 
> (Unsure) When I search for FBTC in the Deposit flow, I cannot see the asset--this is expected. However I also cannot find it on production, even though I'd expect to be able to... and I don't understand why it's not there but other BTC variants are there.... Possible due to not having any price??
> 
> 

gotcha! I will work on these fixes.
I'm unsure about the last point though 🤔

JeremyParish69:
> gotcha! I will work on these fixes. I'm unsure about the last point though 🤔


Thanks!

I think we can consider the "unsure" point to be out-of-scope for now--it sounds like a different issue. I'll try to find a different token to confirm that the Deposit flow's Asset Search no longer shows assets with no liquidity.

I also want to clarify that FBTC.axl should still appear on Portfolio page if the connected wallet has a balance of it on Osmosis, even when there is no liquidity. This is important for allowing new teams to confirm that their new token has arrived on Osmosis and they can set up liquidity.

PR #4111, opened 2025-07-16
Remove superfluid option from pool filters

fabryscript:


fabryscript:
Also, @jonator , I was wondering if I should remove all the definitions of "superfluid" from pool incentives types etc...as this would also change some server code

jonator:
To start, let's focus on removing it from UI. <br><br>As @johnny.wyles said, if they already have superfluid we may need to keep it

PR #4119, opened 2025-07-24
Update asset details querying to new format

fabryscript:


JeremyParish69 (approved):
Asset Detail pages (checked multiple assets) appear to continue to show all the relevant data (description, socials, etc.), and also is able to switch to other languages.

PR #4120, opened 2025-07-24
Redirect to `coinMinimalDenom` URL after accessing Assets page with a symbol

fabryscript:


packages/web/pages/assets/[denom].tsx
```diff
@@ -77,7 +77,19 @@ const AssetInfoView: FunctionComponent<AssetInfoPageStaticProps> = observer(
...
+    const {
+      title,
+      details,
+      coinGeckoId,
+      asset,
+      denom: routerDenom,
+    } = useAssetInfo();
+
+    useEffect(() => {
+      if (routerDenom !== asset?.coinMinimalDenom) {
```
Comment (jonator) [inline packages/web/pages/assets/[denom].tsx]: Per coderrabbit I agree we should check for null to ensure the check is what we expect

PR #4125, opened 2025-07-29
[Blocked] [V30] Make all denoms available as quote in Supercharged Pool Creation

fabryscript:


fabryscript:
> Flow looks good - unable to test full creation since Quotes are still permissioned until v30 is released. Message appears to be generated correctly though as error is correct. This should not be released to main before then.

Blocking this until v30 on Linear

PR #4129, opened 2025-08-01
Use `coinMinimalDenom` on Swap/Limit tool

fabryscript:


jonator (approved):

JeremyParish69 (approved):
Trade page functions correctly.
All "shortcut" tokens (USDT, USDC, ETH, BTC, etc.) work fine.
selecting any asset in the From or To positions will add the appropriate URL params using base denom (with actual slashes--no need to escape the slashes).
Entering a URL with the updates params functions correctly, even with slashes.
Swapping the To and From token functions correctly and also updates the URL accordingly.

Edit: Using symbols still appears to work; unable to test how a collision would be handled because duplicate symbols don't exist yet.

PR #4130, opened 2025-08-01
Accept "/" in asset detail pages

fabryscript:


JeremyParish69 (approved):
Functions correctly.
Visiting asset detail pages now uses the base denom with its slashes (no longer escaped).
Entering specific URLs for an asset detail page using the base denom with slashes goes to the correct page.

PR #4131, opened 2025-08-01
Fix blurry logos and typeface issue

fabryscript:


Cosmos-Harry (approved):
lgtm 

Screenshot 2025-08-04 at 12 03 05 PM

PR #4134, opened 2025-08-04
feat: use coinMinimalDenom in bridging

fabryscript:


JeremyParish69 (approved):
Deposit/Withdraw URL is generated with the base denom instead of Symbol, both from the Portfolio page, and from an Asset Detail page. Substituting the escaped slash (%2F) in the URL with an actual slash ('/') or with the symbol also works as expected.

PR #4140, opened 2025-08-07
Switch to new chainlist location

fabryscript:


fabryscript:
> > I'm unable to suggest a chain. I removed Archway from Keplr and went to Connect it to Keplr again to complete Deposit or Withdraw, but nothing happens when I click image Each click triggers this console error: image
> 
> Strange, I can connect archway image

cc @JeremyParish69

fabryscript:
re-opening post force push

PR #4141, opened 2025-08-08
Fix language querying in asset details

fabryscript:


JeremyParish69 (approved):
I no longer see the query for the asset detail file repeated over a dozen times in the Network tab. Although I see a little bit of repetition under Console errors. 
image

PR #4154, opened 2025-08-12
Switch to new chainlist

fabryscript:


JeremyParish69 (changes requested):
Current currency data is missing some data:
image
The Suggest Chain method allows for more data including coinGecko (and coinImageUrl), and also gasPrices for feeCurrencies
image

PR #4176, opened 2025-09-17
Prioritize origin address over minDenom on deposits

fabryscript:


fabryscript:
> Tested with TIA, ATOM, AKT, unable to test with ETH though as I get an error which may or may not be CORS related

Oh, can you please share a screenshot or update linear pls 🙏🏻

PR #4308, opened 2026-03-18
fix(e2e): optimize stablecoin swap amounts to reduce USDC.eth.axl drain

jasbanza:
**TL;DR:** Monitoring accounts drain USDC.eth.axl because the swap test spends 0.55 on both legs of a round-trip, but receives less due to fees. Fix: reduce return-leg to 0.54 so each run is net-positive, and bump topup target from 1.8 to 3.

## Linear
[MER-171: fix(e2e): optimize stablecoin swap amounts to reduce USDC.eth.axl drain](https://linear.app/osmosis/issue/MER-171/fixe2e-optimize-stablecoin-swap-amounts-to-reduce-usdcethaxl-drain)

## Summary

- **Asymmetric swap amounts**: Forward legs (USDC->X) stay at 0.55, return legs (X->USDC) reduced to 0.54. On-chain analysis shows the SQS router sometimes picks multi-hop routes through OSMO that lose ~0.4% per swap vs ~0.01% on direct routes. The 0.01 difference ensures each round-trip leaves a small surplus instead of draining scarce tokens.
- **Bump USDC.eth.axl warnAmount**: 1.2 -> 2 across all monitoring accounts (SG/EU/US), so topups fill to 3 instead of 1.8 for more runway.

## On-chain data

Analyzed recent swaps from the US monitoring account (osmo10tv5djyyj0492ulgq2kv66n90mhfyakwupy9tv):

| Route | Loss per swap |
|-------|---------------|
| Direct pool 1223 (USDC/USDC.eth.axl) | ~0.01% |
| Multi-hop via OSMO (pools 1464->1133) | ~0.4% |

With 0.55/0.54 asymmetry, even the worst-case OSMO route leaves a +0.006 surplus per run.

## Test plan

- [ ] Verify monitoring swap tests still pass with the new amounts
- [ ] Confirm topup script respects the updated warnAmount (dry run)

### DIFF
--- a/packages/e2e/README.md
+++ b/packages/e2e/README.md
@@ -177,7 +177,7 @@ All three monitoring accounts (SG, EU, US) share the same thresholds:
 | USDC | 7 | 8.4 | token | market buy + limit buy + swap stables |
 | OSMO | $2 | $5 | USD | market sell + limit sell OSMO |
 | BTC | $1.60 | $5 | USD | market sell BTC |
-| USDC.eth.axl | 1 | 1.2 | token | swap stables |
+| USDC.eth.axl | 1 | 2 | token | swap stables |
 | USDT | 1 | 1.2 | token | swap stables |
 
 ---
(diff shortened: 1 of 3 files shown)


jasbanza:
@JohnnyWyles R4R 🙏

JohnnyWyles (approved):

PR #4317, opened 2026-03-24
fix(ui): standardise EntityImage circular masking across all components

JohnnyWyles:
Apply consistent `overflow-hidden rounded-full` wrapper to all EntityImage usages to ensure token logos render as circles throughout the app.

Previously ~75% of usages had no circular masking, relying on the logo file itself to appear circular. The canonical pattern (used in token-select.tsx) wraps EntityImage in a sized div with overflow-hidden and rounded-full, which correctly clips any image regardless of its shape, aspect ratio, or fallback rendering.

Upgraded existing partial implementations (rounded-full without overflow-hidden) and added new wrappers across assets table, bridge screens, transaction history, swap tool, pool detail, orders, portfolio, and navbar components.

Intentionally excluded: pool-assets-icon.tsx and price-selector default quotes row (overlapping stacked icon designs that require square bounds).

### DIFF
--- a/packages/types/src/asset-types.ts
+++ b/packages/types/src/asset-types.ts
@@ -189,13 +189,6 @@ export interface Asset {
 
   pegMechanism?: "algorithmic" | "collateralized" | "hybrid";
 
-  /**
-   * When true, the logo image uses its full bounding box — content extends near
-   * the edges and should not be clipped to a circle. Set in osmosis-labs/assetlists
-   * and flows through codegen automatically.
-   */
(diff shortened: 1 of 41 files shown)


jasbanza (approved):
LGTM. Some logos get cut off though, but that would be a logo design requirement issue

JohnnyWyles:
> LGTM. Some logos get cut off though, but that would be a logo design requirement issue

Still looking into this to see if there is another solution before merging

PR #4320, opened 2026-03-25
fix(ci): prevent false Slack balance alerts when check script crashes

jasbanza:
When the check-balances script crashes (network error, dependency issue, etc.) rather than detecting a real low balance, all three e2e workflows fire a false Low Balance Warning to Slack because they only gate on `steps.balance-check.outcome == 'failure'`.

Added an `is_balance_issue` output to the Extract balance details step (`true` when `balance-report.json` exists, `false` on script crash) and gated the Slack alert on it.

**Files:** `frontend-e2e-tests.yml` (1), `prod-frontend-e2e-tests.yml` (1), `monitoring-limit-geo-e2e-tests.yml` (8 instances).

---

> [!NOTE]
> **Low Risk**
> Low risk workflow-only change that narrows when Slack alerts fire; main risk is misconfigured outputs/conditions causing missed notifications.
> 
> **Overview**
> Prevents **false low-balance Slack alerts** when `check-balances` fails due to a script crash rather than an actual balance shortage.
> 
> Across the E2E GitHub Actions workflows, the balance-details extraction step now emits an `is_balance_issue` output based on whether `packages/e2e/balance-report.json` exists, and the Slack alert step is additionally gated on `is_balance_issue == 'true'` (with the crash case reported as `outcome=error`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 50783d8d9aa7ecfe9e7fe0c458fc5f5f8730330a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/frontend-e2e-tests.yml
+++ b/.github/workflows/frontend-e2e-tests.yml
@@ -66,12 +66,14 @@ jobs:
             echo "summary=${SUMMARY}" >> "$GITHUB_OUTPUT"
             echo "address=${ADDRESS}" >> "$GITHUB_OUTPUT"
             echo "outcome=${OUTCOME}" >> "$GITHUB_OUTPUT"
+            echo "is_balance_issue=true" >> "$GITHUB_OUTPUT"
           else
-            echo "summary=Balance report not available." >> "$GITHUB_OUTPUT"
-            echo "outcome=warn" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


jasbanza:
@JohnnyWyles this is just to avoid false alerts. Please approve when you can🙏

JohnnyWyles (approved):

PR #4321, opened 2026-03-25
fix: deduct swap fees when using MAX option

Adraca:
## What is the purpose of the change?
This PR ensures that when a user selects the 'MAX' option in the swap interface, the estimated swap fees are deducted from the input amount to prevent transaction failures due to insufficient balance.

## Testing and Verifying
- Manually verified the derivation logic for 'swapFeeAmount' in packages/web/hooks/use-swap.tsx.
- Confirmed 'useAmountInput' correctly subtracts 'swapFeeAmount' from the max available tokens.

## Linear Task Archive
N/A - External Security/Logic Contribution via Adraca AI Research.

### DIFF
--- a/packages/web/hooks/use-swap.tsx
+++ b/packages/web/hooks/use-swap.tsx
@@ -848,6 +848,17 @@ export function useSwap(
     [tokenInFeeAmountFiatValue, networkFee?.gasUsdValueToPay]
   );
 
+  
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


Adraca:
Regarding the CodeRabbit review: The changes to 'useSwapAssets' and 'useSwap' were necessary dependencies to ensure the 'tokenInFeeAmount' correctly propagates to the calculation layer. This is not out-of-scope; it is the complete plumbing for the fix.

PR #4322, opened 2026-03-31
fix(bridge): break deposit quote feedback loop when selecting Max

jasbanza:

When selecting **Max** on an empty deposit/withdrawal input, the gas-adjustment `useEffect` in `CryptoFiatInput` fires on every quote response because each gas estimate differs slightly. This re-sets the input amount, triggers a new debounced quote fetch, and repeats 2-3 times until values converge — causing unnecessary network requests and UI flicker.

The fix adds a `gasAppliedToMax` ref that gates the gas deduction to run only once per Max activation. The ref resets whenever `isMax` becomes `false` (button toggle or keyboard edit), so subsequent Max presses still apply gas correctly.

**Affected flows:** Both deposits and withdrawals (same `CryptoFiatInput` component). Swaps are unaffected (different input flow).

**Affected assets:** Any asset where the gas fee denom matches the transfer denom — confirmed on Firmachain, Photon, AKT. ATOM was unaffected because its gas estimates are more stable.

### Linear Task

[MER-183: Fix: Deposit quote feedback loop when selecting Max on empty input](https://linear.app/osmosis/issue/MER-183/fix-deposit-quote-feedback-loop-when-selecting-max-on-empty-input)

## Brief Changelog

- Add `gasAppliedToMax` ref to `CryptoFiatInput` to prevent repeated gas-adjustment cycles
- Restructure the max-amount `useEffect` to early-return when gas has already been applied
- Reset the ref when `isMax` goes `false` (Max button toggle or keyboard input)

## Testing and Verifying

- **Deposit** Firmachain: select Max on empty input — quote should fire once (not 3x)
- **Withdrawal** Firmachain: same test
- **Deposit** AKT: same test
- **Deposit** ATOM: confirm no regression (was already unaffected)
- Select Max, then keyboard-edit the amount, then select Max again — gas should be re-applied fresh
- Select Max on a pre-filled input — gas should still be deducted correctly

---

> [!NOTE]
> **Medium Risk**
> Changes the `Max` amount calculation behavior in `CryptoFiatInput`, which can affect how much a user attempts to transfer when gas estimates change or disappear. While scoped and covered by new tests, it touches user-facing amount-setting logic in deposit/withdraw flows.
> 
> **Overview**
> Fixes a quote feedback loop when selecting **Max** by guarding the gas-deduction logic in `CryptoFiatInput` so it only applies once per Max activation, rather than re-adjusting on every slightly-changed `transferGasCost`.
> 
> The guard resets when **Max** is turned off, when the selected asset changes, and when `transferGasCost` becomes `undefined` (setting full balance and allowing gas to be deducted again once quotes return). Adds a focused test suite covering these Max/gas edge cases and regression scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98b09d836d8b256dee629b0abd385fc09af888c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/web/components/control/__tests__/crypto-fiat-input-max-guard.spec.tsx
+++ b/packages/web/components/control/__tests__/crypto-fiat-input-max-guard.spec.tsx
@@ -409,4 +409,85 @@ describe("CryptoFiatInput gasAppliedToMax guard", () => {
       expect(onChangeCryptoInput).toHaveBeenCalledWith(expectedJuno);
     });
   });
+
+  it("re-applies gas after transferGasCost goes undefined then returns", async () => {
+    const balanceRaw = "2000000";
+    const gasRaw1 = "5000";
+    const gasRaw2 = "5200";
+    const onChangeCryptoInput = jest.fn();
+
+    const balanceDec = makeBalance(balanceRaw).toDec().toString();
+    const balance = makeBalance(balanceRaw);
+
+    const baseProps = {
+      currentUnit: "crypto" as const,
+      setCurrentUnit: jest.fn(),
+      setIsMax: jest.fn(),
+      canSetMax: true,
+      transferGasChain: { prettyName: "Cosmos Hub" },
+      assetPrice: undefined,
+      assetWithBalance: {
+        denom: "ATOM",
+        address: "uatom",
+        decimals: 6,
+        amount: balance,
+      },
+      fiatInput: "",
+      onChangeFiatInput: jest.fn(),
+      isInsufficientBal: false,
+      isInsufficientFee: false,
+    };
+
+    // 1) Gas applied initially
+    const { rerender } = render(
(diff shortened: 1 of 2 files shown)


JoseRFelix:
@jasbanza Thank you for the PR! Please make sure to add unit tests so that we don't add any regressions.

jasbanza:
In order to properly test, needs rebase to include endpoint fallback fixes

jasbanza:
@JoseRFelix R4R - tested now after rebase, works as expected with AKT etc. 

With regards to unit tests, this is what was done:
image

JoseRFelix (approved):
LGTM! Thank you for adding the unit tests

PR #4334, opened 2026-04-21
feat: add orderbook pool type and filterable cosmwasmpool types

JohnnyWyles:
Adds filtering and sorting for several pool types that were included all the time before.
Allows Orderbook pools to be forwarded to an intermediary page before direction to Celatone.

## Brief Changelog

  - Added cosmwasm-orderbook as a recognised pool type using code ID 885, alongside the existing cosmwasm subtypes. Orderbook pools get their own icon (open-book), colour (text-wosmongton-300), and pool page link. Spread factor is not displayed for orderbook pools since it's not applicable.
  - Moved Astroport PCL and White Whale out of a server-side alwaysIncludedTypes bypass that was causing them to appear in the pools table regardless of filter selection. They now respect the filter dropdown like all other types.
  - Added Orderbook, Astroport, and White Whale as explicit options in the pools table filter dropdown, in the order: Supercharged, Orderbook, Weighted, Stableswap, Astroport, White Whale, Transmuter. All are selected by default except Transmuter (unchanged from before).
  - Added localisation strings for the three new filter labels.

## Testing and Verifying

### DIFF
--- a/packages/server/src/queries/complex/pools/env.ts
+++ b/packages/server/src/queries/complex/pools/env.ts
@@ -8,6 +8,8 @@ export const TransmuterPoolCodeIds = IS_TESTNET ? ["3084"] : ["148"];
 export const AlloyedPoolCodeIds = getAlloyedPoolCodeIds(IS_TESTNET);
 const AstroportPclPoolCodeIds = IS_TESTNET ? ["8611"] : ["842"];
 const WhitewhalePoolCodeIds = IS_TESTNET ? ["?"] : ["503", "641"];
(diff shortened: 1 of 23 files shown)


jasbanza:
"Localization files do not have the same keys"
Missing keys: [ 'orderbook', 'astroport', 'whitewhale' ]

jasbanza:
@JohnnyWyles I made a commit - it was still necessary to add the new keys to all files, even though it's not translated. Otherwise it bugs out and shows the key path (doesn't fallback to en file automatically)

### Before:

image

### After:

image

jasbanza (approved):

PR #4337, opened 2026-04-24
fix: fetch low-liquidity pools before join tx if not in cache

JohnnyWyles:
Mirrors the fetchRemainingPools fallback already used in sendCreateConcentratedLiquidityPositionMsg so that weighted/stable pools with zero or low liquidity (filtered out on initial load) can be found when the user tries to add liquidity.

This prevented generating the transaction to add liquidity to low-liquidity pools unless they were in the cache.


JohnnyWyles:
@coderabbitai review

jasbanza (approved):

PR #4362, opened 2026-05-22
feat(bridge): surface Wormhole governor daily-limit delay in redeem widget

jasbanza:

Surfaces the Wormhole chain-governor "daily limit exceeded" status in the in-app redeem widget at `/wormhole`. Previously, when a user pasted a stuck-transfer hash whose VAA was being held by the Wormhole guardians' 24h rate limiter, the widget would just throw `"VAA not yet available. The Wormhole guardians may still be processing this transaction."` with no further explanation - so users typically assumed something was broken and either retried in a tight loop or gave up. Wormholescan shows a clear `DAILY LIMIT EXCEEDED` badge in this case; we now mirror that locally so users on Osmosis see the same context (with a countdown when available) without bouncing to wormholescan.io.

## Brief Changelog

- Add `fetchGovernorDelay(operation)` helper in `packages/web/components/bridge/wormhole-redeem.tsx` that combines three Wormholescan endpoints to decide whether a VAA-less operation is being held by the chain governor:
  1. `/v1/governor/is_vaa_enqueued/{chain}/{emitter}/{seq}` - definitive per-VAA answer.
  2. `/api/v1/governor/enqueued_vaas/{chain}` - used to populate a `releaseTime` countdown when the row is available.
  3. `/api/v1/governor/notional/max_available/{chain}` - heuristic fallback for the propagation window where the operation has been observed but the guardians haven't enqueued the VAA yet (compares the transfer's USD value to the best guardian's remaining headroom).
- All governor lookups are wrapped in `fetchWithTimeout` and swallow failures - a flaky guardian endpoint can never block the existing redeem flow; we just fall through to the original "VAA not yet available" error.
- Add a new `governor_delayed` status, paint it with a rust-bordered warning card that uses the official Wormhole copy ("This transfer will take up to 24 hours to complete as Wormhole has reached the daily transfer limit for {chain}..."), and surface a "Check again" button so the user can re-poll without retyping the hash.
- Soft "likely-enqueued" variant during the propagation window includes the source chain's remaining headroom for transparency.
- Add `formatGovernorReleaseCountdown` + minute-granularity ticker so the badge shows "Releases in 3h 25m" / "Releases shortly" while the user is on the page.
- Tests: 11 new specs covering `fetchGovernorDelay` (enqueued / missing row / heuristic / fits-in-limit / endpoint failure / non-numeric USD) and `formatGovernorReleaseCountdown` (hours+minutes / minutes-only / sub-minute / no release time). Full suite passes locally (`32/32` in `wormhole-redeem.spec.tsx`).

## Testing and Verifying

This change has been tested locally by running the `wormhole-redeem.spec.tsx` Jest suite (`32/32` passing including the new specs), prettier and ESLint over the touched files (both clean), and reasoning through the render branches against real Wormholescan API responses (verified with chain 3104 / token-bridge emitter against `/v1/governor/is_vaa_enqueued`, `/api/v1/governor/enqueued_vaas/3104`, and `/api/v1/governor/notional/max_available/3104`).

End-to-end manual verification still wants a live VAA currently held by the governor - happy to walk through it with someone holding such a hash. In the meantime the new flow degrades cleanly: any governor-endpoint failure -> existing `"VAA not yet available"` error path, unchanged for users.

### DIFF
--- a/packages/web/components/bridge/__tests__/wormhole-redeem.spec.tsx
+++ b/packages/web/components/bridge/__tests__/wormhole-redeem.spec.tsx
@@ -49,7 +49,9 @@ jest.mock("@mysten/wallet-standard", () => ({
 
 import {
   checkIfRedeemed,
+  fetchGovernorDelay,
   findWormchainRecvTx,
+  formatGovernorReleaseCountdown,
   lookupOsmosisIbcPacket,
   resolveOperation,
   WormholeRedeem,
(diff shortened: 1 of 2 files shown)


jasbanza:
@JohnnyWyles part of a batch of 2 PRs I whipped up a while back for somebody for wormhole recovery - worked fine in preview. Lost track of it but it should be merged. Not super urgent...

JohnnyWyles (commented):
Ran a full review on this (correctness + the osmosis-frontend convention pass). Solid feature overall: good test discipline, correct `PricePretty`/`formatPretty` reuse, honest effect deps, and the swallow-and-fallthrough genuinely can't block the redeem path. `Dec` round-trips are safe (verified the constructor truncates >18dp and parses exponent strings). One thing to flag:

---

**🔴 The new `governor_delayed` state makes a pre-existing unguarded `SuiRedeemPanel` reachable with an empty VAA.**

The fix is in untouched code (the Sui panel guard at L1527), so I'm raising it here rather than inline, but this PR is what triggers it, so it's worth resolving before merge.

The Sui block at L1527 is the only one of the three redeem panels gated on a denylist (`status !== "already_redeemed" && status !== "success"`) rather than `status === "ready"`. The Solana block (L1484) and NonSolana block (L1545) both require `status === "ready"`.

Previously a VAA-less operation threw before `operation`/`status`/`resolvedTxHash` were set, so the Sui block never mounted without a signed VAA. This PR adds a path that, for a VAA-less op, sets `operation` + `resolvedTxHash` + `status = "governor_delayed"` and returns instead of throwing. That new state satisfies the Sui block's denylist, so `SuiRedeemPanel` now mounts alongside the "Daily limit exceeded" card. The panel doesn't inspect the VAA: it renders "VAA is signed and ready. Connect a Sui wallet..." plus a working **Redeem on Sui** button. Clicking it calls `executeSuiRedeem({ vaaBase64: operation.vaa.raw, ... })` (L1083) with `operation.vaa.raw === ""`, which decodes to empty bytes and fails deep in the Sui/Wormhole deserialize with a confusing low-level error. (Solana is safe: its block is gated on `status === "ready"`; this is Sui-only.)

Two ways to fix, your call on scope:
- **In-scope (preferred):** keep the `governor_delayed` state from leaving a Sui op renderable, e.g. the new branch shouldn't set the panel-render preconditions, or the Sui block excludes `governor_delayed`.
- **Pre-existing line:** tighten the Sui guard at L1527 to `status === "ready"` to match the other two panels (the `status !== ...` lines then become redundant). One line, strictly correct, but it's outside this diff so arguably a separate follow-up.

A render test mounting in `governor_delayed` with a Sui-destination op would have caught this and is worth adding alongside whichever fix.

jasbanza:
@JohnnyWyles Good catch, confirmed valid. Traced it end-to-end: the new VAA-less path sets `operation` + `resolvedTxHash` + `status = governor_delayed` and returns, which satisfied the Sui panel's status denylist at L1527 and mounted `SuiRedeemPanel` with `operation.vaa.raw === ""` and a live Redeem-on-Sui button.

Fixed in-scope (your preferred option): the Sui guard now excludes `governor_delayed`, so the panel can't mount in that state. I went with this rather than tightening the guard to `status === "ready"` because the Sui panel intentionally stays mounted during `signing`/`submitting` to drive its own button-label transitions — switching to `ready` would unmount it mid-redeem and break the live Sui flow.

Also added the render test you suggested: it mounts in `governor_delayed` with a Sui-destination op and asserts the delay badge renders while no redeem panel does (both the native Sui panel and its Wormholescan fallback advertise "VAA is signed and ready", so its absence proves neither mounted).

Pushed as 7fba8ec. prettier/eslint clean, `wormhole-redeem.spec.tsx` 33/33.

JohnnyWyles (approved):

PR #4363, opened 2026-05-23
fix: default bridge input unit to crypto instead of fiat

Olek97:
When opening the deposit or withdraw bridge modal, the amount input
defaulted to fiat (USD). Changed the default to crypto so the user
sees the underlying token unit instead.

### DIFF
--- a/packages/web/components/bridge/amount-screen.tsx
+++ b/packages/web/components/bridge/amount-screen.tsx
@@ -185,7 +185,7 @@ export const AmountScreen = observer(
     const [pendingChainApproval, setPendingChainApproval] = useState(false);
     const [wishesToProceed, setWishesToProceed] = useState(false);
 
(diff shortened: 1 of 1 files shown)


JohnnyWyles:
Thanks for the PR, @Olek97,  appreciate you taking the time to dig into the bridge input.

This is a deliberate design choice on our end, not a bug. We default the deposit/withdraw amount input to fiat (USD) so users lead with the value of what they're moving rather than the raw token amount. The fiat-first framing is intentional and consistent with how we surface value across the bridge flow, so we'd like to keep it as the default.

The toggle to switch to the crypto unit is still there for anyone who prefers to enter a token amount, so both paths remain available; we're just keeping value-first as the entry point. Realise that token-first is likely a more "crypto-native" approach, but the default aims to make the token more abstract to users who are not.

Going to close this for that reason, but please don't let it put you off contributing. If you spot something that looks like an actual bug or have an idea for a UX improvement, we welcome it.

PR #4377, opened 2026-06-10
fix(web): prevent MoonPay and Swapped signing oracle abuse

JoseRFelix:
## Summary

- Close the MoonPay signing oracle: `/api/integrations/moonpay/sign-url` no longer signs arbitrary client-supplied URLs and instead builds URLs server-side from validated params.
- Harden Swapped signing: `/api/swapped-signature` now requires a valid Osmosis (`osmo`) bech32 address before signing.
- Add regression tests (15 total) that document the prior attack vector and verify the fixed behavior.

## Reasoning

Both fiat on-ramp signing endpoints were public, unauthenticated HMAC signing oracles:

1. **MoonPay** accepted any `{ url }` in the request body and signed it with Osmosis's MoonPay private key. An attacker could craft a URL with their own `walletAddress`, obtain a valid Osmosis signature, and distribute a phishing link that delivers purchased OSMO to the attacker's wallet while appearing to come from Osmosis's official integration.

2. **Swapped** accepted any `walletAddress` string with no validation and returned a fully signed widget URL.

MoonPay URL signing is meant to prevent third-party tampering with pre-filled sensitive params (wallet address, amount, etc.). The integrator's server is the trust boundary — it must only sign URLs it intends to authorize. Passing through client-built URLs violated that boundary.

### What this PR changes

| Before | After |
|---|---|
| MoonPay endpoint signs any parseable URL | Rejects `{ url: ... }` requests with 400 |
| Client controls full query string | Server builds URL from allowlisted params + env api key |
| Swapped accepts any wallet string | Validates `osmo` bech32 before signing |
| No security tests | 15 tests covering oracle rejection + legitimate signing |

### Frontend change

The MoonPay widget callback now sends the **connected wallet address** plus currency params extracted from the SDK URL, rather than forwarding the raw URL to the signing endpoint.

### Remaining limitation

A determined attacker can still call the API directly with `{ walletAddress: "osmo1attacker..." }` without wallet ownership proof. This does not enable silent theft of connected in-app users (the destination address is visible in the MoonPay UI), but phishing of new/off-platform users remains possible without cryptographic wallet binding. That is a follow-up hardening item (rate limiting, monitoring, optional wallet signature challenge).

## Test plan

- [x] `cd packages/web && npx jest --config jest.integrations.config.js`
- [ ] Verify MoonPay on-ramp still loads and signs correctly when wallet is connected in the app
- [ ] Verify Swapped on-ramp iframe still loads for a connected Osmosis wallet
- [ ] Confirm `POST /api/integrations/moonpay/sign-url` with `{ url: "..." }` returns 400
- [ ] Confirm `POST /api/swapped-signature` with `{ walletAddress: "not-a-wallet" }` returns 400


packages/web/server/integrations/moonpay/sign-moonpay-url.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  if (params.baseCurrencyCode) {
+    searchParams.set("baseCurrencyCode", params.baseCurrencyCode);
+  }
+
+  if (params.baseCurrencyAmount) {
+    searchParams.set("baseCurrencyAmount", params.baseCurrencyAmount);
+  }
+
+  const host = publicKey.startsWith("pk_live")
```
JohnnyWyles, line 85: Non-blocking, confirm intent: the server-built URL drops `baseCurrencyAmount` and picks host off the `pk_live` prefix. Assuming that's deliberate from moving URL construction server-side. Not urgent since MoonPay is flag-dark in prod right now (the LD `moonpay` flag gates the UI), so there's no live checkout flow this affects, but worth a sanity check before the flag is ever flipped back on, that a pre-filled buy amount isn't expected to carry through. The endpoint itself is still HTTP-reachable regardless of the flag, which is exactly why hardening it here is the right call.

JohnnyWyles (approved):

packages/web/server/integrations/moonpay/sign-moonpay-url.ts
JoseRFelix: The old MoonPay flow never pre-filled a purchase amount. it only set the currency `baseCurrencyCode` and the default currency `defaultCurrencyCode`. When we moved URL building to the server, we only included the buy amount if the client happened to send one. That didn't break anything relative to before, but it meant the signed URL could differ from what the widget was actually requesting. We fixed this in commit `b5a962f83` by defaulting to $200 USD on both the server and the widget, so the two stay in sync.

Furthermore, checking for the pk_live prefix was a quick stopgap. The follow-up switches to the MoonPay SDK's `url.generate()`, so whether we hit sandbox or production is determined by the secret key rather than the public key's prefix.

PR #4407, opened 2026-07-08
fix(utils): bound the EVM RPC fallback chain and drop the hanging merkle endpoint

jasbanza:

Stop degraded Ethereum RPCs from timing out bridge quotes. One hanging endpoint could consume the whole 12s tRPC procedure budget under viem's default timeouts/retries, turning a recoverable estimate failure into a user-visible "something went wrong". Independently shippable hardening per the investigation in MTN-214 (Johnny's comment).

### Linear Task

[MTN-214](https://linear.app/osmosis/issue/MTN-214/bridge-deposits-max-amount-skip-quotes-fail-because-axelar-bridge-fee)

## Brief Changelog

- Bound each `http()` transport in `getEvmRpcTransport` to a 3s timeout with no retries, and disable fallback-level retries, so the worst-case walk of the URL list stays well inside the 12s procedure budget.
- Replace `eth.merkle.io` (currently hangs ~15s on `eth_estimateGas` while answering other methods fast) with `eth.drpc.org` in the mainnet RPC list, and document that listed endpoints must support state-override gas estimation.

## Testing and Verifying

Endpoint behavior measured 2026-07-08: merkle `eth_estimateGas` hangs 15.5s (its `eth_blockNumber` answers in 0.3s); publicnode, keplr, and drpc all answer the same estimate — including with state overrides — in 0.2–1.8s. Utils package builds and all 108 tests pass. Full measurements in MTN-214.

---

> [!NOTE]
> **Medium Risk**
> Changes RPC failover behavior and mainnet endpoint list for bridge quoting; mis-tuned timeouts could cause more estimate failures, but client-side receipt polling keeps viem defaults.
> 
> **Overview**
> **Hardens bridge quotes** against slow or stuck Ethereum RPCs by making server-side viem clients fail over quickly instead of burning the whole tRPC budget on one URL.
> 
> `getEvmRpcTransport` now accepts optional **`timeout`** and **`retryCount`**, applied to each `http()` transport and to fallback-level retries. **Skip** and **Squid** pass **3s timeout** and **zero retries** wherever they build public clients for EVM gas estimation and ERC-20 approval checks.
> 
> On Ethereum mainnet, **`eth.merkle.io` is removed** from the RPC list (its `eth_estimateGas` was hanging ~15s) and **`eth.drpc.org` is added**. Comments note listed endpoints must support state-override gas estimation and respond promptly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 595027ad5c1864d6235b2d9180d4a38d90c35936. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bridge/src/skip/index.ts
+++ b/packages/bridge/src/skip/index.ts
@@ -565,7 +565,10 @@ export class SkipBridgeProvider implements BridgeProvider {
 
     const provider = createPublicClient({
       chain: evmChain,
-      transport: getEvmRpcTransport(evmChain),
+      transport: getEvmRpcTransport(evmChain, {
+        timeout: 3_000,
+        retryCount: 0,
+      }),
     });
 
     return provider;
@@ -812,7 +815,10 @@ export class SkipBridgeProvider implements BridgeProvider {
 
       const provider = createPublicClient({
         chain: evmChain,
-        transport: getEvmRpcTransport(evmChain),
+        transport: getEvmRpcTransport(evmChain, {
+          timeout: 3_000,
+          retryCount: 0,
+        }),
       });
 
       const estimatedGas = await this.estimateEvmGasWithStateOverrides(
--- a/packages/bridge/src/squid/index.ts
+++ b/packages/bridge/src/squid/index.ts
@@ -479,7 +479,10 @@ export class SquidBridgeProvider implements BridgeProvider {
 
       const fromTokenContract = createPublicClient({
         chain: evmChain,
-        transport: getEvmRpcTransport(evmChain),
+        transport: getEvmRpcTransport(evmChain, {
+          timeout: 3_000,
+          retryCount: 0,
+        }),
       });
 
       approvalTx = await this.getApprovalTx({
--- a/packages/utils/src/ethereum.ts
+++ b/packages/utils/src/ethereum.ts
@@ -180,20 +180,24 @@ export const EthereumChainInfo = [
  * Builds a viem fallback transport from a chain's default RPC URLs.
  * Falls through each URL in order on failure.
  *
- * Each attempt is bounded (short timeout, no retries) so a degraded RPC
- * can't stall the chain: server-side bridge quotes run under a 12s tRPC
- * procedure timeout, and viem's defaults (10s timeout, 3 retries per
- * transport plus fallback-level retries) let one hanging endpoint consume
- * the entire budget before the remaining URLs are tried.
+ * By default each attempt uses viem's timeout/retry behavior, which suits
+ * long-lived client-side calls like waiting for a transaction receipt —
+ * failing fast there mislabels an in-flight transaction as failed. Server-side
+ * bridge quotes run under a 12s tRPC procedure timeout, where those same
+ * defaults (10s timeout, 3 retries per transport plus fallback-level retries)
+ * let one hanging endpoint consume the entire budget before the remaining
+ * URLs are tried — those call sites should pass a short timeout and zero
+ * retries.
  */
-export function getEvmRpcTransport(chain: {
-  rpcUrls: { default: { http: readonly string[] } };
-}) {
+export function getEvmRpcTransport(
+  chain: {
+    rpcUrls: { default: { http: readonly string[] } };
+  },
+  opts?: { timeout?: number; retryCount?: number }
+) {
   return fallback(
-    chain.rpcUrls.default.http.map((url) =>
-      http(url, { timeout: 3_000, retryCount: 0 })
-    ),
-    { retryCount: 0 }
+    chain.rpcUrls.default.http.map((url) => http(url, opts)),
+    { retryCount: opts?.retryCount }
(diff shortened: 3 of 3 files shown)


jasbanza:
@JohnnyWyles this works now, but need to merge this to rebase the skip fix too
image

JohnnyWyles:
### Review: premise confirmed, one blocker before merge

Ran the reviewer pass plus live probes against all four endpoints. The server-side motivation is verified and worth shipping: the 12s procedure budget is real (`packages/trpc/src/api.ts:89`), and merkle's hang is exactly as described.

**Empirical probe (2026-07-09, consumer broadband; cross-checks the measurements in the PR body):**

| Endpoint | plain `eth_estimateGas` | with state override |
| --- | --- | --- |
| `eth.merkle.io` (dropped) | timeout >10s | timeout >12s |
| `eth.drpc.org` (replacement) | 1.19s | 1.12s, override honored |
| `evm-1.keplr.app` | 1.47s | 1.63s, override honored |
| `ethereum-rpc.publicnode.com` (primary) | unreachable (TLS, two stacks) | unreachable |

The merkle diagnosis reproduces exactly, and drpc supports `eth_estimateGas` with state overrides (which the sibling Skip fee PR depends on). Note the primary URL, publicnode, was unreachable from my location via both curl/schannel and Node fetch while drpc and keplr worked fine from the same network. Possibly regional or transient, but worth one check from the server environment. Mitigating irony: this PR's own bound is what makes a degraded primary survivable.

### Blocker: the 3s/no-retry policy leaks into the client-side receipt wait

`getEvmRpcTransport` has four consumers on both sides of the wire. The server ones (skip/squid quote estimation) are the intended target and the policy is right there. But `use-bridge-quotes.ts:672-731` uses the same transport for `waitForTransactionReceipt` on the user's already-broadcast approval/send txs, and `wagmi.ts:17` inherits it too.

Mechanics of the regression:

- viem's `waitForTransactionReceipt` treats a transport error during a poll as fatal: it rejects the whole wait immediately (only "receipt not found" keeps polling; verified in viem 2.23.5 source, `withRetry` wraps only `getTransaction`/`getBlock`).
- Our catch handler then fires the "transactionFailed" toast and skips `trackTransferStatus`, while the tx is confirming onchain unaffected.
- Every viem default chain except mainnet has exactly one RPC URL, so on Arbitrum/Base/Optimism/etc. there is no fallback to act as the retry.
- Before this change, each poll got up to 4 attempts x 10s with backoff (fallback-level retries), so tripping the false toast needed ~40s of sustained RPC failure. After, a single 3s blip or one 429 (routine on free public L2 RPCs) trips it.

So the mislabeling code path is pre-existing, but this change moves the trigger from "rare sustained outage" to "single transient blip" on exactly the flaky endpoints the PR is hardening against. Fail-fast is correct for a speculative, retryable quote; it is the wrong trade for a post-broadcast wait where a false negative tells the user their in-flight deposit failed.

**Suggested fix (~5 lines, preserves the intent exactly):** parameterize the policy, keep viem defaults by default, and pass the tight values at the two server call sites:

```ts
export function getEvmRpcTransport(
  chain: { rpcUrls: { default: { http: readonly string[] } } },
  opts?: { timeout?: number; retryCount?: number }
) { ... }
```

Separate follow-up (pre-existing, not this PR): the receipt-wait error handler equates "wait errored" with "transaction failed" when the truth is "status unknown". It should still kick off transfer tracking (Skip status API, not RPC-dependent) or show an unable-to-confirm state.

jasbanza:
@JohnnyWyles Addressed in 595027ad5 — `getEvmRpcTransport` now takes an optional `{ timeout, retryCount }` and defaults to viem's built-in behavior, so the client-side receipt wait (`use-bridge-quotes.ts`) and the wagmi config keep the old resilience. The tight `{ timeout: 3_000, retryCount: 0 }` policy is passed only at the server call sites — note there are **three**, not two: `skip/index.ts` has a second consumer in `getViemProvider` alongside `estimateGasFee`, plus the squid approval path.

On publicnode: probed from a different network today — `eth_blockNumber` answers in ~0.28s (drpc ~0.30s from the same network), so the unreachability looks regional/transient to your location rather than an endpoint problem.

The pre-existing "wait errored ≠ transaction failed" issue is now tracked as [MTN-219](https://linear.app/osmosis/issue/MTN-219/bridge-evm-receipt-wait-errors-are-shown-as-transaction-failed-while) with a full repro scenario.

JohnnyWyles (approved):
Blocker resolved in 595027ad5: the fail-fast policy is now scoped to the three server-side quote call sites and the client receipt wait / wagmi config keep viem defaults. Re-verified: all five call sites audited, utils + bridge typecheck clean, CI green. Remaining notes (doc phrasing on the old retry mechanism, optional policy-pinning unit test, publicnode/drpc environment checks) are non-blocking.

PR #4424, opened 2026-07-30
feat(navbar): denom-gated asset alert banners

JohnnyWyles:
## What

Urgent **asset alerts**: top banners shown only to connected wallets holding specific denoms, for deadline-driven actions like a bridge or protocol sunsetting. Alerts are authored in fe-content (`cms/asset-alerts.json`, schema PR: osmosis-labs/fe-content — linked below) and go live from a CMS merge with no frontend release. Everything is gated behind a new `asset-alerts` LaunchDarkly flag (off in prod; needs creating before enablement).

First consumer: the **BitFrost (Int3face) bridge sunset** — withdrawal period ends 2026-08-10, affecting 11 int3-bridged assets.

Linear: MTN-240.

image

## How it works

- **Holdings**: reuses the portfolio passthrough query the alloy-variants toast already makes. `getPortfolioAssets` now also returns `heldAssets` (denom + assetlist symbol) extracted from the sidecar `total-assets` breakdown, which is composed in underlying denoms across bank, staked, unstaking, in-locks, unclaimed rewards, and pooled (sidecar explodes GAMM shares and CL positions; alloys are deliberately never decomposed — alerting alloy holders requires listing the alloy denom explicitly).
- **Matching**: `getActiveAssetAlerts` does a strict exact-minimal-denom intersection (never symbols) and pairs each alert with the held assets that triggered it.
- **Copy**: localized alert text may use a `{symbols}` placeholder → the matched symbols, capped at 5 with a locale-neutral `+N` suffix; unlisted denoms fall back to a shortened denom.
- **Layout**: banners (announcement + up to `MAX_STACKED_BANNERS` alerts) render in one fixed, measured stack. A ResizeObserver publishes the rendered height through the nav-bar zustand store; the in-flow spacer and the trade page's viewport-fixed content both consume it, so wrapped/multi-line banners can't overlap content. This replaces the previous hardcoded 52px/`h-[124px]` assumption and **also fixes a pre-existing ~19px overlap** where the current prod announcement banner clips the swap ads banner on the trade page.
- **Lifecycle**: per-alert localStorage dismissal (read at derive time — no flash of previously dismissed banners), `startDate`/`endDate` windows re-evaluated via a timer at the next boundary so alerts appear/expire mid-session, and derived visibility ignores cached query data the moment the flag flips off or the wallet disconnects.

## Testing

- 17 specs for the alert utilities (exact-denom matching incl. rejection of symbol/partial matches, date windows incl. boundary semantics, matched-asset collection, `{symbols}` interpolation, symbol cap, shortened-denom fallback).
- 12 nav-bar store specs (banner height setter, reset survival).
- 10 server portfolio specs (`getHeldAssets` symbol pairing, unlisted-denom fallback, pooled-only holding present in total-assets but not user-balances; fixtures match the real sidecar response shape).
- Typecheck, eslint, prettier clean on touched packages.
- Manually verified on a local dev server with stubbed CMS + spoofed holdings: single alert, two simultaneous alerts + announcement banner stacking, dismissal persistence, long symbol lists, trade-page offset.

## Notes for review

- With the flag off (current state) nothing changes; the announcement banner renders as before.
- Date checks in the new util use plain `Date` rather than the navbar's dayjs — deliberate: self-contained and directly unit-testable.
- Deferred with follow-ups: e2e for stacking/dismissal (once the CMS file is live), mobile-specific banner heights (pre-existing).

## Risks

- Alert visibility depends on the sidecar total-assets breakdown; when a sidecar sub-fetch fails (`is_best_effort`) a holder may briefly not see the banner (fails quiet, next refetch recovers).
- Trade-page offset now derives from measured height at runtime; verified locally across banner add/dismiss/resize.

### DIFF
--- a/packages/server/src/queries/complex/portfolio/__tests__/assets.spec.ts
+++ b/packages/server/src/queries/complex/portfolio/__tests__/assets.spec.ts
@@ -1,7 +1,7 @@
 import { AssetLists as assetLists } from "../../../__tests__/mock-asset-lists";
 import { PortfolioAssetsResponse } from "../../../sidecar";
 import { calculatePercentAndFiatValues, getAllocations } from "../assets";
-import { checkAssetVariants } from "../assets";
+import { checkAssetVariants, getHeldAssets } from "../assets";
 
 const MOCK_DATA: PortfolioAssetsResponse = {
   categories: {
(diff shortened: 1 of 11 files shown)


JohnnyWyles:
Feature flag flipped on edgenet to show on vercel preview - awaiting https://github.com/osmosis-labs/fe-content/pull/129 checks and merge

jasbanza (approved):

PR #4423, opened 2026-07-30
chore(deps): re-resolve in-range dependabot advisories (tier 2)

jasbanza:
## What is the purpose of the change:

Tier 2 of the Dependabot backlog triage: re-resolve the runtime and dev transitives whose patched version already sits inside their existing semver range. Closes 34 open alerts (17 high, 14 medium, 3 low) with no manifest, range, or code changes — `yarn.lock` only.

### Linear Task

MTN-232 — https://linear.app/osmosis/issue/MTN-232/dependabot-tier-2-runtime-app-deps-in-range-re-resolves

## Brief Changelog

Runtime:

- `@trpc/server` 10.45.2 → 10.45.4 — prototype pollution ([#249](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/249)). Highest-value item here: the app uses tRPC directly on the server.
- `h3` 1.11.1 → 1.15.11 ([#315](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/315), [#320](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/320), [#319](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/319), [#316](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/316), [#254](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/254)).
- `defu` → 6.1.7 ([#349](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/349)), `flatted` → 3.4.4 ([#318](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/318), [#314](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/314)), `path-to-regexp` → 1.9.0 / 6.3.0 ([#146](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/146)).
- `picomatch` → 2.3.2 / 4.0.5 ([#327](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/327), [#330](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/330), [#328](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/328)), `micromatch` → 4.0.8 ([#143](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/143)), `socket.io-parser` → 4.2.7 ([#317](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/317)).
[...]

PR #4420, opened 2026-07-23
chore(deps): patch dependabot crypto/signature advisories (tier 1)

jasbanza:
## What is the purpose of the change:

Tier 1 of the Dependabot backlog triage: patch the cryptographic and signature libraries, which are the highest-impact open advisories for a wallet/DeFi frontend (private-key extraction, signature forgery, address-validation bypass). All fixes are in-range, patch-level lockfile re-resolves — no direct-dependency or code changes.

Closes **15 alerts** (13 high, 2 medium).

### Linear Task

MTN-235 — https://linear.app/osmosis/issue/MTN-235

## Brief Changelog

- `secp256k1` 4.0.3 → 4.0.4 and 5.0.0 → 5.0.1 — private-key extraction over ECDH ([#155](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/155), [#154](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/154)).
- `node-forge` 1.3.1 → 1.4.0 — Ed25519/RSA-PKCS signature forgery, basicConstraints bypass, modInverse DoS ([#335](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/335), [#334](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/334), [#333](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/333), [#332](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/332), [#247](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/247), [#246](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/246), [#245](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/245)).
- `browserify-sign` 4.2.1 → 4.2.6 — dsaVerify signature forgery ([#118](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/118)).
- `base-x` 3.0.9 → 3.0.11, 4.0.0 → 4.0.1, 5.0.0 → 5.0.1 — homograph validation bypass ([#190](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/190), [#189](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/189), [#188](https://github.com/osmosis-labs/osmosis-frontend/security/dependabot/188)).
[...]

Issue #598, opened 2022-06-06
Revisit "My Pools" section in pools page


jonator:
[Notion meeting notes](https://www.notion.so/osmosiszone/6-6-2022-Pools-page-49ccb1cf9a4644e988ff5faf313e020b)
[Figma](https://www.figma.com/file/wQjMyxY0EnEk29gBzGDMe5/Osmosis-Component?node-id=5042%3A31893)

We've received a lot of feedback that it's confusing to see pools at the top of the assets page (as pool cards). We should revisit how we display owned pools (LP positions) in the assets page.

- [ ] Move "My Pools" section (as it is) below assets table & IBC transfers table
- [x] Move hide zero balances switch to left of search bar in assets table
- [x] Add link to support site
- [ ] Add highly engaging link to frontier site (address concerns with finding non-scammy links)

CMarshall92:
If your happy for me to pick this up I could implement the changes fairly quickly i think.

Will post a PR soon

CMarshall92:
Can we link the new PR thats now not a draft https://github.com/osmosis-labs/osmosis-frontend/pull/653

PR #4427, opened 2026-08-04
Release to production

JohnnyWyles:

Release `stage` to production. This release ships denom-gated asset alert banners (a CMS-driven tool for deadline-driven holder warnings, live without a frontend release), an incentive gauge creation and top-up interface that replaces CLI-only `MsgCreateGauge` access (shipping **behind a feature flag, off by default**), high-loss bridge acknowledgement enforced on the signing surface, a bridge max-amount fix for gas fee drift, two display fixes (alloy variants no longer duplicated on the New screen, price chart y-axis label clipping), removal of the add-liquidity quick action for cosmwasm pools, a full Dependabot backlog triage plus a protestware override, and E2E monitoring cost reductions with Slack fleet reporting.

### Linear Task

Per-change tasks are linked in the individual PRs below (MTN-43, MTN-137, MTN-199, MTN-214, MTN-230, MTN-232, MTN-234, MTN-235, MTN-240, MTN-242).

## Brief Changelog

**Features**

- #4424 — feat(navbar): urgent asset alert banners shown only to connected wallets holding specific denoms, for deadline-driven actions like a bridge sunsetting. Authored in fe-content (`cms/asset-alerts.json`) and live from a CMS merge with no frontend release.
- #4401 — feat(pools): guided UI for external incentivisers to fund a new incentive gauge on a pool (`MsgCreateGauge`), which had no frontend integration before this (no codecs, no send method, no UI).
- #4429 — feat(pools): gate that gauge creation entry behind a new `incentivizePool` flag, **default off**. It is a new signing surface with a 50 OSMO non-refundable fee, so it needs a kill switch that does not require reverting a release. Both the entry button and the modal are gated on all three pool-detail surfaces, since the modal opens off separate state. **The `incentivize-pool` LaunchDarkly flag exists and is off in every environment, so gauge creation ships dark.**
- #4403 — feat(bridge): a high-loss bridge withdrawal can no longer be signed without an explicit, still-fresh acknowledgement. The checkbox previously lived only on the amount screen, was never reset when the quote drifted or the provider auto-switched, and was absent from the review screen that actually signs.

**Fixes**

- #4410 — fix(bridge): widen the max-amount gas reserve to 3x and stop the max clamp trusting a stale gas quote while the fee market moves underneath it. Follow-up to #4406, where field-tested max ETH deposits still overshot Keplr's reserve at 2x.
- #4426 — fix(assets): exclude non-canonical alloy variants from the New category, so a newly listed constituent no longer appears alongside the alloy it backs. Surfaced by `cbDOGE.axl` (an `allDOGE` constituent).
- #4425 — fix(pools): hide the add-liquidity quick action on the pools table for cosmwasm pools (transmuter/alloyed, orderbook, Astroport PCL, Skeleton Swap, generic cosmwasm).
- #4415 — fix(chart): size the price chart's left y-axis gutter to the widest label instead of a hardcoded 36px, which clipped the leading digit (allBTC/USDC showed `18,000` instead of `118,000`).

**Dependencies** (Dependabot backlog triage, lockfile-only unless noted)

- #4420 — tier 1: cryptographic and signature libraries, the highest-impact advisories for a wallet frontend (private-key extraction, signature forgery, address-validation bypass).
- #4423 — tier 2: re-resolve runtime and dev transitives already patched inside their existing semver range. Closes 34 alerts (17 high, 14 medium, 3 low).
- #4418 — patch protobufjs, axios, js-yaml and brace-expansion without disturbing older majors that intentionally coexist in the lockfile.
- #4419 — patch sharp, brace-expansion and shell-quote (alerts #598-601, surfaced by the rescan after #4418).
- #4416 — bump axios from 0.27.2 to 0.33.0.
- #4430 — alias `es5-ext` to the `@unes/es5-ext` fork, which strips the upstream postinstall that geolocates the host by timezone and prints an anti-war message (Socket High protestware alert, Nexus advisory sonatype-2022-2248). Build tooling only, via a `@cosmology/telescope` devDependency, so it never shipped to users; what it removes is a script that ran on developer machines and CI runners.

**CI / infrastructure** (no production code)

- #4411 — ci(e2e): run the two expensive US monitoring suites 12-hourly rather than hourly. Flow accounting (Mar-Jul 2026) measured ~297 USDC/month consumed almost entirely by those suites.
- #4412 — ci(e2e): cut monitoring trade sizes ~3x and add a surplus-sweep workflow, targeting conversion leakage that scales with trade size (~$550 parked across the wallets).
- #4414 — ci(e2e): periodic Slack report of total assets across all five E2E accounts valued at SQS prices, with burn-rate and runway trend. Read-only.

## Security alerts on this release

Socket raised two High-severity Warn alerts. Both were triaged; see the [triage comment](https://github.com/osmosis-labs/osmosis-frontend/pull/4427#issuecomment-5179659459) for the detail.

- **`es5-ext` protestware: real, and fixed** by #4430, now included in this release.
- **`node-forge` 1.3.1 → 1.4.0 "90% likely obfuscated": verified false positive.** Both published tarballs were diffed (the 1.4.0 SHA-512 matches this release's lockfile). It is not minified, let alone obfuscated: `jsbn.js` averages ~27 characters per line, and its 902-line diff collapses to 64 under `diff -w -B` because the bulk is `} else {` reformatting. Every substantive change is a fix, three of them security fixes (ed25519 signature malleability, x509 CAs missing `basicConstraints`, pkcs12 unvalidated MAC). No new network, `eval`, or exfiltration primitives in any added line, and no install-time scripts. 1.4.0 is strictly more secure than 1.3.1, so no override was applied.

## Testing and Verifying

Each change was tested and verified in its own PR (see the linked PRs). The post-merge suite on `stage` passed at `396d09af0` (Jest, Lint and Format, Server E2E, Preview frontend E2E), and CI is re-running against the two later merges (#4430 at `713bc020d`, #4429 at `7f309ccf9`).


JohnnyWyles:
Triaging the two High-severity Socket Warn alerts above (https://github.com/osmosis-labs/osmosis-frontend/pull/4427#issuecomment-5177404278). Both checks pass, so neither blocks, but both were introduced by this release and both are worth a decision.

**Summary: one is real and fixed elsewhere, one is a verified false positive.**

## 1. `es5-ext` 0.10.62 → 0.10.64, protestware: REAL

Upstream ships `"postinstall": "node -e \"try{require('./_postinstall')}catch(e){}\" || exit 0"`. `_postinstall.js` geolocates the host by timezone and prints an anti-war message, and the `try/catch` plus `|| exit 0` makes it fail silently by design. Nexus Firewall quarantines the package (sonatype-2022-2248).

**Never a production exposure.** It reaches us only through `@cosmology/telescope`, a devDependency of `packages/proto-codecs`, so it has never been in the web bundle. What it does is run a script on developer machines and CI runners.

Fixed in #4430 by aliasing to the `@unes/es5-ext` community fork (verified: fork has `"scripts": {}`, and `_postinstall.js` is absent from `node_modules` after install). That PR is intentionally separate so this release is not blocked on it.

## 2. `node-forge` 1.3.1 → 1.4.0, "90% likely obfuscated": FALSE POSITIVE

Both published tarballs were pulled and diffed rather than reasoned about. The 1.4.0 SHA-512 matches this release's lockfile integrity hash, so this covers the exact artifact that would ship.

- **Not obfuscated, not even minified.** `jsbn.js`, the file driving the large diff, averages **~27 characters per line** in both versions.
- **The big diff is a reformat.** 902 changed lines in `jsbn.js` collapse to **64** under `diff -w -B`; the bulk is `} else {` rewritten as `}\nelse {`. The heuristic likely reacted to that churn plus the natural density of a bignum library.
- **Every substantive change is a fix**, and three are security fixes: `ed25519.js` rejects non-canonical signature scalars (**signature malleability**), `x509.js` rejects CAs missing `basicConstraints` (**certificate validation**), `pkcs12.js` throws instead of silently accepting unvalidated MAC data (**MAC bypass**). Plus a jsbn `square()` API and infinite-loop guard, one added OID, and comment typo corrections across `aes`/`pbe`/`tls`/`util`/`xhr`/`prime.worker`.
- **No exfiltration surface added.** Every added line across all 13 changed files was grepped for `eval`, `new Function`, `fetch`, `XMLHttpRequest`, http/https URLs, `child_process`, `process.env`, base64 decoding, WebSocket, and `.send(`. Zero hits.
- **No install-time scripts** in either version (only `prepublish`, which never runs on consumers).
- 1.4.0 only **removes** files (CHANGELOG, legacy Flash shim) and adds none.

**1.4.0 is strictly more secure than 1.3.1.** Pinning back to silence the heuristic would remove three real cryptographic fixes, and node-forge sits four levels deep in the wallet stack (`@cosmos-kit/core` → WalletConnect → wagmi → node-forge) on the user fund path, so an override there carries real risk for no benefit. Suggest marking it acceptable risk in the Socket dashboard so it stops recurring on every PR.

Full detail in #4430.

## Unrelated note on the dependency diff table

Socket reports `npm/axios@1.14.0 ⏵ 0.33.0`, which reads as a downgrade. It is not. `master` carries five separate axios entries; `stage` dedupes them (`axios@^1.4.0, axios@^1.6.0` merged, and `^0.27.2, ^0.33.0` merged), and Socket paired entries across the reorganised lockfile. The 1.x line is untouched; the 0.x line moved 0.27.2 → 0.33.0 as intended by #4416.

jasbanza (approved):

dogemos (approved):

PR #4434, opened 2026-08-10
fix(assets): dismiss the alloyed-assets toast per variant group

JohnnyWyles:

Make the convert-to-alloy toast reach users when a new alloy lists, instead of staying silent for anyone who dismissed it once before.

Dismissing the toast wrote a single global localStorage boolean, `do-not-show-alloyed-assets-toast`. Once set, it suppressed the toast for every alloy, permanently. So when a new alloy lands and the symbol and `variantGroupKey` move onto the alloy denom, every previously-dismissed holder of a variant would never be prompted to convert. That is the group most worth reaching: they held a variant and saw the prompt before.

The old checkbox made the problem bigger than deliberate opt-outs would suggest. It read "Remind me later", and leaving it unchecked, the default, was the permanent-suppression option. Anyone who closed the toast without reading the checkbox opted out of every future alloy.

Dismissal is now recorded per variant group, and the checkbox is "Don't show again" with suppression opt-in. Background and the migration context it serves: see the Linear task.

### Linear Task

https://linear.app/osmosis/issue/MTN-255/alloyed-assets-toast-dismissal-is-global-so-no-previously-dismissed

## Brief Changelog

Each change with the one-line reason it exists:

- **`use-alloyed-assets-toast-dismissal.ts` (new)** stores a set of dismissed variant group keys (the canonical asset's own `coinMinimalDenom`) instead of one boolean, so declining one alloy cannot hide another.
- **The legacy key is ignored, not migrated.** Ignoring the old key re-offers a few previously-declined alloys once, which is the cheaper mistake.
- **State lives in a zustand store, not `useLocalStorage`.** The toast renders in react-toastify's tree from `_app.tsx` while the gate lives in the conversion modal's tree, so they are separate hook instances. `react-use`'s `useLocalStorage` is per-instance `useState` with no cross-instance sync, so a dismissal written by one was invisible to the other, and each merged onto its own stale snapshot. A probe confirmed two writers lost one write outright.
- **Every mutation re-reads storage before merging**, so writers within a tab accumulate rather than clobber, and a write picks up what another tab already stored. This is best-effort across tabs, not atomic: `getItem` then `setItem` is not a transaction, so two tabs writing in the same instant can still lose one update. There is no lost-update primitive for localStorage and the cost is one re-offered toast, so it stays uncoordinated and the docstring says so.
- **Checkbox inverted to "Don't show again"** (`alloyedAssets.dontShowAgain`, all 17 locales), so the default action just closes the toast. `remindMeLater` is removed with no remaining consumers.
- **`shouldPersistDismissalOnClose` (new, named on purpose)** holds the polarity, so the old inverted behaviour cannot return without failing a test.
- **`use-alloyed-assets-toast-session.ts` (new)** moves once-per-session suppression to `sessionStorage`, so it survives a reload instead of resetting on every page load. Its hydration flag doubles as the SSR mount gate.
- **`alloyed-assets-toast-policy.ts` (new)** holds eligibility and visibility as pure functions, lifted out of the tRPC `onSuccess` callback so the release-critical rules are assertable without mounting a component or a query client. The callback is now wiring only.
- **The proactive toast is alloy-only.** `variantGroupKey` also groups non-alloy canonical assets (34 of 65 live groups, including legacy consolidations like `MARS.old` and today's ibc-denom USDC with its ten bridged variants), and volunteering those unprompted is noise. Conversion stays available for every group through the portfolio Convert button and the modal, so this narrows the unsolicited toast only and deliberately does not gate `checkAssetVariants`. It self-resolves for USDC: when that canonical becomes an alloy, it starts qualifying with no code change.
- **`displayToast` gains overloads** so the type system enforces that only the alloyed toast carries `variantGroupKeys`, and it throws on an empty list rather than letting a dismissal silently do nothing.

### Manual verification

Two things to know before looking for the query, both of which make it invisible on a preview build:

- `local.*` procedures do not use HTTP. `localLink` builds a tRPC caller and runs the resolver in the browser, so there is **no Network entry** for `portfolio.getPortfolioAssets`.
- The logger link is error-only outside development. Its `enabled` predicate is `NODE_ENV === "development" || (direction === "down" && result instanceof Error)`, so on a Vercel preview or production build **only failed calls reach the console**. A successful call runs silently.

So the toast appearing is itself the signal that the query succeeded and at least one alloyed group was eligible. Verify the rest through storage:

1. Dismiss the toast with the box ticked, then in Application, Local Storage check `dismissed-alloyed-assets-toast-groups`. It should list exactly the alloy denoms you hold variants of (`factory/.../alloyed/...`) and nothing else. A non-alloy canonical appearing here, for example an `ibc/` denom, means the eligibility filter regressed.
2. Reload. The toast should not return for those groups, and should still appear if you hold a variant of some other alloy.
3. Dismiss with the box unticked and confirm nothing is written. This is the polarity fix: the default action must not persist anything.
4. `seen-alloyed-assets-toast-this-session` lives in **sessionStorage**, not localStorage. It now survives a reload, so clear it or open a new tab to make the toast eligible again within a session.
5. Any pre-existing `do-not-show-alloyed-assets-toast` should remain untouched and no longer suppress anything.

To inspect the payload itself, either run locally with `NODE_ENV=development` so the logger prints both directions (`result.data.assetVariants[]`, where each entry has `canonicalAsset.coinMinimalDenom` as the group key and `canonicalAsset.isAlloyed` as the eligibility flag), or read the persisted cache in Application, IndexedDB, `tanstack-query-cache`.

Note a disconnected wallet produces a `Fetch error. Not Found.` for this query from the other call sites, which do not gate on the alloyed-assets flag. That is the sidecar rejecting `passthrough/portfolio-assets/` with an empty address, is pre-existing on `stage`, and is unrelated to this change.

## Behaviour change worth calling out

Users who deliberately declined allBTC or allETH get re-prompted once. Given the old default suppressed permanently on a plain close, most of those dismissals were not deliberate, so one re-prompt against a dismissal that now works per group is the tradeoff.


jasbanza:
@JohnnyWyles please can you check if the copilot comments are valid?

jasbanza (approved):

PR #4440, opened 2026-08-19
Release to production

JohnnyWyles:

Release `stage` to production. This release ships the Int3face bridge provider removal now that the bridge has shut down permanently, a correctness fix that restores the intended zero-slippage guard on 1:1 variant conversions, warn-accept flags stamped into the transaction auth memo so a tx hash alone shows what a user was warned about, accurate messaging for timed-out Wormhole packets in place of a misleading "try again in a minute", per-variant-group dismissal for the alloyed-assets toast, and removal of one bundled copy of axios 0.21.4 from the client bundle.

### Linear Task

Per-change tasks are linked in the individual PRs below (MTN-261, MTN-257, MTN-137).

### Bridge
- Remove the Int3face bridge provider at sunset, including the quote provider, transfer-status polling and the alloy `allTokenMinimalDenom` route (#4432). Deposits were already halted via assetlists #3570; this lands only after withdrawals closed, so the in-app exit route was never pulled out from under users.
- Stamp warn-accept flags into the tx auth memo, e.g. `OsmosisFE/warn:slip=99.89,pi=12.40` (#4409, MTN-137). Builds on the signing-surface acknowledgement shipped in the previous release (#4403).
- Explain timed-out Wormhole packets on the `/wormhole` recovery widget instead of telling every user with no Wormchain receive to wait (#4436, MTN-257). A timed-out packet was refunded and will never be relayed, so the widget now distinguishes in-flight from refunded before choosing what to say.

### Pools
- Demand exact output only for genuinely 1:1 conversion routes (#4433). The slippage guard in `getConvertVariantMessages` was dead code: `&&` short-circuited on an out-of-bounds index, so every variant conversion used the 5% allowance instead of the intended 0. Since `maxSlippage` feeds `tokenOutMinAmount` directly, a 1:1 transmuter conversion was signing a transaction willing to accept 95% of the quoted output.

### Assets
- Dismiss the alloyed-assets toast per variant group rather than via a single global `do-not-show-alloyed-assets-toast` boolean (#4434). Previously one dismissal permanently suppressed the prompt for every future alloy, silencing exactly the holders most worth reaching when a new alloy lists.

### Dependencies
- Keep one of the two copies of axios 0.21.4 out of the production `_app` chunk (#4439, MTN-261). Removed at the import site (`ChainType` is a runtime enum, so importing it as a value dragged the whole SDK in) rather than by forcing resolutions.

## Testing

Stage CI is green. Unit tests accompany the memo stamping, alloyed-toast dismissal, variant-conversion and Wormhole redeem changes.

## Notes for review

- #4433 changes quote direction handling on conversion routes. Worth a targeted check on stage that variant conversions still quote and sign correctly before promoting.
- Int3face disappearing from the bridge picker and the Wormhole timeout copy are the two user-visible changes in this release.
- 64 files changed, +1959/-964, of which 17 are localization files.


JohnnyWyles:
Superseded by #4441. The previous release (#4427) was squash-merged, which left master and stage with divergent histories: this PR renders against a stale July merge base (133 files instead of the real 64) and its merge conflicts cannot be resolved, since resolution would push a merge commit to stage, which stage's linear-history protection forbids. #4441 carries the identical tree (byte-for-byte stage) with current master merged in as a parent, so it displays the true unreleased delta, merges conflict-free, and restores shared ancestry for future releases.

PR #4450, opened 2026-08-31
feat(bridge): support multi-tx Skip routes (CCTP deposits into alloys)

JohnnyWyles:
## What

Adds multi-transaction route support to the Skip bridge provider, behind a new `multiTxBridgeRoutes` LaunchDarkly flag (default off in prod).

This unlocks USDC deposits from the CCTP L2s (Arbitrum, Base, Optimism, Polygon, Avalanche) into alloyed USDC, which currently have **no single-transaction route at all**: the CCTP leg lands on Noble, whose forwarding cannot carry the swap memo, so the final leg into the alloy must be signed separately on Noble.

Linear: MTN-287

Current State:
image

New State:
image

## How it works

**Provider (`packages/bridge`)**
- Single-tx routes are preferred, but not at any price: with the flag on, both route variants are quoted in parallel and the single-tx route is kept unless the multi-tx route's output beats it by more than 0.5%. Preferring single-tx unconditionally was wrong in practice: QA on Avalanche USDC showed a single-tx axelar+swap route paying ~21% less than the multi-tx CCTP route (15.68 vs 19.96 out of 20), matching what Skip's own site quotes. With the flag off, only the single-tx request is made (byte-identical to the pre-multi-tx shape). Multi-tx quoting is deposit-only: the executor supports exactly an EVM first tx plus one cosmos step, so shapes it cannot execute are never quoted.
- Multi-tx quotes carry ordered `transactionSteps` (first step doubles as `transactionRequest`, so single-tx consumers are untouched), the quoted route snapshot (`multiTxRouteData`), and the later steps' network fees (`intermediateGasFees`, priced into the quote's fee total).
- Address lists are built from `required_chain_addresses` (which can repeat chains) rather than `chain_ids`.
- **Key-derivation gate**: multi-tx routes through an intermediate chain that is not secp256k1 / coin type 118 are refused at quote time. Quote-time messages derive intermediate addresses by bech32 conversion, which on e.g. Injective (ethsecp256k1, coin type 60) yields an address the user does not control, and the first tx would route funds through it. Injective-intermediate routes stay off until a quote-time wallet-address mechanism exists (tracked as MTN-294); their single-tx Axelar alternative still serves them.
- `getTransactionStep` rebuilds a later step for signing by replaying `/msgs` for the originally quoted route with the wallet's own account as sender (asserted), a fresh timeout, and a fresh gas estimate. It never re-routes. Stored relay fee quotes (`smart_relay_fee_quote`) are stripped before the replay: Skip validates their ~30 minute expiration on submission even though the relay fee was already collected by the first transaction, so an unstripped replay stops building about half an hour after quoting.
- **Fee sizing for the intermediate step**: the default safe-side high gas price can exceed the flat amount the route reserves for fees on the intermediate chain (Skip holds back 0.02 USDC on Noble, while high-priced Noble gas comes to ~0.028), so the step's fee is sized to a budget: the route's reserved amount, improved at sign time by the account's live balance (topping up the fee token genuinely raises it). Over budget, the fee reprices at the chain's minimum gas price; still over, the gas limit is capped into the budget, refused with actionable copy only when a real simulation proves the step cannot fit.

**Execution (`packages/web`)**
- Preflight before the irreversible first tx: connect the wallet on the intermediate chain, verify its account matches the address the quoted transactions were built against, and check the final step's fee-token balance when it is not paid from the arriving funds (failing closed on an unreadable balance). Nothing is signed until all pass.
- Step flow: sign the EVM tx (existing ERC20 approval sub-flow), persist a resumable pending-step history entry at broadcast, poll Skip status until funds reach the intermediate chain, then rebuild and sign the final step there. The review screen shows the step notice and live phase copy; provider rows show a step-count badge. Reverted EVM receipts (which viem returns rather than throws) fail the entry instead of leaving it pending.
- The resumable entry is awaited to durable storage before polling or signing, and the flow stops loudly if it could not be recorded: the final step's signing guard and the only recovery surface both read persisted storage, so proceeding without it would dead-end with funds on the intermediate chain. Duplicate terminal callbacks from a status provider are ignored before they can mutate a snapshot, so a late duplicate cannot clobber a restored pending step.
- Resume: closing the app mid-flow is safe. The history entry persists the quoted route, and transfer rows in Transactions / Recent activity show a **Continue** button that re-checks arrival (with immediate in-transit feedback), connects the intermediate wallet, and prompts the final signature. Mid-flow entries are never handed to the status provider (whose first-leg "success" would misreport the transfer as complete); tracking hands over on the final signature, polling on the intermediate chain.
- **Failed final step recovery**: if the signed final step fails onchain or its transfer times out, the funds are still on the intermediate chain, so the entry restores its pending step and offers Continue again with copy saying where the funds are, instead of turning into a terminal failed row. A step version counter orders advance vs restore in the cross-session persistence merge. Unresolved mid-flow entries outlive the normal history expiry (Continue is the recovery path), bounded by a 30-day long stop.
- Named failures carry specific recovery copy in both the live flow and resume: an expired stored route (funds safe, route no longer signable) and a fee shortfall (top up gas on the intermediate chain, then Continue). The client matches these on the error message, which is all that survives the tRPC boundary; the literals are mirrored in `packages/web/utils/multi-tx.ts` because importing a value from `@osmosis-labs/bridge` in client code pulls that package (and its Node-only LaunchDarkly dependency) into the browser bundle and breaks the web build. A test pins the mirrored literals to the bridge constants so a reword on either side fails CI rather than silently dropping the copy.

## Known limitations
- The cross-session double-sign guard (persisted history + Web Locks) is per browser profile. From a second device the remaining guard is the expected-arrival balance check, which passes if the intermediate account happens to hold unrelated funds of at least the expected amount; a second signature would sweep those to the user's own Osmosis address via the memo swap. Self-custody on both ends, so no loss of custody, but worth knowing.

## Testing
- [x] Manual QA: Avalanche USDC -> allUSDC deposit on the preview deploy
- [x] Manual QA: resume flow (close app after tx1, continue from Transactions)
- [x] Real-funds review round: findings on relay-quote expiry and Noble fee headroom addressed (quote stripping, budget-sized fees)
- [x] Unit coverage: Skip provider fee-budget sizing (reprice, cap, refusal, live-balance top-up) and route-replay quote stripping; transfer-history advance/restore/merge-ordering, long-stop expiry, and duplicate-callback handling
- [ ] Fresh real-funds pass on the resume path against the updated preview
- [ ] Cross-tab double-sign lock retest (was blocked by the expiry bug)


packages/web/stores/transfer-history.tsx
```diff
@@ -335,14 +390,99 @@ export class TransferHistoryStore implements TransferStatusReceiver {
...
+        snapshot.firstStepTxHash === firstStepHash ||
+        snapshot.sendTxHash === firstStepHash
+    );
+    if (!storedSnapshot) return "missing";
+
+    if (storedSnapshot.pendingStep?.stale) {
+      this.markPendingStepStale(sendTxHash);
+      return "stale";
+    }
+    if (storedSnapshot.pendingStep) return "resumable";
```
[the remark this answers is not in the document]
JohnnyWyles: Already satisfied as of 107510bcc (this comment was raised against an earlier commit). The atomic claim lives inside `signFinalStep` itself, which is the single choke point BOTH signing paths go through (the live flow after arrival, and history resume): it holds a cross-tab exclusive Web Lock keyed by the first step's hash for the whole critical section, performs the authoritative persisted-state recheck INSIDE the lock, and releases only after the broadcast advance has been durably written via an awaited persist (`persistNow`), so the next lock holder's check is guaranteed to observe it. A session finding the lock taken yields (`ifAvailable`), since the holder is actively signing that very step. Where Web Locks are unavailable, the storage recheck remains as documented best effort.

96524c37c additionally made the persistence layer monotonic (union merge that never regresses an advanced entry), which closes the full-array-overwrite half of this concern.

packages/web/components/bridge/use-bridge-quotes.ts
```diff
@@ -693,96 +846,219 @@ export const useBridgeQuotes = ({
...
-          : undefined,
-      });
-
-      setIsBroadcastingTx(true);
+      // The funds this step will move on the intermediate chain, for the
+      // resume-time sanity check. Replay protection does NOT come from
+      // balances (shared state that changes for unrelated reasons): it
+      // comes from the persisted history store, which the signing session
+      // updates at final-step broadcast and any stale session re-reads
+      // before signing (syncPendingStepFromStorage).
```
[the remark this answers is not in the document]
JohnnyWyles: Declining the `preArrivalBalance` mechanism: it was implemented and then deliberately removed after review, because an aggregate balance is shared state that moves for unrelated reasons. A baseline delta both strands valid transfers (spending unrelated funds while tx1 is in flight makes an executable step look stale) and passes falsely (unrelated funds arriving after completion satisfy the threshold). Replay protection instead comes from transaction-level cross-session state: the cross-tab exclusive lock in `signFinalStep` plus the persisted history recheck inside it, with the advance durably written before the lock releases (107510bcc, hardened in 96524c37c).

On the gas point: the remaining balance comparison is explicitly a sanity check, not a spend proof. A fee shortfall that slips past it fails safely on-chain (insufficient funds), and the fulfillment handler marks the entry failed; the rebuilt step's gas fee is only known after `/msgs` rebuild, so pre-checking it would not be authoritative anyway. The residual case where the user manually moved the arrived funds but holds enough of the same denom results in their own initiated transfer completing from their own account, which is not a double-spend.

packages/web/components/bridge/use-bridge-quotes.ts
```diff
@@ -693,96 +846,219 @@ export const useBridgeQuotes = ({
...
-          hash: approveTxHash,
+        return;
+      }
+      // When the final step's fee isn't paid from the arriving funds (e.g.
+      // INJ on Injective), the account there must already hold the fee token.
+      const stepGasFee = finalStep.gasFee;
+      if (
+        stepGasFee &&
+        fromAsset &&
+        stepGasFee.denom.replace(/^u/, "").toLowerCase() !==
```
jasbanza: This check is skipped on Noble, which is the one case that needed it.

`stepGasFee.denom.replace(/^u/, "")` turns `uusdc` into `usdc`, which matches the transferred `USDC`, so the fee-balance preflight never runs on a Noble final step.

The comparison is correct about the denominations. The assumption underneath it is that when the fee token is the transferred asset, the arriving funds necessarily cover the fee — and finding 2 in my review is the case where that's false by about 0.008 USDC. Skip reserves a flat 0.02, the wallet wants 0.028216, and nothing catches it before the user signs.

A headroom check comparing the reserve against an actual fee estimate would catch this; the denom equality check can't.

packages/web/components/bridge/use-bridge-quotes.ts
```diff
@@ -939,6 +1231,41 @@ export const useBridgeQuotes = ({
...
+      /**
+       * True when the final step's fee token loosely differs from the
+       * transferred asset (uusdc vs USDC matches; inj vs USDC doesn't) — the
+       * arriving funds then can't pay the step's own gas, so the user's
+       * account on the intermediate chain must hold the fee token.
+       */
+      finalStepGasWarning: Boolean(
+        gasFee &&
+          fromAsset &&
+          gasFee.denom.replace(/^u/, "").toLowerCase() !==
```
jasbanza: Same `replace(/^u/, "")` comparison as the preflight above, so `finalStepGasWarning` can never fire on Noble either.

Worth flagging because it hides the problem twice over: I'd noted the absence of a gas warning during QA as correct behaviour, when it's actually this suppression. Both guards against an underfunded final step are off on the chain where the shortfall happens.

jasbanza (commented):
I ran this on the Vercel preview with real funds rather than reading it, including the interrupted-transfer paths. Two findings from that are worth your attention before this goes out, both hit during a single $10 Avalanche USDC into allUSDC deposit that ended with the funds sitting on Noble and no way to move them from inside the app. I recovered them with a manual IBC in the end.

Everything I verified as working is at the bottom, and the feature itself is clearly worth having — I measured the multi-tx CCTP route returning 19.96 of $20 against 16.38 for the single-tx Axelar alternative, so the 0.5% preference threshold is picking the right route.

## 1. The final step's relay fee quote expires, and nothing checks it

The persisted route embeds Skip's `smart_relay_fee_quote` with an `expiration` timestamp. Mine was `2026-09-11T23:53:05Z`, roughly 35 minutes after the quote was taken. `getTransactionStep` replays `/v2/fungible/msgs` against the originally quoted route, which is the right call per section 2 of MTN-287, but once that timestamp passes Skip rejects the replay.

- **Repro**
  - Start an Avalanche (or any L2) USDC deposit into allUSDC and sign the first CCTP transaction.
  - Let the funds land on Noble, then leave it alone for an hour.
  - Press Continue on the history row.
  - The wallet never opens and a generic error toast appears.
- **What comes back**
  ```
  500 Fetch error. relay fee quote has expired.
      path: bridgeTransfer.getTransactionStepByBridge
  ```
- **Why it's hard to recover from**
  - Neither `smart_relay_fee_quote` nor `expiration` is referenced anywhere in `packages/bridge/src/skip`, so nothing checks the timestamp before offering Continue and nothing handles the specific failure.
  - The error lands in the catch-all at the bottom of `useMultiTxResume.resume`, so the user sees `somethingIsntWorking` / `sorryForTheInconvenience`. Nothing indicates the route expired and there's no suggested action.
  - Continue stays enabled, so the natural response is to keep pressing it.
  - `isLive` exempts unresolved mid-flow entries from the 3-day prune, on the reasoning that Continue is the only recovery path. That's sound in principle, but it means the button is offered for three days after it stopped working.
- **Topping up gas doesn't help**, which I'd flag because it's the obvious first thing to try. I retried two days later after funding the Noble account well above the fee, and it failed identically with no wallet prompt — the msgs replay runs before any fee is computed and before anything is offered for signature.
- **Possible directions**, though you'll know the Skip API better than I do:
  - Stop offering Continue once `expiration` has passed, and say the route expired rather than showing the generic toast.
  - Or, if Skip will re-quote the relay fee for an in-flight route without re-routing it, refresh on resume. Much better for users if the API allows it.

## 2. The final step's fee reserve has no headroom

Skip's route reserves a flat 0.02 USDC for the Noble leg regardless of transfer size, and the fee Keplr proposes by default exceeds it.

- **Repro**
  - Start an Avalanche USDC deposit into allUSDC and sign the first transaction.
  - When the funds arrive on Noble, accept the fee Keplr offers by default.
  - Signing fails with "Insufficient balance to cover the fees" and the transfer stalls with the funds on Noble.
  - Opening Keplr's Fee Options and picking any tier gives 0.014107, which fits. That's the only reason my $20 test completed.
- **The numbers**
  - At $20: 19.98005 arrived, the msg sends 19.96, leaving 0.02005 for fees. Keplr's default is 0.028216.
  - At $10: same 0.02 reserve, same shortfall.
  - The gap is a constant of roughly 0.008, so small transfers are hit as hard as large ones and a sufficiently small one may not be completable at all.
- **On the mechanism**, I think this is the intermediate account not existing at quote time rather than us supplying a bad estimate:
  - `estimateCosmosStepGasFee` returns `undefined` when simulation fails, and its doc comment anticipates exactly this — the account may not be funded yet because the funds arrive with the prior step.
  - `signFinalStep` then passes `fee: undefined`, and since `preferNoSetFee` is `Boolean(gasFee)` it also passes `preferNoSetFee: false`, handing fee selection to the wallet.
  - That matches what I saw: the fee was freely editable and defaulted to Keplr's own number rather than ours. Worth confirming on your side.
- I've left two inline comments on the guards that would otherwise catch this. Both compare the fee denom against the transferred asset, so both are skipped on Noble.

## Smaller observations

- The review screen labels the combined estimate as the "Avalanche network fee" when it also covers the Noble leg, which makes the fee look like it belongs to one chain.
- Pressing Continue before the funds have arrived waits rather than prompting, which is correct, but it does so silently for several seconds before the "still in transit" toast. A pending state on the button would show the click registered.
- Switching Keplr accounts removes the Continue row entirely, since history is account-scoped. So the resume wrong-account guard is defence in depth rather than a reachable path — not a problem, just noting it can't be exercised from the UI.

## Not tested

The cross-tab double-signature lock. I opened two tabs and pressed Continue in both, but they failed on the expired relay quote before either reached a signature prompt, so the lock was never exercised. I'd rather retest that after the expiry fix than claim any coverage of it now.

## One question

The preflight validates the final step's signer against the connected wallet, but I didn't find anything checking that the first leg's CCTP mint recipient is that same account. It looks consistent by construction today since both come from the same wallet lookup, so this is a question rather than a finding — is there a path where they could diverge? The failure mode if they did is funds landing on an account the user can't sign from.

## What I verified working

- **The stale-funds guard is exactly right.** After I recovered the funds manually, Continue detected the shortfall against `expectedArrival`, called `markPendingStepStale`, and showed the specific "expected funds are no longer in your Noble account" copy instead of failing generically. The entry stops offering Continue and drops out of the prune exemption, so the row clears itself.
- **The first leg isn't tracked as the completed transfer.** My $20 transfer's history row flipped to success only after the Noble step, not at the CCTP burn. That's the coderabbit comment on `transfer-history.tsx` genuinely satisfied in the shipped code.
- **The refactored shared EVM sender is fine on the ordinary path.** With the flag off I ran Arbitrum USDC into Noble USDC, which routed via Squid rather than Skip, and got both the ERC20 approval and the transfer prompt: `0x657f332017f94c252f9f739153f7049e927a80ec87f430e64a7ff5c7fde58346`. That exercises the approval branch on a non-Skip provider.
- **Flag plumbing is correct.** I probed LaunchDarkly's client eval endpoint directly rather than trusting the dashboard: preview serves the flag on, production serves `false`. One aside, not this PR's behaviour on preview — the `defaultFlags` entry only applies when `NODE_ENV === "development"` with no client ID set, so it can't mask the flag on preview, but it does silently force the feature on in local dev.
- **Persistence at broadcast works.** Closing the tab at broadcast preserves a working Continue.
- **The happy path completes.** $20 Avalanche USDC into allUSDC, Noble tx `31C0197A0E11D0DD1ABF898F63316B8CF447FAB47CDD044405CFB0059A0ECAFC`, code 0, allUSDC arrived on Osmosis.

packages/bridge/src/interface.ts
```diff
@@ -613,6 +692,78 @@ const txSnapshotSchema = z.object({
...
+        })
+        .optional(),
+      /**
+       * `expectedArrival.denom` balance the intermediate account held
+       * BEFORE the first transaction (minimal denom units). Makes the
+       * resume balance check replay-proof: the arrived funds must be
+       * present on top of this, so an account that already held enough of
+       * the denom can't pass the check after the step was completed from
+       * another session.
+       */
```
jasbanza: This doc comment describes the `preArrivalBalance` field, which was removed — it now sits directly above `stale`'s own doc block and reads as though it documents it.

Copilot flagged this on 10 Sep and I agree; safe to delete.

jasbanza (changes requested):

packages/web/stores/transfer-history.tsx
```diff
@@ -301,6 +381,36 @@ export class TransferHistoryStore implements TransferStatusReceiver {
...
+                "transfer.multiTxStepFailed",
+                { chain: snapshot.pendingStep.prettyName },
+              ],
+            },
+            ToastType.ERROR,
+            { updateToastId: sendTxHash }
+          );
+          // make the restore durable before another session's storage
+          // check can mistake it for an already-advanced entry
+          await this.persistNow();
```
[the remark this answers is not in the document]
JohnnyWyles: Applied in 543897911. You are right that the status write preceded the resolved-hash check: `receiveNewTxStatus` now returns before touching `snapshot.status`/`reason` when the hash is already resolved, and the failed-final-step restore adds the failed attempt's hash to `_resolvedTxStatusKeys` (a retry advances under a new final hash, so nothing legitimate is suppressed). The three per-case `break` guards that became redundant are removed, and a regression test covers a duplicate `failed` callback after a restore.

packages/web/components/bridge/use-bridge-quotes.ts
```diff
@@ -786,6 +910,248 @@ export const useBridgeQuotes = ({
...
+      // ---- Step 1: the EVM transaction ----
+      // Persist the resumable entry (with the quoted route, so the final
+      // step can be rebuilt after a reload) the moment the wallet returns a
+      // hash: the funds are en route from broadcast, so waiting for the
+      // receipt to record it would lose the resume record if the app
+      // closes during confirmation.
+      let entryRecorded = false;
+      const sendTxHash = await sendEvmBridgeTx(firstStep, (broadcastHash) => {
+        firstStepBroadcastHash = broadcastHash;
+        entryRecorded = trackTransferStatus({
```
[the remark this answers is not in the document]
JohnnyWyles: Applied in 543897911. `entryRecorded` did only prove the in-memory push, and both consumers that matter (the final step's signing guard via `syncPendingStepFromStorage`, and the recovery surface if the session ends) read persisted storage. The flow now awaits `transferHistoryStore.persistNow()` immediately after the first tx broadcasts, before arrival polling, and fails with the same recorded-failure copy if that write throws.

packages/bridge/src/skip/index.ts
```diff
@@ -507,6 +704,428 @@ export class SkipBridgeProvider implements BridgeProvider {
...
+
+    // Fee budget for this step: the route's reserved amount is the floor,
+    // improved by the account's live balance when that reads higher — the
+    // balance includes the arrived funds plus any top-ups the user made, so
+    // topping up the fee token genuinely raises what can be spent on fees.
+    // A short or failed balance read never shrinks the budget below the
+    // route's reserve (the read can lag the arrival).
+    const feeDenom = this.ctx.chainList.find(
+      (c) => c.chain_id === step.chainId
+    )?.feeCurrencies?.[0]?.coinMinimalDenom;
```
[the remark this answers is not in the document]
JohnnyWyles: By design, no change. A chain only accepts fees in its registered fee tokens, so pricing the fee in whatever denom the message happens to move would build a fee the chain rejects; `getDefaultGasPrice` selecting a registered fee currency is the correct source for the fee denom, and the budget is deliberately only applied when that denom matches what the step spends.

The case you describe (step moves a token that is not the chain's fee token) is not silently mispriced: `getStepFeeBudget` returns undefined, the fee stays at the normal safe-side estimate, and the preflight in `use-bridge-quotes.ts` checks the account's balance of that fee denom on the intermediate chain and refuses the transfer before anything irreversible, failing closed when the balance cannot be read. Multi-tx intermediates are also currently restricted to slip44-118 chains, so Noble (where the fee token is the arriving asset) is the only live shape.

packages/web/components/bridge/use-bridge-quotes.ts
JohnnyWyles: Fixed in 8f1873f1d. You were right that the denom munging disabled this on the one chain that needed it. The `replace(/^u/, "")` comparison against the source asset symbol is gone: the preflight now compares the fee denom against the denom of the token the drafted final msg actually moves, both minimal denoms on the same chain, so `uusdc` vs `uusdc` matches exactly and `inj` vs `uusdc` does not.

When they match, the fee is paid from the arriving funds, and the headroom you asked for is enforced where the fee is built rather than here: the step's fee is sized to the route's reserve (arrival minus what the msg sends), repriced at the chain's minimum gas price if the default safe-side price overshoots, and the gas limit capped into the reserve, refusing outright when a real simulation proves it cannot fit. At sign time the budget is raised by the account's live balance, so topping up the fee token now genuinely helps, which it did not before.

When they differ, this balance check still runs, and it now fails closed on an unreadable balance instead of proceeding.

packages/web/components/bridge/use-bridge-quotes.ts
JohnnyWyles: Fixed in 8f1873f1d, same change as the preflight guard above: `finalStepGasWarning` now compares the fee denom against the drafted msg's token denom instead of the u-prefix-stripped source symbol, so it fires exactly when the arriving funds cannot pay the step's gas (an INJ-style intermediate) and stays quiet when they can.

The warning's `{denom}` also renders the display denom now (USDC.n rather than uusdc); it was interpolating the minimal denom straight into user-facing copy.

packages/bridge/src/interface.ts
JohnnyWyles: Deleted in 8f1873f1d. The block was left behind when `preArrivalBalance` was removed, and it did read as `stale`'s documentation. The step schema has also been extracted into a named `multiTxStepSchema` since, so the step's fields are documented in one place and reused by the new `advancedStep` field.

JohnnyWyles:
@jasbanza thank you for running this with real funds, and for recovering those Noble funds manually rather than just reporting a stuck transfer. Both findings are fixed, along with all three inline comments and the smaller observations. Details below, plus an answer to your question and two things I would like you to retest.

## 1. Relay fee quote expiry

Fixed in 8f1873f1d. I probed the API before choosing an approach, and the fix is simpler than either option you suggested.

Skip validates `smart_relay_fee_quote.expiration` only on the payload actually submitted to `/v2/fungible/msgs`. Replaying the stored operations **without** the quote field is accepted, and returns a Noble msg identical to the original (same amount, sender, receiver, memo and min_out; only the timeout timestamps refresh). The relay fee was already collected by the first transaction, so the quote has no job left to do at that point.

So `getTransactionStep` now strips every `smart_relay_fee_quote` from the stored operations before replaying, at any nesting depth and for any operation shape, not just CCTP. There is no expiry check to get stale, and no re-quote needed. As a backstop, if Skip ever rejects a replay as expired anyway, that surfaces as a named error mapped to specific copy ("this transfer's saved route has expired, your funds are safe in your {chain} account") instead of the generic toast.

Because the strip happens at replay time, transfers already persisted with an expired quote are rescued too, which is the state your stuck entry was in.

## 2. Fee reserve headroom

Fixed in 8f1873f1d. Your numbers reconstruct exactly, and the mechanism turned out to be slightly different from the one we both assumed.

The estimate was never `undefined`. Intermediate steps are priced through `getDefaultGasPrice`, which deliberately uses the registry's **high** gas price step. Noble's steps are low/average 0.1, high 0.2. Your observed 0.028216 is 141,080 gas at 0.2, which is our own pinned fee, and the Keplr tier that worked, 0.014107, is the same gas at Noble's actual minimum of 0.1. So the safe-side price doubling pushed us past Skip's flat 0.02 reserve; "freely editable" was just Keplr's Fee Options being available on a dapp-suggested fee.

The step's fee is now sized to a budget rather than priced blind:

- budget = what the route reserves on the intermediate chain (arrival minus what the step's msg sends), derived generically from the operations, not hardcoded for Noble
- if the safe-side fee exceeds it, reprice at the chain's minimum gas price (registry low step, or the live price on fee-market chains)
- if still over, cap the gas limit into the budget, but only with margin over what a real simulation measured; a cap that would guarantee an out-of-gas failure refuses instead, with copy telling the user to top up gas on that chain
- at sign time the budget is raised by the account's **live balance**, so topping up genuinely helps now. That directly fixes the "topping up gas doesn't help" note in your report, which was true before: the msgs replay ran ahead of any fee computation.

Your $20 case: 180,000 gas at 0.1 is 0.018, inside the 0.02 reserve, with the transaction no longer depending on the user finding Keplr's fee editor.

## Inline comments and smaller observations

- Both fee guards: the `replace(/^u/, "")` comparisons are gone, replaced by exact minimal-denom comparison against the token the drafted msg moves. Replies on both threads.
- Orphaned `preArrivalBalance` doc block: deleted.
- Combined fee labelled as one chain's: multi-chain quotes now show a plain "Network fees" instead of "Avalanche network fee".
- Continue silent for several seconds: the arrival poller now fires the "still in transit" toast on the first unsuccessful check rather than after the whole ~30s budget, so the click registers immediately. (The spinner you expected was already there, just too subtle to notice for that long.)
- Account switch removing the Continue row: agreed, noted as defence in depth rather than a reachable path.

## Your question about the CCTP mint recipient

Safe by construction, and it cannot diverge. Both come from the same single source: the route's `required_chain_addresses` produces one Noble entry in `address_list`, and that one entry is simultaneously the mint recipient encoded into the tx1 burn and the sender of the drafted tx2. There is no second lookup that could disagree.

Two checks sit on top of that before anything irreversible: the preflight refuses unless the connected wallet's own Noble account equals that entry, and at sign time `getTransactionStep` refuses to return a step whose built sender is not the wallet address. Multi-tx intermediates are also gated to slip44-118 chains, because on a non-118 chain the bech32-derived address would be one the user does not control, which is the real version of the risk you were pointing at.

## Also fixed since your review

A broader review pass turned up three things worth knowing, all fixed:

- A failed or timed-out **final** step used to leave a terminal "failed" row with copy implying the deposit failed, while the funds actually sat on Noble. It now restores the pending step and offers Continue again, with copy saying where the funds are.
- A **reverted first** EVM transaction left an immortal pending row with a dead Continue (viem returns rather than throws on revert, so the status was never checked). Now it fails the entry, and the expiry exemption has a 30-day long stop.
- Multi-tx quoting is gated to deposits, since the executor only supports an EVM-first-plus-one-cosmos-step shape.

## What I would like you to retest

Both on the updated preview, since the fixes changed exactly the paths you exercised:

1. **The cross-tab double-sign lock**, which you could not reach before because both tabs died on the expired quote.
2. **A fresh real-funds resume pass**: deposit, leave it well past 35 minutes so the quote would previously have expired, then Continue.

Unit coverage was added for the fee sizing (reprice, cap, refusal, live-balance top-up), the quote stripping, and the history advance/restore/merge ordering, but none of that substitutes for a real run on the path you found broken.

jasbanza:
@JohnnyWyles 


Both retests ran on the updated preview with real funds: a $10 Avalanche USDC deposit into allUSDC, resumed 75 minutes after the burn.

**What happened**

- Burn `0x2a6b8406ad0680d59a63c05abacb9acb81251ca287c72a8447e4b7be1ecae6a1` on Avalanche at 13:15:55Z. 9.98 USDC minted to `noble17mlflt…v3fr2` at 13:19:34Z, with the 0.02 relay fee taken at burn time — which matches your account of the quote having no job left at replay.
- Final step `791D2D90B219E63122363F7CC11863AC8A8BC8B069BE624519A2D23E381469B9` on Noble at 14:36:25Z, code 0, 9.96 USDC onward. allUSDC arrived on Osmosis.

**Relay quote expiry — fixed.** 75 minutes against a roughly 30-minute TTL, and the replay rebuilt and broadcast with no expiry error and no generic failure. Stripping the quote rather than re-quoting is a better answer than either option I suggested.

**Fee reserve — fixed.** The step's fee came out at 14,105 uusdc, 141,042 gas at Noble's minimum 0.1, pre-filled, and I never opened Keplr's Fee Options. At the safe-side 0.2 that same limit is 28,209, which is the 0.028216 I hit the first time, so the reprice branch is doing exactly what you describe.

**Cross-tab lock — works.** Two tabs on the same pending transfer: clicking Continue in the second while the first held an open signature prompt produced no second prompt, and once the first signed, the second moved to "deposited" rather than offering Continue again. That is the piece of this PR that had no coverage at all when I reviewed it.

**One thing I would like your read on**

This run also measures how far simulation under-reports gas on that Noble step: simulated 94,028 (the 141,042 limit divided by the 1.5 multiplier), actually used 119,829. That is 27% under.

The cap branch in `estimateCosmosStepGasFee` accepts a capped limit when `cappedGas >= ceil(simulatedGas * 1.1)`. With those numbers the guard accepts any cap from 103,431 upward, while the transaction needed 119,829. A cap in that range passes the check and then runs out of gas, which still charges the fee, consumes the whole reserve, and leaves the funds on the intermediate chain with nothing left to fund a retry — the outcome the rest of this change exists to prevent.

It could not fire here, because a 20,000 reserve permits a 200,000 cap. It becomes reachable when the reserve is small relative to gas: a different route shape, or an intermediate chain with a higher minimum gas price.

Requiring the same 1.5 margin the normal path already uses, rather than 1.1, would close it.

**Minor, and I would leave it as it is:** the review screen quoted 0.02 for the Noble leg where the real fee was 0.0141, because the quote-time estimate prices the fallback limit and then caps it to the reserve. Erring high on a fee estimate is the right direction, so this is fine to keep — I am noting it only so it is not mistaken for a discrepancy later.

I have left the review as it stands for the gas-cap margin. Everything else here I am satisfied with.

JohnnyWyles:
You are right, and the 1.1 was a number I picked without evidence to back it. Your measurement is the argument: 94,028 simulated against 119,829 used is 27% under, which is most of the way to the 1.5 the normal path applies, so a 1.1 guard sits inside the range where the transaction still runs out of gas. Fixed in 2d223e077.

Taking your suggestion had one consequence worth flagging, because it changed the shape of the code rather than just the constant. Once a cap has to clear the full `gasMultiplier`, the "capped" limit **is** the full limit, and the branch above already returns whenever the budget covers that. So for a simulated step the cap branch could only ever refuse.

Rather than leave a condition that can never accept, I made it explicit:

- **Simulated limit: never capped.** If the reserve cannot cover the full multiplied limit at the floor price, it refuses with the top-up copy. No path remains where a simulation-derived limit is trimmed to fit.
- **Fallback limit: still capped.** There is no simulation to under-report there, and the fallback is a deliberate overestimate of a known message shape (250,000 for an IBC transfer), so capping it stays safe and keeps quotes working while the intermediate account is still unfunded.

Tests follow the same split: the case that previously asserted a simulated cap now asserts refusal, and I retargeted the neighbouring test to the boundary your numbers describe, where the reserve covers the simulated gas itself but not the multiplied limit. That is the range that would have slipped through before, and it is the one most likely to appear first on a chain with a higher minimum gas price.

Agreed on the review screen quoting 0.02 against a real 0.0141, and agreed on leaving it. The quote-time estimate prices the fallback limit before any capping, and a fee estimate that errs high is the right direction.

Thank you for the retests, particularly the cross-tab lock. That was the one piece of this PR with no coverage of any kind when you first reviewed it, and it is not something I could have exercised without a second real signature.

jasbanza (approved):
Approving.

I re-read the four commits plus the gas-cap fix. The restore path, the cross-tab guard, the reverted-transaction handling and the new copy all hold up, and the tests land on the new branches rather than around them. Thanks for the quick turnaround on the gas cap, and for explaining why the cap branch had to change shape rather than just its constant.

One note, not blocking.

The provider and the client read "what this step moves" differently. A cosmos message can carry the amount in one of two places: `token` for an IBC transfer, or `funds` for a contract call. `getMsgSpend` in the Skip provider handles both. The client only reads `token` — once in the preflight, once in the gas warning.

So if an intermediate step ever arrives as a contract call, the client sees no token, concludes the fee must come from the account's own balance, finds that account empty (the funds have not arrived yet), and refuses the transfer before the first transaction is signed. The provider, meanwhile, had sized that fee correctly from the route reserve.

Nothing that quotes today reaches this: multi-tx is deposits only, one cosmos step, and Skip's step for a CCTP deposit is an IBC transfer. I am noting it because the provider was written to handle the contract shape and the client was not, so the two would drift apart the first time such a route appeared. Mirroring the same `token` then `funds` read on the client would settle it, whenever suits.

PR #3816, opened 2024-09-02
Add Skip API key

jonator:
[Context](https://osmosis-network.slack.com/archives/C046V54UZC6/p1724714961228419)

Skip has requested that we start using an API key. The `SKIP_API_KEY` has been added to Vercel, and @osmosis-labs/front-end devs need to add it to their `.env.local`. Reach out to @niccoloraspa for the key.

PR #4455, opened 2026-09-01
fix(bridge): send a real slippage tolerance on Skip routes (MTN-252)

jasbanza:

Osmosis → EVM withdrawals were signed with a destination floor set to exactly the quoted output, so any adverse move between quote and execution made the destination swap unfillable and the bridged token was refunded on the far chain as axlUSDC — a denom most wallets do not display, which is why this reaches us as "my funds disappeared". This forwards a real tolerance to Skip, and keeps the Osmosis leg covering what the destination swap is going to spend.

### Linear Task

[MTN-252](https://linear.app/osmosis/issue/MTN-252/skip-bridge-drops-the-users-slippage-and-sends-0percent-tolerance-so)

## The issue

A withdrawal is two swaps with a bridge in the middle. Each swap carries a minimum it will accept, and the bridge takes a fee in transit.

```mermaid
flowchart LR
    A["allUSDT<br/>on Osmosis"] -->|"swap<br/>accepts ≥ min_asset"| B["axlUSDC"]
    B -->|"Axelar bridge<br/>minus fee"| C["axlUSDC<br/>on destination"]
    C -->|"swap spends amount_in<br/>accepts ≥ amountOutMin"| D["USDT delivered"]
    C -.->|"either bound not met"| E["refunded as axlUSDC<br/>invisible in the wallet"]
```

Two defects, both fixed here.

**1. The tolerance was dropped.** `getQuote` destructured the caller's `slippage`, used it only in the cachified key, and never forwarded it; `client.ts` turned that omission into an explicit `"0"`. With zero tolerance Skip sets the destination `amountOutMin` to its full quoted output, so the floor equals the prediction and absorbs nothing. Because `amount_in` is pinned, the destination output is a pure function of pool state — it clears the floor only while that pool has not moved at all, and reverts the moment it has.

**2. Forwarding a tolerance alone is not enough.** Skip scales the Osmosis leg's `min_asset` down by the tolerance but pins the destination swap's `amount_in` to the *untoleranced* quote. An Osmosis output landing between the two clears the Osmosis leg and still leaves the executor holding less than its calldata spends, so the swap cannot run and the token is refunded anyway. Raising `min_asset` back to `amount_in + fee_amount` — exactly what Skip returns at zero tolerance — closes that gap, so the tolerance reaches the destination bound only.

## Brief Changelog

- Default `slippage` to 0.5% in `SkipBridgeProvider.getQuote` and forward it as `slippage_tolerance_percent` on `messages()`. `route()` is left alone — Skip's v2 route request has no slippage field.
- Drop the `?? "0"` fallback in `SkipApiClient.messages`.
- Add `raiseMinAssetToDestinationInput`, which lifts the signed `min_asset` to `evm_swap.amount_in + axelar_transfer.fee_amount`. No-ops on routes without a destination EVM swap, so Cosmos-to-Cosmos and same-token paths are untouched.
- Find that floor in both shapes Skip returns: `msg.swap_and_action` on a `MsgExecuteContract`, and `forward.next.wasm.msg.swap_and_action` inside the packet-forward memo on a `MsgTransfer`, at any depth. The memo is re-serialized after editing, without which the raise is silently dropped.
- Guard that `evm_swap.denom_in` and `axelar_transfer.fee_asset.denom` are the same asset before summing them, and warn when a route that swaps before bridging carries no floor at all, rather than returning a quiet no-op.
- Add the `evm_swap` and `cctp_transfer` variants to `SkipOperation`, which were missing from the union, and type the CCTP smart relay fee quote.
- Correct the `getQuote` comment's claim that 0.5% sits "inside the app's own 2% and 6% gates". The figures are 10% and 6%, and neither gate bounds the signed tolerance — they compare the quote's realized loss and price impact.
- Correct seven spec fixtures passing `slippage: 0.01` under a percent-typed field to `1`.
- Add outgoing-request-body assertions and unit coverage for the raise.

The displayed quote does not move: `expectedOutput` is read from `route.amount_out`, which neither the tolerance nor the raise touches. Only the signed floors change.

## Testing and Verifying

Full `packages/bridge` suite passes; `tsc` and eslint clean. The request-body assertions were confirmed to fail against the unfixed code (`Expected "0.5"`, `Received "0"`).

### Real transfers

Four withdrawals on 2026-09-14, production against this branch's preview, paired on each chain a couple of minutes apart. All four completed. The difference is the margin they completed with.

| run | destination floor | delivered | **margin at fill** | floor vs quoted output |
|---|---|---|---|---|
| Optimism, production | 19,515,603 | 19,515,603 | **0** | 0.000% |
| Optimism, this PR | 19,630,288 | 19,699,240 | **68,952** | 0.350% |
| Avalanche, production | 5,487,233 | 5,487,233 | **0** | 0.000% |
| Avalanche, this PR | 5,466,571 | 5,485,772 | **19,201** | 0.350% |

Both production runs delivered *exactly* their floor, to the unit. That is not luck twice — with `amount_in` pinned, the destination output is deterministic given pool state, so a floor set to the quote is met precisely while nothing moves and missed entirely once something does. Every production withdrawal is that coin flip. This branch puts 0.35% between the floor and the prediction, on both chains, which is the entire point.

Osmosis txs: Optimism [B28FBE78…7181](https://www.mintscan.io/osmosis/transactions/B28FBE78728EB8A6AD975F99C645BF98B05EBBE4B86F8728CEF586BB3FA97181) / [670D8C38…FBD8](https://www.mintscan.io/osmosis/transactions/670D8C383438F16FE9B5DBD4EC0724A33994635573F9338AF743A77871C3FBD8), Avalanche [E8AE857E…76A2](https://www.mintscan.io/osmosis/transactions/E8AE857E333D6A6C87879F1F859FC4B3BE37C2E8AF84E9261A2710A53DAD76A2) / [613E727E…73B6](https://www.mintscan.io/osmosis/transactions/613E727E5CF68C8ADE5C0126D1222F9CCA6CCC62C1C39AB3BBE605EDE24573B6). All `STATE_COMPLETED_SUCCESS`, native USDT delivered, no axlUSDC left on either destination. On every run `min_asset` equalled `amount_in + fee` exactly — on production because zero tolerance makes them identical, on this branch because the raise puts them back.

### Regression runs on the rewritten helper

Two further withdrawals on 2026-09-15, after the raise was rewritten to walk both message shapes. Both completed, both verified by decoding the signed message rather than reading the UI.

| run | shape | signed floor | result |
|---|---|---|---|
| 20 USDT → Avalanche | destination swap | raised to 19,964,994, exactly `amountIn + fee` | delivered 19.915905 USDT, **0.351% margin** |
| 20 USDT → Ethereum | no destination swap | left at 19,900,000, tolerance only | delivered 18.775136 USDT |

[D5647081…86DB3](https://www.mintscan.io/osmosis/transactions/D5647081903E0227FB9719C6AD859610AEC84D083F8E5EBF68B7BCE3DB286DB3) and [D259CF59…73C5](https://www.mintscan.io/osmosis/transactions/D259CF597ACB4EB2CF7D7B1375A414B87C81499546317BC45B0AC111176273C5). Both released the native destination asset. On the second the GMP payload was decoded directly to confirm it carries no `ead2fe5b` swap selector, so the untouched path is right rather than merely assumed.

### The failures that motivated each half

Both on Osmosis → Optimism, earlier the same day.

**Production reproduced the reported bug.** Destination floor 59,406,333 against 59,400,844 arriving; the swap ran, returned less than the floor, reverted, and the executor refunded 59.400844 axlUSDC to the EOA where it was invisible. [97C87578…B48E](https://www.mintscan.io/osmosis/transactions/97C8757852D26B92AA98289A8555D63D023259E3A219547867A165695702B48E)

**A tolerance-only build still failed, which is how defect 2 surfaced.** Its destination floor had 0.34% to spare and it reverted anyway, because the Osmosis leg came in **13 units** under the pinned `amount_in`. [6512C204…A588](https://www.mintscan.io/osmosis/transactions/6512C2045515E837F347960403322E1053D7551A2EED017A4EC3CB558489A588)

### What is not covered

- **None of the four successful runs encountered pool drift** — every Osmosis leg delivered its quote to the unit. They prove the floors are set correctly and that a withdrawal completes; the drift case is prevented by construction rather than demonstrated live.
- **A short Osmosis leg still fails, on Osmosis** — gas lost, nothing bridged, nothing stranded. That is the better of the two failures, and it is unavoidable on our side while Skip's executor spends a quote-time constant rather than the balance it actually received. **Recommend escalating to Skip**, with two asks: have the destination executor spend its received balance rather than the pinned `amount_in`, or failing that apply the tolerance to `amount_in` as well as `amountOutMin`. Nothing has been filed with them yet. Detail in MTN-252.
- **`amountOutMin > amount_in` is not impossible by construction**, contrary to the original framing in the ticket. The destination swap can yield above parity — in one run it turned 19,695,404 axlUSDC into 19,699,240 USDT. The defect is that the floor carries no tolerance, not that it is arithmetically unreachable.
- **The packet-forward memo path is not reachable through our UI today, so the raise on that shape is insurance rather than a live repair.** Every pair the UI offers returns either a `MsgExecuteContract` with a top-level floor or a CCTP `MsgTransfer` with no floor and no destination swap: `allUSDT` to all five destinations, `USDC` to all eleven. The memo shape appeared only for OSMO, TIA, INJ and ATOM as source assets, none of which are withdrawable — OSMO has an empty `counterparty` list. Forcing it with size does not work either; at $1,000 and $10,000 the route still goes through Osmosis. That branch is therefore covered by unit tests plus a real captured Skip response replayed through the helper (floor 7,802,201 raised to the required 7,813,934), not by an on-chain run. It is kept because Skip picks routes dynamically and the asset list changes, and the failure it prevents is the invisible refund this PR exists to remove.
- Choice of 0.5% over Squid's 1%, the ~0.30/0.70 apportionment, and the split-route question: see MTN-252.

### DIFF
--- a/packages/bridge/src/skip/index.ts
+++ b/packages/bridge/src/skip/index.ts
@@ -63,7 +63,17 @@ export class SkipBridgeProvider implements BridgeProvider {
       toChain,
       fromAddress,
       toAddress,
-      slippage,
+      /**
+       * Percent, per `getBridgeQuoteSchema`. No caller supplies this today, so
+       * the default is what every real request uses. Skip splits the tolerance
+       * across the route's swap legs rather than applying it to each, so
+       * end-to-end exposure equals this number. 0.5% clears the 0.02%-0.14%
+       * quote-time margins measured on Osmosis to EVM stable routes and sits
+       * inside the app's own 2% and 6% gates. Deliberately tighter than Squid's
+       * 1%: Skip has no recommended value to defer to, and its tolerance
+       * reaches a swap on Osmosis rather than only the vendor's own route.
+       */
+      slippage = 0.5,
     } = params;
 
     return cachified({
@@ -219,6 +229,7 @@ export class SkipBridgeProvider implements BridgeProvider {
           amount_in: route.amount_in,
           amount_out: route.amount_out,
           operations: route.operations,
+          slippage_tolerance_percent: slippage.toString(),
         });
 
         const transactionRequest = await this.createTransaction(
--- a/packages/bridge/src/skip/__tests__/skip-bridge-provider.spec.ts
+++ b/packages/bridge/src/skip/__tests__/skip-bridge-provider.spec.ts
@@ -15,7 +15,7 @@ import {
   GetBridgeQuoteParams,
 } from "../../interface";
 import { SkipBridgeProvider } from "..";
-import { SkipMsg } from "../types";
+import { SkipMsg, SkipMsgsRequest } from "../types";
 import {
   ETH_EthereumToOsmosis_Msgs,
   ETH_EthereumToOsmosis_Route,
@@ -129,7 +129,7 @@ describe("SkipBridgeProvider", () => {
       toChain: { chainId: 1, chainName: "Ethereum", chainType: "evm" },
       fromAddress: "osmo107vyuer6wzfe7nrrsujppa0pvx35fvplp4t7tx",
       toAddress: "0x7863Ec05b123885c7609B05c35Df777F3F180258",
-      slippage: 0.01,
+      slippage: 1,
     });
 
     expect(quote).toBeDefined();
@@ -241,7 +241,7 @@ describe("SkipBridgeProvider", () => {
       fromChain: { chainId: 1, chainName: "Ethereum", chainType: "evm" },
       toAddress: "osmo107vyuer6wzfe7nrrsujppa0pvx35fvplp4t7tx",
       fromAddress: "0x7863Ec05b123885c7609B05c35Df777F3F180258",
-      slippage: 0.01,
+      slippage: 1,
     });
 
     expect(quote).toBeDefined();
@@ -348,7 +348,7 @@ describe("SkipBridgeProvider", () => {
     fromChain: { chainId: 1, chainName: "Ethereum", chainType: "evm" },
     toAddress: "osmo107vyuer6wzfe7nrrsujppa0pvx35fvplp4t7tx",
     fromAddress: "0x7863Ec05b123885c7609B05c35Df777F3F180258",
-    slippage: 0.01,
+    slippage: 1,
   };
 
   const useEthereumToOsmosisRouteWithFeeBehavior = (fee_behavior: string) =>
@@ -457,7 +457,7 @@ describe("SkipBridgeProvider", () => {
       toChain: { chainId: 1, chainName: "Ethereum", chainType: "evm" },
       fromAddress: "osmo107vyuer6wzfe7nrrsujppa0pvx35fvplp4t7tx",
       toAddress: "0x7863Ec05b123885c7609B05c35Df777F3F180258",
-      slippage: 0.01,
+      slippage: 1,
     });
 
     expect(quote.transferFee.isAdditive).toBe(false);
@@ -484,7 +484,7 @@ describe("SkipBridgeProvider", () => {
       toChain: { chainId: 1, chainName: "Ethereum", chainType: "evm" },
       fromAddress: "0xabc",
       toAddress: "0xdef",
-      slippage: 0.01,
+      slippage: 1,
     };
 
     await expect(provider.getQuote(params)).rejects.toThrow(
@@ -535,7 +535,7 @@ describe("SkipBridgeProvider", () => {
       toChain: { chainId: 1, chainName: "Ethereum", chainType: "evm" },
       fromAddress: "0xabc",
       toAddress: "0xdef",
-      slippage: 0.01,
+      slippage: 1,
     };
 
     const txData: BridgeTransactionRequest = {
@@ -587,7 +587,7 @@ describe("SkipBridgeProvider", () => {
       toChain: { chainId: 1, chainName: "Ethereum", chainType: "evm" },
       fromAddress: "osmo1ABC123",
       toAddress: "0xdef",
-      slippage: 0.01,
+      slippage: 1,
     };
 
     const txData: BridgeTransactionRequest = {
@@ -945,6 +945,80 @@ describe("SkipBridgeProvider", () => {
         denom: "USDC",
         transferTypes: ["quote"],
       });
+    });
+  });
+
+  describe("slippage tolerance", () => {
+    const osmosisToEthereumQuoteParams: GetBridgeQuoteParams = {
+      fromAmount: "10000000000000000000",
+      fromAsset: {
+        denom: "ETH",
+        address:
+          "ibc/EA1D43981D5C9A1C4AAEA9C23BB1D4FA126BA9BC7020A25E0AE4AA841EA25DC5",
+        decimals: 18,
+      },
+      fromChain: {
+        chainId: "osmosis-1",
+        chainName: "osmosis",
+        chainType: "cosmos",
+      },
+      toAsset: {
+        denom: "WETH",
+        address: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
+        decimals: 18,
+      },
+      toChain: { chainId: 1, chainName: "Ethereum", chainType: "evm" },
+      fromAddress: "osmo107vyuer6wzfe7nrrsujppa0pvx35fvplp4t7tx",
+      toAddress: "0x7863Ec05b123885c7609B05c35Df777F3F180258",
+    };
+
+    /**
+     * The tolerance only exists in the outgoing request — Skip echoes none of
+     * it back, and the returned quote reports `route.amount_out` either way, so
+     * a test that inspects the quote cannot see this bug at all.
+     */
+    const captureMsgsRequest = () => {
+      const captured: { body?: SkipMsgsRequest } = {};
+
+      server.use(
+        rest.post(
+          "https://api.skip.money/v2/fungible/route",
+          (_req, res, ctx) => res(ctx.json(ETH_OsmosisToEthereum_Route))
+        ),
+        rest.post(
+          "https://api.skip.money/v2/fungible/msgs",
+          async (req, res, ctx) => {
+            captured.body = await req.json();
+            return res(ctx.json(ETH_OsmosisToEthereum_Msgs));
+          }
+        )
+      );
+
+      (estimateGasFee as jest.Mock).mockResolvedValue({
+        gas: "420000",
+        amount: [{ denom: "uosmo", amount: "1232" }],
+      });
+
+      return captured;
+    };
+
+    it("sends the default tolerance when no caller supplies one", async () => {
+      const captured = captureMsgsRequest();
+
+      await provider.getQuote(osmosisToEthereumQuoteParams);
+
+      expect(captured.body?.slippage_tolerance_percent).toBe("0.5");
+    });
+
+    it("prefers an explicit caller tolerance over the default", async () => {
+      const captured = captureMsgsRequest();
+
+      await provider.getQuote({
+        ...osmosisToEthereumQuoteParams,
+        slippage: 1.5,
+      });
+
+      expect(captured.body?.slippage_tolerance_percent).toBe("1.5");
     });
   });
 });
--- a/packages/bridge/src/skip/client.ts
+++ b/packages/bridge/src/skip/client.ts
@@ -78,10 +78,7 @@ export class SkipApiClient {
       `${this.baseUrl}/v2/fungible/msgs`,
       {
         method: "POST",
-        body: JSON.stringify({
-          ...options,
-          slippage_tolerance_percent: options.slippage_tolerance_percent ?? "0",
-        }),
+        body: JSON.stringify(options),
       }
     );
   }


JohnnyWyles:
Reviewed the Skip slippage fix. The tolerance forwarding looks right and I'd land it as-is: the `?? "0"` traces back to #3247 via a rename in #3816, so it was never deliberate, `index.ts:213` is the only caller of `messages()`, and the `0.01` → `1` fixture correction matches the percent-typed schema.

The `min_asset` raise is the part I'd hold. I pulled a live quote for your exact pair and it comes back as an IBC transfer with the floor in the memo, where the helper doesn't look, so the raise no-ops on that route. Details and a suggested fix inline.

Related: the two `getQuote` tests use a route with no `evm_swap` and a `MsgTransfer` with no `swap_and_action`, so the raise isn't covered end to end, and the three unit tests assert against a message shape Skip may not return for this route. I have a captured request/response pair that would work as a fixture.

Minor: `cctp_transfer` is a real operation kind missing from `SkipOperation`, same gap as the `evm_swap` omission. Not blocking, the CCTP routes I sampled have no `evm_swap`.

packages/bridge/src/skip/index.ts
```diff
@@ -41,7 +41,68 @@ import {
...
+  if (!evmSwap || !axelarTransfer) return msgs;
+
+  const required =
+    BigInt(evmSwap.amount_in) + BigInt(axelarTransfer.fee_amount);
+
+  return msgs.map((message) => {
+    if (!("multi_chain_msg" in message)) return message;
+
+    const parsed = JSON.parse(message.multi_chain_msg.msg);
+    const minAsset = parsed?.msg?.swap_and_action?.min_asset?.native;
```
JohnnyWyles: On a live `allUSDT → Optimism USDT` quote (same pair as your verified runs), Skip returns the Osmosis leg as `/ibc.applications.transfer.v1.MsgTransfer`, not a `MsgExecuteContract`. The floor is inside the memo:

```
memo → forward.next.wasm.msg.swap_and_action.min_asset.native.amount = 19947367
```

`parsed?.msg?.swap_and_action` is `undefined` there, so the message returns unchanged. `amount_in 18908318 + fee 1069046 = 19977364`, so the raise would have added 29,997 units but didn't. Did your four runs take the contract shape? If Skip picks per-route, this path is unprotected.

A recursive search handles both shapes, and the memo has to be re-stringified or the edit is dropped on serialization:

```ts
function raiseIn(node: unknown, required: bigint): boolean {
  if (!node || typeof node !== "object") return false;
  if (Array.isArray(node)) return node.map((n) => raiseIn(n, required)).some(Boolean);
  const obj = node as Record<string, any>;
  let touched = false;
  const native = obj.swap_and_action?.min_asset?.native;
  if (typeof native?.amount === "string" && BigInt(native.amount) < required) {
    native.amount = required.toString();
    touched = true;
  }
  for (const [k, v] of Object.entries(obj)) {
    if (typeof v === "string" && (k === "memo" || k === "msg")) {
      let inner;
      try { inner = JSON.parse(v); } catch { continue; }
      if (raiseIn(inner, required)) { obj[k] = JSON.stringify(inner); touched = true; }
    } else if (raiseIn(v, required)) touched = true;
  }
  return touched;
}
```

Verified against the captured response: floor goes to 19977364. Happy to send the JSON as a fixture.

Worth splitting the two no-ops too: "already covered" is fine, but "found an `evm_swap`, found no `min_asset`" is the bug this PR fixes and currently passes silently.

packages/bridge/src/skip/index.ts
```diff
@@ -41,7 +41,68 @@ import {
...
+    (
+      operation
+    ): operation is Extract<SkipOperation, { axelar_transfer: unknown }> =>
+      "axelar_transfer" in operation
+  )?.axelar_transfer;
+
+  if (!evmSwap || !axelarTransfer) return msgs;
+
+  const required =
+    BigInt(evmSwap.amount_in) + BigInt(axelarTransfer.fee_amount);
```
JohnnyWyles, line 169: `denom_in` and `fee_asset` aren't checked before adding. They match on the routes I sampled, so this is defensive rather than a live defect, but `ETH_EthereumToOsmosis_Route` in `mocks.ts` has an 18-decimal `ethereum-native` fee against a `weth-wei` transfer, so the mismatched shape does exist in Skip's responses (deposit-side only, so unreachable here today). Returning `msgs` untouched when `fee_asset.denom !== evm_swap.denom_in` is cheap insurance against writing an 18-decimal fee into a 6-decimal floor.

packages/bridge/src/skip/index.ts
```diff
@@ -63,7 +124,17 @@ export class SkipBridgeProvider implements BridgeProvider {
...
       fromAddress,
       toAddress,
-      slippage,
+      /**
+       * Percent, per `getBridgeQuoteSchema`. No caller supplies this today, so
+       * the default is what every real request uses. Skip splits the tolerance
+       * across the route's swap legs rather than applying it to each, so
+       * end-to-end exposure equals this number. 0.5% clears the 0.02%-0.14%
+       * quote-time margins measured on Osmosis to EVM stable routes and sits
+       * inside the app's own 2% and 6% gates. Deliberately tighter than Squid's
```
JohnnyWyles: There's no 2% gate. `trade-warnings.ts` has `HighPriceImpactGate = 0.1` and `HighSlippageGate = 0.06`, and neither applies to the Skip tolerance. Reads as 10% and 6%.

packages/bridge/src/skip/index.ts
jasbanza: You're right, and I reproduced it. A live `messages()` call returns a `MsgTransfer` whenever the Cosmos-side swap happens somewhere other than Osmosis, with the floor at `memo.forward.next.wasm.msg.swap_and_action.min_asset` exactly as you described. On a captured Arbitrum route the signed floor was 7,802,201 against a required 7,813,934 — an 11,733 unit shortfall the old helper passed through untouched.

Fixed in 49f8adcb0. The helper now walks both shapes to arbitrary `forward.next` depth and re-serializes the memo, which as you flagged is the part that silently drops the edit if missed. I verified it by running the real captured route response through the real helper rather than a synthetic fixture, so your offer of a request/response pair is covered — thank you for it though, it was the right instinct.

One correction to my own PR description, which I've now amended. I originally presented this as closing a live hole. I don't think it is. Checking every pair our UI actually offers, the memo shape never appears: `allUSDT` to all five destinations and `USDC` to all eleven come back as either `MsgExecuteContract` with a top-level floor, or a CCTP `MsgTransfer` with no floor and no `evm_swap`. The memo shape only showed up for OSMO, TIA, INJ and ATOM as source assets, none of which are withdrawable in our UI — OSMO has an empty `counterparty` list entirely. I also tried forcing it with size, on the theory that a large trade might exceed Osmosis liquidity and route via Neutron; at $1,000 and $10,000 it still routes through Osmosis.

So the raise is defensive rather than a fix for something users are hitting today. I still think it's worth carrying, because Skip picks routes dynamically and the asset list changes, and the failure mode if it ever does route that way is the invisible refund this whole ticket exists to remove. But it's your call whether that justifies the extra complexity, and I'd rather you weigh it knowing it's insurance rather than a live repair.

packages/bridge/src/skip/index.ts
jasbanza: Added in 49f8adcb0. The two amounts are summed, so they now have to be the same asset before the sum is used, and the mismatch is logged rather than passing silently.

Your `mocks.ts` example checks out: `ETH_EthereumToOsmosis_Route` pairs `denom_in: 0xC02aaA39...` (WETH) with `fee_asset.denom: "ethereum-native"`, so the mismatched shape is real and not hypothetical. On the live routes I sampled they always matched, including the multi-hop Neutron one where the floor is denominated in the Neutron IBC representation while the fee is quoted on the EVM side — Skip's own zero-tolerance identity holds there too, so the sum stays meaningful across that boundary.

packages/bridge/src/skip/index.ts
jasbanza: You're right, and it was wrong in two ways rather than one. Corrected in 49f8adcb0.

The figures are `HighPriceImpactGate` 10% and `HighSlippageGate` 6%, as you say. The 2% I quoted is `HighSlippageToleranceGate`, which does exist but only on the unmerged MTN-150 branch, and there it gates the slippage a user types on the trade surfaces — it has never applied to a bridge transfer.

The more substantive error is the word "inside". Neither gate bounds the tolerance we sign: `HighSlippageGate` compares the quote's realized input-to-output fiat loss, and `HighPriceImpactGate` the quote's price impact. Neither reads `slippage_tolerance_percent`. They catch a bad route, not a generous tolerance.

QA happened to demonstrate this. A 20 USDT withdrawal to Ethereum signed with a 0.5% tolerance tripped the high-slippage gate, because the 1.224864 USDT Axelar relayer fee is 6.12% of the transfer. Same transaction, two quantities differing by more than tenfold — which is exactly what the old comment obscured by implying one bounded the other.

I've also corrected the same claim in the 1 September comment on MTN-252.

jasbanza:
Thanks for the review — all four points were right, and two of them changed the code rather than just the comments.

**What changed**

- `49f8adcb0` — the memo walk, the `denom_in`/`fee_asset` guard, the split no-op paths, `cctp_transfer` on the operation union, and the gate-figure correction.
- `3076f01a2` — two CodeRabbit follow-ups on that commit: `findSwapAndAction` no longer stops at a `swap_and_action` that carries no floor, and the CCTP smart relay fee quote is now typed.

**On splitting the no-op paths**

I keyed the distinction on whether the route has a `swap` operation rather than on whether an `evm_swap` was found, because "has `evm_swap`, has no floor" turns out to be legitimate and common: a route with no Cosmos-side swap comes back as a plain `MsgTransfer` with no `min_asset` anywhere, and there is genuinely nothing to raise. What should never happen is a route that swaps before bridging and still carries no floor, so that is the case that now warns.

**QA**

Two withdrawals on the preview, both completed, both verified against the signed message rather than the UI:

| Route | Shape | Floor | Result |
|---|---|---|---|
| 20 USDT → Avalanche | destination swap | raised to 19,964,994, exactly `amountIn + fee` | delivered 19.915905 USDT, 0.351% margin |
| 20 USDT → Ethereum | no destination swap | left at 19,900,000, tolerance only | delivered 18.775136 USDT |

Hashes [D5647081…86DB3](https://www.mintscan.io/osmosis/transactions/D5647081903E0227FB9719C6AD859610AEC84D083F8E5EBF68B7BCE3DB286DB3) and [D259CF59…73C5](https://www.mintscan.io/osmosis/transactions/D259CF597ACB4EB2CF7D7B1375A414B87C81499546317BC45B0AC111176273C5). Both released the native destination asset, so neither left a wrapped token stranded. On the second I checked the GMP payload directly rather than trusting the route response — it carries no `ead2fe5b` swap selector, so the early return is correct there.

**What QA could not cover**

The memo path itself, for the reachability reason in the thread above: no pair our UI offers produces that shape, so it rests on unit tests plus a real captured Skip response run through the real helper. If you disagree with that reachability finding I'd genuinely like to know, since it's the one thing standing between "this is insurance" and "this is a live fix" — and it's the question I'd most like a second pair of eyes on.

packages/bridge/src/skip/index.ts
```diff
@@ -41,7 +41,142 @@ import {
...
+ * of `forward.next` hops. Both shapes occur on live Osmosis withdrawals, so
+ * reading only the first left the second signing an unraised floor.
+ */
+function findSwapAndAction(node: unknown): SkipSwapAndAction | undefined {
+  if (!node || typeof node !== "object") return undefined;
+
+  const record = node as Record<string, unknown>;
+
+  if (record.swap_and_action)
+    return record.swap_and_action as SkipSwapAndAction;
```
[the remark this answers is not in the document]
jasbanza: Added in `b3a3703`. A new route fixture carries `swap` + `axelar_transfer` + `evm_swap` operations alongside the tolerance-reduced floor Skip signs for that shape, and the test asserts the floor in the built `transactionRequest` has been raised to `evm_swap.amount_in + axelar_transfer.fee_amount`.

Verified it is a real guard rather than a passing assertion: replacing the `raiseMinAssetToDestinationInput(msgs, route.operations)` call in `getQuote` with a bare `msgs` turns exactly one test red, and it is this one. 45 passing, up from 44.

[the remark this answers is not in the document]
jasbanza: Correction: the reply above belongs to CodeRabbit's separate "exercise the minimum-floor adjustment through `getQuote`" finding, which arrived as an outside-diff-range note with no inline thread to reply on. This thread — continuing the search past a floorless `swap_and_action` — was already addressed in `3076f01`. Full response posted as a PR comment.

jasbanza:
Both outstanding CodeRabbit findings from the 15 September review, resolved in `b3a3703`.

## Exercising the floor raise through `getQuote` — done

The finding was correct, and it was the more useful of the two. `raiseMinAssetToDestinationInput` had six unit tests, but every provider-level fixture routed without a destination swap: `ETH_OsmosisToEthereum_Route` carries `axelar_transfer` alone and its `MsgTransfer` has no `swap_and_action`, so the helper returned at its first guard and the call in `getQuote` was untested.

`b3a3703` adds `ETH_OsmosisToEthereum_DestinationSwap_Route` — `swap` + `axelar_transfer` + `evm_swap` — and the matching `MsgExecuteContract` carrying the floor Skip signs for that shape at a 0.5% tolerance. The new test asserts the floor in the built `transactionRequest` has been raised to `evm_swap.amount_in + axelar_transfer.fee_amount`.

I checked it is a real guard rather than an assertion that happens to pass: replacing the `raiseMinAssetToDestinationInput(msgs, route.operations)` call with a bare `msgs` turns exactly one test red, and it is this one. 45 passing, up from 44.

## Smart Relay fee quote on `SkipCctpTransfer` — landed in `3076f01`, with one part declined

The field is there, and I verified the shape against [Skip's documentation](https://docs.skip.build/go/general/smart-relay) rather than taking the suggestion on trust — `fee_amount`, `fee_denom`, `fee_payment_address`, `relayer_address` and `expiration` all match.

Worth recording that the Major severity overstates it, though: there is no live defect here and there never was one. `getQuote` forwards `operations: route.operations` into `/msgs` verbatim, so every field Skip sends survives whether or not our type knows about it. The field is insurance against a future refactor that rebuilds operations field by field, which is what the doc comment on it says.

What I did not take is the discriminated union making the quote required when `smart_relay` is `true`. This type describes a response we parse, not a request we build, so a union forces every read site to narrow on `smart_relay` before touching any field — a real cost for something nothing currently consumes. It would also promise more than we check: Skip can return `smart_relay: true` with an expired or absent quote and we do not validate the response, so the union would be accurate in the type and wrong at runtime. A plain optional field says exactly what we know.

jasbanza:
@tac0turtle I need to rebase and resolve conflicts (and sanity check) on this before it can be included in stage release - I'd prefer to include it in a smaller prod release next week.

PR #4466, opened 2026-09-14
chore(keplr-hooks): drop unused fee, gas, and recipient configs

tac0turtle:

Continue unused-code cleanup after #4463 / #4464. Trim `@osmosis-labs/keplr-hooks` to the amount/stake configs the app actually uses.

Independent of #4462 (pools package decoupling). Does not touch `@osmosis-labs/pools`, `/api/pools*`, or MobX pool queries.

### Linear Task

n/a

## Brief Changelog

- Delete unused `FeeConfig`, `GasConfig`, `RecipientConfig`, `GasSimulator`, and `send-types`
- Remove unused React `use*Config` wrappers (`useFeeConfig`, `useGasConfig`, `useRecipientConfig`, `useGasSimulator`, `useAmountConfig`, `useStakedAmountConfig`)
- Keep `AmountConfig`, `StakedAmountConfig`, `IFeeConfig`, `FeeType`, `DefaultGasPriceStep`, and amount errors
- Type `IFeeConfig.toStdFee()` with `@cosmjs/amino` instead of `@cosmjs/launchpad`
- Drop unused keplr-hooks deps: React 16, react-router, launchpad, ethers, axios, and several `@keplr-wallet/*` packages

## Testing and Verifying

This change has been tested locally by typechecking `@osmosis-labs/keplr-hooks` against a built `@osmosis-labs/keplr-stores`, running prettier on the package, and confirming remaining app imports still resolve (`AmountConfig`, `StakedAmountConfig`, `IFeeConfig`, amount errors). Full Next.js rebuild was not run.


JohnnyWyles:
`toStdFee()` is dead after this PR. Its only caller was the `autorun` in `tx/gas-simulator.ts:328`, which this PR deletes. What remains is the declaration on `IFeeConfig` and the `FakeFeeConfig` implementation in `@osmosis-labs/stores` — nothing calls either. Given the PR trims this package to what the app actually uses, an uncalled method on a kept interface seems worth removing too.

Side effect rather than the reason: `StdFee` is the only thing `keplr-hooks` imports from `@cosmjs/amino`, so dropping the method also drops the dependency this PR adds. Flagging that it's specific to this package and doesn't cut against #4468, which is deliberately moving keplr-stores/stores onto amino.

It does mean editing `packages/stores/src/ui-config/fake-fee-config.ts`, so it widens this to a second package — happy for it to be a follow-up if you'd rather keep the PR to one.

tac0turtle:
Good catch — `toStdFee()` had no remaining callers after `GasSimulator` went away.

Pushed that onto this PR rather than a follow-up:
- dropped `IFeeConfig.toStdFee` and the `FakeFeeConfig` implementation
- dropped `@cosmjs/amino` from keplr-hooks (it was only there for `StdFee`)

Does not change #4468, which still moves keplr-stores/stores onto amino for the real `StdFee`/`StdSignDoc`/`AminoMsg` usages.

JohnnyWyles (approved):

PR #4478, opened 2026-09-14
refactor(web): unify Button on ui/button without restyling

tac0turtle:

There were two Button components. They are **not** visually equivalent, so this does not map \`mode=\"primary\"\` onto \`variant=\"default\"\`.

Old \`~/components/buttons\`: \`mode\` + \`size\`, \`w-full\`, 2px border hover, solid disabled fills.
New \`~/components/ui/button\`: \`variant\` + \`size\`, not full-width, opacity disabled, ring focus.

This PR puts both class sets on \`ui/button\`, discriminated by API:

- \`mode\` set → original classes (including default size \`normal\` / h-56)
- \`variant\` or neither → existing shadcn classes (unchanged)

Then deletes the duplicate file and points leftover call sites at \`ui/button\`.

### Linear Task

n/a

## Brief Changelog

- Move the pre-shadcn Button CVA onto \`components/ui/button.tsx\` as \`mode\`
- Migrate review-order, toast, order-modal, base-pool, 1-click session, IconButton
- Delete \`components/buttons/button.tsx\`
- Do **not** replace old IconButton with ui IconButton (different defaults: rounded-xl osmoverse-700 vs rounded-full h-12)

## Testing and Verifying

This change has been tested locally by comparing old vs new CVA class lists (copied verbatim) and running web lint-staged. Confirm visually: confirm order, alloyed convert toast, pool details toggle, 1-click session period, and navbar icon buttons.


JohnnyWyles (approved):
Tested visually and by clicking: alloyed assets toast (dismiss + convert), order history modal, 1-Click Trading session period screen, pool detail show-more toggle, swap review modal, and navbar icon buttons. No visual change on any of them.

Verified the class lists are copied verbatim, so `mode` renders exactly as the old component did. The only behavioural change is `mode="primary"` now being explicit on the alloyed convert button, which is correct: the old Button defaulted to `mode: "primary"`, and the unified component only applies legacy classes when `mode` is set.

The 12 dropped modes and 4 dropped sizes are all unused (checked each against stage). `size="sm"`/`"xs"` hits belong to the shadcn scale, not the legacy one, and `icon-social` was never what rendered the social buttons - those are `IconLink` anchors in `components/cards/icon-link.tsx`.

`preview-trade-tests` is the known harness gap, not this PR: the Keplr popup never appeared for the cancel step so nothing was signed, and the helpers in `packages/e2e/pages/transactions-page.ts` wait on the WebSocket toast alone without the REST-poll fallback that `TradePage.startTxConfirmation()` uses.

PR #4479, opened 2026-09-15
fix(e2e): confirm limit-order txs via REST poll and surface chain failures

JohnnyWyles:

Preview e2e has been failing intermittently on transactions that actually succeeded, and failing opaquely on transactions that genuinely didn't. Four PRs hit this over two days (#4463, #4464, #4473, #4478), each costing a re-run and a manual log read to classify.

Three distinct gaps, all in how the e2e layer decides a transaction is done:

1. **The limit-order helpers never got the REST fallback.** `cancelLimitOrder`, `claimAndCloseAny` and `claimAll` in `transactions-page.ts` waited on the in-app "Transaction Successful" toast alone, with a 40s bound. That toast is driven by the app's WebSocket `TxTracer`; when the socket stalls, the assertion fails even though the tx was included on-chain. `TradePage.startTxConfirmation()` already solves exactly this by racing the toast against an LCD poll, and `packages/e2e/README.md` documents that as the guard — the transactions-page helpers were simply never given it.

2. **Deterministic failures were reported as generic timeouts.** `Promise.any` only rejects once *every* branch rejects, so a tx included with a non-zero code left the toast branch running its 120s budget against Playwright's 90s test timeout. The chain error was guaranteed to be swallowed.

3. **CheckTx rejections were worse.** `captureBroadcastHash()` logged a non-zero broadcast code and returned the hash anyway. That tx never enters a block, so the LCD poll can't report a "failed on-chain" error either — both waits ran out and CI showed only `Test timeout of 90000ms exceeded`. This is the #4473 preview-swap failure: `code=32` "account sequence mismatch" from concurrent suites sharing a wallet, where the ante handler had already explained the problem and the harness discarded it.

### Linear Task

n/a

## Brief Changelog

- Give `cancelLimitOrder` / `claimAndCloseAny` / `claimAll` the same racing confirmation `TradePage` uses (capture broadcast hash → poll LCD from Node, which does not go through the browser proxy)
- Arm confirmation *before* the action click, which also absorbs the separate stale-toast dismissal
- `TransactionsPage` retains the hash it captured and exposes `getLastTxUrl()`; both cancel tests prefer it, since `TradePage.getTransactionUrl()` falls back to a hash that is never set for actions driven from this page
- Reject immediately on a tx included with a non-zero code (both page objects)
- Throw at broadcast on a CheckTx rejection, carrying code / codespace / raw_log (both page objects)

## Testing and Verifying

- `npx playwright test --list` enumerates all 58 tests across 11 files (this typechecks the suite)
- Prettier clean; `git diff --check` clean
- Simulated the confirmation race directly against all three paths: a CheckTx rejection and an on-chain failure each reject immediately with the chain's own message, while a not-yet-indexed tx still waits — so transient LCD lag is unaffected and the toast can still win

Wallet e2e was **not** run locally (it spends real funds and is gated in CI). The preview suites on this PR are the real check.

### Known limitation

This does not fix every red trade test. It covers the missing-toast case (#4463, #4464) and both timeout-hiding cases (#4473). It does **not** cover #4478, where the Keplr popup never appeared so nothing was signed — there is no transaction to confirm by either signal.


packages/e2e/pages/transactions-page.ts
```diff
@@ -170,6 +171,158 @@ export class TransactionsPage extends BasePage {
     }
   }
 
+  /**
+   * Capture the tx hash from the app's `/api/broadcast-transaction` response.
+   * Mirrors `TradePage.captureBroadcastHash()`; arm before the action click so
+   * a fast broadcast isn't missed.
+   */
+  private async captureBroadcastHash(timeout: number): Promise<string> {
```
tac0turtle, line 179: **[suggestion]** `captureBroadcastHash` and `startTxConfirmation` are copied from `TradePage` rather than extracted next to `pollTxOnChain` in `packages/e2e/utils/tx-confirm.ts`. The copies already diverged in this PR: TradePage's CheckTx error includes `codespace` (and logs it), TransactionsPage's throw at line 214 does not. The fail-fast matcher, abort/hash-clobber guard, and toast/REST race will have to be patched twice the next time this flakes.

**Suggestion:** Lift hash capture + confirmation race into `tx-confirm.ts` (or a shared page mixin) and have both page objects call it. Keep `lastTxHash` / `getLastTxUrl()` as the TransactionsPage-specific surface.

packages/e2e/pages/transactions-page.ts
```diff
@@ -170,6 +171,158 @@ export class TransactionsPage extends BasePage {
...
+   *   2. Fallback (REST): capture the broadcast hash and poll the LCD from
+   *      Node, which does not go through the browser proxy.
+   *
+   * Whichever confirms first wins; the loser is aborted. Rejects early on a
+   * deterministic failure (CheckTx rejection at broadcast, or the REST poll
+   * seeing the tx included with a non-zero code), otherwise only if BOTH fail.
+   *
+   * Arm this BEFORE the action click so the broadcast response isn't missed.
+   */
+  private startTxConfirmation(timeout = 120_000): Promise<void> {
```
tac0turtle, line 238: **[suggestion]** Cancel/claim previously waited 40s for the toast, which failed inside the 90s Playwright test timeout (`packages/e2e/playwright.config.ts`) with a visible assertion error. They now use TradePage's 120s budget, including on `waitForResponse`. These helpers only wait 15s for Keplr and do not have TradePage's 60–75s approval-retry window, so a missing broadcast (the documented #4478 case: popup never appears, nothing is signed) still dies as `Test timeout of 90000ms exceeded` instead of `broadcast response contained no txhash` / toast timeout. CheckTx and on-chain `code !== 0` fail fast; the no-tx path does not.

**Suggestion:** Use a confirmation budget that can actually reject before the test timeout (e.g. 40–60s, or remaining time until the 90s deadline). Keep the 120s default on TradePage if the Keplr retry window still needs it.

packages/e2e/pages/trade-page.ts
```diff
@@ -831,14 +843,36 @@ export class TradePage extends BasePage {
...
+    // A non-zero on-chain code is deterministic: the tx is included and failed,
+    // so the toast will never arrive. Racing that rejection against the pair
+    // below surfaces the chain error immediately instead of letting the toast
+    // branch run out its budget (which exceeds Playwright's test timeout and
+    // would report a generic timeout instead of the real reason).
+    const chainFailure = restPromise.then(
+      () => new Promise<never>(() => {}),
+      (err: Error) => {
+        if (
+          err?.message?.includes("failed on-chain") ||
```
tac0turtle, line 855: **[suggestion]** Fail-fast depends on `err.message.includes("failed on-chain")` / `"rejected at CheckTx"` in both page objects. A wording tweak in `pollTxOnChain` or `captureBroadcastHash` silently restores the old "toast branch runs until the test timeout" behavior. The outer `.catch` then wraps even those deterministic errors as `Transaction not confirmed via WebSocket toast or on-chain REST poll: …`, so the chain reason is buried in a message that claims confirmation never happened. Same shape at `packages/e2e/pages/transactions-page.ts:278` and the wrapper at line 306.

**Suggestion:** Throw a small typed error (`ChainTxError` / `{ kind: "check-tx" | "deliver-tx" }`) from the capture/poll helpers and `instanceof`-check it in `chainFailure`. Rethrow those errors as-is instead of wrapping them in the generic confirmation timeout.

packages/e2e/pages/transactions-page.ts
```diff
@@ -272,9 +409,7 @@ export class TransactionsPage extends BasePage {
       await approveBtn.click();
       await successPromise;
     } else {
-      console.log(
-        "1CT or pre-approved; success toast received (claim all)."
-      );
+      console.log("1CT or pre-approved; success toast received (claim all).");
```
tac0turtle, line 412: **[nit]** The 1CT branch still logs `success toast received` in `claimAll` (and the unchanged cancel / claim-and-close logs). After this PR that branch also wins when the REST poll confirms and the toast never rendered, which is the case the PR is fixing.

**Suggestion:** Log `transaction confirmed` (or include the winner already printed by `startTxConfirmation`) so CI logs do not imply a toast that was not there.

tac0turtle (approved):
left some nits with grok otherwise looks good

packages/e2e/pages/transactions-page.ts
JohnnyWyles: Scope expansion, can fix if we see this test also fail frequently

packages/e2e/pages/trade-page.ts
JohnnyWyles: Would fix at the same time as these wordings change rather than reinforcing now

PR #4481, opened 2026-09-15
perf(web): upgrade Next.js 16 with Turbopack

tac0turtle:
## What is the purpose of the change:

Upgrade `@osmosis-labs/web` from Next 14.2 / webpack / React 18 to Next 16.3 / Turbopack / React 19. Stay on Pages Router.

Cold `next build` on this machine (same generated lists, empty `.next`):

| | stage (14.2 webpack) | this PR (16.3 Turbopack) |
|---|---|---|
| Compile | 27–34s | 14s |
| TypeScript | 20s | 16s |
| **Wall** | **88s** | **38s** |

Warm Turbopack compile (FS cache) was 3.2s. CI will stay closer to the cold number unless `.next` is cached between jobs.

### Linear Task

n/a

## Brief Changelog

- Next `^16.3.5`, React 19, `nuqs` v2 (`NuqsAdapter` in `_app`)
- Turbopack is default (`next dev --turbopack` / `next build --turbopack`); webpack kept for `ANALYZE=true`
- Port libsodium / bip39 / launchpad aliases and `sprite.svg` to `turbopack`
- `middleware.ts` → `proxy.ts`
- `next lint` → `eslint .` (`eslint-config-next` stays on 14 until ESLint 9 / flat config)
- React 19 type/API fixes (`useRef` init, nullable `RefObject`, Stepper no longer uses `defaultProps`)

## Testing and Verifying

This change has been tested locally by rebuilding the website (`next build --turbopack` succeeds). Not clicked through in the browser; SSG still logs missing local Upstash Redis (`KV_STORE_*`), same as stage.

Follow-ups that are **not** in this PR: App Router, React Compiler, ESLint 9, CI cache of `.next`.

PR #4484, opened 2026-09-15
fix(web): transform nuqs ESM in Jest

tac0turtle:

Stacked on the Edge runtime PR. nuqs v2 (from #4481) is ESM-only. Jest was failing to parse it for any test that imports the swap tool / adapter.

### Linear Task

n/a

## Brief Changelog

- Add `nuqs` to web Jest `transformIgnorePatterns`

## Testing and Verifying

`yarn --cwd packages/web` jest: 55 suites, 621 tests passed.


tac0turtle:
Jest nuqs transform landed on #4481 instead so this PR is redundant.

